PR #82, opened 2021-06-30
[Password page] Add default blocks

### LINKED ISSUE #3: Password template: Blocks missing in Email Sign Up section
https://screenshot.click/21-06-cm2sm-li704.png

**Problem**
Blocks are missing in the email sign up section that is installed by default on Password template. 

**Expected**
All the blocks should display by default.

ludoboludo:

Fixes #3 .

Added the block settings to the `.json` template. It was using some settings as if the section didn't have blocks. 

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/120822071318/editor)

**Checklist**

### DIFF
--- a/templates/password.json
+++ b/templates/password.json
@@ -19,12 +19,18 @@
             "paragraph": "<p>Be the first to know when we launch.</p>"
           }
         },
-        "email": {
+        "email_form": {
           "type": "email_form"
         }
       },
-      "block_order": ["heading", "subheading", "email"]
+      "block_order": [
+        "heading",
+        "subheading",
+        "email_form"
+      ]
     }
   },
-  "order": ["main"]
+  "order": [
+    "main"
+  ]
 }
--- a/templates/password.json
+++ b/templates/password.json
@@ -4,13 +4,27 @@
     "main": {
       "type": "newsletter",
       "settings": {
-        "heading": "Opening soon",
-        "subheading": "<p>Be the first to know when we launch.</p>",
         "full_width": false
-      }
+      },
+      "blocks": {
+        "heading": {
+          "type": "heading",
+          "settings": {
+            "heading": "Opening soon"
+          }
+        },
+        "subheading": {
+          "type": "paragraph",
+          "settings": {
+            "paragraph": "<p>Be the first to know when we launch.</p>"
+          }
+        },
+        "email": {
+          "type": "email_form"
+        }
+      },
+      "block_order": ["heading", "subheading", "email"]
     }
   },
-  "order": [
-    "main"
-  ]
+  "order": ["main"]
 }
--- a/templates/password.json
+++ b/templates/password.json
@@ -13,7 +13,7 @@
             "heading": "Opening soon"
           }
         },
-        "subheading": {
+        "paragraph": {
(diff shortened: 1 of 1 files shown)


templates/password.json
```diff
@@ -4,13 +4,27 @@
...
+          "type": "email_form"
+        }
+      },
+      "block_order": ["heading", "subheading", "email"]
     }
   },
-  "order": [
-    "main"
-  ]
+  "order": ["main"]
```
chrisberthe, line 29: Can you preserve the formatting? I assume you have auto-formatting on in VS Code.

templates/password.json
```diff
@@ -4,13 +4,27 @@
...
+            "heading": "Opening soon"
+          }
+        },
+        "subheading": {
+          "type": "paragraph",
+          "settings": {
+            "paragraph": "<p>Be the first to know when we launch.</p>"
+          }
+        },
+        "email": {
```
chrisberthe, line 22: Same as my other comment. I have a personal preference for `email_form`—it describes the block.

templates/password.json
```diff
@@ -4,13 +4,27 @@
...
-      }
+      },
+      "blocks": {
+        "heading": {
+          "type": "heading",
+          "settings": {
+            "heading": "Opening soon"
+          }
+        },
+        "subheading": {
```
chrisberthe, line 16: Perhaps we should remain consistent with the block names defined in [`sections/newsletter.liquid`](https://github.com/Shopify/dawn/blob/main/sections/newsletter.liquid#L117).

templates/password.json
```diff
@@ -4,13 +4,27 @@
...
+          "type": "paragraph",
+          "settings": {
+            "paragraph": "<p>Be the first to know when we launch.</p>"
```
chrisberthe, line 19: Can we not instead use the [default value](https://github.com/Shopify/dawn/blob/main/sections/newsletter.liquid#L124) that's already defined?

templates/password.json
ludoboludo: the name of the block is `subheading` that's why I thought it would make sense to keep it this way. [Screenshot](https://screenshot.click/21-06-paxbx-whf5g.png)

templates/password.json
ludoboludo: I don't think so as this is more specific to the password page. Otherwise the default text would be: `Be the first to know about new collections and exclusive offers.`. Which isn't too bad but not as on point I think. 

I can chat it over with a designer if needed 🤔

templates/password.json
chrisberthe: I did some scanning across the theme and we're pretty consistently naming this the block type. Also it requires you to navigate to `*.schema.json` to know where this is coming from.

templates/password.json
Guillaumegranger1: I think `Be the first to know when we launch.` was suggested by @gretchen-willenberg I'm ok with using the other sentence if we have to, but I think this one's better for the password  page.

chrisberthe (approved):

sofiamatulis (approved):
Looks good 👍 💯

PR #87, opened 2021-06-30
add title logic in the <head>

### LINKED ISSUE #77: [Global] Page title identical throughout
## Issue summary

The page `title` content is the same for each page throughout: "OS 2.0." The `title` is the first bit of content conveyed to screen reader users. As a result, screen reader users rely on unique page `title` content to immediately confirm the current position within the site.

In addition, context-aware `title` content is also required to front-load critical details of the current state. Without this context, users may be unaware of the current state of a page and need to search around the page content to understand the current context.

<details>
Screenshot of Dawn theme product page. Browser tab content of OS 2.0 is highlighted. Web inspector reveals the title element which is also highlighted.

</details>

#### HTML

```html
<title>OS 2.0</title>
```

## Steps to reproduce

1. Go to a product page, using Dawn with any supported browser
2. Review the `title` content in the browser tab
3. With an active screen reader, navigate through the main navigation and load a new page

### Expected

- `title` content to reflect the current page context.

### Actual

- `title` content is the same throughout.

## Recommendation

Adjust the `title` content to be based on the current page content and state.

Include context-aware state to front-load information right away for screen reader users.

#### HTML

```html
<!-- Homepage -->
<title>OS 2.0</title>

<!-- Product page -->
<title>{Product title} – OS 2.0</title>

<!-- Search results page with results found -->
<title>9 results found for "sandal" – Search results – OS 2.0</title>

<!-- Search results page with no results found -->
<title>No results found for "glasses" – Search results – OS 2.0</title>

<!-- Contact page in error state -->
<title>Error – Contact – OS 2.0</title>
```

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **Component:** Global
- **WCAG SC:** [2.4.2 Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
- **Severity:** High
- **Effort:** Medium

### PR DESCRIPTION
**Pull request by ludoboludo**

About ->  #77 .

Added logic to have something similar to what's shown in the issue

```html
<!-- Homepage -->
<title>OS 2.0</title>

<!-- Product page -->
<title>{Product title} – OS 2.0</title>

<!-- Search results page with results found -->
<title>9 results found for "sandal" – Search results – OS 2.0</title>

<!-- Search results page with no results found -->
<title>No results found for "glasses" – Search results – OS 2.0</title>

<!-- Contact page in error state -->
<title>Error – Contact – OS 2.0</title>
```

**Other considerations**

Should we also add the filtering tags in the title ? When looking at Debut, it had:
```liquid
{%- if current_tags -%}
  {%- assign meta_tags = current_tags | join: ', ' -%} &ndash; {{ 'general.meta.tags' | t: tags: meta_tags -}}
{%- endif -%}
```
Since now we're using a different [filtering approach](https://shopify.dev/api/liquid/objects/filter), it would need some new logic. 

I could do something like this: 
```liquid 
if request.page_type == 'collection'
  for filter in collection.filters
    for value in filter.values
      if value.active
        echo ' &ndash; '
        echo value.label
      endif
    endfor
  endfor
endif
```

And I would need to update the JS for filters so it updates the `title` the same way it currently updates the `url` with `updateURLHash()`.

---

***In the desired output above,*** we would add `Error` in the title to reflect an error in form submission. `form.errors` is something that can only be used within a form. I could use it to do an assign and check against that assign in `theme.liquid` but that seem like a scrappy solution. 

It would be ideal if we didn't have to hard code the logic for the page title as it usually won't differ from theme to theme. I'm curious to hear what @Thibaut thinks about it. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=120824234006)
- [Editor](https://os2-demo.myshopify.com/admin/themes/120824234006/editor)

**Checklist**

### DIFF
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -12,7 +12,33 @@
       <link rel="preconnect" href="https://fonts.shopifycdn.com" crossorigin>
     {%- endunless -%}
 
-    <title>{{ shop.name }}</title>
+    {%- liquid
+      capture seo_title
+      if request.page_type == 'search' and search.performed == true
+        if search.results_count == 1
+          echo 'general.search.results_with_count.one' | t: count: search.results_count, terms: search.terms
+        elsif search.results_count > 1
+          echo 'general.search.results_with_count.other' | t: count: search.results_count, terms: search.terms
+        else
+          echo 'general.search.no_results' | t: terms: search.terms
+        endif
+      else
+        echo page_title
+      endif
+
+      if current_page != 1
+        echo ' &ndash; '
+        echo 'general.meta.page' | t: page: current_page
+      endif
+
+      assign escaped_page_title =  page_title | escape
+      unless escaped_page_title contains shop.name
+        echo ' &ndash; '
+        echo shop.name
+      endunless
+    endcapture -%}
+
+    <title>{{ seo_title | strip }}</title>
 
     {% if page_description %}
       <meta name="description" content="{{ page_description | escape }}">
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -15,26 +15,9 @@
     {%- liquid
       capture seo_title
       if request.page_type == 'search' and search.performed == true
-        if search.results_count == 1
-          echo 'general.search.results_with_count.one' | t: count: search.results_count, terms: search.terms
-        elsif search.results_count > 1
-          echo 'general.search.results_with_count.other' | t: count: search.results_count, terms: search.terms
-        else
-          echo 'general.search.no_results' | t: terms: search.terms
-        endif
+        echo 'general.search.results_with_count' | t: count: search.results_count, terms: search.terms
       else
         echo page_title
-      endif
-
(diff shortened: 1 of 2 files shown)


layout/theme.liquid
```diff
@@ -12,7 +12,44 @@
...
+      if request.page_type == 'collection'
+        for filter in collection.filters
+          for value in filter.values
+            if value.active
+              echo ' / '
+              echo value.label
+            endif
+          endfor
+        endfor
+      endif
```
ludoboludo, line 38: Added this to see how it would work for the filtering to show up. I would still need to add something in the JS to update the page title as it updates the `updateURLHash()` if we decide to do it this way.

chrisberthe: Seems really granular to have theme devs update the title like this, especially as we release new features that could potentially impact it, e.g. collection filtering. I would remove for now until we have a better general sense of how we want to improve our approach to titles in our themes.

PR #83, opened 2021-06-30
Update color variables for rgb. Remove unused color variables

KaichenWang:
**Why are these changes introduced?**
- Update definition of CSS RGB color variables in `Liquid` to leverage updated `ColorDrop`
- Fix issue with invalidated CSS when setting a theme color setting to `none`

Fixes https://github.com/Shopify/dawn/issues/52

**What approach did you take?**
- Define RGB colors using `ColorDrop` `red`, `green` and `blue` properties: 
```
--color-base-text-rgb: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
```
- Remove any unreferenced CSS color variables (clean up)

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=120822431766)
- [Editor](https://os2-demo.myshopify.com/admin/themes/120822431766/editor)

**Checklist**
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [x] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [x] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [x] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #97, opened 2021-07-05
Update color CSS variables in Liquid to only use RGB

KaichenWang:

Address https://github.com/Shopify/dawn/pull/83#discussion_r661669932

1. Simplify CSS variables in Liquid - Use only RGB definition of each color setting:

#### Before
``` css
/* Define color and its RGB equivalent in Liquid */
--color-base-text: {{ settings.colors_text }};
--color-base-text-rgb: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};

/* Usage in CSS */
color: var(--color-base-text); /* Opacity = 1 */
color: rgba(var(--color-base-text-rgb), 0.5); /* Opacity < 1 */
```
#### After
``` css
/* Define only RGB color in Liquid */
--color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};

/* Usage in CSS */
color: rgb(var(--color-base-text)); /* Opacity = 1 */
color: rgba(var(--color-base-text), 0.5); /* Opacity < 1 */
```
2. Refactor usage of CSS variables to accommodate simplification mentioned in point 1
3. Use same Liquid color settings in `giftcard` template and `password` layout
4. [Additional updates](https://github.com/Shopify/dawn/pull/97#issuecomment-875043775)

**Other considerations**

Also explored (but did not include in this PR) a solution for addressing:
> When color setting is set to none the R,G,B property values are 0,0,0, which is "black"

Potential solution adds some unwanted complexity to color system, so a separate issue will be created to allow for further exploration and discussion.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=120836161558)
- [Editor](https://os2-demo.myshopify.com/admin/themes/120836161558/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2,52 +2,52 @@
 
 :root,
 .color-background-1 {
-  --color-foreground: var(--color-base-text);
-  --color-foreground-rgb: var(--color-base-text-rgb);
-  --color-background: var(--color-base-background-1);
-  --color-background-rbg: var(--color-base-background-1-rgb);
-  --payment-terms-background-color: var(--color-base-background-1);
+  --color-foreground: rgb(var(--color-base-text));
+  --color-foreground-rgb: var(--color-base-text);
+  --color-background: rgb(var(--color-base-background-1));
+  --color-background-rbg: var(--color-base-background-1);
+  --payment-terms-background-color: rgb(var(--color-base-background-1));
 }
 
 .color-background-2 {
-  --color-background: var(--color-base-background-2);
-  --color-background-rbg: var(--color-base-background-2-rgb);
+  --color-background: rgb(var(--color-base-background-2));
+  --color-background-rbg: var(--color-base-background-2);
 }
 
 .color-inverse {
-  --color-foreground: var(--color-base-background-1);
-  --color-foreground-rgb: var(--color-base-background-1-rgb);
-  --color-background: var(--color-base-text);
-  --color-background-rbg: var(--color-base-text-rgb);
+  --color-foreground: rgb(var(--color-base-background-1));
+  --color-foreground-rgb: var(--color-base-background-1);
+  --color-background: rgb(var(--color-base-text));
+  --color-background-rbg: var(--color-base-text);
 }
 
 .color-accent-1 {
-  --color-foreground: var(--color-base-solid-button-labels);
-  --color-foreground-rgb: var(--color-base-solid-button-labels-rgb);
-  --color-background: var(--color-base-accent-1);
-  --color-background-rbg: var(--color-base-accent-1-rgb);
+  --color-foreground: rgb(var(--color-base-solid-button-labels));
+  --color-foreground-rgb: var(--color-base-solid-button-labels);
+  --color-background: rgb(var(--color-base-accent-1));
+  --color-background-rbg: var(--color-base-accent-1);
 }
 
 .color-accent-2 {
-  --color-foreground: var(--color-base-solid-button-labels);
-  --color-foreground-rgb: var(--color-base-solid-button-labels-rgb);
-  --color-background: var(--color-base-accent-2);
-  --color-background-rbg: var(--color-base-accent-2-rgb);
+  --color-foreground: rgb(var(--color-base-solid-button-labels));
+  --color-foreground-rgb: var(--color-base-solid-button-labels);
+  --color-background: rgb(var(--color-base-accent-2));
+  --color-background-rbg: var(--color-base-accent-2);
 }
 
 .color-foreground-outline-button {
-  --color-foreground: var(--color-base-outline-button-labels);
-  --color-foreground-rbg: var(--color-base-outline-button-labels-rgb);
+  --color-foreground: rgb(var(--color-base-outline-button-labels));
+  --color-foreground-rbg: var(--color-base-outline-button-labels);
 }
 
 .color-foreground-accent-1 {
-  --color-foreground: var(--color-base-accent-1);
-  --color-foreground-rbg: var(--color-base-accent-1-rgb);
+  --color-foreground: rgb(var(--color-base-accent-1));
+  --color-foreground-rbg: var(--color-base-accent-1);
 }
 
 .color-foreground-accent-2 {
-  --color-foreground: var(--color-base-accent-2);
-  --color-foreground-rbg: var(--color-base-accent-2-rgb);
+  --color-foreground: rgb(var(--color-base-accent-2));
+  --color-foreground-rbg: var(--color-base-accent-2);
 }
 
 :root,
@@ -73,8 +73,8 @@
 
 :root,
 .color-background-1 {
-  --color-link: var(--color-base-outline-button-labels);
-  --color-link-hover: rgba(var(--color-base-outline-button-labels-rgb), 0.85);
+  --color-link: rgb(var(--color-base-outline-button-labels));
+  --color-link-hover: rgba(var(--color-base-outline-button-labels), 0.85);
 }
 
 .color-background-2,
@@ -89,10 +89,10 @@
 
 :root,
 .color-background-1 {
-  --color-button-background: var(--color-base-accent-1);
-  --color-button-border: var(--color-base-accent-1);
-  --color-button-text: var(--color-base-solid-button-labels);
-  --color-button-text-rgb: var(--color-base-solid-button-labels-rgb);
+  --color-button-background: rgb(var(--color-base-accent-1));
+  --color-button-border: rgb(var(--color-base-accent-1));
+  --color-button-text: rgb(var(--color-base-solid-button-labels));
+  --color-button-text-rgb: var(--color-base-solid-button-labels);
 }
 
 .color-background-2,
@@ -109,9 +109,9 @@
 
 .button--secondary {
   --color-button-background: transparent;
-  --color-button-border: var(--color-base-outline-button-labels);
-  --color-button-text: var(--color-base-outline-button-labels);
-  --color-button-text-rgb: var(--color-base-outline-button-labels-rgb);
+  --color-button-border: rgb(var(--color-base-outline-button-labels));
+  --color-button-text: rgb(var(--color-base-outline-button-labels));
+  --color-button-text-rgb: var(--color-base-outline-button-labels);
 }
 
 .color-background-2 .button--secondary,
@@ -127,9 +127,9 @@
 
 .button--tertiary {
   --color-button-background: transparent;
-  --color-button-border: rgba(var(--color-base-outline-button-labels-rgb), 0.2);
-  --color-button-text: var(--color-base-outline-button-labels);
-  --color-button-text-rgb: var(--color-base-outline-button-labels-rgb);
+  --color-button-border: rgba(var(--color-base-outline-button-labels), 0.2);
+  --color-button-text: rgb(var(--color-base-outline-button-labels));
+  --color-button-text-rgb: var(--color-base-outline-button-labels);
 }
 
 .color-background-2 .button--tertiary,
@@ -161,16 +161,16 @@
 :root,
 .color-background-1,
 .color-background-2 {
-  --color-card-hover: rgba(var(--color-base-text-rgb), 0.06);
+  --color-card-hover: rgba(var(--color-base-text), 0.06);
 }
 
 .color-inverse {
-  --color-card-hover: rgba(var(--color-base-background-1-rgb), 0.06);
+  --color-card-hover: rgba(var(--color-base-background-1), 0.06);
 }
 
 .color-accent-1,
 .color-accent-2 {
-  --color-card-hover: rgba(var(--color-base-solid-button-labels-rgb), 0.06);
+  --color-card-hover: rgba(var(--color-base-solid-button-labels), 0.06);
 }
 
 /* base */
@@ -848,7 +848,7 @@ summary::-webkit-details-marker {
 }
 
 .title-wrapper-with-link a {
-  color: var(--color-base-outline-button-labels);
+  color: rgb(var(--color-base-outline-button-labels));
   margin-top: 0;
   flex-shrink: 0;
 }
@@ -1593,8 +1593,8 @@ input[type='checkbox'] {
 
 .cart-count-bubble {
   position: absolute;
-  background-color: var(--color-base-accent-1);
-  color: var(--color-base-solid-button-labels);
+  background-color: rgb(var(--color-base-accent-1));
+  color: rgb(var(--color-base-solid-button-labels));
   height: 1.7rem;
   width: 1.7rem;
   border-radius: 100%;
--- a/assets/base.css
+++ b/assets/base.css
@@ -42,8 +42,7 @@
 :root,
 .color-background-1 {
   --color-link: var(--color-base-outline-button-labels);
-  --color-link-hover: var(--color-base-outline-button-labels);
-  --alpha-link-hover: 0.85;
+  --alpha-link: 0.85;
 }
 
 .color-background-2,
@@ -51,8 +50,7 @@
 .color-accent-1,
 .color-accent-2 {
   --color-link: var(--color-foreground);
-  --color-link-hover: var(--color-foreground);
-  --alpha-link-hover: 0.7;
+  --alpha-link: 0.7;
 }
 
 /* Primary button */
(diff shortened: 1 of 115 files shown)


layout/theme.liquid
```diff
@@ -45,20 +45,13 @@
...
+        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
+        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};       
+        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};       
+        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};      
+        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};     
+        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};      
+        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```
chrisberthe, line 54:
```suggestion
        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};
        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};
        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};
        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};
        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};
        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```

templates/gift_card.liquid
```diff
@@ -44,14 +44,13 @@
...
+        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
+        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};       
+        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};       
+        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};      
+        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};     
+        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};      
+        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```
chrisberthe, line 53:
```suggestion
        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};
        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};
        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};
        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};
        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};
        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```

layout/password.liquid
```diff
@@ -42,20 +42,13 @@
...
+        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
+        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};       
+        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};       
+        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};      
+        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};     
+        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};      
+        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```
chrisberthe, line 51:
```suggestion
        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};
        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};
        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};
        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};
        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};
        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
```

assets/collage.css
```diff
@@ -397,7 +397,7 @@
   left: 0;
   overflow: auto;
   width: 100%;
-  background: rgba(var(--color-base-text-rgb), 0.2);
+  background: var(--color-foreground-20);
```
Thibaut, line 400: Not a fan of these "percentage" variables. I think `rgba(var(--color-base-text-rgb), 0.2)` is more readable. What do we gain by using variables that mimic `rgba`?

Thibaut (commented):
:+1: overall, nice cleanup!

assets/base.css
```diff
@@ -2,52 +2,52 @@
 
 :root,
 .color-background-1 {
-  --color-foreground: var(--color-base-text);
-  --color-foreground-rgb: var(--color-base-text-rgb);
-  --color-background: var(--color-base-background-1);
-  --color-background-rbg: var(--color-base-background-1-rgb);
-  --payment-terms-background-color: var(--color-base-background-1);
+  --color-foreground: rgb(var(--color-base-text));
+  --color-foreground-rgb: var(--color-base-text);
```
chrisberthe, line 6: I find this a bit confusing—I'd expect `-rgb` to be the one with the `rgb` function attached. Thoughts on instead having a single color, e.g. `--color-foreground: var(--color-base-text)` for both so we can rely on a single custom property for all references?

assets/section-password.css
```diff
@@ -339,5 +339,5 @@ details.modal .modal__toggle-open {
   width: 7rem;
   height: 2rem;
   vertical-align: top;
-  color: var(--color-text);
+  color: rgb(var(--color-base-text));
```
ludoboludo, line 342: This needs to be `--color-forground` instead otherwise the icon isn't updating it's color:

assets/base.css
ludoboludo: Then the one thing about this approach as I was thinking about it is that we need to add `rbg(...)` around the `var(--color-base-text)` variables in a few files which then adds more lines/characters overall. 

But I do find it clearer to have one single custom property rather than 2.

assets/collage.css
KaichenWang: Mostly to make the CSS a bit less verbose
``` css
div {
  color: rgba(var(--color-foreground-rgb), 0.2);
}

div {
  color: var(--color-foreground-20);
}
```

But yes I think we can use the former instead of the latter if it's less ambiguous

assets/base.css
```diff
@@ -3,96 +3,65 @@
 :root,
 .color-background-1 {
   --color-foreground: var(--color-base-text);
-  --color-foreground-rgb: var(--color-base-text-rgb);
   --color-background: var(--color-base-background-1);
-  --color-background-rbg: var(--color-base-background-1-rgb);
-  --payment-terms-background-color: var(--color-base-background-1);
```
KaichenWang, line 9: Anyone know where this was being used?

ludoboludo: ah It might be for installments 🤔  Cause I remember there was something specific that was needed for the icon used in the payment terms to change itself and keep a good color contrast. So even though it looks like it's not used anywhere in the theme, it is by the installment API. 

To test it, you just need to change the currency in the test store to USD and the installments' info will show up. Then try and change the background color to a similar purple of the icon and it should automatically change.

KaichenWang:
## Update

Following feedback, these updates were made:

### 1. All color variables are now in `R,G,B` format

#### Before
There were 2 types of color variables:
1. with suffix `-rgb` - meant to be wrapped with `rgb(...)` or `rgba(...)`
2. without suffix `-rgb` - meant to be used as-is

#### After
- Simplification with removal of suffix `-rgb` variables
- **All** color variables now **need** to be wrapped with `rgb(...)` or `rgba(...)`

### 2. Removed variables for foreground colors with opacity

#### Before
Variables `--color-foreground-XX` (with `XX` denoting opacity percentage) were defined and could be used throughout the theme's CSS
``` css
/* example: foreground color with 20% opacity */
color: var(--color-foreground-20);
```
#### After
This abstraction was removed in favour of simplicity. Instead, use `rgba()` in combination with `--color-foreground` variable.
``` css
/* example: foreground color with 20% opacity */
color: rgba(var(--color-foreground), 0.2);
```

assets/base.css
KaichenWang: @chrisberthe Updated. [See details](https://github.com/Shopify/dawn/pull/97#issuecomment-875043775)

assets/collage.css
KaichenWang: @Thibaut Updated. [See details](https://github.com/Shopify/dawn/pull/97#issuecomment-875043775)

assets/base.css
chrisberthe: Essentially what Ludo said—here's some additional context for what we're checking for:
1. Check the parent element of the banner for a CSS variable called --payment-terms-background-color
2. Check the document root for a CSS variable called --payment-terms-background-color
3. Check the document root for a CSS variable called --color-body
4. If none are present the logo will default to purple

Something I completely forgot about though is that the CSS custom properties above _need_ to be specified in hex, so let's move this custom property to all layout files and update to `--payment-terms-background-color: {{ settings.colors_background_1 }}`.

chrisberthe (approved):

assets/base.css
```diff
@@ -1,151 +1,106 @@
...
-  --color-foreground-55: rgba(var(--color-foreground-rgb), 0.55);
-  --color-foreground-50: rgba(var(--color-foreground-rgb), 0.5);
-  --color-foreground-40: rgba(var(--color-foreground-rgb), 0.4);
-  --color-foreground-30: rgba(var(--color-foreground-rgb), 0.3);
-  --color-foreground-20: rgba(var(--color-foreground-rgb), 0.2);
-  --color-foreground-10: rgba(var(--color-foreground-rgb), 0.1);
-  --color-foreground-8: rgba(var(--color-foreground-rgb), 0.08);
-  --color-foreground-4: rgba(var(--color-foreground-rgb), 0.04);
-  --color-foreground-3: rgba(var(--color-foreground-rgb), 0.03);
-}
```
ludoboludo, line 70: Removing those I guess makes it a little longer to update if ever needed when we go back and forth with designers.

templates/gift_card.liquid
```diff
@@ -44,14 +44,14 @@
...
+        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
+        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};
+        --color-base-background-2: {{ settings.colors_background_2.red }}, {{ settings.colors_background_2.green }}, {{ settings.colors_background_2.blue }};
+        --color-base-solid-button-labels: {{ settings.colors_solid_button_labels.red }}, {{ settings.colors_solid_button_labels.green }}, {{ settings.colors_solid_button_labels.blue }};
+        --color-base-outline-button-labels: {{ settings.colors_outline_button_labels.red }}, {{ settings.colors_outline_button_labels.green }}, {{ settings.colors_outline_button_labels.blue }};
+        --color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};
+        --color-base-accent-2: {{ settings.colors_accent_2.red }}, {{ settings.colors_accent_2.green }}, {{ settings.colors_accent_2.blue }};
+        --payment-terms-background-color: {{ settings.colors_background_1 }};
```
ludoboludo, line 54: I don't think this is needed here as you can't add any section to the gift card page. So there shouldn't be a need for the `payment-terms-background-color` variable.

ludoboludo (commented):
Looks good to me 👌 

Left a comment to remove a variable where it's not needed. 

I do feel like we're losing in ease of customization by hard coding the alpha value in quite a few places vs having css variables set for the different alpha values we need across the theme (like we had with `var(--color-foreground-20)`). 
I did find the variable names straight forward and if you'd need to make a change then you can update it across the theme in one place rather than having to search for each `0.2` values for example. 
Not sure how often we will come across the situation where the opacity will need to be changed (designers choose to make a change to the theme or theme support customization for individual merchants) but something to keep in mind 🤔

templates/gift_card.liquid
KaichenWang: Removed

ludoboludo (dismissed):

chrisberthe (dismissed):

ludoboludo (dismissed):

chrisberthe (approved):

bakura10:
Hi !

I really like the approach of using one variable, that is used either in rgb() or rgba() filter based on the use case. Here are a few additions that would make it simpler to work with colors:

* A new "rgb" variable that would output all the three components: {{ color.rgb }}

* It would be extremely useful to be able to extend the default filter to work with transparent color:

```
{% assign my_color = section.settings.background | default: settings.background }}
```

We are using a lot of sections where we want to fallback on default color if none has been set.

PR #98, opened 2021-07-05
Fixed active facets tap target and drawer scrolling

tyleralsbury:

First part of faceted filtering fast follow fixes. #100

_Tap targets_

Changed up the markup a bit for the pills so that they now have a wrapper `<a>` with a `<span>` inside that have the pill styles.  The `<a>` has padding to ensure that it will always be at least `44px` (actually `45px` given the current line heights, etc...) to follow accessible tap target principles

> Equivalent targets: If there is more than one target on a screen that performs the same action, only one of the targets need to meet the target size of 44 by 44 CSS pixels.

[Source - w3.org](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html)

_Scrolling issue_

When a drawer is open, the content behind should not be scrollable.  The `menu-drawer` custom element was already adding a class for this, but the class wasn't doing anything. I added `overflow-y: hidden` to prevent vertical scrolling when the class is on the body.

**Other considerations**

The `menu-drawer` custom element is also used for the main menu and the overflow breakpoints are different for the two.  I've needed to add a mobile version and a tablet version of the overflow hidden class.

**Testing**

[Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=120835735574)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2072,3 +2072,7 @@ details-disclosure > details {
     transform: translateY(0);
   }
 }
+
+.overflow-hidden-mobile {
+  overflow-y: hidden;
+}
--- a/assets/base.css
+++ b/assets/base.css
@@ -2076,17 +2076,17 @@ details-disclosure > details {
 
 .overflow-hidden-mobile,
 .overflow-hidden-tablet {
-  overflow-y: hidden;
+  overflow: hidden;
 }
 
 @media screen and (min-width: 750px) {
   .overflow-hidden-mobile {
-    overflow-y: auto;
+    overflow: auto;
   }
 }
 
 @media screen and (min-width: 990px) {
   .overflow-hidden-tablet {
-    overflow-y: auto;
+    overflow: auto;
   }
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -2061,6 +2061,7 @@ details[open] > .header__icon--menu .icon-hamburger {
 details-disclosure > details {
   position: relative;
 }
+
 @keyframes animateMenuOpen {
   0% {
     opacity: 0;
@@ -2073,6 +2074,19 @@ details-disclosure > details {
   }
 }
 
-.overflow-hidden-mobile {
+.overflow-hidden-mobile,
+.overflow-hidden-tablet {
   overflow-y: hidden;
+}
+
+@media screen and (min-width: 750px) {
+  .overflow-hidden-mobile {
+    overflow-y: auto;
+  }
+}
+
+@media screen and (min-width: 990px) {
+  .overflow-hidden-tablet {
+    overflow-y: auto;
+  }
 }
--- a/assets/template-collection.css
+++ b/assets/template-collection.css
@@ -349,35 +349,62 @@ button.facets__button {
   grid-row: 2;
 }
 
-a.active-facets__button {
+.active-facets__button {
+  padding: 1.1rem 0.5rem;
+  text-decoration: none;
+}
+
+.active-facets__button:first-child {
+  margin-left: 0.5rem
+}
+
+span.active-facets__button-inner {
   color: var(--color-foreground);
   box-shadow: 0 0 0 0.1rem var(--color-foreground);
   border-radius: 2.6rem;
-  font-size: 1.2rem;
+  font-size: 1rem;
   min-height: 0;
   min-width: 0;
   padding: 0.5rem 1rem;
-  margin: 0 0 1rem 1rem;
+  display: flex;
+  align-items: center;
 }
 
-a.active-facets__button {
-  box-shadow: 0 0 0 0.1rem var(--color-foreground);
+@media screen and (min-width: 990px) {
+  .active-facets__button {
+    padding: 0;
+    margin: 0 0 1rem 1rem;
+  }
+
+  .active-facets__button:first-child {
+    margin-left: 1rem;
+  }
+
+  span.active-facets__button-inner {
+    font-size: 1.2rem;
+  }
 }
 
-a.active-facets__button:hover {
+.active-facets__button:hover .active-facets__button-inner {
   box-shadow: 0 0 0 0.2rem var(--color-foreground);
 }
 
-a.active-facets__button--light {
+.active-facets__button--light .active-facets__button-inner {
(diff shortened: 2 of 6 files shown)


assets/base.css
```diff
@@ -2072,3 +2073,20 @@ details-disclosure > details {
     transform: translateY(0);
   }
 }
+
+.overflow-hidden-mobile,
+.overflow-hidden-tablet {
+  overflow-y: hidden;
+}
+
```
sofiamatulis, line 2081: Since the original class doesn't target x or y (just `overflow`). What do you think in terms of naming? 🤔 

Maybe `overflow-vertical-hidden`? Or what do you think?

.overflow-hidden {
  overflow: hidden;
}

assets/template-collection.css
```diff
@@ -349,35 +349,62 @@ button.facets__button {
...
+.active-facets__button:first-child {
+  margin-left: 0.5rem
+}
+
+span.active-facets__button-inner {
   color: var(--color-foreground);
   box-shadow: 0 0 0 0.1rem var(--color-foreground);
   border-radius: 2.6rem;
-  font-size: 1.2rem;
+  font-size: 1rem;
```
sofiamatulis, line 365: In terms of accessibility, I dont think font sizes are supposed to be smaller than 12px 🤔 

Source: https://accessibility.psu.edu/fontsizehtml/#:~:text=For%20traditional%20computer%20monitors%2C%20a,can%20be%20zoomed%20to%20200%25.

assets/template-collection.css
```diff
@@ -447,7 +474,7 @@ a.active-facets__button:focus {
```
sofiamatulis, line 474: Not related to this PR, but facet filtering. I noticed that the button has an uneven focus at the top and bottom:

https://screenshot.click/05-42-rfdgj-17eqk.png

assets/template-collection.css
```diff
@@ -447,7 +474,7 @@ a.active-facets__button:focus {
 
 .mobile-facets__wrapper {
   margin-left: 0;
-  margin-bottom: 3rem;
+  margin-bottom: 1rem;
 }
```
sofiamatulis, line 479: Even though we cant scroll in the back anymore, I am still able to click on the header menu and other items:

https://screenshot.click/05-49-c7xxh-4exik.mp4

sections/main-collection-product-grid.liquid
```diff
@@ -361,20 +367,28 @@
...
           {%- if filter.min_value.value != nil or filter.max_value.value != nil -%}
             <a class="active-facets__button active-facets__button--light button button--tertiary js-facet-remove" href="{{ filter.url_to_remove }}">
-              {%- if filter.min_value.value -%}{{ filter.min_value.value | money }}{%- else -%}{{ 0 | money }}{%- endif -%}-{%- if filter.max_value.value -%}{{ filter.max_value.value | money }}{%- else -%}{{ filter.range_max | money }}{%- endif -%}
-              {% render 'icon-close-small' %}
+              <span class="active-facets__button-inner button button--tertiary">
+                {%- if filter.min_value.value -%}{{ filter.min_value.value | money }}{%- else -%}{{ 0 | money }}{%- endif -%}-{%- if filter.max_value.value -%}{{ filter.max_value.value | money }}{%- else -%}{{ filter.range_max | money }}{%- endif -%}
+                {% render 'icon-close-small' %}
+              </span>
             </a>
           {%- endif -%}
```
sofiamatulis, line 393: On Safari, the background is still scrollable: https://screenshot.click/05-51-wiacu-ueec2.mp4

sofiamatulis (commented):
Did a first pass and added some comments 👍

Oliviammarcello:
Looks good @tyleralsbury. 

Just have a few notes related to the active facets tap targets; drawer scrolling looks great. FYI All other UI notes have just been added to a [polish review here.](https://github.com/Shopify/dawn/issues/100#issuecomment-874265744)

- [ ] The active facets are outside of the margins on tablet and mobile. [Screenshot](https://screenshot.click/21-53-1dau2-f69z9.png)
- [ ] The 'X's to remove pills are slightly too high on mobile (desktop is perfect). [Screenshot](https://screenshot.click/21-48-9ofok-xq7tw.png)

assets/base.css
tyleralsbury: Maybe it makes sense to just target both in reality.  Changed it to `overflow: hidden`.

assets/template-collection.css
tyleralsbury: I'll add it to the issue as a minor bug.

assets/template-collection.css
tyleralsbury: Fixed - was a z-index issue.

sections/main-collection-product-grid.liquid
tyleralsbury: That's weird. Version?  It seems to be working as far as I can tell.

assets/template-collection.css
tyleralsbury: Spoke to @Oliviammarcello and we've decided to keep it at `10px`.  The user can just pinch to zoom and make the size bigger if they need to.  This is easy to do on mobile.

tyleralsbury:
Should be ok for a second round.

* We kept the font size at 1rem (checked with Olivia)
* I can't reproduce the Safari scrolling issue... gonna dig a bit and see.

sections/main-collection-product-grid.liquid
sofiamatulis: I haven't been able to reproduce it 🤔

tyleralsbury: 🎩 Magic!
🧚

sofiamatulis (approved):
Retested and it looks good :)

Haven't been able to reproduce the safari issue

sections/main-collection-product-grid.liquid
ludoboludo: Can't replicate anymore on my iPad either. 👍

ludoboludo (approved):

Oliviammarcello (approved):

tyleralsbury:
Merged - thanks for the review, y'all!

PR #105, opened 2021-07-06
Add variant picker sold out UI

### LINKED ISSUE #45: UI Polish Review: Product Template (Part 2)
# Part 2
# UI review findings (WIP)

### Border on media
- [x] All media need to have a border like on the product card, inside the product template and the modal. 🙏 
- [x] Remove background on all media except video

### Settings
- [x] [Move up](https://screenshot.click/21-07-wfgi1-gbv9x.png) `Enable sticky product information on large screens` setting to let the media related settings together. 

### Modal layout
- [x] We need to revert back the max-width change that we did. Seems like having the closing icon on the corner right over the media is not scalable for all media. It could overlay on top of important controls for buyers. [Screenshot](https://screenshot.click/21-17-mbkrl-hl5e9.png)
    - [x] [Reference](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=7086%3A75028) of the close position

**[Large breakpoint](https://screenshot.click/21-07-p5x5l-rq472.png)**
- [x] Media width 

> 
        max-width: 100%;
        padding: 2rem 10rem 0;

- [x] Close button position

> 
         top: 2.2rem;

**[Medium breakpoint (750px I believe)](https://screenshot.click/21-07-l96fa-czu5n.png)**

- [x] Media width

> 
        max-width: 100%;
        padding: 2rem 11rem 0;

- [x] Close button position

> 
         top: 2.2rem;

**Mobile**
- [x] Keep as is, but open minded if it requires some changes if I didn't see the full picture properly

### No JS
- [x] When I am tabbing, I can't get to the product images, just to the product information column. Wondering if this is expected? ([Video recording](https://screenshot.click/21-07-1apn8-92ymx.mp4))
- [x] Dynamic checkout button is not showing when JS is disabled and the [space is kept. The button](https://screenshot.click/21-07-hf14c-wkbeh.png) `Add to cart` could be styled as the primary button in this case instead of the outline secondary button.
- [x] Video media, could we display the video icon to announce the media type? ([Video recording](https://screenshot.click/21-07-ioflt-bte38.mp4))
- [ ] To check if this is still happening, but it worked when I tested on the latest demo you shared with me. https://github.com/Shopify/dawn/issues/179

### A11y
- [x] Dynamic checkout button focus style is not the same as the theme ([Video recording](https://screenshot.click/21-07-u2mky-rhg0l.mp4))
- [ ] The page title (the browser tab title for the page; not the main content heading or h1) does not describes the topic or purpose of the page. but should be fixed by https://github.com/Shopify/dawn/pull/87

### General

**Product modal**
- [x] Mobile/tablet: When I open the modal on mobile, if I play with the position of the image with my finger, and close the modal, opening another image opens the modal to the position I left the previous modal. I am wondering if we can align the media in the modal, top-center to have more success to be zoomed in a detail area of the image? I considered center-center, but would like to hear your thoughts.
- [ ] Media sizes inside the modal are won't always look great inside the modal, both on Mobile. MELI TO PROVIDE SIZING

**Variant picker**
- [ ] The button type doesn't [display the sold out UI](https://screenshot.click/21-06-u7m7g-ei8sn.png), [example](https://screenshot.click/21-06-qcoz9-7uvtd.png) of a sold out combination.

# 🎩 My review tasks

**_Note for myself; Quick list of things I am looking to review:_**
- [x] Rich media
- [x] No JS
- [x] Product modal with various ratio

Validate the sizing of each media type with designers. 
   - We need to create a Product with media that meets the following requirements:
      - [x] Multiple aspect ratios (horizontal and vertical)
      - [x] High and low resolution images
      - [x] Transparent and non-transparent background
      - [x] Matching and non-matching media posters
      - [x] Very tall portrait images (Snowboard)

_Suggestion from Ludo:_ 
> I wonder if maybe we could use a multiplier of the actual image width instead. The reason I say that is if I'm using a shitty image. That is going to make it very pixelated! So maybe if we grabbed the `image.width | times: 2` it might be ok whatever the media you're using.

## Interaction review
- [x] Try the no-JS flow

## A11Y review
- [x] Test the keyboard navigation and the tab order
- [x] The page title (the browser tab title for the page; not the main content heading or `h1`) describes the topic or purpose of the page.
- [x] All text that looks and acts like a heading is marked as a heading.
- [x] Heading levels are chosen to convey correct hierarchical order in content, not for visual styling.
- [ ] Decorative images are identified as not requiring alternative text. 
- [ ] N/A - Forms control markup is written so it can be used with Autofill. 
- [x] Touch target size of main links and buttons are large enough and far enough apart from each other to activate easily with a finger.
- [x] An alternative method to activate custom swipe actions or gestures is available (ex., zoom or pan a 3D model.)

#### Navigation and links
- [x] Content reading and focus/tab order, as determined by the code order, is logical and intuitive (top-to-bottom, left-to-right.) 
- [x] Actionable elements have a clear, visible focus state for non-mouse users.
- [x] Link text clearly describes the purpose or destination of the link. (Design annotation kit + PR, for generic links)
- [ ] N/A - Icon-only links have a text alternative. (Design annotation kit + PR)
- [ ] Links which open a new tab/window are accompanied by an aural and visual warning. (PR)

#### Magnification and Responsive Design
- [x] Low vision users can magnify or zoom in on the content in the browser on any device, including desktop and mobile.
- [ ] The design is optimized for all zoom states, eliminating horizontal scrolling.
- [x] The user can use the website in their preferred orientation (landscape/portrait.)

#### Dynamic Content
- [ ] Users are made aware of content that is dynamically inserted on a page. New content is positioned next in the DOM from the activator, in the logical reading order / tab order of the page.
- [x] All keyboard-only and touch screen interactions follow expected patterns so users know how to interact with widgets on the page.
- [ ] When users activate scripted functionality (buttons, form submissions, etc.), they know whether the action was successful or not, through a success/error message. The feedback must be available to sighted users, screen reader users, low-vision users, and all other user groups.

cfxd:
Add variant picker sold out UI, just a fist iteration to get the ball rolling

Fixes #45 (Variant picker).

Enable/disable options as available, including an option to allow disabled combinations to remain enabled for use with third party stock notification apps.

**Demo links**
Password: `fleomo`
- [Product with one option](https://cfx-theme-playground.myshopify.com/products/billowing-glade)
- [Product with two options](https://cfx-theme-playground.myshopify.com/products/ancient-water)
- [Product with three options](https://cfx-theme-playground.myshopify.com/products/autumn-hill)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -573,8 +573,11 @@ class VariantSelects extends HTMLElement {
         const html = new DOMParser().parseFromString(responseText, 'text/html')
         const destination = document.getElementById(id);
         const source = html.getElementById(id);
+        const variantPickerDestination = document.querySelector('variant-radios') || document.querySelector('variant-selects');
+        const variantPickerSource = html.querySelector('variant-radios') || html.querySelector('variant-selects');
 
         if (source && destination) destination.innerHTML = source.innerHTML;
+        if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
 
         document.getElementById(`price-${this.dataset.section}`)?.classList.remove('visibility-hidden');
         this.toggleAddButton(!this.currentVariant.available, window.variantStrings.soldOut);
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -168,10 +168,13 @@ fieldset.product-form__input .form__label {
   color: rgb(var(--color-background));
 }
 
-.product-form__input input[type='radio']:disabled + label {
+.product-form__input input[type='radio']:disabled + label, .product-form__input input[type='radio'].disabled + label {
   border-color: rgba(var(--color-foreground), 0.1);
   color: rgba(var(--color-foreground), 0.4);
   text-decoration: line-through;
+}
+.product-form__input input[type='radio'].disabled:checked + label, .product-form__input input[type='radio']:disabled:checked + label {
+  color: rgba(var(--color-background),.4);
 }
 .product-form__input input[type='radio']:focus-visible + label {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -10,6 +10,11 @@
 
 <script src="{{ 'product-form.js' | asset_url }}" defer="defer"></script>
 
+{%- assign variants_available_arr = product.variants | map: 'available' -%}
+{%- assign variants_option1_arr = product.variants | map: 'option1' -%}
+{%- assign variants_option2_arr = product.variants | map: 'option2' -%}
+{%- assign variants_option3_arr = product.variants | map: 'option3' -%}
+
 {%- assign first_3d_model = product.media | where: "media_type", "model" | first -%}
 {%- if first_3d_model -%}
   {{ 'component-product-model.css' | asset_url | stylesheet_tag }}
@@ -185,20 +190,38 @@
               {%- if block.settings.picker_type == 'button' -%}
                 <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
                   {%- for option in product.options_with_values -%}
-                      <fieldset class="js product-form__input">
-                        <legend class="form__label">{{ option.name }}</legend>
-                        {%- for value in option.values -%}
-                          <input type="radio" id="{{ section.id }}-{{ option.name }}-{{ forloop.index0 }}"
-                                name="{{ option.name }}"
-                                value="{{ value | escape }}"
-                                form="product-form-{{ section.id }}"
-                                {% if option.selected_value == value %}checked{% endif %}
-                          >
-                          <label for="{{ section.id }}-{{ option.name }}-{{ forloop.index0 }}">
-                            {{ value }}
-                          </label>
-                        {%- endfor -%}
-                      </fieldset>
+                    <fieldset class="js product-form__input">
+                      <legend class="form__label">{{ option.name }}</legend>
+                      {%- for value in option.values -%}
+                        {%- assign option_disabled = true -%}
+                        {% for option1_name in variants_option1_arr %}
+                          {% case option.position %}
+                            {% when 1 %}
+                              {% if variants_option1_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                            {% when 2 %}
+                              {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                            {% when 3 %}
+                              {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == product.selected_or_first_available_variant.option2 and variants_option3_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                          {% endcase %}
+                        {% endfor %}
+                        <input type="radio" id="{{ section.id }}-{{ option.name }}-{{ forloop.index0 }}"
+                              name="{{ option.name }}"
+                              value="{{ value | escape }}"
+                              form="product-form-{{ section.id }}"
+                              {% if option.selected_value == value %}checked{% endif %}
+                              {% if option_disabled %}class="disabled"{% endif %}
+                        >
+                        <label for="{{ section.id }}-{{ option.name }}-{{ forloop.index0 }}">
+                          {{ value }}
+                        </label>
+                      {%- endfor -%}
+                    </fieldset>
                   {%- endfor -%}
                   <script type="application/json">
                     {{ product.variants | json }}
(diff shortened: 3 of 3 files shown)


tyleralsbury:
Thanks for the contribution, cfxd.

Based on our code principles for Dawn, there are some issues with this Pull Request.  See [the contributing guidelines](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles) for the full document on code principles.  In particular, this PR has an issue with the server rendered principle.  The goal in Dawn, and in theme development going forward, is for Liquid to handle all of the rendering tasks and for complex client-side logic for rendering to be avoided.

This PR brings up an interesting question, though - is there a way for the Shopify platform, through Liquid alone, to help with this sort of thing?

An example would be something along the lines of having something available in the value in `{%- for value in option.values -%}` for the second option onwards that would say whether or not the option should be enabled, based on the `selected_value` of the other options.  Then we could just fetch the variant picker part of the page from the server and render it directly with Liquid handling all the complex logic.

My impression is that right now the platform doesn't exactly provide what would be needed for this to be possible, but it feels like a potential road forward for this functionality to be implemented without needing to do so much heavy lifting within the JS.

cfxd:
Right on, I figured that's what you were going for based on #45 and I just wanted to get this out into the universe somehow. I do actually believe it's possible using liquid and I'm working on it!

tyleralsbury:
> Right on, I figured that's what you were going for based on #45 and I just wanted to get this out into the universe somehow. I do actually believe it's possible using liquid and I'm working on it!

Very interesting!  Feel free to update here as you go.  Thanks, @cfxd .

cfxd:
@tyleralsbury check it out, just forced a push with a refactor in Liquid. Some question marks but at least it's a mvc.

cfxd:
Thanks @gregjotau. I believe the delay you're experiencing is due to the `fetch()` call which is definitely slower than using client side data and logic, but if you're doing this on a non-live theme then I do believe it should speed up a bit when the theme is published live.

gregjotau:
@cfxd @tyleralsbury 

I pushed it live on: https://famme.no/products/long-sleeve?variant=37560099471534

It works good enough for us, but of course, before merging should get some attention and cleanup :)

gregjotau (dismissed):
Works fine on famme.no

cfxd:
@tyleralsbury would love a code review when you have a chance please!

cfxd:
@tyleralsbury (and @gregjotau ) I just merged this with latest changes and squashed the commits. Would love a code review please!

gregjotau:
I have used the code for some time now on famme.no, also up to date with the latest version/commits. So 👍 from me.

cfxd:
Hmm, this is a complex problem. If I do what some propose and simply hide or disable the unavailable options, then in this scenario a shopper would be forever stuck on Blue / Small and they couldn't switch the color or the size.

Even if they could only switch the color option, then you're right—the rendering never occurs because the initial selection is not an existing configuration.

Quite a conundrum indeed.

tyleralsbury:
> Hmm, this is a complex problem. If I do what some propose and simply hide or disable the unavailable options, then in this scenario a shopper would be forever stuck on Blue / Small and they couldn't switch the color or the size.

I agree.  I think crossing them out but still letting them be clickable is the right way to approach this, as you mentioned.

> Even if they could only switch the color option, then you're right—the size re-rendering never occurs because the initial selection is not an existing configuration.
> 
> Quite a conundrum indeed.

It's tricky for sure, given the current reliance on the variant ID.  The fact that these variants don't exist means that they can't have an ID, so we have nothing to base the update on.  This is the main reason that I'm not 100% sure that Liquid provides everything we need to make this work currently.

---

My idea would be that we'd need the platform to have something like "each subsequent option has an awareness of the previous ones".  Then we would be able to use the selected state of the first to determine the available choices for the second, and the combination of those two to determine the third.

This would also mean that we would need to have URL parameters for each of them, so that when we query the URL for the page, we would be able to say which of the choices were selected for each option.  So instead of it relying on the variant to determine which options are supposed to be selected, it could rely on the options themselves.

So, going back to this table....


Option 1 - Colour
Option 2 - Size

|               | Color - Red | Color - Blue |
|---------------|-------------|--------------|
| Size - Small  | Available   | Unavailable  |
| Size - Medium | Unavailable | Available    |

When the page is loaded, "Red / Small" would be selected, so the page would load with Medium crossed out.  This is already possible, but now let's say we changed the colour to "Blue", then there's no associated variant so currently the page has nothing to reference with regards to how to properly update.

With this proposal, we'd have a property on Option 2 (in this case Size) that would have an awareness of the current selection of Option 1 and could be handled.  It would be something like `value.has_associated_previous_options` (naming is hard) which would be a boolean.

There is a big assumption here that I'm not 100% sure about, but it's the fact that the options are actually sequential, or in other words, that the platform even has a concept of "Colour is the first option, Size is the second option".  I don't know if it does.

**Some downsides...**

* It adds complexity.  There's now two sources of truth for what a page should render like - the current variant **and** the currently selected options.  That might create some confusing theme code, since we would always want a valid variant ID to be taken as the primary driver for how a page is rendered if there is one.  Theme code should be clear and easy to read, so this kind of complexity being added to an already complex part of the code might be overwhelming
* Canonical URLs. I don't think this is an issue, since the canonical URL for a product is just the product itself without any URL parameters.  In general URL parameters are bad for SEO, but with canonical URLs I think it's okay.

gregjotau:
@tyleralsbury any update on this? We do not want to have custom code for this as it requires more time when merging new changes in.

gregjotau:
Are there any pending platform changes being worked upon that fixes this? 🙏 Would be great if one checked that, since if no changes is being worked upon, maybe accept some JS soultion.

gregjotau:
Does anyone have a JS solution as a fork btw? (that is bullet proof and addresses the concerns of @tyleralsbury above)

gregjotau:
@tyleralsbury @ludoboludo any update on this, we have a bug related to the sold out UI from this branch that is annoying:

https://github.com/gregjotau/dawn-famme/issues/46

So we *really* need this built in and *solid*. I would argue this is much more important than basically everything else like mega menu (which is also awesome btw!) and small UI fixes.

gregjotau:
@cfxd do you have a JS fix on another branch which is "bullet proof"? Would love to know if you have researched this more 🙏

cfxd:
@gregjotau @mohinht  check 1407

kdubbelaar (commented):
Sold Out variant picker

kdubbelaar:
Yes, you can.


Van: greg taube ***@***.***>
Verzonden: dinsdag 28 juni 2022 06:46
Aan: Shopify/dawn ***@***.***>
CC: Kees Dubbelaar ***@***.***>; Comment ***@***.***>
Onderwerp: Re: [Shopify/dawn] Add variant picker sold out UI (#105)


I guess this one can be closed since 1407 duplicates it?

ludoboludo:
I'll close this PR since we have another one open and being reviewed. 👍 
We can reopen if needed.

PR #128, opened 2021-07-06
Clear rte page content

### LINKED ISSUE #2: Collapsible Tab: Page content is not contained within tab container.
**Problem**
When I added content from a custom page to a Collapsible Tab block it rendered on top of the rest of the page. 

**Expected**
It should create space to render the tab contents when opening the tab, like it does with other content types. 

**Steps to replicate**
To replicate, make a custom page with an image right-aligned. Then link the contents of that page to the Collapsible Tab block in Dawn's product pages. Move the tab above the product description. Open and close the tab. 

image

Video: https://screenshot.click/2021-06-1642-39_vendk-t7tfo_667_16r.mp4

You can see the problem this test store, by opening and closing the second tab on the page: https://macaroni-loaf.myshopify.com/products/lego-mac-and-cheese-not-vegan

kmeleta:

Fixes #2 

Added a clearfix (via `:after`) to the rte class. Right aligned media in the page content rte gets `float: right`. If that content isn't cleared by something else in that page, the content can bleed out of its container and/or cause other layout issues (i.e. collapsible tabs on the product page).

**Other considerations**

The original issue was scoped for page content within collapsible tabs on product pages, but really this same sort of issue could apply anywhere RTE page content can be displayed.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/120841142294/editor)

**Checklist**

### DIFF
--- a/assets/component-rte.css
+++ b/assets/component-rte.css
@@ -1,3 +1,9 @@
+.rte:after {
+  clear: both;
+  content: "";
+  display: block;
+}
+
 .rte > p:first-child {
   margin-top: 0;
 }
--- a/assets/component-rte.css
+++ b/assets/component-rte.css
@@ -1,6 +1,6 @@
 .rte:after {
   clear: both;
-  content: "";
+  content: '';
   display: block;
 }
 
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -248,6 +248,12 @@ a.product__text {
   padding: 0 1rem;
 }
 
+.product__accordion .accordion__content:after {
+  clear: both;
+  content: "";
+  display: block;
+}
+
 .product .price {
   align-items: flex-start;
 }
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -248,12 +248,6 @@ a.product__text {
   padding: 0 1rem;
 }
 
-.product__accordion .accordion__content:after {
-  clear: both;
-  content: "";
-  display: block;
-}
-
 .product .price {
   align-items: flex-start;
 }


assets/section-main-product.css
```diff
@@ -248,6 +248,12 @@ a.product__text {
   padding: 0 1rem;
 }
 
+.product__accordion .accordion__content:after {
+  clear: both;
+  content: "";
+  display: block;
+}
```
ludoboludo, line 255: I think you're right in your description. It would need to be added for the `.rte` class I think. Cause if you go on the home page or anywhere else you can add a section, then add a `page` section and use `page with floated image` then you will get the same issue:

kmeleta: True, that will come up more often then. Moved the same fix to the rte class now. I can't think of a reason why this wouldn't be ok, but maybe there's a rte case out there I'm not thinking of. Curious if anyone thinks there's a safer way to handle this.

ludoboludo: I guess we could have a `rte-page` specific class but I don't think having it to the more general `rte` one should create any problems. It's adding content in places where it might not be needed but not sure if we should avoid that or not. We can check with the second reviewer.

tyleralsbury: I think it should be fine.

tyleralsbury: Also I haven't seen a clearfix in a long time. This brings me back!

tauthomas01 (approved):
Works well. Tested on:

- Product page: page content
- Page section
- Page template

On Chrome/Firefox/Safari, no issues.

I don't think this will break in the future, since we're not really using `float` anymore unless it's RTE or an app is using it, but that's a separate topic.

assets/section-main-product.css
kmeleta: Yeah I'm ok with the content as is. I'll leave it simple as is if you don't have a strong objection Ludo

assets/component-rte.css
```diff
@@ -1,3 +1,9 @@
+.rte:after {
+  clear: both;
+  content: "";
```
chrisberthe, line 3: For consistency with other references:
```suggestion
  content: '';
```

chrisberthe (approved):

PR #134, opened 2021-07-06
Footer image block

### LINKED ISSUE #41: Footer: Image Block
**User Story**
As a merchant, I can add an image file to my footer, so I can display badges (like FDA approved, Certified organic, etc) 

**UI**

image

**Editor Settings**
image

[Figma](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=8722%3A81182)

ludoboludo:

Fixes #41 

Added the new block settings following the figma design. I'm actually wondering what should be the max width we allow. I set it to `200px` for now. 
Also added a new class to deal with the styling. 

**Other considerations**

We talked about vertical alignment in the issue linked but it was decided not to over complicate the block settings. 

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -73,7 +73,7 @@
                     </ul>
                   {%- endif -%}
                 {%- when 'image' -%}
-                  <div class="footer-block__details-content footer-block-image{% if block.settings.alignment != 'top' %} footer-block-image--{{ block.settings.alignment }}{% endif %}">
+                  <div class="footer-block__details-content footer-block-image{% if block.settings.alignment != 'left' %} {{ block.settings.alignment }}{% endif %}">
                     {%- if block.settings.image != blank -%}
                       <img
                         src="{{ block.settings.image | img_url: '400x' }}"
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -82,7 +82,7 @@
                         width="{{ block.settings.image.width }}"
                         height="{{ block.settings.image.height }}"
                         class=""
-                        style="max-width: {{ block.settings.image_width }}px; height: auto;"
+                        style="max-width: min(100%, {{ block.settings.image_width }}px); height: auto;"
                       >
                     {%- else -%}
                       {{ 'image' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
@@ -438,16 +438,16 @@
           "label": "Image alignment on large screen",
           "options": [
             {
-              "value": "top",
-              "label": "Top"
+              "value": "left",
+              "label": "Left"
             },
             {
               "value": "center",
               "label": "Center"
             },
             {
-              "value": "bottom",
-              "label": "Bottom"
+              "value": "right",
+              "label": "Right"
             }
           ],
           "default": "center"
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -72,6 +72,22 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'image' -%}
+                  <div class="footer-block__details-content footer-block-image{% if block.settings.alignment != 'top' %} footer-block-image--{{ block.settings.alignment }}{% endif %}">
+                    {%- if block.settings.image != blank -%}
+                      <img
+                        src="{{ block.settings.image | img_url: '400x' }}"
+                        alt="{{ block.settings.image.alt | escape }}"
+                        loading="lazy"
+                        width="{{ block.settings.image.width }}"
+                        height="{{ block.settings.image.height }}"
+                        class=""
+                        style="max-width: {{ block.settings.image_width }}px; height: auto;"
+                      >
+                    {%- else -%}
+                      {{ 'image' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
+                    {%- endif -%}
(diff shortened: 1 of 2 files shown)


sections/footer.liquid
```diff
@@ -72,6 +72,22 @@
...
+                {%- when 'image' -%}
+                  <div class="footer-block__details-content footer-block-image{% if block.settings.alignment != 'left' %} {{ block.settings.alignment }}{% endif %}">
+                    {%- if block.settings.image != blank -%}
+                      <img
+                        src="{{ block.settings.image | img_url: '400x' }}"
+                        alt="{{ block.settings.image.alt | escape }}"
+                        loading="lazy"
+                        width="{{ block.settings.image.width }}"
+                        height="{{ block.settings.image.height }}"
+                        class=""
```
tauthomas01, line 84: Can we remove this?

sections/footer.liquid
```diff
@@ -72,6 +72,22 @@
...
+                        style="max-width: min(100%, {{ block.settings.image_width }}px); height: auto;"
```
tauthomas01, line 85: I'm trying to understand the usage of `min` here. Do you know when is the `100%` width used? The max width the image can get is 200px and most mobile device resolution have a minimum of 320px so I'm trying to think in which case the 100% width is used.

tauthomas01 (approved):
LGTM, just left one question about the CSS usage.

sections/footer.liquid
KaichenWang: When the image is part of a row with 4 columns, the width of its column might be < 200px

Kapture 2021-07-07 at 10 22 41

ludoboludo: ^ exactly. So images don't end up overlapping on certain breakpoints when the container ends up being less than `200px`.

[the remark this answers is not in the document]
ludoboludo: You're right. Moved it into the `section-footer.css` file.

sections/footer.liquid
```diff
@@ -72,6 +72,22 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'image' -%}
+                  <div class="footer-block__details-content footer-block-image{% if block.settings.alignment != 'left' %} {{ block.settings.alignment }}{% endif %}">
```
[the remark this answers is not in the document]
ludoboludo: Good point

sections/footer.liquid
tauthomas01: Thank you both for the explanation 👍

sections/footer.liquid
```diff
@@ -72,6 +72,22 @@
...
+                    {%- if block.settings.image != blank -%}
+                      <img
```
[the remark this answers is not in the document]
ludoboludo: So the only reason I didn't have it here is because it didn't seem to make a big difference. Since the max-width can be up to `200px` so up to `400px` on high res screens. Not sure if it's worth the logic for a 200px difference in terms of image definition. What do you think ? I could add something like that though if you think it's beneficial: 

```
{%- if block.settings.image.width >= 200 -%}{{ block.settings.image | img_url: '200x' }} 200w,{%- endif -%}
{%- if block.settings.image.width >= 400 -%}{{ block.settings.image | img_url: '400x' }} 400w{%- endif -%}
```

ludoboludo:
> > I'm actually wondering what should be the max width we allow. I set it to 200px for now.
> 
> Looks to me like the container can be up to 252px (when there are 4 columns), which would be kinda weird as a max setting just visually... I'm wondering if instead of pixel values, the setting can represent a percentage, maybe somewhere from 50% to 100%.
> 
> That way, it can have its size depend on the size of the container itself.

Intersting. @melissaperreault I'm not sure who would be the designer in charge of reviewing those footer changes. We're wondering how should the width slider behave. Should it be pixel value, or percent of it's container 🤔  And if we go with pixel value, is 200px an ok limit like I have it right now ? 
[Settings screenshot from figma](https://screenshot.click/21-07-78zkz-u0zu9.png)

sections/footer.liquid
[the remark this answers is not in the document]
KaichenWang: I agree the difference wouldn't be that noticeable, but loading the correct image size would be good as a best practice. I think we could leverage the `block.settings.image_width` setting to set the resolution of the image (similar to header logo):

``` html
{%- assign image_size = block.settings.image_width | append: 'x' -%}
<img srcset="{{ block.settings.image | img_url: image_size }} 1x, {{ block.settings.image | img_url: image_size, scale: 2 }} 2x"
  src="{{ block.settings.image | img_url: image_size }}"
  ...
>
```

tauthomas01 (approved):

melissaperreault:
> Intersting. @melissaperreault I'm not sure who would be the designer in charge of reviewing those footer changes. We're wondering how should the width slider behave. Should it be pixel value, or percent of it's container 🤔 And if we go with pixel value, is 200px an ok limit like I have it right now ?

Just for the record, going with % would be confusing as you wouldn't know if the % applies to the column width for instance.

I suggest to monitor the merchant usage and put trust in how they can design their footer.

The experience is looks good to me!

PR #135, opened 2021-07-07
Update ja.json

Yuya-Maeda:
Fixed a grammatical error and some consistency.

Fixes #0.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/locales/ja.json
+++ b/locales/ja.json
@@ -103,8 +103,8 @@
   "products": {
     "product": {
       "add_to_cart": "カートに追加する",
-      "description": "説明",
-      "on_sale": "Sale",
+      "description": "商品説明",
+      "on_sale": "セール",
       "product_variants": "商品バリエーション",
       "quantity": {
         "label": "数量",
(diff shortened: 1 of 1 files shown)


chrisberthe:
Thanks for the contribution @Yuya-Maeda. The unfortunate news is that our translation system, as its stands, won't allow us to directly edit these files. So if we were to merge your change, it would be overwritten with the previous version.

Your contribution won't go to waste however—we'll ensure we bring your suggestions over to the translation team so we can get this updated on our side.

Yuya-Maeda:
Thanks for the comment. I'm a native speaker of Japanese. When I'm trying out new Dawn theme, I found that some Japanese translation is not natural. It's minor mistakes though.

iCynara:
Thank you @Yuya-Maeda ,  I have passed your editing version to our Japanese language specialist. They will review your suggestions and once it gets approved, we will make notes to our translators and the system will be updated with new version. 

Thank you so much for caring about Language Quality.

PR #140, opened 2021-07-07
Fixed font preconnect to also include a check for a non-system body font

tyleralsbury:

Just a quick fix to ensure we preconnect to the font CDN for non-system body fonts as well as heading fonts.

**Checklist**


KaichenWang (approved):

kmeleta (approved):

chrisberthe:
Shouldn’t this be `or`, not `and`?

KaichenWang:
> Shouldn’t this be `or`, not `and`?

"We need preconnect, unless header is a system font **_and_** body is a system font"

So `and` seems correct here

PR #143, opened 2021-07-07
Add Icelandic locale

baldurh:

Adds Icelandic translation

I updated `en.default.json` line by line

### DIFF
--- a/locales/is.json
+++ b/locales/is.json
@@ -1,0 +1,380 @@
+{
+  "general": {
+    "password_page": {
+      "login_form_heading": "Aðgangur með lykilorði:",
+      "login_password_button": "Opna með lykilorði",
+      "login_form_password_label": "Lykilorð",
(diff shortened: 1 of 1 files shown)


ludoboludo:
Hello @baldurh,

First of all, thank you so much for your effort, and your care for our Icelandic-speaking buyers and merchants. :green_heart:

Unfortunately, we cannot accept your Pull Request.
The cost of maintaining languages is too great and we’re unable to accept pull requests that add languages we cannot fully support internally at this time.

Shopify is always looking for ways to improve the experience for merchants and buyers around the world.

Dawn is provided as a source available repository, partly to allow others to maintain downstream forks that support the full range of commerce. If you’re interested, perhaps maintaining an Icelandic fork would be appropriate.

The work you've done could also be shared on our [forum](https://community.shopify.com/c/Shopify-Community/ct-p/en) once Dawn is launched and feature complete.

PR #151, opened 2021-07-08
Search template design polish

KaichenWang:

Fixes #0.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/locales/fr.schema.json
+++ b/locales/fr.schema.json
@@ -1276,7 +1276,7 @@
         "show_secondary_image": {
           "label": "Afficher la deuxième image en survol"
         },
-        "enable_image_padding": {
+        "add_image_padding": {
           "label": "Ajouter une marge intérieure à l'image"
         },
         "show_vendor": {
--- a/snippets/pagination.liquid
+++ b/snippets/pagination.liquid
@@ -9,14 +9,13 @@
     Accepts:
     - paginate: {Object}
     - anchor: {String} (optional) This can be added so that on page reload it takes you to wherever you've placed your anchor tag.
-    - class: {String} (optional) Appended to container element's class attribute
 {% endcomment %}
 
 <link rel="stylesheet" href="{{ 'component-pagination.css' | asset_url }}" media="print" onload="this.media='all'">
 <noscript>{{ 'component-pagination.css' | asset_url | stylesheet_tag }}</noscript>
 
 {%- if paginate.parts.size > 0 -%}
-  <div class="{{ class | default: 'pagination-wrapper' }}">
+  <div class="pagination-wrapper">
     <nav class="pagination" role="navigation" aria-label="{{ 'general.pagination.label' | t }}">
       <ul class="pagination__list list-unstyled" role="list">
       {%- if paginate.previous -%}
--- a/snippets/product-card.liquid
+++ b/snippets/product-card.liquid
@@ -65,7 +65,7 @@
             </div>
           </div>
         {%- else -%}
-          <div class="card__content"><h2 class="card__text h2">{{ product_card_product.title }}</h2></div>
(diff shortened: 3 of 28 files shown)


snippets/product-card.liquid
```diff
@@ -65,7 +65,7 @@
             </div>
           </div>
         {%- else -%}
-          <div class="card__content"><h2 class="card__text h2">{{ product_card_product.title }}</h2></div>
+          <div class="card__content"><h2 class="card__text h3">{{ product_card_product.title }}</h2></div>
```
tauthomas01, line 68: Should we add `escape` filter as well? I'm seeing HTML tags 

image

snippets/pagination.liquid
```diff
@@ -9,14 +9,13 @@
     Accepts:
     - paginate: {Object}
     - anchor: {String} (optional) This can be added so that on page reload it takes you to wherever you've placed your anchor tag.
-    - class: {String} (optional) Appended to container element's class attribute
```
tauthomas01, line 12: On `main-collection-product-grid.liquid` file, can we remove the `class` parameter as well?

image

tauthomas01 (commented):
I'll do a first pass at the code. Overall looks good

locales/fr.schema.json
```diff
@@ -1277,7 +1277,7 @@
           "label": "Afficher la deuxième image en survol"
         },
         "add_image_padding": {
-          "label": "Ajouter une marge intérieure à l'image"
+          "label": "Ajouter un remplissage d'image"
```
melissaperreault, line 1280: This sounds weird 🤔

`Ajouter de l'espace au pourtour de l'image` would be more accurate. Thoughts? cc. @ludoboludo

ludoboludo: Im not sure if we should be changing those manually though 🤔  but I agree that their translation doesn't convey what the setting is doing. What you mention makes more sense to me as well.

melissaperreault (dismissed):
Hey Kai! Thanks a lot! 🎉 

Some observations:

- [ ] I can't tell why the [border appears darker](https://screenshot.click/21-07-9xc0y-9pq8q.png) on the blog and page cards compared to the product one

Other than that and the translation comment, looks good! 🚢

snippets/pagination.liquid
KaichenWang: Updated

snippets/product-card.liquid
KaichenWang: This is not being rendered by the `product-card` snippet, but rather inside `main-search.liquid`
```
<h2 class="card__text h3">{{ item.title | truncate: 50 | escape }}</h2>
```
The reason the HTML tags, is because the title is being escaped. We actually would need to remove `escape` filter to fix this.

KaichenWang:
>  I can't tell why the [border appears darker](https://screenshot.click/21-07-9xc0y-9pq8q.png) on the blog and page cards compared to the product one

This is due to semi-transparent colors overlapping. Blog and page cards have a background color of  `4% opacity text (foreground) color`. The border also has a color of `4% opacity text color`. Together, the actual perceived color of the border is the sum of the two - `8% opacity text color`

To address this, the border is now only applied to cards which do not also have a background color.

snippets/product-card.liquid
tauthomas01: Sounds good, thanks for sharing context! We can do this separately.

tauthomas01 (approved):
LGTM.

snippets/product-card.liquid
tyleralsbury: I think we're better off outputting the html than it being added to the page as html, otherwise `</div>` in the output would just blow up the layout entirely.

tyleralsbury (approved):
Looked at the code and tested and it looks good to me.

Issue #100, opened 2021-07-05
[Filtering] Fast Follow Fixes + UI Polish


nicklepine:
## Purpose

In order to ensure the feature was merged on time, we merged Faceted filters with some known issues.  The idea now is to prioritize and tackle them one-by-one to try and get as many fixed as possible in the meanwhile and all of them fixed before the merchant launch

### Visual

* [ ] The animation doesn't feel super smooth when the filtering sheet is invoked and then closed. / * Animation is broken when opening/closing the drawer
* [ ]  The checkboxes don't 100% match the style in the Figma file (I'm not sure where else in the theme we encounter checkboxes (customer pages?), but we should make sure the pattern is consistent throughout. The hover style is to be adjusted, and the the label could have its opacity increased to 100%. 

### a11y

* [x] On tablet and mobile, the pills should wear a different smaller style , and comply with the accessible tap target
* [ ] Tabbing can send you behind the drawer
* [x] Can scroll the elements behind the open drawer.  This scrolling should be blocked.

### Browser specific (may not be fixable):

* [ ] On my iPhone, the price filter still allows non numeric values.
* [ ] On my phone, with disabled JS, tapping sorting triggers a layout shift for a slit second.

### Missing feature

* [ ]  Sorting doesn't sync between desktop and mobile
* [ ]  The range slider on the price range

### Refactor

* [ ]  Make a custom element out of the active filters / clear all / reset button to avoid the need to re-bind events to them

Oliviammarcello:
### WIP UI polish review:

- [ ] Align dropdown arrows to the same line as the text. [Screenshot](https://screenshot.click/21-34-6osvv-lwq9t.png)
- [ ] Move checkboxes slightly to the left to appear more aligned on both desktop and mobile. [Desktop screenshot](https://screenshot.click/21-35-a1j4z-xo8i2.png) | [Mobile screenshot](https://screenshot.click/21-38-eut2b-quey5.png)
- [ ] On mobile the "Clear" and "Apply" buttons should fix to the bottom of the screen. [Screenshot](https://screenshot.click/21-38-ak92r-kbz3h.png)
- [ ] The price filtering should have the "$" inside the input field. [Screenshot](https://screenshot.click/21-40-c3xty-3m152.png) | [Figma](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=7155%3A1783)
[...]

PR #153, opened 2021-07-08
Filtering styles and animation improvements

tyleralsbury:

A few more of the points in https://github.com/Shopify/dawn/issues/100 are fixed here

* Animation when opening/closing mobile filters
* Checkbox styles (Excluded hover state)
* Sorting now syncs between desktop and mobile
* Small refactor to remove the need for any logic in the JS for the reset, clear all, and active filter remove buttons

The only thing where the approach changed significantly was the reset, clear all, and active filter buttons.  They are now using a custom element so that when they re-render they automatically get their events bound instead of needing to call a function to loop and bind the events.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=120851431446)

**Checklist**

### DIFF
--- a/assets/collection-filters-form.js
+++ b/assets/collection-filters-form.js
@@ -45,14 +45,14 @@ class CollectionFiltersForm extends HTMLElement {
       const filterDataUrl = element => element.url === url;
 
       this.filterData.some(filterDataUrl) ?
-        this.renderSectionFromCache(filterDataUrl, section, event) :
-        this.renderSectionFromFetch(url, section, event);
+        this.renderSectionFromCache(filterDataUrl, event) :
+        this.renderSectionFromFetch(url, event);
     });
 
     if (updateURLHash) this.updateURLHash(searchParams);
   }
 
-  renderSectionFromFetch(url, section, event) {
+  renderSectionFromFetch(url, event) {
     fetch(url)
       .then(response => response.text())
       .then((responseText) => {
--- a/assets/collection-filters-form.js
+++ b/assets/collection-filters-form.js
@@ -63,7 +63,7 @@ class CollectionFiltersForm extends HTMLElement {
       });
   }
 
-  renderSectionFromCache(filterDataUrl, section, event) {
+  renderSectionFromCache(filterDataUrl, event) {
     const html = this.filterData.find(filterDataUrl).html;
     this.renderFilters(html, event);
     this.renderProductGrid(html);
--- a/assets/collection-filters-form.js
+++ b/assets/collection-filters-form.js
@@ -10,8 +10,6 @@ class CollectionFiltersForm extends HTMLElement {
 
     this.querySelector('form').addEventListener('input', this.debouncedOnSubmit.bind(this));
     window.addEventListener('popstate', this.onHistoryChange.bind(this));
-
-    this.bindActiveFacetButtonEvents();
   }
 
   onSubmitHandler(event) {
@@ -93,7 +91,7 @@ class CollectionFiltersForm extends HTMLElement {
     });
 
     this.renderActiveFacets(parsedHTML);
-    this.renderMobileElements(parsedHTML);
+    this.renderAdditionalElements(parsedHTML);
 
     if (countsToRender) this.renderCounts(countsToRender, event.target.closest('.js-filter'));
   }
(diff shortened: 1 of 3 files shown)


tauthomas01 (approved):
Looks good, desktop and mobile.

I left one small question

assets/collection-filters-form.js
```diff
@@ -190,3 +181,16 @@ class PriceRange extends HTMLElement {
 }
 
 customElements.define('price-range', PriceRange);
+
+class FacetRemove extends HTMLElement {
+  constructor() {
+    super();
+    this.querySelector('a').addEventListener('click', (event) => {
```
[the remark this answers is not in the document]
tyleralsbury: Good point - it might be a good idea to disable that link when there are no active filters for that category.  I'll make a note of it and fix it in another PR.

sections/main-collection-product-grid.liquid
```diff
@@ -367,29 +393,35 @@
...
                 {% render 'icon-close-small' %}
               </span>
             </a>
+          </facet-remove>
+        {%- endfor -%}
+
+        {% if filter.type == "price_range" %}
+          {%- if filter.min_value.value != nil or filter.max_value.value != nil -%}
+            <facet-remove>
+              <a class="active-facets__button active-facets__button--light button button--tertiary" href="{{ filter.url_to_remove }}">
```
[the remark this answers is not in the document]
tyleralsbury: Absolutely correct. I missed that one.

sections/main-collection-product-grid.liquid
```diff
@@ -58,8 +62,20 @@
                               <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
```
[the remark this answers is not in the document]
ludoboludo: yeah, the fact that it works in Safari for mac but not iOS is... weird.

assets/template-collection.css
```diff
@@ -227,6 +227,10 @@
   color: var(--color-foreground-75);
 }
 
+.disclosure-has-popup[open] > .facets__summary::before {
+  z-index: 3;
+}
```
ludoboludo, line 232: I wonder if we should handle the `click outside` to close the menus with JS. And keep this CSS approach for a no JS experience. 
The drawer menu doesn't seem to have a `click outside closes the menu` kinda thing and when using the `facets__summary` it could reduce the amount of clicks to open others. 
But might be overkill 🤷  Just think it would like a smoother experience.

assets/template-collection.css
```diff
@@ -283,7 +287,7 @@
   width: 1.6rem;
   height: 1.6rem;
   top: 0.7rem;
-  left: 0.8rem;
+  left: 1rem;
```
ludoboludo, line 290: It also needs to be updated for `input.mobile-facets__checkbox`. When on my iPad, it looks fine on a large screen now but when it goes to the drawer I get the same thing. It's not ideal to tweak it like this as now the checkbox receiving focus isn't quite aligned with the svg but 🤷 

So from what I was testing it looks like left `4.31rem` is what worked to hide the default checkboxes behind the svg on mobile for iOS.

ludoboludo (dismissed):
LGTM, just left one comment that can be tweaked before merging.

assets/template-collection.css
```diff
@@ -576,13 +582,21 @@ details[open] .mobile-facets__open {
   width: 4.4rem;
   height: 4.4rem;
   z-index: 101;
+  opacity: 0;
+  transition: opacity var(--duration-short) ease;
 }
 
-details[open] .mobile-facets__close--no-js {
+.mobile-facets__close svg {
+  margin-left: 1rem;
```
tauthomas01, line 590: Is this used? I couldn't find it because it's been overwritten by `margin: 0`

image

tauthomas01 (dismissed):

assets/template-collection.css
tyleralsbury: I think this might be too much complexity on the JS side, personally.  There's also more risk of a "time-warp" sort of effect if we do it this way.  Remember that when filters are changed with JS enabled, the other details elements get re-rendered.  If we were to re-render an element after they've clicked the element to open it, then it will close when it re-renders.

We might be able to get around that concern by re-rendering the contents of the details element instead of the whole element, which might be one option. 🤔 

We can add it to the issue as an idea and think about it.

assets/template-collection.css
tyleralsbury: You're right - removing.

assets/template-collection.css
ludoboludo: Sounds good 👍

ludoboludo (approved):
🚀

PR #158, opened 2021-07-08
Hide variant images setting

### LINKED ISSUE #30: Product Page: New Setting: Show only images for currently selected variants
## Purpose

On the product page, we should only show the images in the gallery that correspond to the currently selected variant.

## Approach

Since the way the variant image is currently being put to the forefront is by putting it to the front of the list of gallery images, adding a modifier class on it like `--variant-image` when it's one of the variant featured images and then using CSS like the following:

```
--variant-image {
  display:none;
}

--variant-image:first-child {
  display: block;
}
```

... might be a viable way to handle this such that it works without the need to add any JS logic.

Offer this as a setting: the setting would be what sets the `class` on the variant images.  If it's disabled, the class just wouldn't be there and all the images would be visible.

Problem explained here: https://screenshot.click/21-05-e113w-2ictb.mp4

### LINKED ISSUE #191: Variant image switching

When changing the variant on the product page, it updates the image (if the variant has an image associated to it) and prepends it to the `<ul>`. 

The checks being done in `updateMedia()` in the `variant.js` file doesn't check if the current media that is first in the list is already that same media. So for example, I was trying to add something to scroll back up to the top of the list when the media changes. The problem is that it gets triggered even when the image is actually not changing.

I think `updateMedia()` should only run if the media changes. If the media for `White 32` is the same for `White 36` we should run the function. 

- Dawn Version: 1.1.0
- macOS Version 11.3.1

**Possible solution**

Compare `this.currentVariant?.featured_media` with the first item in the `parent`'s first element.

[Video]([video](https://screenshot.click/21-07-oy1j1-fl6fb.mp4))

ludoboludo:

Fixes #30 
Fixes #191 

I used CSS to hide variant images as mentioned in the issue. I had to also add some JS to deal with it on mobile where I needed to tweak how the total of slides was calculated. 
So I'm creating a new array that regroup all `li` elements that have a `clientWidth > 0`. This is my way for not including hidden elements (using `display: none;`) within the array we use to calculate our `totalPages`.

**Other considerations**

One thing I'm noticing as I was playing with it, is that when you change the variant, it doesn't scroll you back up on the new image. I see in the JS: `window.setTimeout(() => { parent.scroll(0, 0) });` but it doesn't seem to do anything. [Video](https://screenshot.click/21-07-uopfw-e136z.mp4).

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=120951734294)
- [Editor](https://os2-demo.myshopify.com/admin/themes/120951734294/editor)

**Testing**

Go to the Editor linked above, to the product page (Shoes > Louise Slide Sandal) and click on the `Product information` heading and check the `Hide inactive variants` box.  

**Checklist**


assets/slider.js
```diff
@@ -19,15 +20,15 @@ class SliderComponent extends HTMLElement {
   }
 
   initPages() {
-    if (!this.sliderItems.length === 0) return;
-    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItems[0].clientWidth);
-    this.totalPages = this.sliderItems.length - this.slidesPerPage + 1;
+    if (!this.sliderItemsToShow.length === 0) return;
```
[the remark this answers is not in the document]
ludoboludo: Good point, I didn't touch it to keep it the way it had been but it makes sense to remove the `!` as it was not returning when needed 👌

sections/main-product.liquid
```diff
@@ -724,6 +725,12 @@
       "default": true,
       "label": "t:sections.main-product.settings.enable_sticky_info.label"
     },
+    {
+      "type": "checkbox",
+      "id": "hide_variants",
+      "default": false,
+      "label": "Hide inactive variants"
```
melissaperreault, line 732:
```suggestion
      "default": true,
      "label": "Show variant media"
```

@gretchen-willenberg we need to add a setting in the Product template because we want to give the ability to only show the images in the gallery that correspond to the currently selected variant.

In Narrative, we have: `Hide variant media in gallery`

I am wondering if we should give the positive form of the setting? By default, merchants would want to show all their images so this setting would be true by default if this is the direction we take.

gretchen-willenberg: In deciding if it should be positive (Show) or negative (Hide), I think most of our checkboxes are opt-in/unchecked. (?) So...if merchants by default want to show all their images, we want to give them the choice to show only the selected variant's.  I think that can still use the positive Show and narrow down what is being shown. [All in the Google doc, too](https://docs.google.com/document/d/14Aa-twJrAWSo95v7RwTF0M2Pk5xJE3t10GJfmcesWBU/edit?usp=sharing)

`Show media for only the selected variant`
`Show only the selected variant’s image(s)`
`When a variant is selected, show only its media`  _I like how this explains the action._

Tried a few similar versions with `Hide` just to see:
`Hide media of unselected variants`
`Hide other variants’ media when a variant is selected`

ludoboludo: The use of `only` worries me a bit. As when checked, the setting will show not only the selected variant's image but the other medias that aren't linked to variants as well. 
I find `Hide media of unselected variants` to be pretty clear. Just my two cents 😅

assets/slider.js
```diff
@@ -47,7 +48,7 @@ class SliderComponent extends HTMLElement {
 
   onButtonClick(event) {
     event.preventDefault();
-    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderItems[0].clientWidth : this.slider.scrollLeft - this.sliderItems[0].clientWidth;
+    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderItemsToShow[this.sliderItemsToShow.length - 1].clientWidth : this.slider.scrollLeft - this.sliderItemsToShow[this.sliderItemsToShow.length - 1].clientWidth;
```
kmeleta, line 51: Small nitpick. I see the last item being referenced in 3 places. Can we alias this in the constructor so this line is more readable? Something like `this.lastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1]`

kmeleta (commented):
Still testing, but thought I'd leave a couple quick notes.

> One thing I'm noticing as I was playing with it, is that when you change the variant, it doesn't scroll you back up on the new image. I see in the JS: window.setTimeout(() => { parent.scroll(0, 0) }); but it doesn't seem to do anything.

Seems like this line will successfully reset the horizontal scrolling on mobile when the slider is displayed. But on desktop, the scrollable element isn't the parent element of the media, it's the document/window.

assets/slider.js
ludoboludo: The reason is because when switching variants, it will put the feature image of that variant first in line. Because it has a `display: none;` before it send it to the first place, the JS sometimes tries to calculate the width of the element that is yet to be shown. I think it happens in terms of milliseconds and doing the last seemed safer in that sense.

melissaperreault (dismissed):
> One thing I'm noticing as I was playing with it, is that when you change the variant, it doesn't scroll you back up on the new image.

From a UX perspective, it would be expected to scroll back up when the buyer interacts with the variant. If they have the interest to explore the variant list and that merchants assigned a media to it, it's relevant in the purchase flow to have focus on the right visual.

Other than that, looks good to me!  🎉 🎉

kmeleta (approved):
This looks good Ludo. Changing sizes now doesn’t scroll the featured image into view but color variant changes do. Good work getting that included.

On mobile though I notice a couple inconsistencies that I'll describe here, but I'm fine investigating them as a separate issue. They seem to be related to how the slider interacts with the variant picking in general.

- Changing the variants still often doesn’t scroll to the top. This seems to mostly happen when the hide variant setting is disabled OR when it is enabled but you scrolled the slider to another image. https://screenshot.click/15-08-bzap0-dhhkb.mp4 

- The slider counter also counts down as the scroll position is reset to 0 upon changing the variant. This isn't a big deal but it'd be nice to try and fix. https://screenshot.click/15-12-v284p-uh7qr.mp4 

- I have also seen the slider lose track of it's position, which is kind of weird and hard to reproduce reliably https://screenshot.click/15-43-thbkj-8acf5.mp4

assets/slider.js
```diff
@@ -19,15 +19,17 @@ class SliderComponent extends HTMLElement {
...
+    if (this.sliderItemsToShow.length === 0) return;
+    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItemsToShow[0].clientWidth);
+    this.totalPages = this.sliderItemsToShow.length - this.slidesPerPage + 1;
     this.update();
   }
 
   update() {
     if (!this.pageCount || !this.pageTotal) return;
-    this.currentPage = Math.round(this.slider.scrollLeft / this.sliderItems[0].clientWidth) + 1;
+    this.currentPage = Math.round(this.slider.scrollLeft / this.sliderLastItem.clientWidth) + 1;
```
LucasLacerdaUX, line 32: https://user-images.githubusercontent.com/37168033/125852029-e550ddb9-54ea-410f-bf52-4c6bec17704b.mp4

I find it weird how the numbers jump from 3 to 1 every time I click the slider buttons. Not sure if there's anything you can do about this, though.

sections/main-product.liquid
```diff
@@ -24,15 +24,16 @@
...
+            <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if variant_images contains media.src and section.settings.hide_variants %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
               {% render 'product-thumbnail', media: media, position: 'featured', loop: section.settings.enable_video_looping, modal_id: section.id %}
             </li>
           {%- endif -%}
           {%- for media in product.media -%}
             {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
-              <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
+              <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if variant_images contains media.src and section.settings.hide_variants %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
```
LucasLacerdaUX, line 36: Is it the expectation that these images are visible INSIDE the product modal? I find the setting naming a little confusing if that's the case - I would expect it to hide everywhere.

https://user-images.githubusercontent.com/37168033/125852374-a46b13c8-c046-4005-9f77-240ae8c52756.mp4

Right now I can see all the variant images when I open the modal.

sections/main-product.liquid
```diff
@@ -24,15 +24,16 @@
           {{ "accessibility.skip_to_product_info" | t }}
         </a>
         <ul class="product__media-list grid grid--peek list-unstyled slider slider--mobile" role="list">
+          {%- assign variant_images = product.images | where: 'attached_to_variant?', true | map: 'src' -%}
           {%- if product.selected_or_first_available_variant.featured_media != null -%}
             {%- assign media = product.selected_or_first_available_variant.featured_media -%}
-            <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
+            <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if variant_images contains media.src and section.settings.hide_variants %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
```
LucasLacerdaUX, line 30: Nitpicky: would it make sense to switch the order of these conditions (for readability purposes only)?

I feel like the "first requirement" for this logic is to check if the setting is enabled, so I expected to see that condition first.

```suggestion
            <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
```

sections/main-product.liquid
LucasLacerdaUX: Same here.

locales/en.default.schema.json
```diff
@@ -1273,6 +1273,9 @@
         "enable_sticky_info": {
           "label": "Enable sticky product information on large screens"
         },
+        "hide_variants": {
+          "label": "Hide other variants’ media when a variant is selected"
```
LucasLacerdaUX, line 1277: If the expectation is to have all variant media visible inside the modal, does this setting naming still makes sense? My expectation is that turning this on will hide other variants' media EVERYWHERE.

LucasLacerdaUX (commented):
Tested and it works as expected. I have a question about the expectation for modals, though.

sections/main-product.liquid
ludoboludo: I wonder 🤔 @melissaperreault how do you think it should work in the case mentioned above ^ Should we also hide them in the modal ?

assets/slider.js
kmeleta: Yes, sorry for the delay. I documented this as part of #242

sections/main-product.liquid
melissaperreault: The expectation is to hide everywhere when the setting is true, I believe it was the case when I tested if I recall. Otherwise it was overlooked, when this setting is enabled, it applies to all devices! 🙌

ludoboludo: Made some changes in this commit to deal with it: https://github.com/Shopify/dawn/pull/158/commits/263b8919929565718b5df33359138911e6b5b831

assets/global.js
```diff
@@ -533,10 +541,17 @@ class VariantSelects extends HTMLElement {
...
+    const newMediaModal = modalContent.querySelector( `[data-media-id="${this.currentVariant.featured_media.id}"]`);
+
     if (!newMedia) return;
     const parent = newMedia.parentElement;
+    if (parent.firstChild == newMedia) return;
+    modalContent.prepend(newMediaModal);
     parent.prepend(newMedia);
-    window.setTimeout(() => { parent.scroll(0, 0) });
+    parent.dispatchEvent(this.mediaChangeEvent);
+    window.setTimeout(() => { parent.scrollLeft = 0; parent.scrollIntoView({behavior: "smooth", inline: "start"}); });
```
kmeleta, line 553: The issue with it not scrolling vertically and horizontally does indeed appear to be a chrome bug, specifically with the 'smooth' behavior. A normal 'auto' behavior doesn't appear to have the issue, though it's not as slick as the smooth scrolling effect.

I did play with this a bit though and it seems like potentially targeting the first product media item (li) instead of the parent (ul) might allow both scrolls [to work](https://screenshot.click/09-56-mybot-cxnqn.mp4)

I used `document.querySelector('li.product__media-item').scrollIntoView({behavior: "smooth"});`, basically re-querying for the first item each time the media changes. Maybe you found an issue with this approach already, but just in case I wanted to mention that I had some success with it.

assets/global.js
```diff
@@ -533,10 +541,17 @@ class VariantSelects extends HTMLElement {
...
+    window.setTimeout(() => { this.header.productMediaIsChanging = false;}, 800);
```
kmeleta, line 554: The 800ms does seem reliable enough, but obviously it's a little fragile because it’s at the mercy of each browser's scrollintoview speed, which we can’t control.

If it bothers anyone else, we could consider removing this bit and using a "scrolling is finished" timeout check in the existing onScroll method in header.liquid and reset the `productMediaIsChanging` flag there. But otherwise I'm ok with this for now.

assets/global.js
```diff
@@ -496,7 +502,9 @@ customElements.define('slider-component', SliderComponent);
 class VariantSelects extends HTMLElement {
   constructor() {
     super();
+    this.header = document.querySelector('sticky-header');
     this.addEventListener('change', this.onVariantChange);
+    this.mediaChangeEvent = new Event('mediaChange', {"bubbles": true});
```
kmeleta, line 507: Nitpick, single quotes

sections/header.liquid
```diff
@@ -437,15 +437,21 @@
...
+      this.productMediaScroll = this.productMediaChanged.bind(this);
 
+      this.productSection?.addEventListener('mediaChange', this.productMediaScroll)
       window.addEventListener('scroll', this.onScrollHandler, false);
 
       this.createObserver();
     }
 
     disconnectedCallback() {
+      this.productSection?.removeEventListener('mediaChange', this.productMediaScroll)
```
kmeleta, line 454: Nitpick, semis

kmeleta (commented):
Wanted to share a couple thoughts on potential improvements, but I think it'd be nice to get this work merged despite some little scroll quirks.

assets/global.js
ludoboludo: Hey @kmeleta, i've updated it here, let me know if you see anything I should/could do differently: https://github.com/Shopify/dawn/pull/158/commits/224dced7e7aae5fdee1d12a156ce2e2517e21ab0

kmeleta (approved):
Thanks for making those changes! I think it's working pretty smoothly now.

LucasLacerdaUX (changes requested):
I believe the JS needs some refactoring. Let me know if you have any questions or if there is something I can add to make it clearer.

Overall, I feel like this interaction between `VariantSelects` and `StickyHeader` can be improved by removing the `.product-section` interface between the two and concentrating the scroll prevention logic in the `StickyHeader` component.

locales/en.default.schema.json
melissaperreault:
```suggestion
          "label": "Hide other variants’ media after selecting a variant"
```

Hey @gretchen-willenberg the reason why we didn't see yet this line of text is because the PR was not merged yet. I suggested the change based on your audit!

assets/global.js
```diff
@@ -533,10 +539,16 @@ class VariantSelects extends HTMLElement {
     const newMedia = document.querySelector(
       `[data-media-id="${this.dataset.section}-${this.currentVariant.featured_media.id}"]`
     );
+    const modalContent = document.querySelector(`#ProductModal-${this.dataset.section}`);
+    const newMediaModal = modalContent.querySelector( `[data-media-id="${this.currentVariant.featured_media.id}"]`);
```
LucasLacerdaUX, line 543: **nitpick:** Could these queries be moved under `if (parent.firstChild == newMedia) return;`?

assets/global.js
```diff
@@ -456,15 +456,17 @@ class SliderComponent extends HTMLElement {
   }
 
   initPages() {
-    if (!this.sliderItems.length === 0) return;
-    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItems[0].clientWidth);
-    this.totalPages = this.sliderItems.length - this.slidesPerPage + 1;
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
```
LucasLacerdaUX, line 459: I think this should be a `const`. This value is not used by any other function in this class.

```suggestion
    const sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
```

sections/header.liquid
```diff
@@ -437,15 +437,21 @@
...
       this.headerBounds = {};
       this.currentScrollTop = 0;
+      this.preventReveal = false;
+
+      window.preventHeaderReveal = new Event('preventHeaderReveal', {'bubbles': true});
 
       this.onScrollHandler = this.onScroll.bind(this);
+      this.hideHeaderOnScrollUp = this.setPreventReveal.bind(this);
 
+      document.addEventListener('preventHeaderReveal', this.hideHeaderOnScrollUp);
```
LucasLacerdaUX, line 447: I'm not so sure about this `preventHeaderReveal` listener on a `document` level + use of bubbling, because that means the event is being propagated on multiple levels...

Sure, from my limited knowledge, it seems like the performance impact is probably insignificant. But from a code design perspective, I feel like this event should be dispatched directly to the `sticky-header` from `VariantSelects` so this `StickyHeader` JS can easily add the eventListeners to itself.

Thoughts?

sections/header.liquid
```diff
@@ -458,18 +464,33 @@
       observer.observe(this.header);
     }
 
+    setPreventReveal() {
+      this.preventReveal = true;
+    }
```
LucasLacerdaUX, line 469: Does this need to be a function? I believe all you'd need is the `eventHandler` as a named function, not the function it calls.

So Line 445 would become:
```css
this.hideHeaderOnScrollUp = () => this.preventReveal = true;
```

sections/header.liquid
```diff
@@ -458,18 +464,33 @@
...
+
     onScroll() {
       const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
 
-      if (scrollTop > this.currentScrollTop && scrollTop > this.headerBounds.bottom) {
+      window.clearTimeout(this.isScrolling);
+
+      this.isScrolling = setTimeout(() => {
+        this.preventReveal = false;
+      }, 66);
```
LucasLacerdaUX, line 478: Is this needed when `if (this.preventReveal == false)`? If not, we can include it inside `if (this.preventReveal)` below.


Also, where is this `66` timeout coming from? 👀

sections/header.liquid
```diff
@@ -458,18 +464,33 @@
...
+
+      if (this.preventReveal) {
         requestAnimationFrame(this.hide.bind(this));
-      } else if (scrollTop < this.currentScrollTop && scrollTop > this.headerBounds.bottom) {
-        requestAnimationFrame(this.reveal.bind(this));
-      } else if (scrollTop <= this.headerBounds.top) {
-        requestAnimationFrame(this.reset.bind(this));
-      }
+        this.currentScrollTop = scrollTop;
```
LucasLacerdaUX, line 482: This line is both on the `if` and on the `else` blocks. Does it need to happen AFTER the `requestAnimationFrame`? If so, ignore this suggestion.

If not, move it outside the `if`/`else` block.

Also, **nitpick:** an early return could probably help with legibility.

LucasLacerdaUX (commented):
Mostly nitpicks, but there are some code-design related questions / discussions that I'd like to hear your thoughts.

sections/header.liquid
```diff
@@ -437,15 +437,21 @@
...
+      this.productMediaScroll = this.productMediaChanged.bind(this);
 
+      this.productSection?.addEventListener('mediaChange', this.productMediaScroll);
       window.addEventListener('scroll', this.onScrollHandler, false);
 
       this.createObserver();
     }
 
     disconnectedCallback() {
+      this.productSection?.removeEventListener('mediaChange', this.productMediaScroll);
```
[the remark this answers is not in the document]
LucasLacerdaUX: I added some comments below. I'm not fully convinced that `window.preventHeaderReveal` is needed, though.

Same for the element that receives the event. Not sure if dispatching it to parent and bubbling it to document is the best way to go, but I'd like to hear your thoughts on that on the comment below.

sections/header.liquid
ludoboludo: Done 👍  As for the `66` it was just something I used because of [the article reference](https://gomakethings.com/detecting-when-a-visitor-has-stopped-scrolling-with-vanilla-javascript/) I was using.

sections/header.liquid
ludoboludo: Ah good point 👍

assets/global.js
ludoboludo: 👌  done

assets/global.js
```diff
@@ -533,10 +539,16 @@ class VariantSelects extends HTMLElement {
...
+    const modalContent = document.querySelector(`#ProductModal-${this.dataset.section}`);
+    const newMediaModal = modalContent.querySelector( `[data-media-id="${this.currentVariant.featured_media.id}"]`);
+
     if (!newMedia) return;
     const parent = newMedia.parentElement;
+    if (parent.firstChild == newMedia) return;
+    modalContent.prepend(newMediaModal);
     parent.prepend(newMedia);
-    window.setTimeout(() => { parent.scroll(0, 0) });
+    parent.dispatchEvent(window.preventHeaderReveal);
```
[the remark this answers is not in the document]
ludoboludo: No, I just wanted to make sure it would be dispatch and could be grabbed from the sticky header. 
My idea was that you could reuse `dispatchEvent(window.preventHeaderReveal);` anywhere else it might be needed in the theme (if it's ever something that we need to call elsewhere) rather than declare new events every time. But if you say it's okay to do so, then I can change it 👌

assets/global.js
```diff
@@ -484,7 +486,11 @@ class SliderComponent extends HTMLElement {
...
     event.preventDefault();
-    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderItems[0].clientWidth : this.slider.scrollLeft - this.sliderItems[0].clientWidth;
+    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderLastItem.clientWidth : this.slider.scrollLeft - this.sliderLastItem.clientWidth;
+    this.slider.scrollTo({
+      left: slideScrollPosition
+    });
+
     this.slider.scrollTo({
       left: slideScrollPosition
     });
```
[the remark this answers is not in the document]
ludoboludo: Good spotting! thank you 👌

kmeleta (dismissed):
:shipit:

assets/global.js
```diff
@@ -533,10 +535,17 @@ class VariantSelects extends HTMLElement {
     const newMedia = document.querySelector(
       `[data-media-id="${this.dataset.section}-${this.currentVariant.featured_media.id}"]`
     );
+
     if (!newMedia) return;
+    const modalContent = document.querySelector(`#ProductModal-${this.dataset.section}`);
+    const newMediaModal = modalContent.querySelector( `[data-media-id="${this.currentVariant.featured_media.id}"]`);
     const parent = newMedia.parentElement;
+    if (parent.firstChild == newMedia) return;
+    modalContent.prepend(newMediaModal);
```
LucasLacerdaUX, line 544: I just realized that this `modalContent` element is not the right one.

`#ProductModal-${this.dataset.section}` points to the external modal, not its content wrapper.

image

This is causing the element to be prepended to the wrong container:
image

image

LucasLacerdaUX (changes requested):

assets/global.js
ludoboludo: 🎯

LucasLacerdaUX (approved):
image

kmeleta (approved):

Issue #49, opened 2021-06-25
New section: Video


nicklepine:
**User story**
As a merchant, I can add a video section, so I can convey messaging around my business/products in video format

**Context**
Merchants will have the ability to surface a video by adding a collage section, with a sole video block. However, this flow is not the most discoverable. We're contemplating the idea of adding a video section so merchants can more easily find and add this type of section to their online store.

tyleralsbury:
Note for whoever works on this.  We need to use the deferred media pattern from the product gallery and collage sections to ensure that any assets loaded as a result of adding a video are loaded only when requested by the user, preventing a ton of loading up front.

PaulNewton:
A thing to note is the underlying issues here exists upstream that will lead to similar section with pared down feature requirements.
 Mainly the theme designer in shopify does not have conditional settings or conditional presets in the schema that would let developers simply reuse [/sections/collage.liquid](https://github.com/Shopify/dawn/blob/main/sections/collage.liquid) with a **limited** group of settings effectively making it a section for use with a single piece of media through retitling and block restrictions.
If that ever changes sections like this could be consolidated.

For example in the below fakeup image all I've done is change "Collage" title to "Video and Media" and it's default blocks to a single video block. 
![image](https://user-images.githubusercontent.com/47228209/125731372-0f955e7a-f9ca-4cb3-848f-70d87793c0a0.png)

PR #178, opened 2021-07-13
Video section

ludoboludo:

Adds the Video section.

Figma reference: [video section](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=10789%3A83361)

Reused a most of the code from the Collage section. And added some new styling specific to this section.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/sections/video.liquid
+++ b/sections/video.liquid
@@ -1,0 +1,136 @@
+{{ 'video-section.css' | asset_url | stylesheet_tag }}
+
+<link rel="stylesheet" href="{{ 'component-deferred-media.css' | asset_url }}" media="print" onload="this.media='all'">
+
+<div class="video-section{% unless section.settings.full_width %} page-width{% endunless %}">
+  <h2 class="video-section__title title">{{ section.settings.heading }}</h2>
+  <div class="video-section__wrapper{% if section.settings.image_padding %} video-section__media--padding{% endif %}">
+    <noscript>
+      <a href="{{ section.settings.video_url }}" class="card{% if section.settings.cover_image == blank %} video-section__placeholder{% endif %}"
+        {% if section.settings.cover_image != blank %}
+          style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
+        {% endif %}
+      >
+        <div class="video-section__media media media--transparent">
+          {%- if section.settings.cover_image != blank -%}
+            <img
+              srcset="{%- if section.settings.cover_image.width >= 500 -%}{{ section.settings.cover_image | img_url: '500x' }} 500w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 730 -%}{{ section.settings.cover_image | img_url: '730x' }} 730w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 1440 -%}{{ section.settings.cover_image | img_url: '1440x' }} 1440w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 990 -%}{{ section.settings.cover_image | img_url: '990x' }} 990w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 2080 -%}{{ section.settings.cover_image | img_url: '2080x' }} 2080w{%- endif -%}"
+              src="{{ section.settings.cover_image | img_url: '533x' }}"
+              sizes="(min-width: 990px)990px, (min-width: 750px) calc(100vw - 100px), calc(100vw - 30px)"
+              alt="{{ section.settings.description | escape }}"
+              loading="lazy"
+              width="{{ section.settings.cover_image.width }}"
+              height="{{ section.settings.cover_image.height }}"
+              class="video-section__image"
+            >
+          {%- else -%}
+            {{ 'collection-2' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
+          {%- endif -%}
+        </div>
+      </a>
+    </noscript>
+    <deferred-media class="video-section__media deferred-media no-js-hidden" data-media-id="{{ section.settings.video_url.id }}"
+      {% if section.settings.cover_image != blank %}
+        style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
+      {% else %}
+        style="padding-bottom: 56.25%;"
+      {% endif %}
+    >
+      <button
+        id="Deferred-Poster-Modal-{{ section.settings.video_url.id }}"
+        class="video-section__poster media deferred-media__poster"
+        type="button"
+      >
+        {%- if section.settings.cover_image != blank -%}
+          <img
+            srcset="{%- if section.settings.cover_image.width >= 500 -%}{{ section.settings.cover_image | img_url: '500x' }} 500w,{%- endif -%}
+            {%- if section.settings.cover_image.width >= 730 -%}{{ section.settings.cover_image | img_url: '730x' }} 730w,{%- endif -%}
+            {%- if section.settings.cover_image.width >= 1440 -%}{{ section.settings.cover_image | img_url: '1440x' }} 1440w,{%- endif -%}
+            {%- if section.settings.cover_image.width >= 990 -%}{{ section.settings.cover_image | img_url: '990x' }} 990w,{%- endif -%}
+            {%- if section.settings.cover_image.width >= 2080 -%}{{ section.settings.cover_image | img_url: '2080x' }} 2080w{%- endif -%}"
+            src="{{ section.settings.cover_image | img_url: '533x' }}"
+            sizes="(min-width: 990px)990px, (min-width: 750px)calc(100vw - 100px), calc(100vw - 30px)"
+            alt="{{ section.settings.description | escape }}"
+            loading="lazy"
+            width="{{ section.settings.cover_image.width }}"
+            height="{{ section.settings.cover_image.height }}"
+            class="video-section__image"
+          >
+        {%- else -%}
+          {{ 'collection-2' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
+        {%- endif -%}
+        <span class="deferred-media__poster-button motion-reduce">
+          {%- render 'icon-play' -%}
+        </span>
+      </button>
+      <template>
+        {%- if section.settings.video_url.type == 'youtube' -%}
+          <iframe src="https://www.youtube.com/embed/{{ section.settings.video_url.id }}?enablejsapi=1" class="js-youtube" allow="autoplay; encrypted-media" allowfullscreen title="{{ section.settings.description | escape }}"></iframe>
+        {%- else -%}
+          <iframe src="https://player.vimeo.com/video/{{ section.settings.video_url.id }}" class="js-vimeo" allow="autoplay; encrypted-media" allowfullscreen title="{{ section.settings.description | escape }}"></iframe>
+        {%- endif -%}
+      </template>
+    </deferred-media>
+  </div>
+</div>
+
+{% schema %}
+{
+  "name": "Video",
+  "tag": "section",
+  "class": "spaced-section spaced-section--full-width",
+  "settings": [
+    {
+      "type": "text",
+      "id": "heading",
+      "default": "Video",
+      "label": "Heading"
+    },
+    {
+      "type": "image_picker",
+      "id": "cover_image",
+      "label": "Cover image"
+    },
+    {
+      "type": "video_url",
+      "id": "video_url",
+      "accept": [
+        "youtube",
+        "vimeo"
+      ],
+      "default": "https:\/\/www.youtube.com\/watch?v=_9VUPq3SxOc",
+      "label": "Link",
+      "placeholder": "Use a YouTube or Vimeo link",
+      "info": "Video will play inside the page"
+    },
+    {
+      "type": "text",
+      "id": "description",
+      "default": "Describe the video for customers using screen readers",
+      "label": "Video alt text"
+    },
+    {
+      "type": "checkbox",
+      "id": "image_padding",
+      "label": "Add image padding",
+      "default": false,
+      "info": "Select image padding if you don't want your images to be cropped"
+    },
+    {
+      "type": "checkbox",
+      "id": "full_width",
+      "label": "Make section full width",
+      "default": false
+    }
+  ],
+  "presets": [
+    {
+      "name": "Video"
+    }
+  ]
+}
+  {% endschema %}
--- a/sections/video.liquid
+++ b/sections/video.liquid
@@ -1,17 +1,22 @@
 {{ 'video-section.css' | asset_url | stylesheet_tag }}
-
-<link rel="stylesheet" href="{{ 'component-deferred-media.css' | asset_url }}" media="print" onload="this.media='all'">
+{{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
 
 <div class="video-section{% unless section.settings.full_width %} page-width{% endunless %}">
   <h2 class="video-section__title title">{{ section.settings.heading }}</h2>
   <div class="video-section__wrapper{% if section.settings.image_padding %} video-section__media--padding{% endif %}">
     <noscript>
-      <a href="{{ section.settings.video_url }}" class="card{% if section.settings.cover_image == blank %} video-section__placeholder{% endif %}"
+      <div class="video-section__media"
         {% if section.settings.cover_image != blank %}
           style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
+        {% else %}
+          style="padding-bottom: 56.25%;"
         {% endif %}
       >
-        <div class="video-section__media media media--transparent">
+        <a href="{{ section.settings.video_url }}" class="video-section__poster media media--transparent{% if section.settings.cover_image == blank %} video-section__placeholder{% endif %}"
+          {% if section.settings.cover_image != blank %}
+            style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
+          {% endif %}
+        >
           {%- if section.settings.cover_image != blank -%}
             <img
               srcset="{%- if section.settings.cover_image.width >= 500 -%}{{ section.settings.cover_image | img_url: '500x' }} 500w,{%- endif -%}
@@ -28,10 +33,10 @@
               class="video-section__image"
             >
           {%- else -%}
-            {{ 'collection-2' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
-          {%- endif -%}
-        </div>
-      </a>
+            {{ 'collection-2' | placeholder_svg_tag: 'placeholder-svg placeholder video-section__image' }}
+          {%- endif -%}</div>
+        </a>
+      </div>
     </noscript>
     <deferred-media class="video-section__media deferred-media no-js-hidden" data-media-id="{{ section.settings.video_url.id }}"
       {% if section.settings.cover_image != blank %}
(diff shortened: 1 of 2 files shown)


sections/video.liquid
```diff
@@ -0,0 +1,141 @@
+{{ 'video-section.css' | asset_url | stylesheet_tag }}
+{{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
+
+<div class="video-section{% unless section.settings.full_width %} page-width{% endunless %}">
+  <h2 class="video-section__title title{% if section.settings.full_width %} page-width{% endif %}">{{ section.settings.heading }}</h2>
```
tauthomas01, line 5: Do we need `video-section__title`?

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+  <div class="video-section__wrapper{% if section.settings.image_padding %} video-section__media--padding{% endif %}">
```
tauthomas01, line 6: Do we need `video-section__wrapper`?

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+      ],
+      "default": "https:\/\/www.youtube.com\/watch?v=_9VUPq3SxOc",
+      "label": "Link",
+      "placeholder": "Use a YouTube or Vimeo link",
+      "info": "Video will play inside the page"
+    },
+    {
+      "type": "text",
+      "id": "description",
+      "default": "Describe the video for customers using screen readers",
```
melissaperreault, line 118: Wondering if this should become an "info" below the input so merchants can read the full sentence?

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+      "type": "video_url",
+      "id": "video_url",
+      "accept": [
+        "youtube",
+        "vimeo"
+      ],
+      "default": "https:\/\/www.youtube.com\/watch?v=_9VUPq3SxOc",
+      "label": "Link",
+      "placeholder": "Use a YouTube or Vimeo link",
+      "info": "Video will play inside the page"
```
melissaperreault, line 113:
```suggestion
      "info": "Video will play inside the page."
```

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+      "id": "description",
+      "default": "Describe the video for customers using screen readers",
+      "label": "Video alt text"
+    },
+    {
+      "type": "checkbox",
+      "id": "image_padding",
+      "label": "Add image padding",
+      "default": false,
+      "info": "Select image padding if you don't want your images to be cropped"
```
melissaperreault, line 126:
```suggestion
      "info": "Select image padding if you don't want your cover image to be cropped."
```
Aligned with @gretchen-willenberg for changing this and update on the Collage section as well. This setting is applied to one image at once and it appears to be the cover image. Reading is clearer for the merchant.

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+    },
+    {
+      "type": "video_url",
+      "id": "video_url",
+      "accept": [
+        "youtube",
+        "vimeo"
+      ],
+      "default": "https:\/\/www.youtube.com\/watch?v=_9VUPq3SxOc",
+      "label": "Link",
```
melissaperreault, line 111:
```suggestion
      "label": "URL",
```
Making this aligned with the admin when the only resources are external URLs. Aligned with @gretchen-willenberg

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+      "placeholder": "Use a YouTube or Vimeo link",
```
melissaperreault, line 112:
```suggestion
      "placeholder": "Use a YouTube or Vimeo URL",
```
Same as above!

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+<div class="video-section{% unless section.settings.full_width %} page-width{% endunless %}">
+  <h2 class="video-section__title title{% if section.settings.full_width %} page-width{% endif %}">{{ section.settings.heading }}</h2>
+  <div class="video-section__wrapper{% if section.settings.image_padding %} video-section__media--padding{% endif %}">
+    <noscript>
+      <div class="video-section__media"
+        {% if section.settings.cover_image != blank %}
+          style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
+        {% else %}
+          style="padding-bottom: 56.25%;"
+        {% endif %}
```
tauthomas01, line 13: Small nitpick: can we reduce the number of lines and merge into 1 line? For example

```suggestion
          style="padding-bottom: {% if section.settings.cover_image != blank %}{{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}{% else %} padding-bottom: 56.25{% endif %}%;"
```

Or we can move the `padding-bottom: 56.25%` into `.video-section__media` class as default value. When the `{% if section.settings.cover_image != blank %}` condition is satisfied, it will overwrite the default `56.25%` as it's an inline style, which has higher priority.

sections/video.liquid
```diff
@@ -0,0 +1,141 @@
...
+            >
+          {%- else -%}
+            {{ 'collection-2' | placeholder_svg_tag: 'placeholder-svg placeholder video-section__image' }}
+          {%- endif -%}</div>
+        </a>
+      </div>
+    </noscript>
+    <deferred-media class="video-section__media deferred-media no-js-hidden" data-media-id="{{ section.settings.video_url.id }}"
+      {% if section.settings.cover_image != blank %}
+        style="padding-bottom: {{ 1 | divided_by: section.settings.cover_image.aspect_ratio | times: 100 }}%;"
```
tauthomas01, line 43: Same here.

tauthomas01 (commented):
It works great! I left small nitpick comments regarding the code, otherwise the functionality works well.

sections/video.liquid
ludoboludo: Good point, so it doesn't go away. And have a default be `Video to describe your products or your brand` kind of thing ?

sections/video.liquid
ludoboludo: Sounds good. I will need to add the height limit. Right now it's adapting to whatever the size of the image is.

sections/video.liquid
ludoboludo: Nope! you're right.

sections/video.liquid
ludoboludo: Great idea!

locales/en.default.schema.json
```diff
@@ -1568,6 +1568,36 @@
...
+        "heading": {
+          "label": "Heading"
+        },
+        "cover_image": {
+          "label": "Cover image"
+        },
+        "video_url": {
+          "label": "URL",
+          "placeholder": "Use a YouTube or Vimeo URL",
+          "info": "Video will play inside the page."
```
melissaperreault, line 1584:
```suggestion
          "info": "Video plays in the page."
```
New version from Gretchen!

locales/en.default.schema.json
```diff
@@ -1568,6 +1568,36 @@
...
+        },
+        "description": {
+          "label": "Video alt text",
+          "info": "Describe the video for customers using screen readers"
```
melissaperreault, line 1588:
```suggestion
          "info": "Describe the video for customers using screen readers."
```
Just add a period to the end of info text please!

sections/video.liquid
melissaperreault: I wonder if there should be a default, if so, it would have to be short otherwise merchants can't read the full sentence. Perhaps: `Describe your video` ?

When we edit alt text of an image this is the info helper text:

> Write a brief description of this image to improve search engine optimization (SEO) and accessibility for visually impaired customers. This change will be saved immediately.

@gretchen-willenberg could we improve the Video alt text info text to align with what we have for image alt text (and ignore the SEO part)?

Draft: `Describe the video for visually impaired customers using screen readers.`

ludoboludo: I think the reason why users might choose to use screen readers can be left out. Because it's not necessary always true.

gretchen-willenberg: Sure!
1. Updated the `Video alt text` placeholder copy from repeating `Video alt text` to `Describe the video`
2. Agree with your revision to `Describe the video for customers using screen readers.` so it's `Describe the video for visually impaired customers using screen readers.`

I think I had some fuzzy old memory about not specifying why some one might choose to use a screen reader, but all research I just skimmed includes at least "visually impaired" if not "blind or visually impaired"--so corrected!

sections/video.liquid
ludoboludo: Nah I don't think so.

sections/video.liquid
```diff
@@ -0,0 +1,130 @@
...
+              {%- if section.settings.cover_image.width >= 1440 -%}{{ section.settings.cover_image | img_url: '1440x' }} 1440w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 990 -%}{{ section.settings.cover_image | img_url: '990x' }} 990w,{%- endif -%}
+              {%- if section.settings.cover_image.width >= 2080 -%}{{ section.settings.cover_image | img_url: '2080x' }} 2080w{%- endif -%}"
+              src="{{ section.settings.cover_image | img_url: '533x' }}"
+              sizes="(min-width: 990px)990px, (min-width: 750px) calc(100vw - 100px), calc(100vw - 30px)"
+              alt="{{ section.settings.description | escape }}"
+              loading="lazy"
+              width="{{ section.settings.cover_image.width }}"
+              height="{{ section.settings.cover_image.height }}"
+              class="video-section__image"
```
tauthomas01, line 25: I think we can remove all instances of `.video-section__image` because I don't see it being used. What are your thoughts?

tauthomas01 (approved):

sections/video.liquid
melissaperreault: > I think the reason why users might choose to use screen readers can be left out. Because it's not necessary always true.

@ludoboludo can you elaborate on this please?

melissaperreault (dismissed):
Thank you Ludo! Apart from the last piece of content being discussed, this looks good to me! 🎉 

cc. @nicklepine We didn't include a height setting but would like to observe the usage and feedback before doing so unless we feel strongly about it. Merchants cover image would need to be 16:9. Perhaps we could add the info under the setting... I'll leave it to you!

PaulNewton:
`sections/video.liquid`
As I understand #49 this was made to explicitly have "Video" show in the theme designer to lessen merchant confusion looking to add video to complement the the more complex collage section?

By extension does does this mean were also going to have a separate 3D section, PDF section, file section, etc.
Or in future pulls adding the support for that other media then change this to `sections/media.liquid`(or `video-and-media`* or `video-images-and-media`) , with it's schema name to `Video and Media` to show in the designer (_still observing #49_).
  Or break this back down to block usage?
 _*Using "media" also reflects how apis and liquid organizes content like video under the media concept,  but we would probably still want to have **video** in the filename for merchant discoverability._

As a developer while `sections/video.liquid` at least pares down the code to manage for a specific type of content I wouldn't want to support more than a handful of distinct media section files (3d,image,animation,videos) simply because the [theme architecture doesn't have a method of reusing the existing code in other sections|blocks](https://github.com/Shopify/dawn/issues/49#issuecomment-880394624).

As theme setup support the 'add section' picker does not categorize sections together it only displays them alphabetically making it easy to miss related sections or increasing the time it takes to explore for the right section capabilities as the list grows so the display name of sections should take this into account.
image
.

tauthomas01 (approved):

sections/video.liquid
gretchen-willenberg: Added some [research to the Slack thread](https://shopify.slack.com/archives/C03T74ZTD/p1626296045302700) and an updated version of copy: 
`Describe the video to make it accessible for customers using screen readers.`

sections/video.liquid
```diff
@@ -0,0 +1,128 @@
...
+      ],
+      "default": "https:\/\/www.youtube.com\/watch?v=_9VUPq3SxOc",
+      "label": "t:sections.video.settings.video_url.label",
+      "placeholder": "t:sections.video.settings.video_url.placeholder",
+      "info": "t:sections.video.settings.video_url.info"
+    },
+    {
+      "type": "text",
+      "id": "description",
+      "default": "Describe your video",
```
melissaperreault, line 104:
```suggestion
      "default": "Describe the video",
```
Small change

locales/en.default.schema.json
```diff
@@ -1568,6 +1568,36 @@
...
+          "label": "Cover image"
+        },
+        "video_url": {
+          "label": "URL",
+          "placeholder": "Use a YouTube or Vimeo URL",
+          "info": "Video plays in the page."
+        },
+        "description": {
+          "label": "Video alt text",
+          "info": "Describe the video for visually impaired customers using screen readers."
```
melissaperreault, line 1588:
```suggestion
          "info": "Describe the video to make it accessible for customers using screen readers."
```
New sentence provided by Gretchen

melissaperreault (approved):

tauthomas01 (approved):

svinkle:
In the editor, the "Describe the video for customers using screen readers" content is set as the `value` for the `input`. This causes some issues:

1. People may think the `input` is pre-filled and not need to do anything more
2. Setting this as the `value` forces more work on part of the user to remove the default text
3. Submitting as-is sets the video description as "Describe the video for customers using screen readers" which wouldn't be accurate

This content would ideally be hint text appearing below the `input`, connected via `aria-describedby`. (Also, the `aria-labelledby` is redundant.)

```diff
<input
  id="TextSetting-Label2"
  placeholder=""
  autocomplete="off"
  class="Polaris-TextField__Input_30ock"
  type="text"
- aria-labelledby="TextSetting-Label2Label"
+ aria-describedby="TextSetting-Label2Hint"
  aria-invalid="false"
- value="Describe the video for customers using screen readers"
/>
+ <small id="TextSetting-Label2Hint">Describe the video for customers using screen readers</small>
```

gretchen-willenberg:
Looks like @svinkle has an even better fix. 

That said, I think the update to placeholder text was missed. @melissaperreault and I discussed in the Figma (attached). 
Label: `Video alt text`
Placeholder: `Describe your video`
Support text: `Describe the video to make it accessible for customers using screen readers.`

Screen Shot 2021-07-20 at 4 24 46 PM

sections/video.liquid
```diff
@@ -0,0 +1,128 @@
+{{ 'video-section.css' | asset_url | stylesheet_tag }}
+{{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
+
+<div class="video-section{% unless section.settings.full_width %} page-width{% endunless %}">
+  <h2 class="title{% if section.settings.full_width %} page-width{% endif %}">{{ section.settings.heading }}</h2>
+  <div{% if section.settings.image_padding %} class="video-section-padding"{% endif %}>
+    <noscript>
```
svinkle, line 7: I noticed in the demo store there's an extra, empty `a` being generated somehow:

```html
<noscript>
  <div class="video-section__media" style="padding-bottom: 66.67480468750001%">
    <a
      href="https://vimeo.com/571010526"
      class="video-section__poster media media--transparent"
      style="padding-bottom: 66.67480468750001%"
    >
      <img
        srcset="…"
        src="//cdn.shopify.com/s/files/1/0458/4099/0230/files/a-persons-hands-knitting-with-white-wool_533x.jpg?v=1625665055"
        sizes="(min-width: 990px)990px, (min-width: 750px) calc(100vw - 100px), calc(100vw - 30px)"
        alt="Describe the video for customers using screen readers"
        loading="lazy"
        width="4096"
        height="2731"
        class="video-section__image"
      />
    </a>
  </div>
  <a
    href="https://vimeo.com/571010526"
    class="video-section__poster media media--transparent"
    style="padding-bottom: 66.67480468750001%"
  ></a>
</noscript>
```

sofiamatulis:
> To make this easier on merchants, would it be possible to:
> 
> 1. Detect a video URL from YouTube or Vimeo
> 2. Grab the video `id`
> 3. Ping the respective API to grab the video title, and
> 4. Fill in the `input`
> 
> This would suffice as the video description rather than rely on the merchant to come up with something.

After doing some exploration this is what I am thinking that wouldnt be so good by doing that even if possible:

- The request could fail, so we would need to account for that. What would happen in that case? It would fail silently since the merchant wouldn’t even be aware that we are doing that. 
- It feels like we would be adding this new functionality with JS for not a main necessity. In other words, this is not blocking the merchant or user since we can still allow the merchant to add the title manually. Currently we have no `{% javascript` or `<script` in the video liquid section and we would need to add it
- We would be adding more requests since every time we add a new video we would have to call the API for the new title.
- We would need different approaches and API endpoints for Vimeo and Youtube which means more lines of code.
- After calling the API we would need to insert the title back in HTML which isn’t ideal
- This would only work with JS

What are your thoughts? Maybe this could be created as an issue/enhancement.

> To make this easier on merchants, would it be possible to:
> 
> 1. Detect a video URL from YouTube or Vimeo
> 2. Grab the video `id`
> 3. Ping the respective API to grab the video title, and
> 4. Fill in the `input`
> 
> This would suffice as the video description rather than rely on the merchant to come up with something.

@svinkle After doing some exploration, this would make more sense on the platform side since we are trying to keep JS at a minimum on Dawn and this would require quite some JS to be added.

More context here: https://shopify.slack.com/archives/G0MJPQRSB/p1627326724025000

Created an enhancement/issue here https://github.com/Shopify/shopify/issues/305128 cc @tyleralsbury

sofiamatulis:
> To make this easier on merchants, would it be possible to:
> 
> 1. Detect a video URL from YouTube or Vimeo
> 2. Grab the video `id`
> 3. Ping the respective API to grab the video title, and
> 4. Fill in the `input`
> 
> This would suffice as the video description rather than rely on the merchant to come up with something.

@svinkle After doing some exploration, this would make more sense on the platform side since we are trying to keep JS at a minimum on Dawn and this would require quite some JS to be added. I have created an issue.

tyleralsbury (dismissed):
Looks good for the code and 🎩 

I'm wondering about the phrasing for "Add image padding" and whether it would be better to call it something like "Add media padding" since it affects the poster image as well as the video itself.

cc @melissaperreault @gretchen-willenberg

melissaperreault:
> I'm wondering about the phrasing for "Add image padding" and whether it would be better to call it something like "Add media padding" since it affects the poster image as well as the video itself.

@tyleralsbury I agree, good catch!

melissaperreault (commented):
Thank you Sofia!

Looking at this with fresh eyes, [we'll need to](https://screenshot.click/21-07-fa8lw-tcxdj.png):

- [ ] A. Align the heading with others when `Make section full width` is `true`.
- [ ] B. Finally remove completely the setting Add media padding, it would simplify much more the section and merchants could still have this effect on the Collage section if they really want to have the border effect. The border effect is less appealing when the section is full width too.

locales/en.default.schema.json
```diff
@@ -1568,6 +1568,36 @@
...
+          "label": "URL",
+          "placeholder": "Use a YouTube or Vimeo URL",
+          "info": "Video plays in the page."
+        },
+        "description": {
+          "label": "Video alt text",
+          "info": "Describe the video to make it accessible for customers using screen readers."
+        },
+        "image_padding": {
+          "label": "Add image padding",
```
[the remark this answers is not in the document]
melissaperreault: Changed our mind, we'll remove this setting completely on this Video section.

tyleralsbury (dismissed):
Re-approving after changes.

locales/en.default.schema.json
```diff
@@ -1568,6 +1564,36 @@
...
+          "info": "Select image padding if you don't want your cover image to be cropped."
```
tyleralsbury, line 1588: I think we can get rid of this since you removed the setting.

sofiamatulis: Good catch! I had removed the wrong translation

melissaperreault (approved):
🎉  Looks good, thank you Sofia!

sofiamatulis (approved):

sofiamatulis (approved):

melissaperreault (approved):

tyleralsbury (approved):

PR #190, opened 2021-07-15
Small optimizations to JS loading

tyleralsbury:

Bundling + changing the positions of a few JS files to be picked up earlier by the browser.

Since the JS is deferred, we can put them higher up in the document without them negatively affecting the loading of other files.  We can also bundle the JS a bit tighter - removing `variants.js` and `slider.js` and putting them, instead, into `global.js` to remove some network requests.

JS that has the `defer` attribute will download when it's found in the DOM at low priority, so it won't block the loading of more critical resources, and it will only run after the critical render path is done.  By putting the references to these files higher up, we can benefit from them being picked up sooner by the browser and ensure they're ready to go when the rest of the critical path is done downloading.

https://blog.bluetriangle.com/js-delivery-optimization-for-web-performance

**Checklist**


tyleralsbury:
> We could most likely at least have template-collection.css loaded right away 

It should be for sure.  It's a large file but it contains a lot of things that will cause layout shift if it's excluded.  We'll need to make another pass over the theme to fix layout shift at some point.  It crops back up occasionally as things change and move around.

tauthomas01 (approved):

baldurh:
I thought in today's world of http/2 bundling too much was discouraged. Were there any specific problems with having `variants.js` and `slider.js` separate from `global.js`? Also, isn't this forcing users to download unnecessary code on most pages?

PR #195, opened 2021-07-15
Add properties to cart pop up like debut theme

psppro26:
Simply adding properties to cart pop up like Debut theme :-)

### DIFF
--- a/sections/cart-notification-product.liquid
+++ b/sections/cart-notification-product.liquid
@@ -22,6 +22,13 @@
             {%- endfor -%}
           </dl>
         {%- endunless -%}
+        {% unless line_item.properties == empty %}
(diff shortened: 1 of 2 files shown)


sections/cart-notification-product.liquid
```diff
@@ -22,6 +22,13 @@
             {%- endfor -%}
           </dl>
         {%- endunless -%}
+        {% unless line_item.properties == empty %}
+		<div class="cart-notification-product_properties h4">
```
svinkle, line 26: This would generate invalid HTML. Swap the `div` for a `ul`.

ludoboludo:
This has since been added to all cart layouts (popup, page and drawer). Closing the PR.

PR #210, opened 2021-07-16
Update main-product.liquid to fix no-JS add to cart

### LINKED ISSUE #179: Add to cart does not work with JS disabled
Hi,

As of today Dawn output a <noscript> select: https://github.com/Shopify/dawn/blob/main/sections/main-product.liquid#L235

However another one is set in the product form: https://github.com/Shopify/dawn/blob/main/sections/main-product.liquid#L260

Because you cannot have two input with the same name, in this situation the last one (depending on the block order) will take the control, and as add to cart is typically after selectors, only the add to cart one will win.

Dawn add to cart form with non JS is therefore currently stuck to the first available variant.

I think Dawn should make sure that the name="id" only exists once within the form and be synchronized in JS.

tairli:
Fix problem when only initial variant was added to cart wit Javascript disabled regardless of selection made.

Fixes #179.

 Modified a _form_ property on select
 ID _input_ inside form is initially rendered as _disabled_ and enabled with Javascript. Stays disabled without JS to not interfere with _select_


tairli:
> You need to make sure to wrap the code around a DOMContentLoaded event, or add it _after_ the id element. At this stage the input is still not parsed.

The code was working as it was, but technically you're right, so  instead of moving the code below _form_, I've_ redone the PR by moving the JS part into product-form.js as I considered above.

tyleralsbury (approved):
Looks good to me!  Thanks for the contribution.

- Tested with JS disabled ✅ 
- Tested with button variant switcher with JS enabled ✅ 
- Tested with dropdown variant switcher with JS enabled ✅ 
- Tested product w/o variant with JS enabled and disabled ✅ 
- Tested quantity selector ✅

LucasLacerdaUX (approved):
Tested and the fix works as expected 👍

Thanks for contributing and I apologize for taking so long to review.

You can go ahead and merge this as is. I'll replicate the same fix on the `featured-product` section in a separate PR.

tyleralsbury:
@tairli I believe you need to sign the CLA:

Screen Shot 2021-09-07 at 10 26 00 AM

After that, the PR can be merged.  Thanks again!

tairli:
@chrisberthe Do you have any news on this?
It's back to "You've already signed the agreement"

LucasLacerdaUX:
Hey @tairli. Sorry for the long delay, but I finally merged this PR onto the main branch. I'll now reproduce the same fix for the Featured Product section. Thank you for your contribution and patience to keep this updated! :)

The `cla-needed` tag was being wrongfully added because of `shopify-online-store`'s auto commit, so you were right, there was nothing you could do to resolve it on your end. Sorry about that - we'll work internally to prevent that from happening again.

PR #163, opened 2021-07-11
cart add line_item.properties ajax json support

michaelhaessig:
**Why are these changes introduced?**
Hi Team, 

the current implementation of the productForm.onSubmitHandler does not correctly submit custom line item properties to the Ajax cart api. 

the function serializeForm in global.js returns the following JS object:  ( 2 custom properties defined in the form )

````
{
  "quantity": "1",
  "form_type": "product",
  "utf8": "✓",
  "id": "40307349094593",
  "properties[name]": "michael",
  "properties[upload]": "https://upload.dev.cloudlift.app/s/dev/cloudlift-dev-store2/SVutmqCkG2.png",
  "sections": [
    "cart-notification-product",
    "cart-notification-button",
    "cart-icon-bubble"
  ],
  "sections_url": "/products/upload-lift"
}
````

This object is sent via "fetch" with the content-type "application/json".   The Shopify ajax api will not detect the line item properties in this format.  

The docs state that the properties have to nested in the "properties" field: https://shopify.dev/api/ajax/reference/cart#add-line-item-properties

Lots of apps are using custom line item properties in the product form, the current implementation breaks all of them. 

**What approach did you take?**

This pull request extends the serializeForm function to auto nest the line item properties which results in the following JS object: 

````
{
  "quantity": "1",
  "form_type": "product",
  "utf8": "✓",
  "id": "40307349094593",
  "properties": {
    "name": "michael",
    "upload": "https://upload.dev.cloudlift.app/s/dev/cloudlift-dev-store2/3LuvMdXGbS.png"
  },
  "sections": [
    "cart-notification-product",
    "cart-notification-button",
    "cart-icon-bubble"
  ],
  "sections_url": "/products/upload-lift"
}
````

**Checklist**
- [X] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [X] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [X] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [X] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [X] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #257, opened 2021-07-23
Fix null cart properties and line items

### LINKED ISSUE #243: Unable to add line item properties
These are the HTML inputs I am using on the PDP:

```
<product-form>
   <form method="post" action="/cart/add" id="product-form-template--29384234754__main" accept-charset="UTF-8" class="form" enctype="multipart/form-data" novalidate="novalidate" data-type="add-to-cart-form">
         <div id="cart-line-items">
            <input class="item-contains" type="hidden" name="properties[_item_0]" value="item0-data-here">
            <input class="item-contains" type="hidden" name="properties[_item_1]" value="item1-data-here">
         </div>
   </form>
</product-form>
```

## Current Behavior

Adding line items properties via HTML inputs within the `AddToCart` form does not carry over to `cart.json` upon adding product to the cart.

**Current Cart Screenshot:**
image

**Current `Cart.json`:**

```
...
"items":
      [
           {
              "id": 40230556827839,
              "properties": null, 
               ...
           }
      ]
  ...
```

## Expected behavior

**Expected Cart Screenshot:**
image

**Expected `Cart.json`:**

```
...
"items":
      [
           {
              "id": 40230556827839,
              "properties": {
                  "_item_0": "item0-data-here",
                  "_item_1": "item1-data-here",
               }, 
               ...
           }
      ]
  ...
```

- Dawn Version: 1.0.0
- macOS Version 11.3.1

## Possible solution

This works as expected in other themes (like Debut). This seems like a bug, so I am led to believe that the same functionality needs to be implemented into Dawn as it is in Debut.

The current docs for the Dawn theme are the same as they are for the Debut theme. It seems that either the docs need to be updated to explain the proper usage of inputs for line item properties, or that the Dawn theme has a bug that is blocking this functionality from working.

perryraskin:

Fixes #243.

I noticed that the `/cart/add` API endpoint requires a `properties` object to be passed into the `body`. This was not being generated from the line-item HTML inputs by the theme. The theme was only parsing the product form, resulting in key/value pairs such as the following:

```
"properties[Contains]":"1x Item 0",
"properties[_item_0]":"item0-data-here"
```

Instead, we need something like this:

```
"properties": {
   "Contains": "1x Item 0",
   "_item_0": "item0-data-here"
}
```

To resolve this, I implemented some logic that loops through the keys of the parsed form object in search of keys that begin with `properties[`. From there, I am generating a `properties` object and appending it to the final `body` JSON string, which is later passed to the `/cart/add` API endpoint.

**Other considerations:**

The other option here is to update the docs for Liquid to describe a different way it should be done in the Dawn theme. [As of now it says](https://shopify.dev/api/liquid/objects/line_item#line_item-properties) to add line items in this format:

```
<input id="engraving" type="text" name="properties[Engraving]">
```

If there is a known method of doing so properly in Dawn, please update the docs to let us know! Otherwise, this minor update to the code is a great workaround.

I did however try the following line item input:

```
<input type="text" name="properties" value='{"Contains": "1x Item 0", "_item_0": "item0-data-here"}'>
```

which was unsuccessful, as the API returned a `bad_request` error due to the `properties` field not being formatted properly.

**Checklist**

### DIFF
--- a/assets/product-form.js
+++ b/assets/product-form.js
@@ -1,28 +1,43 @@
 class ProductForm extends HTMLElement {
   constructor() {
-    super();   
+    super();
 
-    this.form = this.querySelector('form');
-    this.form.addEventListener('submit', this.onSubmitHandler.bind(this));
-    this.cartNotification = document.querySelector('cart-notification');
+    this.form = this.querySelector("form");
+    this.form.addEventListener("submit", this.onSubmitHandler.bind(this));
+    this.cartNotification = document.querySelector("cart-notification");
   }
 
   onSubmitHandler(evt) {
     evt.preventDefault();
     this.cartNotification.setActiveElement(document.activeElement);
-    
+
     const submitButton = this.querySelector('[type="submit"]');
 
-    submitButton.setAttribute('disabled', true);
-    submitButton.classList.add('loading');
+    submitButton.setAttribute("disabled", true);
+    submitButton.classList.add("loading");
+    let parsedForm = JSON.parse(serializeForm(this.form));
+    let properties = {};
+    for (const key in parsedForm) {
+      const propertiesKeyMatch = key.substring(0, 11);
+      if (propertiesKeyMatch === "properties[") {
+        const startIndex = key.indexOf("[") + 1;
+        const endIndex = key.indexOf("]");
+        const propertyKey = key.substring(startIndex, endIndex);
+        const propertyValue = parsedForm[key];
+        properties[propertyKey] = propertyValue;
+      }
+    }
+    parsedForm.properties = properties;
 
     const body = JSON.stringify({
-      ...JSON.parse(serializeForm(this.form)),
-      sections: this.cartNotification.getSectionsToRender().map((section) => section.id),
-      sections_url: window.location.pathname
+      ...parsedForm,
+      sections: this.cartNotification
+        .getSectionsToRender()
+        .map((section) => section.id),
+      sections_url: window.location.pathname,
     });
 
-    fetch(`${routes.cart_add_url}`, { ...fetchConfig('javascript'), body })
+    fetch(`${routes.cart_add_url}`, { ...fetchConfig("javascript"), body })
       .then((response) => response.json())
       .then((parsedState) => {
         this.cartNotification.renderContents(parsedState);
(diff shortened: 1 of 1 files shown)


kennethparcell (approved):

carlsmith-wow (approved):
I have tested this fix and can confirm that it works.

The properties field must be an object rather than the array that is usually passed through.

Seems like the routes.cart_add_url has changed on the Dawn theme to /cart/add rather than /cart/add.js

perryraskin:
@liao1204ya my testing confirms that this does in fact work. please make sure you are adding your line items correctly according to the docs

bakura10:
Hi,

after reading this issue I realised actually that this is more problematic than I expected. The main issue is that because the form is constructed inside the « buy_buttons » block, this means that any app block or custom code found by merchants on the web will no longer really work, unless a form attribute is properly added to each line item attribute to relate it to the main form.

the main issue is that there is currently no convention or accurate way for an app, or tutorial, to get a form ID. 

For instance, Dawn use a simple ID that is a bit too optimistic (as it will fail as soon as you have two products on the section ; think of related product shown on the same section, which we have on several of our themes):

https://github.com/Shopify/dawn/blob/main/sections/main-product.liquid#L70

on our themes I have added extra info such as the product ID and block id if it exists to ensure it works in a more reliable way. However there is no convention nor way to enforce it.

this means an app block will have absolutely no way to actually generate an accurate form ID that would allow their line attributes to link to the product form.

this issue drastically reduces the benefit of app block, as I saw the line attributes as the main « usage » of app block in product page, but in its current implementation it makes them pretty unusable.

i wonder if a new {{product.form_id}} variable should not be added to the product drop. Shopify would ensure under the hood to ensure a unique ID for this product in the context of section/block. This would allow to avoid hard coding an ID and give a consistent variable to app to use.

perryraskin:
@bakura10 so how do we proceed here? Can we at least get this merged as a workaround for now?

bakura10:
I just came with the problem @perryraskin , not a solution :D. I will think about it on my end and let Shopify engineers try to find about it as well.

michaelhaessig:
We are seeing more merchants using out apps that encounter this issue. 

I've already provided a pr #163  which is still not merged.   It would be much appricated if anybody at Shopify could take a look at this as soon as possible.   

@bakura10 I've also encountered the same issue with implementing app-block that are added outside the form tag.  
I've had to look for the product form in a non deterministic way as the form id is overwritten in the dawn theme. (Normally we could find the right form via the shopify generated id "#product_form_\<id\>" ) .  As a solution i would also recommend to not overwrite the form id in [main-product.liquid](https://github.com/Shopify/dawn/blob/main/sections/main-product.liquid#L262)

bakura10:
We unforunately have to rewrite it @michaelhaessig . There are several use cases where the generic ID is not enough as you may have featured product section, featured collection that would feature the product... Without overriding it there are lot of cases where we end up with several separate form with the same ID :/.

khanhtv990:
thank you very much!

Issue #255, opened 2021-07-23
Price range input placeholder/max values do not handle comma separated currencies (e.g. Euros)


willbroderick:
**Describe the current behavior**
Store setup:
Use Euros, set money format to "€{{amount_with_comma_separator}}", on a store with products containing prices in the thousands.
The price range filter inputs do not contain correct values.

See max/placeholder here:
![image](https://user-images.githubusercontent.com/1590737/126745571-16237223-88e5-4ad0-bc93-889c48d0fbfe.png)

Referring to the code here: https://github.com/Shopify/dawn/blob/main/sections/main-collection-product-grid.liquid#L147
`placeholder="{{ filter.range_max | money_without_currency | replace: ',', '' }}"`
This goes: 2836.10 -> "2.836,10" -> "2.83610"

**Describe the expected behavior**
The output should be `placeholder="2836.10"`

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 2.0.0-alpha.23
- Any OS/Browser

**Possible solution**
The ideal solution would be for Shopify to handle this conversion, perhaps like this:
`placeholder="{{ filter.range_max | money_to_input_value }}"`

**Additional context/screenshots**
In my opinion the logic is far too complex for theme code. I don't think it's reasonable to expect a theme developer to be aware of the nuances of all currencies and their delimiters, whether or not they have 0/2/3 decimal places, to know what is going to be output by the current shop.money_format formatting, and when to check against cart.currency.iso_code instead.

All of these appear to factor into this conversion, and I've yet to see an example that does this correctly.

Locking this logic down to the the state of the world in July 2021 also seems like an odd choice to me.

bakura10:
I agree. Shopify definitely need to give us a better way to format the pricing.

I don't think also that the decimal value should be exposed at all in the filter, because the increment is usually 1, and can make the prices extremely long.

For now I have been able to create integer values using this, but I have no idea if this will work constantly for all currencies:

```
{%- assign min_value = filter.min_value.value | default: 0.0 | divided_by: 100.0 -%}
{%- assign max_value = filter.max_value.value | default: filter.range_max | divided_by: 100.0 -%}
```

I think Shopify should expose additional variables:

* {{ filter.min_value.formatted_value }}, {{ filter.max_value.formatted_value }}, {{ filter.formatted_range_max }}: this would return a pre-formatted values and max range to be used in the selector, without relying on hard-coded currencies list
[...]

PR #273, opened 2021-07-27
Fixed filtering number inputs

tyleralsbury:

Fixed the bug described in https://github.com/Shopify/dawn/issues/255

However, it doesn't fix the underlying platform feature that could be added to remove all of this complexity.

Just added the check for the type of separators to the places that I'd originally missed.

**Other considerations**

There is some good discussion in the issue around a potential Liquid filter that could get rid of all of the nasty, fragile logic in here.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=120921620502)
- [Editor](https://os2-demo.myshopify.com/admin/themes/120921620502/editort)

**Checklist**

### DIFF
--- a/sections/main-collection-product-grid.liquid
+++ b/sections/main-collection-product-grid.liquid
@@ -118,20 +118,12 @@
                           name="{{ filter.min_value.param_name }}"
                           id="Filter-{{ filter.label | escape }}-GTE"
                           {%- if filter.min_value.value -%}
-                            {%- if uses_comma_decimals -%}
-                              value="{{ filter.min_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
-                            {%- else -%}
-                              value="{{ filter.min_value.value | money_without_currency | replace: ',', '' }}"
-                            {% endif %}
+                            {%- if uses_comma_decimals -%}value="{{ filter.min_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"{%- else -%}value="{{ filter.min_value.value | money_without_currency | replace: ',', '' }}"{% endif %}
                           {%- endif -%}
                           type="number"
                           placeholder="0"
                           min="0"
-                          {%- if uses_comma_decimals -%}
-                            max="{{ filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
-                          {%- else -%}
-                            max="{{ filter.range_max | money_without_currency | replace: ',', '' }}"
-                          {% endif %}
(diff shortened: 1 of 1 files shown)


ludoboludo (approved):
The code looks good to me. I didn't change the store's currency as I would have needed to deactivate Shopify payments to change the currency. 

One limitation I noticed, which isn't introduced in this PR is the fact that when I change region/currency, the price filter is gone: [video](https://screenshot.click/21-08-fbcpk-rp2i7.mp4)

And also, say I'm wanting a max value of `$1,000` when I press enter it will change this value to the value that is associated with the `max` attribute of the input which surprised me at first while using it: [video](https://screenshot.click/21-08-2fhg9-putsu.mp4)

sections/main-collection-product-grid.liquid
```diff
@@ -136,17 +133,18 @@
...
-                              value="{{ filter.max_value.value | money_without_currency | replace: ',', '' }}"
-                            {% endif %}
+                          {%- if filter.max_value.value -%}{%- if uses_comma_decimals -%}value="{{ filter.max_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"{%- else -%}value="{{ filter.max_value.value | money_without_currency | replace: ',', '' }}"{% endif %}
                           {%- endif -%}
                           type="number"
-                          placeholder="{{ filter.range_max | money_without_currency | replace: ',', '' }}"
                           min="0"
-                          max="{{ filter.range_max | money_without_currency | replace: ',', '' }}">
+                          {%- if uses_comma_decimals -%}
+                            placeholder="{{ filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
```
KaichenWang, line 141: We've tried to avoid `{% assign %}` as much as possible in the theme, but could the `filter.range_max`, which is repeated a few times, benefit from using it?

```
{%- if uses_comma_decimals -%}                 
  {%- assign range_max = filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' -%}
{%- else -%}
  {%- assign range_max = filter.range_max | money_without_currency | replace: ',', '' -%}
{%- endif -%}

placeholder="{{ range_max }}"
max="{{ range_max }}"                     
```

KaichenWang (approved):
Code changes look good

sections/main-collection-product-grid.liquid
tyleralsbury: I'd rather avoid it for now.  We're going to revisit this approach anyway with some help from the platform given how fragile it is, so it'll likely be something a lot more manageable that wouldn't justify an assign down the line. See #255 for more details.

PR #327, opened 2021-08-05
Adjust id for displaying dynamic names

melissaperreault:
**Why are these changes introduced?**

Currently Announcement bar [have static names](https://screenshot.click/21-08-xtxfy-kzk48.png) and the experience for merchants isn't great.

Related to this the recent update https://github.com/Shopify/online-store-web/issues/10953#issuecomment-879094000 we need to change the id  for `heading` rather than `text`

- [Store](url)
- [Editor](url)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #372, opened 2021-08-10
Dynamic naming for blocks

### LINKED ISSUE #362: Adjust block IDs for dynamic naming of blocks
Block names can end up being the same for each instances of a same block which ends up being confusing to navigate. 

We're going to set it up to so that blocks have dynamic names pulled from certain settings in the schema settings based on their ID. 

- Dawn Version: 1.0.0
- macOS Version 11.3.1

**Possible solution**
We will need to go through all blocks in the theme and check the IDs currently used and change them to use either: 
- heading
- title
- text

To be picked up and used as the naming of the block. 

More context in [this issue on a private repo](https://github.com/Shopify/online-store-web/pull/11040).

ludoboludo:

Fixes #362 .

Searched through all the `"type": "text"` and `"type": "richtext"`.

Updated some IDs so block naming gets updated based on the content added by the merchant. 

I find that we lose some readability as it's easier to read `block.settings.button_label` vs `block.settings.text`. 
I'm curious to see what reviewers think about not updating the translation path with `text` where the updates were made. 

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=124598091798)
- [Editor](https://os2-demo.myshopify.com/admin/themes/124598091798/editor)

**Checklist**

### DIFF
--- a/sections/image-with-text.liquid
+++ b/sections/image-with-text.liquid
@@ -37,9 +37,9 @@
             {%- when 'text' -%}
               <div class="image-with-text__text rte" {{ block.shopify_attributes }}>{{ block.settings.text }}</div>
             {%- when 'button' -%}
-              {%- if block.settings.button_label != blank -%}
+              {%- if block.settings.text != blank -%}
                 <a{% if block.settings.button_link != blank %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}"{% if block.settings.button_link == blank %} aria-disabled="true"{% endif %} {{ block.shopify_attributes }}>
-                  {{ block.settings.button_label | escape }}
+                  {{ block.settings.text | escape }}
                 </a>
               {%- endif -%}
           {%- endcase -%}
@@ -159,7 +159,7 @@
       "settings": [
         {
           "type": "text",
-          "id": "button_label",
+          "id": "text",
           "default": "Button label",
           "label": "t:sections.image-with-text.blocks.button.settings.button_label.label",
           "info": "t:sections.image-with-text.blocks.button.settings.button_label.info"
--- a/sections/image-with-text.liquid
+++ b/sections/image-with-text.liquid
@@ -37,9 +37,9 @@
             {%- when 'text' -%}
               <div class="image-with-text__text rte" {{ block.shopify_attributes }}>{{ block.settings.text }}</div>
             {%- when 'button' -%}
-              {%- if block.settings.text != blank -%}
+              {%- if block.settings.button_label != blank -%}
                 <a{% if block.settings.button_link != blank %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}"{% if block.settings.button_link == blank %} aria-disabled="true"{% endif %} {{ block.shopify_attributes }}>
-                  {{ block.settings.text | escape }}
+                  {{ block.settings.button_label | escape }}
                 </a>
               {%- endif -%}
           {%- endcase -%}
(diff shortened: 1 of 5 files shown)


sections/rich-text.liquid
```diff
@@ -14,7 +14,7 @@
           <div class="rich-text__text rte" {{ block.shopify_attributes }}>{{ block.settings.text }}</div>
         {%- when 'button' -%}
           <a{% if block.settings.button_link == blank %} aria-disabled="true"{% else %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}" {{ block.shopify_attributes }}>
-            {{ block.settings.button_label | escape }}
+            {{ block.settings.text | escape }}
```
utkarshsaxena93, line 17: In this case I don't think the update is required as there can only ever be one `button` 
Screen Shot 2021-08-11 at 8 41 41 AM

sections/newsletter.liquid
```diff
@@ -124,7 +124,7 @@
       "settings": [
         {
           "type": "richtext",
-          "id": "paragraph",
+          "id": "text",
```
utkarshsaxena93, line 127: Similarly here, if the block limit is `1`, static name should suffice. What do you think?

sections/main-product.liquid
```diff
@@ -718,7 +718,7 @@
       "settings": [
         {
           "type": "text",
-          "id": "link_label",
+          "id": "text",
```
utkarshsaxena93, line 721: I think we can leave this unchanged as we show the name from the resource attached which should be unique across different blocks as shown in the screenshot below. I think it is less likely there will be different popups linking to the same page resource.

Name comes from the `About us` resource attached.
Screen Shot 2021-08-11 at 8 52 05 AM

sections/image-with-text.liquid
```diff
@@ -159,7 +159,7 @@
       "settings": [
         {
           "type": "text",
-          "id": "button_label",
+          "id": "text",
```
utkarshsaxena93, line 162: This change is not required either as only one block of each type can be added which should suffice

Screen Shot 2021-08-11 at 8 41 41 AM

utkarshsaxena93 (commented):
I think we can revert the change made in this PR https://github.com/Shopify/dawn/pull/327/files#diff-250e86a261baea4f8f30f09dbd328ffcb0b7d5d73366b8376fd844a02e195a93R34 as we now consider `text` id when showing the name.

sections/newsletter.liquid
ludoboludo: Yeah that's what I was thinking too. From what I understood, it might be useful when deciding which icons to use next to a block 🤔

sections/main-product.liquid
ludoboludo: Since you can have more than 1 I think it might be beneficial here to have it. So you can see which link/popup is where:

sections/image-with-text.liquid
rozinashopify: Buttons can behave unexpectedly, especially if we have more than one button in a block. Let's keep button blocks static named. We'll improve their iconography to reflect the type better later.

sections/main-product.liquid
ludoboludo: Good point. So it relates more to what you're actually seeing in preview 👌

rozinashopify (approved):
🚢⛴🛳

KaichenWang (commented):
@ludoboludo changes are 👍

Question - in `main-product.liquid` you updated the `popup` block to `"id": "text"`, but for `collapsible_tab` block it's better to leave `"id": "heading"`?

```
{
      "type": "collapsible_tab",
      "name": "t:sections.main-product.blocks.collapsible_tab.name",
      "settings": [
        {
          "type": "text",
          "id": "heading",
...
```

ludoboludo:
> @ludoboludo changes are 👍
> 
> Question - in `main-product.liquid` you updated the `popup` block to `"id": "text"`, but for `collapsible_tab` block it's better to leave `"id": "heading"`?

Ah yeah I can see what you mean. It won't affect how it's picked up to show the name. From what I understand it will look for `IDs` that are `heading`, `title` or `text`. I added `text` just cause I didn't think of it as a heading. You think it's better to have it the same ?

PR #391, opened 2021-08-11
Handle form inputs having brackets (e.g. name="properties[some-prop]")

faubertin:
Hi, as it stands, it seems that Dawn does not support line item properties. I have successfully tested this change on my dev store and it would solve the problem. Happy to discuss if there is a better mechanism to make this work.

Adds support for line item properties.

Regex check in form serialization logic.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -107,7 +107,13 @@ const serializeForm = form => {
   const obj = {};
   const formData = new FormData(form);
   for (const key of formData.keys()) {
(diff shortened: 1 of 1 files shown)


faubertin:
Closing since solved by https://github.com/Shopify/dawn/pull/509

PR #393, opened 2021-08-11
Change separator lines opacity

### LINKED ISSUE #22: Update the opacity of all separator lines
Currently, all separator lines have an opacity of 20%. They should be updated to 8% of the text colour to match the stroke lines used in the collage section and create a more subtle embellishment through the theme.

**What's included:**
- Footer, header, announcement bar
- Accordion on the product page
- Lines between pickup availability options, etc.

**What this doesn't include:**
- Form fields
- Product card borders
- Dropdown borders (i.e navigation, cart notification)
- Drawer borders (i.e navigation on mobile, pickup availability) 

cc: @Guillaumegranger1

ludoboludo:

Fixes #22

Changed some of the existing value to be `0.08` instead of `0.2`.

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=124601073686)
- [Editor](https://os2-demo.myshopify.com/admin/themes/124601073686/editor)

**Checklist**

### DIFF
--- a/assets/customer.css
+++ b/assets/customer.css
@@ -57,7 +57,7 @@
 .customer table {
   table-layout: auto;
   border-collapse: collapse;
-  border-bottom: 0.01rem solid rgba(var(--color-foreground), 0.2);
+  border-bottom: 0.01rem solid rgba(var(--color-foreground), 0.08);
   box-shadow: none;
   width: 100%;
   font-size: 1.6rem;
@@ -67,7 +67,7 @@
 @media screen and (min-width: 750px) {
   .customer table {
     border: none;
-    box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.2);
+    box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.08);
   }
 }
 
@@ -147,7 +147,7 @@
 }
 
 .customer tbody tr {
-  border-top: 0.01rem solid rgba(var(--color-foreground), 0.2);
+  border-top: 0.01rem solid rgba(var(--color-foreground), 0.08);
 }
 
 @media screen and (min-width: 750px) {
@@ -197,7 +197,7 @@
   display: block;
   position: absolute;
   left: 0;
-  background: rgba(var(--color-foreground), 0.2);
+  background: rgba(var(--color-foreground), 0.08);
 }
 
 /* mobile table overrides */


ludoboludo:
> Screen Shot 2021-08-19 at 1 02 25 PM
> 
> On shops with light backgrounds, the line contrast is more defined. On darker backgrounds, the line is almost invisible. In the screenshot I have provided, there are lines around the product cards and a separator line below the header. Are we sure we want to change the line opacity to 8%?

Oof, good point 🤔  cc: @Guillaumegranger1

Guillaumegranger1:
I know it makes the theme less scalable, but I think we should move forward with it because it was requested by stakeholders and I don't feel like fighting that battle. Or can we implement the update as is, and see later if we can slightly bump up the opacity? Is it easy to do? ...like changing a variable in one place? @ludoboludo

ludoboludo:
Yes, it wouldn't be hard to change later. It's not something in one place but a few. Doesn't take long to figure out which ones though.

chrisberthe (dismissed):
Looks good. @melissaperreault would the cart notification border be part of this colour update as well? I noticed this variation so questioning it here:

image

kmeleta (dismissed):
If we're moving forward with this change as is, LGTM

ludoboludo:
> Looks good. @melissaperreault would the cart notification border be part of this colour update as well? I noticed this variation so questioning it here:
> 
> [image](https://user-images.githubusercontent.com/991693/130658533-b3acac47-acd6-4451-8a70-ac11a7515231.png)

I assumed we didn't want to do it based on this: [screenshot from the issue](https://screenshot.click/21-08-48eo3-oz8ix.png). I had tested and it looked pretty faint in contrast with the background but I see how it's not seemless with the rest of the header border

melissaperreault:
> Looks good. @melissaperreault would the cart notification border be part of this colour update as well? I noticed this variation so questioning it here:


> I assumed we didn't want to do it based on this: screenshot from the issue. I had tested and it looked pretty faint in contrast with the background but I see how it's not seemless with the rest of the header border


I actually find it detaches and help emphasize the component so I would keep it as is 👌 We also have the interest to revise the overlay patterns across similar experience from this issue https://github.com/Shopify/dawn/issues/462

melissaperreault (commented):
Thanks Ludo, took a look and looks good!

I think we could also make the change in [My account tables](https://screenshot.click/21-08-5dmzy-g0ew0.png) to have a similar feeling from the Cart. [Example of the outcome](https://screenshot.click/21-08-eabpo-olog2.png).

This would affect the Account order history and the order detail tables I believe.

ludoboludo:
@melissaperreault I was making the changes and thought I'd bring something up. It has some continuity when doing it, so it matches the footer and header but the role of the borders here also plays a role in terms of accessibility. 
It helps keep track of which order you're looking at. So if the lines are pretty faint it will make it harder 🤔 
So it might be worth keeping it as is. 
Let me know what you think, I can loop Scott in if necessary 👍

melissaperreault:
> It has some continuity when doing it, so it matches the footer and header but the role of the borders here also plays a role in terms of accessibility.

I understand the point you raise but believe it would be aligned with the overall changes and minimal aesthetic to harmonize in those templates too. To me, the borders are already not accessible (low contrast) at 20% and we have even fewer separator in the Cart template which wears the 8% already. 

This might be a bigger conversation whether our table are easy to consult through Fable testing to inform how we can improve legibility, but it would require probably more than increasing the opacity of the borders. 

I was not part of the effort to change some of the border opacity so I'd like to loop in @Oliviammarcello and @Guillaumegranger1 , perhaps they did consider keeping them at 20%. 🙏

ludoboludo:
@kmeleta & @chrisberthe I'd need your approval again. I just changed the opacity on the account page for the table or order page

kmeleta (approved):

Guillaumegranger1:
Yes! 🚢🚀

PR #394, opened 2021-08-11
Fix purchase flow for IE11

### LINKED ISSUE #66: [Compatibility] Figure out the form attribute for IE11
## Purpose

We are using the `form` attribute on `<input>` elements in a few spots to facilitate sections everywhere - the cart and the product page so far, and I'm sure its use will be broadened over time as we continue to discover new use cases with Sections Everywhere.

The issue is that it is not supported on IE11, which still needs basic functionality like adding to cart and checkout. [caniuse](https://caniuse.com/?search=form%20attribute) gives the rundown.

I don't think this should stop us from using it, but it might mean we need a polyfill which we had, in principle, decided were evil and not to be used.

tyleralsbury:
**Note** - I'm making this as a PR rather than a draft to get 👁️ 👄 👁️ on it.  I only tested the case with variants and expect it might be broken on non-variant products. Since I'll be off for a little while, hopefully someone can pick this up while I'm gone.

Fixes #66.

While we don't explicitly support IE11, we want to make sure that if a buyer insists on using it, they can still add items to their cart and make their way to checkout.

The issue is that the `form` attribute on an input doesn't work in IE.  Originally, I was going to try a polyfill, but looking around they were pretty much all very complex and also relied on jQuery, so instead I went with the approach of just trying to make sure that the form itself got the inputs put into them instead.

To do this, I check if it's IE, and then if it is, I do some DOM manipulation to ensure that the form is complete when they try to submit it.

It's not the most beautiful approach, but it works.

**Other considerations**

I put this together pretty quickly, so there may be bugs.  Some thoughts...

* This doesn't currently scale to work with Featured product.  It probably needs another pass after that is merged if we want Featured product's add to cart to work as well.  
* I didn't test this on a product with no variants. We probably need an early return for that.
* This, of course, doesn't work JS disabled on IE but I think that's a 100% hopeless cause.  I can't imagine a way for us to support IE11 without JS.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=124598386710)

**Testing cases**

Affected pages - Cart and Product

- [ ] IE - with variants
- [ ] IE - without variants
- [ ] Other browsers - with variants
- [ ] Other browsers - without variants

**Expected behaviour** - The site will not look very good, but you can make your way to a product page, add to cart (it will just add it, not AJAX), and then click Checkout and end up on the Checkout page with the correct items.  You can make changes to the variant on the product page and changes to the quantity on the cart and they will submit properly.

**Checklist**

### DIFF
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -235,7 +235,7 @@
               {%- endif -%}
             {%- endunless -%}
 
-            <noscript>
+            <noscript class="product-form__noscript-wrapper">
               <div class="product-form__input{% if product.has_only_default_variant %} hidden{% endif %}">
                 <label class="form__label" for="Variants-{{ section.id }}">{{ 'products.product.product_variants' | t }}</label>
                 <div class="select">
@@ -449,6 +449,21 @@
 
   customElements.define('product-modal', ProductModal);
 {% endjavascript %}
+
+<script>
+  document.addEventListener('DOMContentLoaded', function() {
+    if (!isIE()) return;
+    const hiddenInput = document.querySelector('product-form input[name="id"]');
+    const noScriptInputWrapper = document.createElement('div');
+    const variantSwitcher = document.querySelector('variant-radios') || document.querySelector('variant-selects');
+    noScriptInputWrapper.innerHTML = document.querySelector('.product-form__noscript-wrapper').textContent;
+    variantSwitcher.outerHTML = noScriptInputWrapper.outerHTML;
+
+    document.querySelector('select[name="id"]').addEventListener('change', function(event) {
+      hiddenInput.value = event.currentTarget.value;
+    });
+  });
+</script>
 
 {%- if first_3d_model -%}
   <script type="application/json" id="ProductJSON-{{ product.id }}">
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -452,6 +452,14 @@
 
 <script>
   document.addEventListener('DOMContentLoaded', function() {
+    function isIE() {
+      const ua = window.navigator.userAgent;
+      const msie = ua.indexOf('MSIE ');
+      const trident = ua.indexOf('Trident/');
+
+      return (msie > 0 || trident > 0);
+    }
+
     if (!isIE()) return;
     const hiddenInput = document.querySelector('product-form input[name="id"]');
     const noScriptInputWrapper = document.createElement('div');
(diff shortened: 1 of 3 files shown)


tauthomas01 (commented):
This works well. I tested with and without variants and I can add any products to the cart.

I'm wondering if we should fix the quantity input in the product page.

ludoboludo:
The whole IE11 experience is brutal: [video](https://drive.google.com/file/d/1l2KCI-S0HfenmNA8BfDr6D9CocfbK1H6/view) 
I didn't realize the password field was there already so took a few min using the inspect too. And then when I get on the product page all videos start to play when the page is fully loaded.

Other wise you can get to checkout if you have enough patience to get and are able to see what you're buying.

sections/main-product.liquid
```diff
@@ -450,6 +450,29 @@
...
+    function isIE() {
+      const ua = window.navigator.userAgent;
+      const msie = ua.indexOf('MSIE ');
+      const trident = ua.indexOf('Trident/');
+
+      return (msie > 0 || trident > 0);
+    }
+
+    if (!isIE()) return;
+    const hiddenInput = document.querySelector('product-form input[name="id"]');
```
ludoboludo, line 464: to be able to copy it over to feat. product I think i'd need to change this selector to be more specific to the section on which it's added. 

So i'm thinking:
```javascript
document.querySelector('#product-form-template-{{ section.id }} input[name="id"]');
```

sections/main-product.liquid
```diff
@@ -450,6 +450,29 @@
...
+    const noScriptInputWrapper = document.createElement('div');
+    const variantSwitcher = document.querySelector('variant-radios') || document.querySelector('variant-selects');
+    noScriptInputWrapper.innerHTML = document.querySelector('.product-form__noscript-wrapper').textContent;
+    variantSwitcher.outerHTML = noScriptInputWrapper.outerHTML;
+
+    document.querySelector('select[name="id"]').addEventListener('change', function(event) {
```
ludoboludo, line 470: And here switch it to:
```javacript
document.querySelector('#Variants-{{ section.id }}').
```

ludoboludo:
I added the same code over onto the featured-product section. Ideally we wouldn't repeat the code (event listeners, etc) 🤔

But for now it's done so that you can add the item to the cart and go to checkout from a feat. product section, main product page and even from a page with multiple product forms. 

Ready for review, I'm happy to change anything to reduce the amount of lines.

ludoboludo (approved):

chrisberthe (approved):

PR #412, opened 2021-08-16
Correcting support content in en.default.schema.json

gretchen-willenberg:
**Why are these changes introduced?** Fixes 2 small copy tense changes in sections. See `Heading required to display the menu` and `Video plays in a pop-up if the section contains other blocks.`

See  Header block and Collage section for these strings. 

- [Editor](https://os2-demo.myshopify.com/admin/themes/124615655446/editor)


melissaperreault (approved):
🎉 🎉

PR #424, opened 2021-08-17
Click to dismiss product modal

### LINKED ISSUE #377: Add click-to-dismiss to images in product page media modal
Add click to dismiss on images in the product page media modal.  Clicking anywhere on the image or the margins should close the modal. 

Notes:
- Only for clicks (desktop), not for touch (mobile/table) 
- Only for images. Clicking videos/ 3d should play the file. 

https://screenshot.click/21-24-wqpn8-126pc.mp4

kmeleta:

Fixes #377 

To exclude touch events from mobile devices, I'm using `PointerEvent.pointerType`. Support for the `PointerEvent` interface looks good enough for us to utilize. Probably a minor detail, but worth noting I'm using the `pointerup` event. It seemed jarring for the modal to close immediately on mousedown.

To exclude media other than images from triggering a dismiss, I'm targeting img elements that have the `data-media-id` attribute. Deferred media types also include a poster img that would otherwise trigger the event.

**Other considerations**

Currently, clicks in the white space around the modal content will also close the modal, but I'm not sure if this is expected behavior or not. If this were a more standard modal with an overlay, clicking the overlay would definitely be expected to close the dialog, but clicking the background of the dialog itself (like it is here) probably wouldn't be. Just curious if anyone has any opinions on that.

I added this feature to our `ModalDialog` class, but do we need to limit the functionality to only the `ProductModal` (which extends ModalDialog)? 

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -378,10 +378,17 @@ class ModalDialog extends HTMLElement {
       'click',
       this.hide.bind(this)
     );
+    this.querySelectorAll('img[data-media-id], [role="document"]').forEach(element => {
+      element.addEventListener('pointerup', (event) => {
+        if (event.pointerType === 'mouse' && event.target === event.currentTarget) {
+          this.hide();
+        }
+      })
+    });
     this.addEventListener('click', (event) => {
       if (event.target.nodeName === 'MODAL-DIALOG') this.hide();
     });
-    this.addEventListener('keyup', () => {
+    this.addEventListener('keyup', (event) => {
       if (event.code.toUpperCase() === 'ESCAPE') this.hide();
     });
   }


assets/global.js
```diff
@@ -378,10 +378,17 @@ class ModalDialog extends HTMLElement {
       'click',
       this.hide.bind(this)
     );
+    this.querySelectorAll('img[data-media-id], [role="document"]').forEach(element => {
```
kmeleta, line 381: These selectors make sense to me, but definitely open to other ideas.

ludoboludo: I wonder if we should then add `role="document"` to the video modal on the collage section so it behaves the same way. I think that would make sense for continuity in the behaviour. 

From what I can see only two `moda-dialog` elements are used, on `main-product.liquid` and `collage.liquid`.

ludoboludo (commented):
I'd like to question the behaviour. 
Clicking on the image to close the modal seems a bit odd to me. Maybe I just haven't visited websites that do this yet but it seems a bit counter intuitive. Clicking outside of the image on the background (white by default) to close the modal seems more appropriate based on other experiences I've encountered. 
cc: @Guillaumegranger1 @melissaperreault  @nicklepine 

And then as a follow up to this maybe, we could look into the overall experience of drawers and modal. I feel like it varies a little too much right now:
- Filter drawer, uses a darkened backdrop that you can click on to close the drawer. 
- Popup link, uses a darkened backdrop that you can click on to close the popup. 
- Pickup availability, doesn't have a darkened backdrop but you can click outside of the drawer to close it.
- Menu nav, doesn't have a darkened backdrop and you can't click outside of it to close the drawer.

kmeleta:
> Clicking on the image to close the modal seems a bit odd to me

Expanding on what @ludoboludo mentioned, I won't say it's common, but I have seen this pattern used for similar use cases to this, where it's less of a content modal and more of an image lightbox. I think being able to get a quick closer look at a product image and then dismiss it without even having to reposition your cursor is kind of nice (particularly from a motor impairment pov). I'm just not convinced users will discover or expect the behavior. We use the `zoom-in` cursor on the image list, maybe we could utilize the `zoom-out` cursor when the modal is open?

Guillaumegranger1:
> And then as a follow up to this maybe, we could look into the overall experience of drawers and modal. I feel like it varies a little too much right now:
>
> Filter drawer, uses a darkened backdrop that you can click on to close the drawer.
Popup link, uses a darkened backdrop that you can click on to close the popup.
Pickup availability, doesn't have a darkened backdrop but you can click outside of the drawer to close it.
Menu nav, doesn't have a darkened backdrop and you can't click outside of it to close the drawer.

I agree there are some UX inconsistencies. Perhaps not a priority for now, but I wouldn't mind revisiting this at some point. the `darkened backdrop` and `clicking outside to dismiss` could be applied on the patterns mentioned above if there where no major technical limitation. As for the product image zoom, I wouldn't mind the darkened backdrop there either but I have less context on whether it has been considered and rejected for good reasons.

melissaperreault (dismissed):
> Currently, clicks in the white space around the media will also close the modal, but I'm not sure if this is expected behavior or not.

Tested on both my desktop and mobile and works as expected! Thanks Ken!

assets/global.js
ludoboludo: Let me know if this shouldn't be happen anyway and we can disregard my comment

sections/collage.liquid
```diff
@@ -313,18 +313,20 @@
...
-                  <deferred-media class="collage-video__modal-video template-popup">
-                    <template>
-                      {%- if block.settings.video_url.type == 'youtube' -%}
-                        <iframe src="https://www.youtube.com/embed/{{ block.settings.video_url.id }}?enablejsapi=1" class="js-youtube" allow="autoplay; encrypted-media" allowfullscreen title="{{ block.settings.description | escape }}"></iframe>
-                      {%- else -%}
-                        <iframe src="https://player.vimeo.com/video/{{ block.settings.video_url.id }}" class="js-vimeo" allow="autoplay; encrypted-media" allowfullscreen title="{{ block.settings.description | escape }}"></iframe>
-                      {%- endif -%}
-                    </template>
-                  </deferred-media>
+                  <div class="collage-video__modal-content-info">
```
kmeleta, line 319: I had to move the padding-top that was on the deferred media element somewhere else so it wouldn't get excluded from the dismiss logic. The other 2 modaldialogs have some form of this content wrapper here so I figured I'd match that structure.

ludoboludo (dismissed):
Works well when tested 👌 

I noticed another thing which isn't related to this PR but in the `DeferredMedia` class in `global.js` we could move `window.pauseAllMedia();` at the top of the `loadContent()` to make sure it pauses the other videos when opening the modal.  Otherwise it only pauses them the first time you open the modal but not the subsequent times. 

Could be done in a separate issue tho 🤔

chrisberthe (dismissed):

melissaperreault (approved):
The Modal cursor updates for closing works super well! 👌 🎉

assets/global.js
```diff
@@ -422,12 +428,12 @@ class DeferredMedia extends HTMLElement {
   }
 
   loadContent() {
+    window.pauseAllMedia();
```
kmeleta, line 431: @ludoboludo moving this out of the loaded conditional below seems to fix the issue like you suggested. Let me know if this isn't what you had in mind.

assets/global.js
```diff
@@ -378,12 +378,18 @@ class ModalDialog extends HTMLElement {
...
       if (event.code.toUpperCase() === 'ESCAPE') this.hide();
     });
+    if (this.classList.contains('media-modal')) {
+      this.addEventListener('pointerup', (event) => {
+        if (event.pointerType === 'mouse' && !event.target.closest('deferred-media, product-model')) this.hide();
+      });
+    } else {
+      this.addEventListener('click', (event) => {
+      if (event.target.nodeName === 'MODAL-DIALOG') this.hide();
+    });
```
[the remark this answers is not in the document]
kmeleta: :doh:

assets/global.js
ludoboludo: Nope that's what I was thinking too 👌

ludoboludo (approved):

assets/global.js
```diff
@@ -378,12 +378,18 @@ class ModalDialog extends HTMLElement {
...
       this.hide.bind(this)
     );
-    this.addEventListener('click', (event) => {
-      if (event.target.nodeName === 'MODAL-DIALOG') this.hide();
-    });
-    this.addEventListener('keyup', () => {
+    this.addEventListener('keyup', (event) => {
       if (event.code.toUpperCase() === 'ESCAPE') this.hide();
     });
+    if (this.classList.contains('media-modal')) {
```
LucasLacerdaUX, line 384: Not sure how I feel about using class conditions for logic. Is there any other use of this on the theme that you can remember?
 
I'm wondering if that wouldn't be a good use for `data-` attributes instead.

LucasLacerdaUX (approved):
Working as expected. 

Just a minor code design related question.

PR #451, opened 2021-08-19
Update webkit autofill styles for field labels/buttons

### LINKED ISSUE #422: Email Sign Up Section Bugs
image

1. When using autofill to fill out the email form, we lose the submit arrow. Seems to be working fine on the email form in footer, but not in the email form on the email sign up section. When using autofill, I should still see the submit arrow icon. 

2. When submitting my email the success message is misaligned. The success (and error) messages should look aligned with the email form. 

Recording: https://screenshot.click/21-24-7x30e-ayyhj.mp4

kmeleta:

Fixes #422 

The problem comes from webkit browsers forcing different styling on inputs when autofilled. This is more notable when the input is on a dark background with light text. It doesn't appear we can override these styles with !important so the best option is to adjust the inline label and submit button icon to just play nice with these forced styles. The browser changes the input text to pure black so I've updated the label and button to display black as well. [See here](https://screenshot.click/19-51-ptueq-olvf8.png)

The issue was noticed on the email signup section, but I made this a global change in base.css because this could occur wherever there is a form and a dark background.

Note: These changes should only apply to webkit browsers and to inputs that have been autofilled.

**Other considerations**

There _is_ a potential hack we could use to hide the browser's styles, which involves utilizing the `box-shadow` css property to create a faux background, but we currently use this property to form the border/outline of inputs, so we'd have to change that, at least for this case. I don't think it would be worth the effort and could be fragile, but if anyone _really_ hates the browser styles we can discuss.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126268375062)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126268375062/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1466,6 +1466,12 @@ details[open] > .share-button__fallback {
   width: 2.5rem;
 }
 
+.field__input:-webkit-autofill ~ .field__button,
+.field__input:-webkit-autofill ~ .field__label,
+.customer .field input:-webkit-autofill ~ label {
+  color: rgb(0,0,0);
+}
+
 /* Text area */
 
 .text-area {


sofiamatulis (approved):
Looks good 🎉 I will let a designer confirm if black is the preferred colour considering that in different browsers it wont be the blue background. In safari it's yellow: https://screenshot.click/24-01-pyjli-nk6ge.png

kmeleta:
> if black is the preferred colour considering that in different browsers it wont be the blue background

Thanks @sofiamatulis! Design should definitely sign off (cc: @melissaperreault) but I think we're more or less stuck with black. Safari's background changes to light yellow, but I believe it forces the input text color to black just as chrome does. That's really what made the decision for me.

melissaperreault:
We can definitely try to scale the explorations of the hack in the future if we don't want these kind of outcome on darker theme scheme [as shown here](https://screenshot.click/21-08-c67vb-qh0gk.png).

PR #489, opened 2021-08-25
Remove all instances of optional chaining

### LINKED ISSUE #266: More customers saying they cannot order

I have an iPhone 6 plus, am not 100 percent sure that safari is the latest but am 100 percent sure that chrome is. And I updated the actual phone yesterday as well. You can search for items but are not allowed to order them or see the overview of what kind of products you have or if they are sold out. Got to order via PC.
She I know has a Samsung A71 which is a newer type (another friend that  cannot order either)

- Dawn Version: 1.0.0
- macOS Version 11.3.1

**Possible solution**

Add any other context about the problem here. If applicable, add screenshots to help explain.

This is a customer saying they and their friend cannot order from their phone.

### LINKED ISSUE #487: Remove optional chaining
Dawn is using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). This is supported in safari 13.1 and later. 

It is causing JS to break and prevents buyers from adding items to their cart and checking out. 
Since we officially need to support the latest two versions of safari, this needs to be removed. 

Related issue https://github.com/Shopify/dawn/issues/266

- Dawn Version: 1.1.0 (all versions)
- Safari (versions < 13.1)

**Possible solution**
Remove optional chaining from Dawn

martinamarien:
We have received some issues related to customers being unable to add items to their cart and checkout.
After looking into this a bit, it seems that [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is likely the reason. 

I am able to replicate this on versions of safari `< 13.1`  

According to our [shopify theme requirements](https://shopify.dev/themes/store/requirements), we require that our themes support the latest two versions of safari in Mac and iOS. Safari's current version is 14.1. Even if we did not need to support it, reverting this change is easy and allows for compatibility with older browser versions without the need of pollyfils, so it makes sense to make this change.

Fixes #266
Fixes #487 

I removed all instances of `?.` in Dawn. Now, we either check if the element is present first, and/or create a variable for the element. 

```js
this.header?.doSomething()
```
becomes
```js
if (this.header) this.header.doSomething()  
```

And
```js
document.getElementById('section-header')?.doSomething()
```
Becomes
```js
const header = getElementById('section-header');
if (header) header.doSomething();
```

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126282825750)

**Testing** 
- [ ] Test every feature that was changed 
- [ ] Test on current versions _and_ Safari 12.1
- [ ] Can customers add to cart? 

**Checklist**

### DIFF
--- a/assets/product-model.js
+++ b/assets/product-model.js
@@ -53,4 +53,6 @@ window.ProductModel = {
   },
 };
 
-window.addEventListener('DOMContentLoaded', () => { if (window.ProductModel) window.ProductModel.loadShopifyXR(); });
+window.addEventListener('DOMContentLoaded', () => { 
+  if (window.ProductModel) window.ProductModel.loadShopifyXR();
+});
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -472,10 +472,10 @@
       activeMedia.scrollIntoView();
 
       const container = this.querySelector('[role="document"]');
-      container.scrollLeft = (activeMedia.width - container.clientWidth) / 2; 
+      container.scrollLeft = (activeMedia.width - container.clientWidth) / 2;
 
       if (activeMedia.nodeName == 'DEFERRED-MEDIA' && activeMediaContent && activeMediaContent.querySelector('.js-youtube'))
-        activeMedia.loadContent(); 
+        activeMedia.loadContent();
     }
   }
 


PaulNewton:
⏳  Future note: specifically for  **liquid templates** (not assets).
 Before Optional Chaining gets reimplemented there's an argument first to consider that Optional Chaining is an **anti-pattern** in the **liquid templates**, of shopify themes. What it's doing is choosing programmers syntactic sugar over readable verbosity; particularly through the lenses of support troubleshooting and the merchant perspective.

What programmer has never missed a dot(.) ?
Even a seemingly simple <kbd>❔</kbd> question mark increases the complexity merchants need to successfully interpret a programmers single-line-of-code "elegance",  and be able to synthesize solutions from these inline functional approaches versus a procedural approach that's following a familiar **pattern** of assigned element variables in the javascript that's similar to the more verbose liquid used in template files.  

Add yet another symbol that merchants, or programmers, can miss in typing or reading... so now standalone lines as relatively simple as `document.querySelector('form')?.submit();` instantaneously become social land mines:
`document.querySelector('form').submit();`
(_for example if a setting turns the form off but not the javascript 💣 💥 !_) 

On the flip side of this _False positive_ Optional chaining can **silently fail** which compounds common troubleshooting steps for templates with even simple typos(either in the javascript or the html).
`document.querySelector('farm')?.submit();` <-- does **not** blow up is that good or bad?  😭 

Given the above vagaries it could be argued further without a robust testing process as part of the dawn repo that Optional Chaining should only be allowed in build processes, if at all.

💭 _In javascript **assets** this consideration could ignored as long as assets are consistent in using Optional Chaining over variable assignment and null checks, intermittent usage should probably be avoided and enforced with linting from [theme check](https://github.com/Shopify/theme-check) being updated with Optional Chaining rules._

PR #509, opened 2021-08-27
Fix line item properties when adding a product to cart

tauthomas01:

Fixes https://github.com/Shopify/dawn/issues/243

The goal of this PR is to add line item properties when adding product to the cart. Previously, it was not possible due to the fact that `properties` object was passing as an `Array` structure instead of an `Object`.

**Actual**
```
"properties[engraving]": "Happy birthday",
"properties[isMatte]": "on"
```

**Expected**
```
"properties": {
   "engraving": "Happy birthday",
   "isMatte": "on"
}
```

image

There are other different approaches such as https://github.com/Shopify/dawn/pull/257 and https://github.com/Shopify/dawn/pull/163 but this approach is saving a few lines of JS.

Also, it is normal that we can't see line item properties in the cart notification. It's missing some Liquid code. There is a PR already made here: https://github.com/Shopify/dawn/pull/195

- Detect if `properties[]` field exist before sending the request to the server by using regex

Note that this doesn’t work is uploaded files (`<input type="file" />`): I assume due to the `Content-Type` being set as 'Content-Type': 'application/json', the form expects to receive only JSON format data, but when I try to switch to `Content-Type: multipart/form-data` in the fetch config (which is the way to upload files), it doesn’t work. I believe because we’re mixing JSON file and uploaded files, but I could be wrong.

If I disable JS and rely on the native form, the form will submit on the `<form>` element which has `enctype="multipart/form-data"` so the browser is natively taking care of everything, and I see uploaded files on the cart

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126290919446)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126290919446/editor)

**Checklist**

**Testing**
1. Checkout the branch
2. Add the following in the `main-product.liquid` form

```
<div class="field">
  <input 
    type="checkbox"
     id="isMatte" 
    name="properties[isMatte]"
    form="product-form-{{ section.id }}"
  >
  <label class="field__label" for="isMatte">Matte option</label>
</div>

<div class="field">
  <input class="field__input"
    type="text"
    id="engraving" 
    name="properties[engraving]"
    form="product-form-{{ section.id }}"
  >
  <label class="field__label" for="engraving">Engraving</label>
</div>
```

3. Add the product and make sure we see line item properties.

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -108,7 +108,7 @@ const serializeForm = form => {
   const formData = new FormData(form);
 
   for (const key of formData.keys()) {
-    const regex = /(?:^(properties\[))(.*?)(?:\])/;
+    const regex = /(?:^(properties\[))(.*?)(?:\]$)/;
 
     if (regex.test(key)) { 
       obj.properties = obj.properties || {};
--- a/assets/global.js
+++ b/assets/global.js
@@ -106,9 +106,18 @@ function debounce(fn, wait) {
 const serializeForm = form => {
   const obj = {};
   const formData = new FormData(form);
+
   for (const key of formData.keys()) {
-    obj[key] = formData.get(key);
+    if(key.indexOf('properties[') !== -1) {
+      obj.properties = obj.properties || {};
+      const lineItem = key.split('[');
+      const propertyKey = lineItem[1].substring(0, lineItem[1].length-1)
+      obj.properties[propertyKey] = formData.get(key);
+    } else {
+      obj[key] = formData.get(key);
+    }
   }
+
   return JSON.stringify(obj);
 };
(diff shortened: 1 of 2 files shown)


LucasLacerdaUX (approved):
Tested and working on both featured product and product sections.

To support `input type="file"`, I think we may need to drop the `FormData` -> `JSON` conversion, if I remember correctly. However, it would cause some incompatibility issues with IE and we'd need to make sure nothing else is broken.

For now, this should be good enough :)

thedeadhand:
This worked great. thank you so much.

PR #544, opened 2021-09-01
Update translation.yml to make all languages blocking

sjahu:

We don't want to allow PRs to be merged before having translations for all languages completed. This config change makes all languages blocking in the branch workflow.


martinamarien (approved):

gregjotau:
I would actually like featured to just be launched as soon as possible, but with 1 exception: You guys need to implement english as a default if a translation is missing.

We have Norwegian as the main language in our store, but we want english to be the default for all other domains than famme.no. If english words showed up instead of Norwegian if a translation is missing it would not be a problem.

PR #563, opened 2021-09-03
Replace ResizeObserver by Resize event

### LINKED ISSUE #510: Safari 12.1 support issues

Some of the things we're using aren't supported in Safari 12.1: 
- ResizeObserver, used for the sliders.
- IntersectionObserver, used for the product recos and the header ([specifically Safari on iOS](https://screenshot.click/21-08-3exet-knbgi.png))

Which breaks some functionalities: 
- local selector in the footer (not opening),
- product modal (not opening),
- cart notification (not showing up)

Ideally all those features mentioned above would work as expected. 

- Dawn Version: 1.0.0
- macOS Version 10.14
- Safari version 12.1

**Possible solution**

When removing mentions of the `ResizeObserver` and `IntersectionObserver` it seem to work as expected. But we didn't look into how to replace what they're doing with a fallback. 

ludoboludo:

Fixes #510 

Replaced the `ResizeObserver` by a `Resize` event which applies to the window instead of just the slider element. 

**Other considerations**

I still need to look into the `IntersectionObserver` but when I tested on a laptop that has Safari 12.1.2 and browserstack iOS 12, I didn't get errors like I did before and language selector, cart notification, slider and product modal do work. 

From what I'm seeing it's just the sticky header and product recos that do not work. So product recos never show up and the header behave like a non sticky one, stays at the top of the page at all time, so you have to scroll back up to get to it. 

Would this be acceptable enough ? I think it prevents us from having too many work arounds/extra code and offer a decent experience on those older browser versions. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126328111126)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126328111126/editor)

**Checklist**

**Things to test**
Use browserstack to test iOS 12 and the Mojave OS with version 12 for Safari.
- local selector in the footer,
- product modal,
- cart notification,
- and general functionality of the theme

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -156,7 +156,7 @@ const serializeForm = form => {
   for (const key of formData.keys()) {
     const regex = /(?:^(properties\[))(.*?)(?:\]$)/;
 
-    if (regex.test(key)) { 
+    if (regex.test(key)) {
       obj.properties = obj.properties || {};
       obj.properties[regex.exec(key)[2]] = formData.get(key);
     } else {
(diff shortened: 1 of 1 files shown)


ludoboludo:
You're right @michaelburtonray 👌  thanks for bringing it up. I will close the PR along with the issue.

PR #577, opened 2021-09-07
Description defaults to true for collection

tyleralsbury:

Very quick PR to make the description on by default on the collection page.  Discussed with @Oliviammarcello and Shawn Rudolph.    This is good for SEO as well as for merchant expectations - they don't need to go to another place to see the collection description when they add one to their collection.

Switched a boolean 🤯 

**Checklist**

### DIFF
--- a/sections/main-collection-banner.liquid
+++ b/sections/main-collection-banner.liquid
@@ -46,7 +46,7 @@
     {
       "type": "checkbox",
       "id": "show_collection_description",
-      "default": false,
+      "default": true,
       "label": "t:sections.main-collection-banner.settings.show_collection_description.label"
     },
     {


sections/main-collection-banner.liquid
```diff
@@ -46,7 +46,7 @@
     {
       "type": "checkbox",
       "id": "show_collection_description",
-      "default": false,
+      "default": true,
       "label": "t:sections.main-collection-banner.settings.show_collection_description.label"
     },
     {
```
sofiamatulis, line 52: I just noticed that even if the setting is on, but the string is empty it still renders the `div`

https://screenshot.click/07-41-0tkji-x4y83.png

Should we also be checking the length of the content?

```
      {%- if section.settings.show_collection_description and  collection.description.length > 0 -%}
        <div class="collection-hero__description rte">{{ collection.description }}</div>
      {%- endif -%}
```

sofiamatulis (approved):
Nice 🚢 Just added one comment

sections/main-collection-banner.liquid
tyleralsbury: That should be fine, since we have `div:empty { display: none }`.  This way, we can hide it without adding complexity to the Liquid.

martinamarien (approved):

Issue #380, opened 2021-08-11
Cart Page: Inconsistent error messaging for low item stock


clyde-baker:
Bug description:

When a buyer attempts to increase the quantity of a line item on the cart page, to a number that exceeds the limit of inventory available, it will reset to max quantity available with no explanation as to what just happened. Only after trying to increase the quantity a second time does an error message display.

The error message should display every time a buyer tries to enter a quantity that exceeds the available limit

https://screenshot.click/21-43-1vk8m-krbl2.mp4

_____

Original Comment from @clyde-baker: 

I haven't seen a fix for this yet but it's a UX disaster that will surely lead to a large number of abandonments.

Users are able to add an unlimited number of items to their cart from the product page, even if there is only 1 in stock.

Only once the user initiates checkout does an error appear stating "Out of stock. Some items are no longer available. Your cart has been updated." Unfortunately, waiting until a user has already committed to checking is way too late. And even if you click "Return to cart", the quantity doesn't actually update and remains at the impossible number.

In addition, if the user attempts to increase the impossible quantity from the cart page, it will reset to lowest quantity available with no explanation as to what just happened. Only after trying to increase the quantity a second time does an error display.

Like with the Debut theme, users should be prevented from adding the impossible quantity to cart and immediately given an error message.

- Dawn Version: 1.1.0
- Chrome Version 92.0.4515.131

ludoboludo:
Here is what I'm observing to add some details to this issue: [video](https://screenshot.click/21-09-kle3k-v1g4q.mp4)

ludoboludo:
There is also an odd issue in Safari brought up here by @Binthout: https://github.com/Shopify/dawn/pull/582#issuecomment-920268987

Here is some early observations: 
>Interesting, from what I'm seeing it's working properly when navigation/updating the quantity via keyboard navigation. But when it's using the mouse, it doesn't detect anything for `document.activeElement.getAttribute('name')` from the `onChange(event)` function. 

>So `if (lineItem) lineItem.querySelector(`[name="${name}"]`).focus();` ends up failing and throwing an error which shows up in the video 🤔

ludoboludo:
Looks like it's specific to Mac. `activeElement` will only return the input if it's a `type="text"` otherwise, focus will be set to the Body or `null`. 

>Note: For iOS, elements that are not <input type="text">, will not get focus assigned to them. The property will then return BODY or null.

[...]

PR #582, opened 2021-09-07
Fix add to cart error messaging

### LINKED ISSUE #398: Product Page: Add quantity limit error messaging
In cart, if a buyer tries to add a quantity of a product that is higher than the available inventory, the buyer is served an error message "You can only add X of this item to your cart" and the quantity gets auto-updated to the quantity limit.

On product pages however, we currently let the buyer add a quantity that exceeds the quantity limit for a specific variant. 
We should introduce the same UX pattern of checking against the quantity limit and serving an error message, and updating the quantity on the quantity selector if a buyer does input a quantity that exceeds the limit. 

Like with the Debut theme, users should be prevented from adding an impossible quantity to cart and immediately given an error message.

https://screenshot.click/21-32-rwwjn-k53pr.mp4

martinamarien:
This PR adds error messaging to product pages. 
This will prevent buyers from adding more items to their cart, than are available.

Fixes #398.

Add the `X-Requested-With` header to allow error messages to be returned. 

```
config.headers['X-Requested-With'] = 'XMLHttpRequest';
```
The error message was added to products and featured products. 

Show error message if:
1. Variant has  insufficient stock - after Add to cart. 

If error message is visible. Hide the message if: 
1. Variant is changed 
2. Item is added to cart again

**Other considerations**
N/A 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126344298518)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126344298518/editor)

**Testing** 
- [ ] If you change variants, the message disappears 
- [ ] If you add a quantity that is more than is available, show error message 
- [ ] Test featured product and product 
- [ ] If an error message is present and you submit form, hide error message 
- [ ] Test with dropdown and button variant types. 

**Checklist**

### DIFF
--- a/assets/product-form.js
+++ b/assets/product-form.js
@@ -10,6 +10,7 @@ if (!customElements.get('product-form')) {
 
     onSubmitHandler(evt) {
       evt.preventDefault();
+      this.handleErrorMessage();
       this.cartNotification.setActiveElement(document.activeElement);
 
       const submitButton = this.querySelector('[type="submit"]');
--- a/assets/product-form.js
+++ b/assets/product-form.js
@@ -17,16 +17,23 @@ if (!customElements.get('product-form')) {
       submitButton.setAttribute('disabled', true);
       submitButton.classList.add('loading');
 
-      const body = JSON.stringify({
+      var config = fetchConfig('javascript');
+      config.headers['X-Requested-With'] = 'XMLHttpRequest';
+      config.body = JSON.stringify({
         ...JSON.parse(serializeForm(this.form)),
         sections: this.cartNotification.getSectionsToRender().map((section) => section.id),
         sections_url: window.location.pathname
       });
 
-      fetch(`${routes.cart_add_url}`, { ...fetchConfig('javascript'), body })
+      fetch(`${routes.cart_add_url}`, config)
         .then((response) => response.json())
-        .then((parsedState) => {
-          this.cartNotification.renderContents(parsedState);
+        .then((response) => {
+          if (response.status) {
+            this.handleErrorMessage(response.description);
+            return;
+          }
+
+          this.cartNotification.renderContents(response);
         })
         .catch((e) => {
           console.error(e);
@@ -35,6 +42,17 @@ if (!customElements.get('product-form')) {
           submitButton.classList.remove('loading');
           submitButton.removeAttribute('disabled');
         });
+    }
+
+    handleErrorMessage(errorMessage = false) {
+      this.errorMessageWrapper = this.errorMessageWrapper || this.querySelector('.product-form__error-message-wrapper');
+      this.errorMessage = this.errorMessage || this.errorMessageWrapper.querySelector('.product-form__error-message');
+
+      this.errorMessageWrapper.toggleAttribute('hidden', !errorMessage);
+
+      if (errorMessage ) {
+        this.errorMessage.textContent = errorMessage;
+      }
     }
   });
 }
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -110,6 +110,19 @@
   display: block;
 }
 
+.product-form__error-message-wrapper:not([hidden]) {
+  display: flex;
+  font-size: 1.2rem;
+  margin-bottom: 1.5rem;
+}
+
+.product-form__error-message-wrapper svg {
+  flex-shrink: 0;
+  width: 1.2rem;
+  margin-right: 0.7rem;
+  margin-top: 0.1rem;
+}
+
 /* Form Elements */
 .product-form__input {
   flex: 0 0 100%;
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -112,6 +112,7 @@
 
 .product-form__error-message-wrapper:not([hidden]) {
   display: flex;
+  align-items: center;
   font-size: 1.2rem;
   margin-bottom: 1.5rem;
 }
@@ -119,6 +120,7 @@
 .product-form__error-message-wrapper svg {
   flex-shrink: 0;
   width: 1.2rem;
+  height: 1.2rem;
   margin-right: 0.7rem;
   margin-top: 0.1rem;
 }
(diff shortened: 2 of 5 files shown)


sections/featured-product.liquid
```diff
@@ -229,6 +229,17 @@
               <div {{ block.shopify_attributes }}>
                 {%- if product != blank -%}
                   <product-form class="product-form">
+                    <div class="product-form__error-message-wrapper" role="alert" hidden>
+                      <span class="visually-hidden">{{ 'accessibility.error' | t }} </span>
```
ludoboludo, line 233:
```suggestion
                      <span class="visually-hidden">{{ 'accessibility.error' | t }}</span>
```

svinkle, line 233: Outputting "Error" isn't required. VO includes a _chime_ when the message is announced. Windows screen readers include, _"Alert"_. Safe to remove this.

sections/main-product.liquid
```diff
@@ -264,6 +264,17 @@
           {%- when 'buy_buttons' -%}
             <div {{ block.shopify_attributes }}>
               <product-form class="product-form">
+                <div class="product-form__error-message-wrapper" role="alert" hidden>
+                  <span class="visually-hidden">{{ 'accessibility.error' | t }} </span>
```
svinkle, line 268: Same here.

svinkle:
I just notice when testing with Windows screen readers, focus is lost and moves to the document body after clicking "Add to cart." This is disrupting the error announcement. Can re refrain from shifting focus from the `button`?

Dawn product page with Chrome DevTools open. NerdeFocus extension is open, its log showing focus moving to the body element after clicking the Add to Cart button.

assets/product-form.js
```diff
@@ -10,23 +10,31 @@ if (!customElements.get('product-form')) {
...
+      this.handleErrorMessage();
       this.cartNotification.setActiveElement(document.activeElement);
 
       const submitButton = this.querySelector('[type="submit"]');
 
       submitButton.setAttribute('disabled', true);
       submitButton.classList.add('loading');
 
-      const body = JSON.stringify({
+      var config = fetchConfig('javascript');
```
tauthomas01, line 21: Should this be `const config` instead of `var config`?

assets/product-form.js
```diff
@@ -10,23 +10,31 @@ if (!customElements.get('product-form')) {
...
+      config.headers['X-Requested-With'] = 'XMLHttpRequest';
+      config.body = JSON.stringify({
         ...JSON.parse(serializeForm(this.form)),
         sections: this.cartNotification.getSectionsToRender().map((section) => section.id),
         sections_url: window.location.pathname
       });
 
-      fetch(`${routes.cart_add_url}`, { ...fetchConfig('javascript'), body })
+      fetch(`${routes.cart_add_url}`, config)
```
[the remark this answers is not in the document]
tauthomas01: From Ludo's video in the cart, I was trying to debug and and fix using JS but after some trials and errors, it looks like we have to use inline script `onsubmit="return false"` on the `<form>` element :

```
 <form action="{{ routes.cart_url }}" class="cart__contents critical-hidden" method="post" id="cart" onsubmit="return false">
```

Let me know if it works. For some odd reason, I tried to intercept using JS but the browser's is taking priority over.

tauthomas01 (commented):
Looks good, I left a comment.

assets/product-form.js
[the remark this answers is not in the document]
ludoboludo: Ah, I was looking into it by creating a new PR as there is a separate issue for it. https://github.com/Shopify/dawn/pull/593

I didn't go with your approach but that might actually be the easiest/quickest way. What I had done is added an event listener on the input and triggered the change event, while preventing the default behaviour (submit). But we can talk about it over there or solve it all within this PR ? 🤷

martinamarien:
@svinkle
What I've done is change our approach to disabling the button _only_ when clicking on add to cart.
We were using `disabled`, which caused the button to lose focus automatically. Now, I change it to uses `aria-disabled`, which allows focus to remain on the button. To prevent buyers from submitting the form again while loading, I've included an early return if the button is already in the loading state. 

Will that solve the issue for the live region?

assets/product-form.js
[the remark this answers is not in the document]
martinamarien: I won't tackle this in this PR, since I don't think it's the same issue, and since Ludo already created a new PR for it.

assets/section-main-product.css
```diff
@@ -110,6 +110,21 @@
   display: block;
 }
 
+.product-form__error-message-wrapper:not([hidden]) {
+  display: flex;
+  align-items: flex-start;;
```
tauthomas01, line 115:
```suggestion
  align-items: flex-start;
```

tauthomas01 (approved):

tauthomas01 (approved):

ludoboludo (approved):
🚀

martinamarien:
I'll be merging this PR and addressing Scott's feedback about Window screen readers in a fast follow PR 
> However, Windows screen readers don't work well with live regions + hidden attribute; the content is not being announced as expected. Can we take a different approach? Perhaps toggle the hidden state for the icon only?

Binthout:
I am still having problems with this issue after updating to 2.1.0.
I tested on safari and chrome. It only occurs in safari. 
Usually my products don't use the stock function in Shopify but are marked: keep selling without stock. 
Anyone having the same issue?

Added a video.

https://user-images.githubusercontent.com/86726970/133488279-4c235410-88c6-4b47-8766-cc232b008ad9.mov

ludoboludo:
Hello @Binthout, the fix in this PR was specific to the product page and feat. product section. Not the cart page. We still have an issue open for the cart and looking into a fix for error messaging: https://github.com/Shopify/dawn/issues/380

Though I don't think what you're showing in this video is mentioned there. I was able to replicate locally.

ludoboludo:
Interesting, from what I'm seeing it's working properly when navigation/updating the quantity via keyboard navigation. But when it's using the mouse, it doesn't detect anything for `document.activeElement.getAttribute('name')` from the `onChange(event)` function. 

So `if (lineItem) lineItem.querySelector(`[name="${name}"]`).focus();` ends up failing and throwing an error which shows up in the video 🤔 

I will add it to the issue I mentioned in my previous comment 👍

PR #593, opened 2021-09-08
Cart quantity update issues

### LINKED ISSUE #380: Cart Page: Inconsistent error messaging for low item stock

ludoboludo:

Fixes #380

Dispatch change on keypress `Enter` and prevent default behaviour. 

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126352752662)

**Checklist**

### DIFF
--- a/assets/cart.js
+++ b/assets/cart.js
@@ -27,6 +27,7 @@ class CartItems extends HTMLElement {
   }
 
   onChange(event) {
+    console.log(event);
     this.updateQuantity(event.target.dataset.index, event.target.value, document.activeElement.getAttribute('name'));
   }
 
--- a/assets/cart.js
+++ b/assets/cart.js
@@ -27,7 +27,6 @@ class CartItems extends HTMLElement {
   }
 
   onChange(event) {
-    console.log(event);
     this.updateQuantity(event.target.dataset.index, event.target.value, document.activeElement.getAttribute('name'));
   }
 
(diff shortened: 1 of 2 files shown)


tauthomas01 (approved):
LGTM

PR #614, opened 2021-09-10
Fixed image quality in product modal

tyleralsbury:

There was an issue with the image quality in the product modal.  It was being caused when an image that was uploaded was in between the smallest sizes in the `srcset` attribute- 550 and 1100.  So for example, if a merchant uploaded an image that was `938px` it would instead render a `550px` image.

Note: the solution was tested on the store on which the issue was originally reported and it did solve the quality issue.

To fix this, I've added the original image as the last element in the `srcset` attribute.  This guarantees that the highest possible quality image is an option as one of the `srcset` values regardless of the size that is uploaded and ensures that we output the best possible quality in cases where lower quality / smaller width images are being uploaded.

I also added a few sizes and fixed up the `sizes` attribute to better reflect the rendered image sizes.

**Other considerations**

I created an issue to do this to the rest of our `srcset` attributes throughout the theme.  I think this is a principled way to ensure the best possible quality outputs in all cases. https://github.com/Shopify/dawn/issues/615

**Checklist**

### DIFF
--- a/snippets/product-media.liquid
+++ b/snippets/product-media.liquid
@@ -18,11 +18,15 @@
   <img
     srcset="{%- if media.preview_image.width >= 550 -%}{{ media.preview_image | img_url: '550x' }} 550w,{%- endif -%}
             {%- if media.preview_image.width >= 1100 -%}{{ media.preview_image | img_url: '1100x' }} 1100w,{%- endif -%}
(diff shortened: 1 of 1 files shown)


tyleralsbury:
cc @drewrothrock @justinlangille thanks for finding and reporting this issue!

ludoboludo (approved):

tyleralsbury:
> Nice! TIL about `img_url: 'master'`

Same!  I would have thought it was `img_url` with no parameter, but instead that gives a tiny little image.

Issue #78, opened 2021-06-30
[Global] Details/summary pattern conveys incorrect semantic meaning 


nicklepine:
## Issue summary

Throughout the JS enabled experience, `details`/`summary` pairings are used as show/hide components. In some AT (VO, for example) in some contexts, this pairing could convey incorrect semantic meaning; hearing "summary" may not be expected.

In addition, some environments do not support `details`/`summary` well (as experience in the [Fable usability test session](https://drive.google.com/file/d/10HzGsftkz54ovUEiukEIUPCmEuve-HDz/view?usp=sharing) when the control did not announce the current state consistently.)

<details>
  <summary>Screenshots</summary>
<img width="595" alt="Dawn main navigation, Info button is highlighted." src="https://user-images.githubusercontent.com/1392632/120340787-5166c400-c2c4-11eb-8234-378376439336.png">
<img width="540" alt="Dawn product page, accordion section controls are highlighted." src="https://user-images.githubusercontent.com/1392632/120340795-5297f100-c2c4-11eb-92d3-9f0930691ab8.png">

</details>

#### HTML

```html
<details>
  <summary>
    <!-- … -->
    <h2 class="h4 accordion__title">Care information</h2>
    <!-- … -->
  </summary>
  <div class="accordion__content">
    <p>
      Cold wash only with similar colors. Tumble dry on low temperature or air
      dry.&nbsp;
    </p>
  </div>
</details>
```

## Steps to reproduce

1. Go to a product page, using Dawn with any supported browser
2. With an active screen reader (VO with Safari or NVDA with Chrome), use the virtual cursor to navigate through the main navigation or product accordion controls

### Expected

- Show/hide controls semantics and state to be conveyed using the [disclosure pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure).

### Actual

- Controls are incorrectly conveyed or feature poor support.

## Recommendation

**In the JS env only:**

1. Add the `role` attribute to the `summary` control, setting its value to `button`.
2. Add the `aria-expanded` attribute to the `summary` control, toggling its value from `false` to `true` on activation.
3. Add The `aria-controls` attribute to the `summary` control, setting its value to the `id` of the `div` content container.

These changes will create a more consistent and expected user experience for all assistive technology users.

#### HTML

```html
<details>
  <summary role="button" aria-expanded="false" aria-controls="accordionContent1">
    <!-- … -->
    <h2 class="h4 accordion__title">Care information</h2>
    <!-- … -->
[...]

PR #628, opened 2021-09-14
Fix details/summary accessibility pattern

tauthomas01:

Fixes https://github.com/Shopify/dawn/issues/78

The goal of this PR is to normalize `details` and `summary` regarding accessibility behavior. The affected components are:
- Collection filters
- Header search modal
- Header menu (mobile)
- Header menu (desktop)
- Product page: tab collapsible page content

- Added `role="button"` attribute to the `summary` control
- Added `aria-expanded` attribute to toggle `true` or `false` depending on the state (open/close)
- Added `aria-controls` attribute
- Add `Esc` key support to close the content and focus back to the "controller"

I created a global code snippet to loop through all instances that started with `#Details-"` (e.g. `<details id="Details-HeaderMenu">`) and added `click` event and `keyup` event.

Some components are already using their custom elements (`details-modal`, `menu-drawer`) so I added the logic there instead so leveraging the global code snippet so I can avoid potential conflicts.

I'm always open to improve the approach.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126384111638)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126384111638/editor)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -6,6 +6,20 @@ function getFocusableElements(container) {
   );
 }
 
+document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
+  summary.addEventListener('click', (event) => {
+    const isOpen = event.currentTarget.closest('details').hasAttribute('open');
+    console.log(isOpen);
+    if(isOpen) {
+      event.currentTarget.setAttribute('aria-expanded', false);
+    } else {
+      event.currentTarget.setAttribute('aria-expanded', true);
+    }
+  });
+
+  summary.addEventListener('keyup', onKeyUpEscape);
+});
+
 const trapFocusHandlers = {};
 
 function trapFocus(container, elementToFocus = container) {
@@ -116,6 +130,7 @@ function onKeyUpEscape(event) {
 
   const summaryElement = openDetailsElement.querySelector('summary');
   openDetailsElement.removeAttribute('open');
+  summaryElement.setAttribute('aria-expanded', false);
   summaryElement.focus();
 }
 
@@ -374,6 +389,7 @@ class MenuDrawer extends HTMLElement {
 
   closeSubmenu(detailsElement) {
     detailsElement.classList.remove('menu-opening');
+    detailsElement.querySelector('summary').setAttribute('aria-expanded', false);
     removeTrapFocus();
     this.closeAnimation(detailsElement);
   }
--- a/assets/global.js
+++ b/assets/global.js
@@ -9,7 +9,7 @@ function getFocusableElements(container) {
 document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
   summary.addEventListener('click', (event) => {
     const isOpen = event.currentTarget.closest('details').hasAttribute('open');
-    console.log(isOpen);
+    
     if(isOpen) {
       event.currentTarget.setAttribute('aria-expanded', false);
     } else {
--- a/assets/global.js
+++ b/assets/global.js
@@ -349,6 +349,7 @@ class MenuDrawer extends HTMLElement {
 
       setTimeout(() => {
         detailsElement.classList.add('menu-opening');
+        summaryElement.setAttribute('aria-expanded', true);
       });
     }
   }
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -345,12 +345,12 @@
             <li>
               {%- if link.links != blank -%}
                 <details-disclosure>
-                  <details>
-                    <summary class="header__menu-item list-menu__item link focus-inset">
+                  <details id="Details-HeaderMenu">
+                    <summary class="header__menu-item list-menu__item link focus-inset" role="button" aria-expanded="false" aria-controls="HeaderMenu-MenuList">
                       <span {%- if link.child_active %} class="header__active-menu-item"{% endif %}>{{ link.title | escape }}</span>
                       {% render 'icon-caret' %}
                     </summary>
-                    <ul class="header__submenu list-menu list-menu--disclosure caption-large motion-reduce" role="list" tabindex="-1">
+                    <ul id="HeaderMenu-MenuList" class="header__submenu list-menu list-menu--disclosure caption-large motion-reduce" role="list" tabindex="-1">
                       {%- for childlink in link.links -%}
                         <li>
                           {%- if childlink.links == blank -%}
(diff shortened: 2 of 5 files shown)


svinkle:
@tauthomas01 Great stuff. One other thing worth mentioning; this is not required for modal window patterns since a modal is its own, separate context (users are moved from the "main document" context (`role="document"`–the default role) to a dialog context (`role="dialog"`, keyboard trap, visually identified by a "window", etc.)

With a modal window pattern, the launcher would have `aria-haspopup="dialog"` as the indicator token. No other hints (context attributes such as `aria-expanded`) are required.

I'm still testing the mobile navigation and collapsable facet patterns. I'm leaning towards converting these to a modal context as well since they both trap keyboard focus, and the collapsable facet specifically 1) looks like a modal, and 2) has a major design flaw where, on desktop, the distance from activator to the content creates a proximity issue for low-vision zoom users.

I'd say let's go with this PR for now, but may need to make adjustments elsewhere.

svinkle (dismissed):

svinkle (dismissed):

sections/main-product.liquid
```diff
@@ -113,7 +113,7 @@
                   </div>
                   {% render 'icon-caret' %}
                 </summary>
-                <div class="accordion__content rte">
+                <div class="accordion__content rte" id="ProductAccordion-{{ block.id }}-{{ section.id }}">
                   {{ block.settings.content }}
                   {{ block.settings.page.content }}
                 </div>
```
ludoboludo, line 119: I wonder if we should do the same approach for the share button and remove the duplicate work being done in `share.js` 🤔

ludoboludo: Looks like it would mean removing an event listener and the `addAccessibilityAttributes` function.

sections/header.liquid
```diff
@@ -345,25 +345,25 @@
...
                           {%- else -%}
-                            <details>
+                            <details id="Details-HeaderSubMenu-{{ forloop.index }}">
                               <summary class="header__menu-item link link--text list-menu__item focus-inset caption-large">
                                 {{ childlink.title | escape }}
                                 {% render 'icon-caret' %}
                               </summary>
-                              <ul class="header__submenu list-menu motion-reduce">
+                              <ul id="HeaderMenu-SubMenuList-{{ forloop.index }}" class="header__submenu list-menu motion-reduce">
                                 {%- for grandchildlink in childlink.links -%}
```
ludoboludo, line 367: Should the same approach be used for the mobile menu ? And we can then remove the `addAccessibilityAttributes(summaryElements)` function from the `MenuDrawer` class.

ludoboludo (commented):
Looking good, I can pick it up to make changes if some are needed. 

I was mostly wondering if we should/could remove some duplicate functionality like mentioned in my comments below.

kmeleta (commented):
I really like the idea of making that logic more reusable, I just have some thoughts on how exactly that's done. Feel free to disagree.

sections/header.liquid
```diff
@@ -345,25 +345,25 @@
...
               {%- if link.links != blank -%}
                 <details-disclosure>
-                  <details>
+                  <details id="Details-HeaderMenu">
                     <summary class="header__menu-item list-menu__item link focus-inset">
                       <span {%- if link.child_active %} class="header__active-menu-item"{% endif %}>{{ link.title | escape }}</span>
                       {% render 'icon-caret' %}
                     </summary>
-                    <ul class="header__submenu list-menu list-menu--disclosure caption-large motion-reduce" role="list" tabindex="-1">
+                    <ul id="HeaderMenu-MenuList" class="header__submenu list-menu list-menu--disclosure caption-large motion-reduce" role="list" tabindex="-1">
```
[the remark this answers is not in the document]
ludoboludo: Good point!

assets/global.js
```diff
@@ -6,6 +6,21 @@ function getFocusableElements(container) {
   );
 }
 
+document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
```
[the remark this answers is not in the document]
svinkle: > Any reason we couldn't just target `summary` elements in general?

Not all instances of `details`/`summary` convey the same meaning. When JS is available, it's used to enhance the user experience by providing greater AT support and more familiar navigational patterns (support for native `details`/`summary` can be inconsistent.) Some instances convey a disclosure pattern, where others instances convey a modal context.

Ideally there'd be only two sources of truth; one for each type of pattern. I'm not overly opinionated on how we target either in the DOM, but it should be obvious to anyone reading the code. And of course, any `id` that's produced must be unique, otherwise there's the possibility of a target mismatch as well as invalid HTML.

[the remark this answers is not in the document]
ludoboludo: I think IDs was the chosen way because we wanted to stir away from using data-attributes too often like we did in previous themes.
 I was trying to see if I could move everything to the `details-disclosure` file. But there are a few side effects. Messes with the styling for the menu drawer, functionality of the facet filtering and also collapsible tabs on the product page. 

So instead I removed the `keyup` listener from `DetailsDisclosure` since it's already in that other function. 

Also I noticed that when you clicked outside of the summary content when it's open, it closes it but doesn't remove `aria-hidden="true"` unless you click on the summary element itself to close it or press the ESC key.

Let me know what you both think 👍

svinkle (approved):

assets/global.js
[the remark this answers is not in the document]
ludoboludo: Yeah I've added something to the `close()` function so it toggles the attribute 👌

sections/header.liquid
```diff
@@ -96,7 +96,7 @@
                     {%- for link in section.settings.menu.links -%}
                       <li>
                         {%- if link.links != blank -%}
-                          <details>
+                          <details id="Details-menu-drawer-menu-item">
```
tyleralsbury, line 99: Since this is in a `for`, I think it will produce a lot of duplicate IDs.

sections/header.liquid
```diff
@@ -116,7 +116,7 @@
                                           {{ childlink.title | escape }}
                                         </a>
                                       {%- else -%}
-                                        <details>
+                                        <details id="Details-menu-drawer-submenu">
```
tyleralsbury, line 119: Same here.

sections/header.liquid
```diff
@@ -345,25 +345,25 @@
             <li>
               {%- if link.links != blank -%}
                 <details-disclosure>
-                  <details>
+                  <details id="Details-HeaderMenu">
```
tyleralsbury, line 348: Here too, I believe.

tyleralsbury (commented):
I think I found some spots where duplicate IDs will come up.

svinkle:
@tauthomas01, @ludoboludo any chance we can get this one pushed through?

tyleralsbury (approved):
Code looks good. Note that I didn't test it, since I saw the ✅ from Scott already.

PR #655, opened 2021-09-20
Remove white space mobile view for product page

### LINKED ISSUE #499: Excessive white space on product pages in mobile view on Desktop/editor
When looking at the mobile view of a page while on desktop, either through the theme editor, or in Inspect Element, there is excessive padding between the media gallery and the product info section, on product pages that include 3d media. 

I suspect its due to the "View in your space" button that typically displays on actual mobile devices. 

Could we look into preventing this white space to display? This may through off merchants editing their themes.
image
 

https://screenshot.click/21-56-20f01-bbq36.mp4

sofiamatulis:

Fixes #499 .

Added `display: none;` when the visibility is set to `hidden`

**Other considerations**

**Demo links**

Test it on the product `Pleated Heel Boot off white` on mobile (by resizing the screen or viewing it in the editor in the mobile view)

You should still see the `view in your space` when you are using a device: https://screenshot.click/20-03-zhf48-0umjv.png

- [Editor](https://os2-demo.myshopify.com/admin/themes/126411571222/editor)

**Checklist**

### DIFF
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -3,6 +3,7 @@
   color: rgb(var(--color-foreground));
   margin: 1rem auto;
   box-shadow: none;
+  display: flex;
 }
 
 .button.product__xr-button:hover {
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -3,7 +3,6 @@
   color: rgb(var(--color-foreground));
   margin: 1rem auto;
   box-shadow: none;
-  display: flex;
 }
 
 .button.product__xr-button:hover {
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -11,6 +11,7 @@
 
 .product__xr-button[data-shopify-xr-hidden] {
   visibility: hidden;
+  display: none;
 }
 
 @media screen and (max-width: 749px) {
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -1,8 +1,9 @@
-.button.product__xr-button {
+.product__xr-button {
   background: rgba(var(--color-foreground), 0.08);
   color: rgb(var(--color-foreground));
   margin: 1rem auto;
   box-shadow: none;
+  display: flex;
 }
 
 .button.product__xr-button:hover {
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -14,7 +14,7 @@
 }
 
 @media screen and (max-width: 749px) {
-  slider-component .product__xr-button:not([data-shopify-xr-hidden]) {
+  slider-component .product__xr-button {
     display: none;
   }
 
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -10,7 +10,7 @@
 }
 
 .product__xr-button[data-shopify-xr-hidden] {
-  display: none;
+  visibility: hidden;
 }
 
 @media screen and (max-width: 749px) {
--- a/assets/component-product-model.css
+++ b/assets/component-product-model.css
@@ -10,7 +10,6 @@
 }
 
 .product__xr-button[data-shopify-xr-hidden] {
-  visibility: hidden;
   display: none;
 }
 
--- a/assets/product-model.js
+++ b/assets/product-model.js
@@ -54,5 +54,8 @@ window.ProductModel = {
 };
 
 window.addEventListener('DOMContentLoaded', () => { 
+  if (Shopify.designMode) {
+    document.querySelectorAll("[data-shopify-xr-hidden]").forEach(element => element.classList.add('hidden'));
+  }
   if (window.ProductModel) window.ProductModel.loadShopifyXR();
 });


assets/component-product-model.css
```diff
@@ -11,6 +11,7 @@
 
 .product__xr-button[data-shopify-xr-hidden] {
   visibility: hidden;
```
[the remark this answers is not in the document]
sofiamatulis: Good point! Removed

ludoboludo (approved):
Seems to work as expected. 👍

kmeleta (dismissed):
Tested on desktop browsers and mobile device 👍

LucasLacerdaUX:
cc @melissaperreault Can I get your 👍  on this, just to confirm that we changed our approach from previous product media decisions?

------

**For context:** The `visibility:hidden` was intentionally set to avoid content jumping while the page is loading. This follows our approach on every previous theme that had 3D media implemented, as we decided (at the time) that having some extra whitespace was a smaller issue than this page jumping:

screencast_2021-09-20_14-48-28

This can be an issue on slower connections (simulate `Slow 3G connection` on Chrome Developer Tools to better understand this), where the page might still be loading and the user starts interacting with it, then the `View in your space` button appears and pushes the content - possibly making the user click a button by accident.

assets/component-product-model.css
```diff
@@ -10,7 +10,7 @@
 }
 
 .product__xr-button[data-shopify-xr-hidden] {
-  visibility: hidden;
+  display: none;
```
LucasLacerdaUX, line 13: cc @melissaperreault Can I get your 👍  on this, just to confirm that we changed our approach from previous product media decisions?

------

**For context:** The `visibility:hidden` was intentionally set to avoid content jumping while the page is loading. This follows our approach on every previous theme that had 3D media implemented. We decided (at the time) that having some extra whitespace was a smaller issue than this page jumping:



This can be an issue on slower connections (simulate `Slow 3G connection` on Chrome Developer Tools to better understand this), where the page might still be loading and the user starts interacting with it, then the `View in your space` button appears and pushes the content - possibly making the user click a button by accident.

melissaperreault: At first glance I am concerned by this page jump happening without user interaction and can lead to a poor UX. It is something we debated on our internal implementation of our Debut theme. (https://github.com/Shopify/debut/pull/1166)

I would like to bring @tylerrowsell and  @sainihas to the conversation as this is affecting a global feature experience and want to bring the right people in the conversation.

I'd  like to check and sync about other alternatives we could have.

cc. @nicklepine

tylerrowsell: From the technical side of things, one thing that comes to mind as a possible solution to explore... We could detect user agents server side and then make that info available in a liquid drop. It would require a list of all mobile user agents but perhaps could be a starting point. Then we could use progressive enhancements in js to further check ones that we don't detect via user agent. 

I am not confident in how reliable that would be though, for that reason I would suggest we consider an alternative button/layout for "view in your space". 

I'll defer to Sai on this one about potential alternative for design.

tylerrowsell: From the technical side of things, one thing that comes to mind as a possible solution to explore... We could detect user agents server side and then make that info available in a liquid drop. It would require a list of all mobile user agents but perhaps could be a starting point. Then we could use progressive enhancements in js to further check ones that we don't detect via user agent.

I am not confident in how reliable that would be though, for that reason I would suggest we consider an alternative button/layout for "view in your space".

I'll defer to Sai on this one about potential alternative for design.

sofiamatulis: Here is the commit: https://github.com/Shopify/dawn/pull/655/commits/87d531ab6fef175e04859db48675e61e94556a56

ludoboludo (approved):
Test on chrome and safari and seems to do the job 👌

melissaperreault (approved):
Glad we found a clean approach for both merchants and buyers experience! 🚢

sofiamatulis:
When reviewing this PR it was noted that there was a bug on master (not related to this PR)

- This whitespace above the arrows is a bug and should be removed. (aka replace visibility:hidden with display:none). The space is there because there are two buttons on mobile and there should only be only one (ipad and mobile differ where the button is situated)
- The second whitespace, below the arrows, is where the button will actually be shown. That whitespace should be kept (outside the editor)
-On the editor, you shouldn’t have any whitespace (add display: none to like it's already done)

Expectations:

1. Mobile view on your laptop: https://screenshot.click/22-37-gvlo4-j5q57.png
2. Editor mobile view: https://screenshot.click/22-38-vmdjs-0hj6n.png
3. On your actual phone: https://screenshot.click/22-53-aosiu-9mt0q.png
4. On your actual ipad: https://screenshot.click/22-54-a267k-axquy.png


@melissaperreault @ludoboludo I have re-requested your reviews

assets/product-model.js
```diff
@@ -54,5 +54,8 @@ window.ProductModel = {
 };
 
 window.addEventListener('DOMContentLoaded', () => { 
+  if (Shopify.designMode) {
+    document.querySelectorAll("[data-shopify-xr-hidden]").forEach(element => element.classList.add('hidden'));
```
LucasLacerdaUX, line 58: For this Editor Mobile view, I feel that second button could be **displayed** inside the editor instead of adding `display: none` to it. This would allow the merchant to see and tweak its colors. 

The weird part about it, though, is that merchants wouldn't be able to launch the actual XR (so the button would act as disabled). AFAIK, the `shopify-xr` script does not work in Editor mode.

I’m also okay with this current solution as it matches the decision we previously made (to have a whitespace).

LucasLacerdaUX (approved):
Looks good to me. 

I'm not so sure about the decision to hide the button instead of force-displaying it on the Editor from a UX perspective, so I just left some comments below.

kmeleta (dismissed):
Meli and Ludo should weigh in, but the latest changes now align with my expectations 👍  Thanks!

assets/product-model.js
ludoboludo: But then not all mobile device support the `view in your space` thing right ? So we also need that mentioned.

ludoboludo (dismissed):

assets/product-model.js
melissaperreault: @LucasLacerdaUX is bringing a good point.

We'll need to keep monitoring merchant's experience when they want to validate how the feature will work/display for them and their customers when editing their theme. If they don't see any button in the editor view, regardless if it is compatible or not, they might think it is a bug and could contact support advisor, etc. I am surprised we don't have any insights on that experience yet, isn't it something that was happening on Debut and Narrative already?

On the other side, we don't want to show an invalid layout to merchants as they would again perceive it as a bug because they wouldn't be able to compare what is the reality for the same reasons.

A potential missed opportunity is to have a fallback CTA label/tooltip in place that informs buyers there is a 3D model available and that they should consider viewing this product on the proper device to have the right experience. cc. @bldelgado Do you recall if this was explored? 🤔

For the short term, I don't mind having this fix until we find the right experience, but in the meantime, I would hope that merchants help docs could inform merchants that they could never see the `View in your space` button in the editor mode. Thoughts? _(I am not sure where in the help docs this information could live, at the [product media type](https://help.shopify.com/en/manual/products/product-media/product-media-types#3d-models) level but more at the theme customization detail level?)_

> This would allow the merchant to see and **tweak its colors**

I don't think this is the biggest concern as the theme color scheme doesn't allow you to tweak granular colors like that.

tylerrowsell: > But then not all mobile device support the view in your space thing right ? So we also need that mentioned.
Unless it shows on any device and takes you to the app store to use a compatible app to view it 

Most Android/Apple devices will support it. It does not require an app and uses the native functionality of  `iOS Quicklook` or `Android Scene Viewer` if the device is new enough.

LucasLacerdaUX (commented):
The alignment fix broke the `display: none` we just added to one of the buttons and re-introduced the whitespace above the slider arrows.

assets/component-product-model.css
```diff
@@ -3,6 +3,7 @@
   color: rgb(var(--color-foreground));
   margin: 1rem auto;
   box-shadow: none;
+  display: flex;
```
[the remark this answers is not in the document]
ludoboludo: It still seems to hide it properly with `display: none !important` applied to it from the `hidden` class. From here: 


So maybe we just don't need that other display none 🤔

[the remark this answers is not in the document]
LucasLacerdaUX: That JS solution is only applied on Design Mode. The issue I mentioned above, related to the CSS property, is applied both inside and outside the Editor.

**Before / After (outside the Theme Editor)**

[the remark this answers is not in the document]
sofiamatulis: I can see needing the class `button` because we have a modifier class right after 🤔 `button--full-width`

Good catch on the flex. I am reverting that and will try to align it on the `ipad` breakdown differently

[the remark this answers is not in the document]
ludoboludo: Ah gotcha, I was focusing on the Editor 🤦  but I see it on the "live" version 👍  good catch.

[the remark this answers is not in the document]
LucasLacerdaUX: Oh, I wasn't suggesting to remove the `button` class from the element. Just the specificity on the CSS selector:

assets/component-product-model.css
```diff
@@ -14,7 +15,7 @@
 }
 
 @media screen and (max-width: 749px) {
-  slider-component .product__xr-button:not([data-shopify-xr-hidden]) {
+  slider-component .product__xr-button {
```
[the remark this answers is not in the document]
sofiamatulis: I just kept it for the `hover` like you suggested :)

melissaperreault (approved):
Good for this PR to 🚢 but observed we could make an extra effort to optimizing vertical spacing in another PR 🙏 

[In this example](https://screenshot.click/21-09-vzxq3-n6uda.png), in blue, I am confident we can reduce some pixels between the CTA and the next block below it but we'll need to check various scenarios. i.e We already identified that the `Title` block would need a little update because the top margin [is way to large right now](https://screenshot.click/21-09-3dsqa-ry4gz.png).

sofiamatulis:
> Should we still see the space when doing a mobile view in our browser ? 🤔 We keep it there so there isn't any layout shift right ? screenshot
> 
> Just getting confused about where and when we should see the white gap 😅

@ludoboludo We do keep it there on mobile on browser :)

All the scenarios are here: https://github.com/Shopify/dawn/pull/655#issuecomment-925152164

ludoboludo (approved):

LucasLacerdaUX (approved):

PR #660, opened 2021-09-21
Predictive search result iOS fix 

### LINKED ISSUE #546: Search result disappear when scrolling
Due to the header disappearing when you scroll the search result on mobile disappear as well

Search results should not disappear when scrolling

- Dawn Version: 2.0.0
- iOS Version 14.7.1 Safari

**Possible solution**
Disable the hiding feature when search results are visible

https://user-images.githubusercontent.com/1823617/131752285-d53982cd-5444-4cd5-8517-00f4a1901065.MOV

ludoboludo:

Fixes #546.

I looked at a few different approach and ended up using a JS one. 
- I check if it's iOS or not and then add an event listener to know when the search is open. 
- When it's opened I save the scroll value, apply it to the body, add a class to the body to fix its position and I dispatch a custom event to the sticky header. I also dispatch an event when closing the search.
- The sticky header has a listener for this event and won't do anything to the header (hide it) while it's open. 

This make it so I can scroll without triggering the `hide` function on the header. 

I was testing it on my iPad and it worked ok. Not as well as it does on Chrome by default but much better than what we currently have on `main`.

[Video demo](https://screenshot.click/21-09-we6h1-7jzrk.mp4)

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -262,7 +262,7 @@
               {%- endif -%}
                 <form action="{{ routes.search_url }}" method="get" role="search" class="search search-modal__form">
                   <div class="field">
-                    <input class="search__input field__input" 
+                    <input class="search__input field__input"
                       id="Search-In-Modal-1"
                       type="search"
                       name="q"
@@ -278,7 +278,7 @@
                         autocorrect="off"
                         autocomplete="off"
                         autocapitalize="off"
-                        spellcheck="false" 
+                        spellcheck="false"
                       {%- endif -%}
                     >
                     <label class="field__label" for="Search-In-Modal-1">{{ 'general.search.search' | t }}</label>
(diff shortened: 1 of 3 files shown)


sections/header.liquid
```diff
@@ -514,11 +514,17 @@
...
 
       this.onScrollHandler = this.onScroll.bind(this);
       this.hideHeaderOnScrollUp = () => this.preventReveal = true;
+      this.setPredictiveSearchState = () => {
+        this.predictiveSearchOpen = !this.predictiveSearchOpen;
+      };
 
       this.addEventListener('preventHeaderReveal', this.hideHeaderOnScrollUp);
+      this.addEventListener('predictiveSearchOpen', this.setPredictiveSearchState);
+      this.addEventListener('predictiveSearchClosed', this.setPredictiveSearchState);
```
ludoboludo, line 527: Should I do a similar iOS check like in the `predictive-search.js` file ?

kmeleta (commented):
Couple notes for now, but I'm still thinking about possible alternatives. That said, it's definitely improved, so even if we can't get it perfect in every case I'm inclined to get this in sooner than later, given the severity of the issue.

Noticed if you're viewing search results while scrolled down the page a bit, you seem to lose the body overflow/fixed effect and the empty search bar ends up off screen https://screenshot.click/05-02-grpgy-ranz9.mp4

sections/header.liquid
ludoboludo: Good point, since I'm calling the same function anyway 👌

PR #668, opened 2021-09-22
Include original image in srcsets

### LINKED ISSUE #615: Add original image to srcset
## Purpose

`srcset` attributes currently have a series of hardcoded image sizes in them that only output if the uploaded image is greater than or equal to their size.  The problem with this is that when a merchant uploads an image that's between two sizes, it only renders the next smallest size.

For an extreme example, let's say two `srcset` values are `550px` and `1100px` and the merchant uploads an image that's `1099px`.  It would only output the `550px` size since it is smaller than the `1100px` size, leading to a drastically lower quality image than the merchant would expect given the quality of their upload.

## Solution

We can add the original image to all of the `srcset` attribute throughout Dawn by adding the following into it:

``` Liquid
{{ image | img_url: 'master' }} {{ image.width }}w
```

This will output the original image URL and provide a `w` value based on the original image size.

In the above case, instead of outputting:

``` html
<img srcset="/image550x 550w," ... >
```

We would instead get:

``` html
<img srcset="/image550x 550w, /imageoriginal 1099w," ... >
```

kmeleta:

Fixes #615

Added a srcset entry for the original image asset (highest possible quality image) to all existing srcsets.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126422220822/editor)

**Checklist**

### DIFF
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -30,7 +30,8 @@
                     {%- if block.settings.image.width >= 1100 -%}{{ block.settings.image | img_url: '1100x' }} 1100w,{%- endif -%}
                     {%- if block.settings.image.width >= 1500 -%}{{ block.settings.image | img_url: '1500x' }} 1500w,{%- endif -%}
                     {%- if block.settings.image.width >= 2200 -%}{{ block.settings.image | img_url: '2200x' }} 2200w,{%- endif -%}
-                    {%- if block.settings.image.width >= 3000 -%}{{ block.settings.image | img_url: '3000x' }} 3000w,{%- endif -%}"
+                    {%- if block.settings.image.width >= 3000 -%}{{ block.settings.image | img_url: '3000x' }} 3000w,{%- endif -%}
+                    {{ block.settings.image | img_url: 'master' }} {{ block.settings.image.width }}w"
                   src="{{ block.settings.image | img_url: '1500x' }}"
                   sizes="(min-width: {{ settings.page_width }}px) {% if section.blocks.size == 1 %}calc({{ settings.page_width }}px - 100px){% else %}{{ settings.page_width | minus: 100 | times: 0.67 | round }}px{% endif %}, (min-width: 750px){% if section.blocks.size == 1 %} calc(100vw - 100px){% else %} 500px{% endif %}, calc(100vw - 30px)"
                   alt="{{ block.settings.image.alt | escape }}"
@@ -70,7 +71,8 @@
                         {%- if featured_media.width >= 1100 -%}{{ featured_media | img_url: '1100x' }} 1100w,{%- endif -%}
                         {%- if featured_media.width >= 1500 -%}{{ featured_media | img_url: '1500x' }} 1500w,{%- endif -%}
                         {%- if featured_media.width >= 2200 -%}{{ featured_media | img_url: '2200x' }} 2200w,{%- endif -%}
-                        {%- if featured_media.width >= 3000 -%}{{ featured_media | img_url: '3000x' }} 3000w,{%- endif -%}"
+                        {%- if featured_media.width >= 3000 -%}{{ featured_media | img_url: '3000x' }} 3000w,{%- endif -%}
+                        {{ featured_media | img_url: 'master' }} {{ featured_media.width }}w"
                       src="{{ featured_media | img_url: '1500x' }}"
                       sizes="(min-width: {{ settings.page_width }}px) {% if section.blocks.size == 1 %}calc({{ settings.page_width }}px - 100px){% else %}{{ settings.page_width | minus: 100 | times: 0.67 | round }}px{% endif %}, (min-width: 750px){% if section.blocks.size == 1 %} calc(100vw - 100px){% else %} 500px{% endif %}, calc(100vw - 30px)"
                       alt="{{ featured_media.alt | escape }}"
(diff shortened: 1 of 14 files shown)


bakura10:
I am not sure this is a very good idea. The max size of images at Shopify is if I remember 8000 px wide (it has often increase over the years). Loading such high resolution image will likely lead to very poor loading time. Also having pages with potentially several images of 5 or 6000 px wide can cause quite a lot of strain on browser memory. I think this should be capped at some reasonable values. The quality should not be at the expense of a potentially poor experience.

kmeleta:
@bakura10 thanks for the feedback. It's definitely a valid concern, but as I understand, it may be somewhat edge case. It is possible we could cap the original image size just to be safe, but I'm not sure that’s necessary at this point. 

The browser should choose the srcset entry most appropriate for the image's rendered size (as defined by the `sizes` attr). If an image only needs to render 1400px wide, the original 6000px source would never be loaded. This really only becomes a possibility for fullscreen sections (like image banner) on extremely high resolution displays.

Is there a specific case/scenario you feel we're not considering enough? Happy to discuss further.

ludoboludo (approved):
Looks good 👍 

Ran some lighthouse test for performances too to make sure it was ok. Checked all `` in the repo. Tested on mobile and Safari. 
Tested with an image that's `1000px` wide for the image banner and it loaded the right image 👌

bakura10:
@kmeleta , here is the approach I am doing on my themes. I have a 'image-attributes' snippet that is used like this:

```
<img loading="lazy" sizes="..." {% render 'image-attributes', image: image, sizes: '200,300,400,800,1200' %}">
```

This image attributes snippet renders general attributes such as alt tag, and all the srcset and src attributes, based on the required sizes. This way, this means that your code asked up to 1200 (it is up to the developer to indicate the needed sizes based on the context).

This way, you know that the max size is 1200. You can then generate appropriate sizes like this:

```liquid
{%- assign desired_sizes = sizes | remove: ' ' | split: ',' -%}
{%- assign supported_sizes = '' -%}

{%- for size in desired_sizes -%}
  {%- assign size_as_int = size | times: 1 -%}

  {%- if image.width <= size_as_int -%}
    {%- assign supported_sizes = supported_sizes | append: image.width -%}
    {%- break -%}
  {%- endif -%}

  {%- assign supported_sizes = supported_sizes | append: size | append: ',' -%}
{%- endfor -%}

{%- comment -%}
  If we have no single size which matches, we at least set the maximum width of the image, so that there at least
  something that is displayed on the screen.
{%- endcomment -%}

{%- if supported_sizes == blank -%}
  {%- assign supported_sizes = image.width -%}
{%- endif -%}

{%- assign supported_sizes = supported_sizes | split: ',' | compact -%}
{%- assign src_list = '' -%}

{%- for supported_size in supported_sizes -%}
  {%- assign size_descriptor = supported_size | append: 'x' -%}
  {%- capture src_list -%}{{ src_list }}{{ image | img_url: size_descriptor }} {{ supported_size }}w{% unless forloop.last %}, {% endunless %}{%- endcapture -%}
{%- endfor -%}
```

tyleralsbury (approved):
Looks good to me.

tyleralsbury:
@bakura10 thanks for the idea.  We're looking to keep the use of Snippets and logic in the theme to a minimum, so I think it's safe for now for us to just include the original image size for simplicity's sake.

As long as the `sizes` attribute is informing the browser the right size image to load in any given context, we shouldn't run into many issues around gigantic images being loaded.  Even if a merchant adds an 8000px wide image, unless it's on a full screen section whose sizes attribute is `100vw` and the user is on a massive screen, it should still not load a crazy huge image.

tyleralsbury:
> I have the question: 550w, 720w, 990w... Who handed us these magic numbers?

Great question.  These numbers are based on the key values at different breakpoints.  This is why they can differ from one section to another or even based on particular settings. Different sections are inevitably going to be slightly different from each other.

So, for example, to simplify, let's say we have a section with two breakpoints.  On one of them, the rendered image can be between 720px and 1000px, and on the other it can be between 400px and 680px.  We also want to concern ourselves with screens with a pixel density of 2.

**In that case, we would have the following image sizes, considering 1x and 2x:**

```
400px, 800px
680px, 1360px
720px / 1440px 
1000px / 2000px
```

So based on that, we can say we've got the following: `400, 680, 720, 800, 1000, 1360, 1440, 2000`

Since we've got some that are pretty close together, we can take the highest quality of the ones that are relatively close to remove some unnecessary noise in the `srcset` output.  So we could get rid of `680` and `1360`.  That means we now have `400, 720, 800, 1000, 1440, 2000`.

But there's now a secondary problem where there are some large gaps in some spots, `400 -> 720`, `1000 -> 1440`, `1440 - 2000`.  So if a merchant uploads an image that is `1860`, they're automatically going to be rendering an image of size `1440`, which is about ~77% of the size of the original image they uploaded.  Now let's say that they will be rendering the section in the context in which `2000` is the most appropriate size... the rendered image is `1440`, rather than the full `1860` that would have been available if we used their original image.  So to account for that, we now include the `master` image as one of the sizes as well.

The main upside of this is that we are providing specific values that are relevant to the specific context of each section and reducing the number of values that would be needed to be output if we were to use `50w` increments.  This is especially the case in a context in which we might care to cater to sizes between, say, `400` to `7680 (4k)`.  We would be outputting over 100 values for the `srcset` if we did that.

The downside is the seemingly "magic" choice of the numbers and a bit more thinking required on the part of the theme developer.

bakura10:
@tyleralsbury thanks for the details. I have a question: when we use the image_tag filter we first have to use the image_url with a width parameter which is, as I understand it, the max width we want for an image. So when the image_tag is generated instead of using master would it not be better to use this width instead?

regardibg generating the srcset, I remember that there was one service doing something smart based on the min width, max width, image quality and number of steps: https://www.responsivebreakpoints.com/

maybe it would be interesting to check their algorithm :)

PR #688, opened 2021-09-24
Add typography heading and body font size scale setting

### LINKED ISSUE #634: Add typography base sizes
## Describe the enhancement you'd like

We want to introduce typography base size. Merchants want to be able to change the base size of their theme to accomplish an aesthetic preference in multiple areas of the theme and this new setting will meet their needs.

Some merchants quotes:

> "_Menu: I would just like the text a bit bigger, I'm not sure of the exact size_"

> _"General: It would simply make it easier to adjust the font size for older customers"_

> _"Product cards: I would like a bigger size font for the description of my products, just a size up to the price will be great"_

Because we want to maintain a great balance, the options we'll offer wouldn't be too extreme as we don't want to break the successful grid and alignment.

Below are my initial suggestions but would like to pair with our dev to optimize the outcome and ensure no other dependencies are broken. ([Reference](https://screenshot.click/21-09-3hbb2-ljo51.png))

### Heading
I estimate the range should be from `26px` to `34px` and have an increment of `2px`.

### Body text
I estimate the range should be from `14px` to `17px` (Maybe 18px) and have an increment of `1px`.

### Visual reference methodology
I prefer with this issue to pair with a dev to see the changes IRL as oppose to try to mock up the changes in various screens as I want us to turn around quickly on the change.

KaichenWang:

Fixes #634

#### 1. `body_scale` - Add theme typography setting for scaling body text size

- The setting value is used as a multiplier in CSS to a scale up `<html>` base font-size
- Since theme CSS uses `rem` units, all elements will be "scaled up" when this multiplier is applied - including all body text, icons, padding, margin, height, width, etc.

| `body_scale` value (%) | Multiplier (`body_scale` / 100%) |`0.1rem` equivalent |
| --------------- | --------------- | --------------- |
| `100` | `1.00`  | `1px`
| `130` | `1.30`  | `1.3px`

#### 2. `heading_scale` - Add theme typography setting for scaling heading text size

- This allows for headings (all text that uses heading font) to scale independently from all other elements
- Since `body_scale` affects all CSS elements including headings, this must be negated for headings (divide by `body_scale`)
- The heading multiplier is therefore `heading_scale` / `body_scale`
- Some icons also scale with `heading_scale`

| `heading_scale` value (%) | Multiplier (`heading_scale` / `body_scale`) if `body_scale = 100%` | `0.1rem` equivalent if `body_scale = 100%` | 
| --------------- | --------------- | --------------- | 
| `100` | `1.00`  | `1px` |
| `150` | `1.50`  | `1.5px` |

**Other considerations**

n/a

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126427398166)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126427398166/editor)

**Checklist**


assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
   font-family: var(--font-heading-family);
```
sofiamatulis, line 354: When we have the font size range to `150%` the text gets cut off in the `collage` https://screenshot.click/28-03-89dbl-7ohm8.png

This happens when the scales are different: https://screenshot.click/28-13-47dfl-m832n.png

assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
...
   font-style: var(--font-heading-style);
   font-weight: var(--font-heading-weight);
-  letter-spacing: 0.06rem;
+  letter-spacing: calc(var(--font-heading-scale) * 0.06rem);
   color: rgb(var(--color-foreground));
   line-height: 1.3;
```
sofiamatulis, line 359: When I update the size, the filter title doesnt seem to change: https://screenshot.click/28-05-hftxj-r8zyo.mp4

Code:
https://screenshot.click/28-08-9zifa-8am6p.png

assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
...
 }
```
sofiamatulis, line 360: The `login` seems to change size, but I believe this is in an `a` tag https://screenshot.click/28-07-2hzor-qptfj.png

assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
...
+  font-size: calc(var(--font-heading-scale) * 2rem);
 }
 
 @media only screen and (min-width: 750px) {
   h2,
   .h2 {
-    font-size: 2.4rem;
+    font-size: calc(var(--font-heading-scale) * 2.4rem);
   }
 }
```
sofiamatulis, line 394: The vendor sizing (wrapped in a `p`) isnt updating its size: https://screenshot.click/28-15-mb2ni-0t7jo.png

Code: https://screenshot.click/28-16-99w0a-ejdjh.png

assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
...
 
 h3,
 .h3 {
-  font-size: 1.7rem;
+  font-size: calc(var(--font-heading-scale) * 1.7rem);
 }
```
sofiamatulis, line 400: Not sure if it's related to this PR, but is it expected that we have a lot of space between the text and icon? I dont seem to see this with the other items: https://screenshot.click/28-20-7e0f0-z06r0.png

assets/base.css
```diff
@@ -354,73 +354,73 @@ h5,
...
   font-style: var(--font-heading-style);
-  font-size: 1.5rem;
+  font-size: calc(var(--font-heading-scale) * 1.5rem);
 }
 
 h5,
 .h5 {
-  font-size: 1.2rem;
+  font-size: calc(var(--font-heading-scale) * 1.2rem);
 }
```
sofiamatulis, line 419: 404 `p` tag isnt updating its size: https://screenshot.click/28-21-r6tak-6tmjf.png

assets/base.css
```diff
@@ -474,6 +474,11 @@ blockquote {
   letter-spacing: 0.04rem;
 }
 
+h2.caption-large,
+h3.caption-large {
+  font-size: calc(var(--font-heading-scale) * 1.3rem);
+}
+
 .color-foreground {
   color: rgb(var(--color-foreground));
```
sofiamatulis, line 483: The price `p` in the gift card doesnt seem to be updating either: https://screenshot.click/28-22-zd38z-zdtqd.png

assets/base.css
```diff
@@ -561,6 +566,10 @@ h6:empty {
   font-family: inherit;
 }
```
sofiamatulis, line 568: From what I understand, inputs shouldnt be changing sizes. Is that correct? https://screenshot.click/28-24-4m7d4-kfyt8.mp4

templates/gift_card.liquid
```diff
@@ -49,6 +49,9 @@
         --font-heading-style: {{ settings.type_header_font.style }};
         --font-heading-weight: {{ settings.type_header_font.weight }};
 
+        --font-body-scale: {{ settings.body_scale | divided_by: 100.0 }};
+        --font-heading-scale: {{ settings.heading_scale | times: 1.0 | divided_by: settings.body_scale }};
```
sofiamatulis, line 53: I am curious as why you are keeping the decimal instead of using `divided_by: 100` and `times: 1`

snippets/facets.liquid
```diff
@@ -71,13 +71,13 @@
                               {% if value.count == 0 and value.active == false %}disabled{% endif %}
                             >
 
-                            <svg width="16" height="16" viewBox="0 0 16 16" aria-hidden="true" focusable="false">
+                            <svg width="1.6rem" height="1.6rem" viewBox="0 0 16 16" aria-hidden="true" focusable="false">
```
sofiamatulis, line 74: Wondering if we'll need to do this to all of the other `svgs`?

sofiamatulis (commented):
This is an exciting update! Added a few comments :)

assets/base.css
KaichenWang: Inputs will change in scale if `body_text` setting is changed. In fact, all padding, margin, height, width in the theme scales with `body_text` setting

templates/gift_card.liquid
KaichenWang: Dividing and multiplying integers using Liquid filter produces an integer result, but for scale, we need decimal values

snippets/facets.liquid
LucasLacerdaUX: While I'm not sure if it should apply to all `svg`s, it should probably apply consistently to icons that are accompanied by text.

For instance, on the product page, the pickup availability one follows the text size, but the collapsible tabs and share button does not.

image

assets/base.css
KaichenWang: This would be considered "body text" and should be scaling with `body_scale` setting

assets/base.css
KaichenWang: Not related to this PR. For `<select>`, AFAIK, there's no way to ensure its width fits the currently selected option

assets/base.css
KaichenWang: This would be considered "body text" and should be scaling with body_scale setting

assets/base.css
KaichenWang: Updated. The filter drawer heading now scales with `heading_scale` setting

Screen_Shot_2021-09-28_at_3_03_12_PM

assets/base.css
KaichenWang: Updated. The login link and its associated icon should scale with `body_scale` and **not** `heading_scale` setting. This has been fixed

snippets/facets.liquid
KaichenWang: Updated https://github.com/Shopify/dawn/pull/688/commits/2ff7cac0ff3b224b861db3ade8f517e66f94bb50:

- Share icon now scales with `body_scale` setting
- Collapsible tab heading icons now scale with `heading_scale` setting

LucasLacerdaUX (approved):
Looks good to me!

Just tested the new changes and all the comments I had were addressed :D

assets/base.css
KaichenWang: Would prefer to address this in a separate PR. The reason being that this would require some refactoring and re-testing of the collage section. To be clear, this issue already exists with collage since longer product/collections titles could always cause overflow, but you are correct that this issue is made worse by the option to increase font scale

melissaperreault (commented):
Thank you a lot Kai! This is massive! Very excited about this upcoming update and what it will open in flexibility for Dawn!

Did a first pass - I need to do another one, here are my first observations:

- [x] [Cart notification](https://screenshot.click/21-09-lcicb-tdyh4.png) product information should use `body_scale` to match the cart
   - [ ] Related to that, something I am wondering is if we should always treat the product title as a heading. I see some different approach on various template, I'd like to have your opinion and perhaps have a wider UX conversation about it.
- [x] Not sure if I saw this covered on another comment but the down-facing caret aren't scaling with the right scaling value for the [Collection filters](https://screenshot.click/21-09-zf7cu-m3438.png) and the [nested navigation](https://screenshot.click/21-09-fh56x-ho4te.png).
- [ ] I don't think this is related but noting in case, [filter options aren't centered vertically](https://screenshot.click/21-09-6pksh-8jzau.png)
- [ ] I noticed a `<p>` tag under the [Pickup availability drawer over here](https://screenshot.click/21-09-fumbp-w3td6.png)
- [ ] [Inside the search combo box](https://screenshot.click/21-09-acqg8-ql6x7.png), the table heading scales with the `heading_scale` value, we are not scaling it by that value in Customer account table and Cart.
   - [ ] This made me realize that styles in Cart, Customer account and Search combo box for the table title aren't the same, I would like to understand the difference when pairing.

assets/base.css
sofiamatulis: Since the issue already exists, we can do that. Thanks for creating the issue :)

sofiamatulis (dismissed):
Retested everything and it looks good 🚢 

thanks for the changes 👍 . Just noticed an internal liquid error on the gift card page (left a comment)

KaichenWang:
@melissaperreault 

Body text `line-height` (initially 1.8) now adjusts based on `body_scale` according to the formula
`line-height = 1 + 0.8 / scale`

| `body_scale` value (%) | Multiplier (`body_scale` / 100%) | Computed `line-height` |
| --------------- | --------------- | --------------- |
| `100` | `1.00`  | `1.8`
| `130` | `1.30`  | `1.62`

melissaperreault:
Awesome Kai, super update! Thanks! 
After testing it out and testing headings, I wish we can include that for when scaling Headings as well to harmonize spacing and balance white space line height. 🙏 

Coming later to put my final observations for margins like we discussed IRL. Thank you!

melissaperreault (commented):
Thank you Kai! 🎉 

Scaling looks amazing among the changes for line height for body and what is upcoming for headings. I believe checking for headings later once you apply the same logic won't affect my review much.

My review focuses on areas some sections margins don't benefit the scaling flexibility + checking the full set of sections in action.

- [x] Rich text section: The margins becomes too wide on mobile when scaling the body font. Leaving the reading area too narrow and unpleasant to read for customers. [Would adjusting the calculation](https://screenshot.click/21-10-zf0a2-lsoid.mp4) fix the outcome? (    `width: calc(100% - 7rem);` instead of    `width: calc(100% - 8rem);` )
   - [ ] Other areas I noticed a similar need for adjustments are:
   - [ ] [Footer](https://screenshot.click/21-10-e7x0o-q0zyg.png), text blocks gets wrapped too quickly because of that
   - [ ] [Email sign up section](https://screenshot.click/21-10-ro73l-u2hj5.png), input field and text could take more space horizontally
   - [ ] Image with text section:  [Padding could be smaller](https://screenshot.click/21-10-ibwb9-0ptw3.png) when the body is scaled
- [ ] Collage section: I would tend to suggest to scale this [section blocks default headings ](https://screenshot.click/21-10-e99lk-gu15d.png) with the `body_scale` instead of the headings one to not create false expectations for merchants that their text wold be so large. Rationale is to improve the outcome on smaller screens, there is little to no place for content because of the inner padding/margins applied to the different blocks. But that might also need a little UI/UX revision to improve the `Collage mobile layout` setting.
   - [ ] Default headings for Collection list too would be inconsistent [to express the official outcome once merchants select a real collection](https://screenshot.click/21-10-jtf41-ohf39.mp4). The harmony between the section title and the block's text size once a collection is linked is much better. We can discuss of what would be a better alternative.
- [ ] Product cards: When scaling body font, there is a perceptible [white space added around the card asset](https://screenshot.click/21-10-frpo8-2nzuu.png)
- [x] Cart template: The table scale is [not successful right now](https://screenshot.click/21-10-m4r3j-cdhj2.mp4). I would suggest if possible not scaling the product image to leave more space for the content. The trash can icon also wraps on a second line, perhaps not scaling the image would keep safe space still.
- [ ] Customer account template: Taking a note but will open another issue about it as I believe this is tied to a global change what we will need to update. Table [heading from the cart](https://screenshot.click/21-10-ham5b-68hbb.png) compared to customer account [is different in size and inconsistent in letter-spacing](https://screenshot.click/21-10-c735o-bxerw.mp4). Although we might not want to have those labels too small, they could get some love to improve the template hierarchy.
- [ ] Featured collection: [Section default content needs](https://screenshot.click/21-10-3aln0-ulhvf.png) more love, vendor name is too close to the card right now compared to the real outcome once you link a collection. (Will open a new issue)
- [ ] Featured product: 
   - [ ] [Default view](https://screenshot.click/21-10-k37ya-o6gq0.png), the Sold out label is not vertically aligned with the price (Will open a new issue)
   - [ ] When the link label [is wrapping on two lines or more](https://screenshot.click/21-10-urp6t-mwn49.png), the arrow doesn't scale anymore
   - [ ] Share block: [the input goes offscreen](https://screenshot.click/21-10-jjqi5-gcll4.png) when scaling body

KaichenWang:
@melissaperreault I've addressed the [feedback](https://github.com/Shopify/dawn/pull/688#pullrequestreview-770457497) and updated the PR. The following has not been addressed, here are some details:

> Collage section: I would tend to suggest to scale this section blocks default headings with the body_scale instead of the headings one to not create false expectations for merchants that their text wold be so large. Rationale is to improve the outcome on smaller screens, there is little to no place for content because of the inner padding/margins applied to the different blocks. But that might also need a little UI/UX revision to improve the Collage mobile layout setting.

>  Default headings for Collection list too would be inconsistent to express the official outcome once merchants select a real collection. The harmony between the section title and the block's text size once a collection is linked is much better. We can discuss of what would be a better alternative.

Agreed, I think the collage section in general would benefit from a slight refactor, especially to accommodate larger text. We can track all fixes in this new issue https://github.com/Shopify/dawn/issues/699

> Customer account template: Taking a note but will open another issue about it as I believe this is tied to a global change what we will need to update. Table heading from the cart compared to customer account is different in size and inconsistent in letter-spacing. Although we might not want to have those labels too small, they could get some love to improve the template hierarchy.

Aligned on opening an issue to tackle headings globally in the theme for more consistency

assets/template-giftcard.css
```diff
@@ -22,7 +22,7 @@ h6:empty {
 }
 
 html {
```
sofiamatulis, line 24: When I adjust the font size in the gift card page I get a Liquid internal error:

https://screenshot.click/05-09-ciwss-eoc1q.mp4

config/settings_schema.json
```diff
@@ -4,8 +4,8 @@
     "theme_name": "Dawn",
     "theme_version": "2.2.0",
     "theme_author": "Shopify",
-    "theme_documentation_url": "https://help.shopify.com/manual/online-store/themes/os20/themes-by-shopify/dawn",
-    "theme_support_url": "https://support.shopify.com/"
+    "theme_documentation_url": "https:\/\/help.shopify.com\/manual\/online-store\/themes\/os20\/themes-by-shopify\/dawn",
+    "theme_support_url": "https:\/\/support.shopify.com\/"
```
sofiamatulis, line 8: Is this change intended?

snippets/product-card-placeholder.liquid
```diff
@@ -8,17 +8,17 @@
...
       {% render 'price' %}
     </div>
   </div>
+
+  <div class="card card--text-only card--soft card--product">
+    <div class="card__inner">
+      <div class="card__content"><h2 class="card__text h3">{{ 'onboarding.product_title' | t }}</h2></div>
+    </div>
+  </div>  
 </div>
```
sofiamatulis, line 24: Small nitpick but on the product page, when scaling the body size, the spacing between some elements feel a bit tight: https://screenshot.click/05-17-vobf4-raq5b.mp4


cc @melissaperreault

assets/component-cart-notification.css
```diff
@@ -58,8 +58,8 @@
 
 .cart-notification__heading .icon-checkmark {
```
sofiamatulis, line 59: Not related to this PR but I am seeing this and not on `main` 🤔 
Are you seeing this as well?

https://screenshot.click/05-20-zzfrg-7d2il.mp4

assets/component-cart-items.css
```diff
@@ -156,7 +156,7 @@ cart-remove-button .icon-remove {
   }
 }
```
sofiamatulis, line 157: In the filter, when the body is set to the highest size it's cutting the text in some languages (screenshot in French):

https://screenshot.click/05-22-khmva-ekybx.png

Also the focus is cutting of the text next to it: https://screenshot.click/05-23-0npqv-8vhyv.mp4

sofiamatulis (commented):
Looks good 👍 Left a few small comments and then I can do a final pass

config/settings_schema.json
KaichenWang: Good catch. Reverted

snippets/product-card-placeholder.liquid
melissaperreault: @KaichenWang Is it possible that in this particular block space that the spacing is not scaled?

sofiamatulis (approved):
Looks great 🎉  thanks for all the changes and the issues created

assets/component-accordion.css
```diff
@@ -32,9 +36,10 @@
 
 .accordion .icon-accordion {
```
LucasLacerdaUX, line 37: **Minor bug and a bit of an edge case** 

Now that accordion icons are based on the size of font heading scale, we'd need to limit the width of `.accordion__title` based on that, or it would look like that with long titles:


**Expected:**



**Solution:**
```css
.accordion__title {
  display: inline-block;
  max-width: calc(100% - 6rem);
```
could be replaced with:
```css
.accordion__title {
  display: inline-block;
  max-width: calc(100% - (var(--font-heading-scale) * 4rem));
```

assets/base.css
```diff
@@ -447,7 +456,7 @@ blockquote {
 .caption {
   font-size: 1rem;
   letter-spacing: 0.07rem;
-  line-height: 1.7;
+  line-height: calc(1 + 0.7 / var(--font-body-scale));
```
LucasLacerdaUX, line 459: **Side note, not related to this PR**: I find this element's font size literally impossible to read on mobile. And my phone has a huuuge almost table-sized screen, so I can't even imagine how hard it is on smaller devices (`1rem` = `10px`)

Is this intentional?

LucasLacerdaUX (approved):
Looks good to me! Tested every page and it seems to be working consistently across different templates. 

Amazing work, Kai! 


My only comment about the accordion title is almost an edge case and can be fixed in a follow-up PR if you prefer.

PR #702, opened 2021-09-29
Fixed translations [quantity] syntax should not be translated

onkbear:

Translation `cart_quantity_error_html` does not work in some languages.

- [Store](https://themes.shopify.com/themes/dawn/styles/default/preview)

For instance in French, when changing quantity to 999999 in their cart page and then following message appears.

`Vous ne pouvez ajouter que [quantité] de cet article à votre panier.`

I think this syntax `[quantity]` should not be translated.

dawn

I have fixed translation files which `[quantity]` is translated.

**Other considerations**

**Demo links**

**Checklist**


ludoboludo:
Hello and thanks for bringing this up @onkbear! 

I think our original approach was too fragile. I'm going to try and think of a best way to go about fixing the issue 🤔 I will keep you posted shortly.

PR #723, opened 2021-10-01
Option to disable default og:images

drnic:

Shopify Apps want to use App Blocks to add bespoke `<meta property="og:image">` properties, such as https://mocralivelinks.com.

Shopify Apps can now disable Dawn's default `og:image` by setting the metafield `shop.metafields.overrides.og_image = true`

An optional metafield on Shop to be set by a Shopify App.

**Other considerations**

There are no other examples of using `shop.metafields` to allow Apps to enable/disable features of this theme. Please let me know what best practices exist for naming the metafield.

**Demo links**

- https://mocra-dev-with-designer.myshopify.com/products/adidas-classic-backpack - the `og:image` and friends are added via an App Block, and the original Dawn `og:image` has been removed
- 
**Checklist**


bakura10:
The issue with this approach is that it will only make it work for Dawn. Unless this metafield is standardized and pushed by Shopify to be used for all themes, this approach won't really work. The number of those "standardized" metafields should really be kept to a minimum in my opinion so I am personally against introducing too many.

I think a better long-term solution here is that Shopify adds a new "Social media" image inside the SEO box of each resource directly into the admin to override the page_image, instead of using a metafield overrides.

drnic:
This patch to Dawn would be the start of a new standard for templates yes.
New standards start here. They take time but they need a point of reference.
-- 
Dr Nic Williams
+61 437 276 076
twitter @drnic

bakura10:
The issue of leaving dead metafield is a big blocker for me as a theme developer and definitely so I am strongly against such a change. How would you a merchant would use back the original system if they uninstall your app? The metafield would stay to true so the merchant could not go back to the original setting without editing the code.

Dead code has always been the biggest pain from a support perspective so I do not wish a similar issue to be created with metafields.

drnic:
I'm trying really hard to get into the spirit of Store 2.0 and "don't ask merchants to edit their themes". 

Dawn, and other themes, all only defer to the contents of Shopify's internal `page_image` variable [1]

> For product pages, collection pages, and blog posts, the `page_image` is the resource's featured image if it exists. For example, a product page's `page_image` is the same as its `product.featured_image`. If a featured image does not exist, then the page_image is based on the store's social sharing image set in the admin.

What are my options to ensure I can override this `page_image` with a generated social image for each page/product/collection/blog article?

I'm not 100% following the argument against metafields -- are Apps not able to delete/cleanup their metafields when the App is removed?

Is there a standardized approach I should be using here? I'm sorry for not knowing a better path here so I would love suggestions. "Ask Shopify core team to change Shopify" is one suggestion. 

[1] https://shopify.dev/api/liquid/objects/page-image

drnic:
> However if a merchant uninstall your app, because metafields are tied to the store and not your app, and that upon uninstall apps are removed the ability to access the store, merchants will be left with the metafield set to true

Ouch. Thanks for clarifying that metafields added by an app are not removed with the removal of the app, and that the app loses its ability to clean up after itself. On top of this, afaik, `shop.metafields` do not have a native UI in Shopify Admin for merchants to DIY cleanup.

I can see that this PR is not appropriate.

Do Shopify staff visit these PRs and participate in these discussions towards the improvement of Shopify API + product? Or where do I escalate this topic to have this discussion?

bakura10:
@drnic : I am not Shopify staff but yes Shopify dev read this repo ;). I am sure they will be able to give you info if this is planned (I agree it would be very useful and a logical addition to what is available now).

chadleap:
It would be nice to have some sort of way, like this, to toggle on/off the entire theme-based structured data. I use a JSON-LD schema app and have to choose between having redundant markups present or editing core theme files to remove it, which then complicates theme updates.

PR #726, opened 2021-10-01
Proposal: Use standard_product_type for schema.org Product category

alexcwatt:
[schema.org's Product model](https://schema.org/Product) has a `category` that could be populated by `standard_product_type`. This field is not yet available in Liquid but I'm putting together a proposal to add it, and wanted to demo what this could look like in Dawn.

Any thoughts on doing this?

### DIFF
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -496,6 +496,9 @@
       "@type": "Thing",
       "name": {{ product.vendor | json }}
     },
+    {%- if product.standard_product_type -%}
+      "category": {{ product.standard_product_type.type_path | json }},
+    {%- endif -%}
     "offers": [
       {%- for variant in product.variants -%}
         {


sections/main-product.liquid
```diff
@@ -496,6 +496,9 @@
       "@type": "Thing",
       "name": {{ product.vendor | json }}
     },
+    {%- if product.standard_product_type -%}
+      "category": {{ product.standard_product_type.type_path | json }},
+    {%- endif -%}
```
tjoyal, line 501: If there is a "custom type" shoud we:
- Keep using `standard_product_type`? 
- Use `custom_product_type` instead?
- Set category to `null`?

alexcwatt: Good question. I think if custom is set, we should keep using `standard_product_type`, which is hierarchical and matches the expectations of schema.org's `category`. I could be wrong, but I don't see much value in injecting `custom_product_type` in here when it could be used in so many merchant-specific ways, and may not follow the format (which specifies using `>` or `/` to build the path).

PR #728, opened 2021-10-03
Move all Assets_url tags to the head

Andrewrico:

Fixes #0.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/snippets/stylesheet.liquid
+++ b/snippets/stylesheet.liquid
@@ -1,0 +1,90 @@
+{% comment%}
+    # loadTemplateStylesheet.
+    # This will load styles for each template files individually.
+{% endcomment %}
+{% assign loadOptimizedTemplateStylesheet = template.name %} 
+{% case loadOptimizedTemplateStylesheet %}
+
+
+{% when 'index' %}
+
+    {% comment%} # image-banner {% endcomment %}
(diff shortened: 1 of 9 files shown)


Andrewrico:
If approved, I can tackle the rest of the theme

bakura10:
I think the reason you are doing this is to avoid the same files being included twice if the section is added more than one time.

I don’t think this fix is very desirable though as it removes the benefit of having everything encapsulated in the same section. Your change also now leads everything to be loaded even if the section is not used, which is what Shopify wanted to avoid I think.

the best option here would be to keep the files in their respective section but having a new filter allowing to add the file to the head once:

{{ ‘foo.css’ | stylesheet_head_tag }}

Even if called several times this would lead to have the file included once only.

snippets/stylesheet.liquid
```diff
@@ -0,0 +1,90 @@
+{% comment%}
+    # loadTemplateStylesheet.
+    # This will load styles for each template files individually.
+{% endcomment %}
+{% assign loadOptimizedTemplateStylesheet = template.name %} 
+{% case loadOptimizedTemplateStylesheet %}
+
+
+{% when 'index' %}
```
bakura10, line 9: Sections can be addd on every pages so I supposed the sections will end up unstyled if you add them on something different than home page

PR #735, opened 2021-10-05
Fix cart error translation

ludoboludo:

There is currently an issue where we're relying on a string to change the quantity amount of the error message in the cart. 

It's looking for a string `[quantity]` which only exists in the english language. It's translated in other languages so the JS won't find the string it's looking for. 

This is a follow up to [this PR](#702) by @onkbear. 

I've changed `[quantity]` to `{{ quantity }}` in the english translation so that in `theme.liquid` I can set `quantity` to be `[quantity]`. 

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126478647318)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126478647318/editor)

**Checklist**

### DIFF
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -148,7 +148,7 @@
 
       window.cartStrings = {
         error: `{{ 'sections.cart.cart_error' | t }}`,
-        quantityError: `{{ 'sections.cart.cart_quantity_error_html' | t }}`
+        quantityError: `{{ 'sections.cart.cart_quantity_error_html' | t: quantity: '[quantity]' }}`
       }
 
       window.variantStrings = {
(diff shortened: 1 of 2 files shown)


movermeyer (dismissed):
Neat. Yeah, that should work from the translation perspective.

LucasLacerdaUX (approved):
👍 Thanks for making this change.

Do we need to request translations before merging?

ludoboludo:
> 👍 Thanks for making this change.
> 
> Do we need to request translations before merging?

Yes we do 😓  I should have requested it right away. It was just in case we wanted to go with a different approach. I will let you know when the translations are back for re approval gentlemen.

ludoboludo:
Hmm weird it says translations are good but there is only one language that came back 🤔

LucasLacerdaUX (approved):

movermeyer (approved):

PR #745, opened 2021-10-06
Support file upload line item properties on Product Forms and Cart Notifications

### LINKED ISSUE #731: Support file upload line item properties on Add To Cart JS
`input type="file"` are not supported by our Add to Cart AJAX. `FormData` properties are being `JSON.stringify`ied and therefore, which does not support this type of input.

This is causing incompatibility issues to apps that adds Line Item properties and to customized themes, and will likely lead to merchants further editing their code.

Our Add to Cart JS should be compatible with `input type="file"` and file uploads should not result in a string or unknown path.

**Possible solution**
Use `FormData` directly instead of serializing the line item properties.
See https://github.com/Shopify/theme-scripts/pull/124 

- Dawn Version: 1.0.0
- macOS Version 11.3.1

LucasLacerdaUX:
Add support for `type="file"` inputs on Product Forms.
Adds line item properties to cart notifications

Fixes #731

Using `FormData`, we are now able to send files to the `cart_add_url` endpoint. Submitting the request as `FormData` is necessary to support file uploads when adding items to cart by AJAX. Otherwise, the browser would replace the `File` object with an invalid string such as `C:\fakepath\filename.png`.

This is how it works:
1. Check if browser is compatible with `FormData`.
2. If so, remove the `Content-Type` as we're no long passing a `json` as the request's body.
3. Convert the form to `FormData`.
4. Add extra attributes used by cart notification, such as `sections` and `sections_url`.
5. Set the `request.body` to use the FormData object.

**How to test** 🎩 
1. Add the following code to `main-product.liquid` and `featured-product.liquid` product forms:
```liquid
<div>
  <input type="file" id="file" name="properties[File]" form="{{ product_form_id }}">
  <label class="field__label" for="file">Custom Image</label>
</div>

<div class="field">
  <input type="checkbox" id="isMatte" name="properties[isMatte]" form="{{ product_form_id }}">
  <label class="field__label" for="isMatte">Matte option</label>
</div>

<div class="field">
  <input class="field__input" type="text" id="engraving" name="properties[engraving]" form="{{ product_form_id }}">
  <label class="field__label" for="engraving">Engraving</label>
</div>
```
2. Test different product types (no variants, multiple variants) and custom properties (engraving / no engraving, matte ON/OFF, file / no file)
3. Test in multiple browsers (Safari, IE11, Chrome, Firefox and mobile browsers)
4. Test on both featured product and product page.

Expected results:

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126483791894)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126483791894/editor)

**Checklist**

### DIFF
--- a/sections/cart-notification-product.liquid
+++ b/sections/cart-notification-product.liquid
@@ -12,14 +12,15 @@
       {% endif %}
       <div>
         <h3 class="cart-notification-product__name h4">{{ item.product.title | escape }}</h3>
-        {%- unless item.product.has_only_default_variant -%}
           <dl>
-            {%- for option in item.options_with_values -%}
-              <div class="product-option">
-                <dt>{{ option.name }}: </dt>
-                <dd>{{ option.value }}</dd>
-              </div>
-            {%- endfor -%}
+            {%- unless item.product.has_only_default_variant -%}
+              {%- for option in item.options_with_values -%}
+                <div class="product-option">
+                  <dt>{{ option.name }}: </dt>
+                  <dd>{{ option.value }}</dd>
+                </div>
+              {%- endfor -%}
+            {%- endunless -%}
             {%- for property in item.properties -%}
               {%- assign property_first_char = property.first | slice: 0 -%}
               {%- if property.last != blank and property_first_char != '_' -%}
@@ -38,7 +39,6 @@
               {%- endif -%}
             {%- endfor -%}
           </dl>
-        {%- endunless -%}
         {%- if item.selling_plan_allocation != nil -%}
           <p class="product-option">{{ item.selling_plan_allocation.selling_plan.name }}</p>
         {%- endif -%}


sections/cart-notification-product.liquid
```diff
@@ -20,6 +20,23 @@
                 <dd>{{ option.value }}</dd>
               </div>
             {%- endfor -%}
+            {%- for property in item.properties -%}
+              {%- assign property_first_char = property.first | slice: 0 -%}
+              {%- if property.last != blank and property_first_char != '_' -%}
+                <div class="product-option">
+                  <dt>{{ property.first }}: </dt>
```
ludoboludo, line 27: I wonder if we should ask the translation team about the use of `:`. Does it mean the same thing everywhere ? 🤔

assets/product-form.js
```diff
@@ -12,7 +12,7 @@ if (!customElements.get('product-form')) {
     onSubmitHandler(evt) {
```
KaichenWang, line 12: Could be handled in a separate issue/PR - In Safari (iOS and macOS) the spinner inside the button when adding-to-cart is not visible.

I believe setting an explicit width to the svg fixes this, but requires more testing in all instances where it's being used.
```css
.loading-overlay__spinner svg {
  width: 100%;
}
```

Kapture 2021-10-07 at 15 16 10
e

sections/cart-notification-product.liquid
```diff
@@ -20,6 +20,23 @@
                 <dd>{{ option.value }}</dd>
               </div>
             {%- endfor -%}
+            {%- for property in item.properties -%}
```
KaichenWang, line 23: Since all this markup is inside the `{%- unless item.product.has_only_default_variant -%}`, does that mean if the product has line item properties then `item.product.has_only_default_variant` will output `false`? Asking because otherwise the line item properties will not be displayed for products with only 1 variant?

KaichenWang (approved):
Works great!

sections/cart-notification-product.liquid
KaichenWang: Good question. We actually use colon the same way in a few areas of the theme, I would be curious to know how best to handle this for i18n

assets/product-form.js
LucasLacerdaUX: I'll fix this in a separate issue. Thanks for sharing a potential fix :D

KaichenWang (approved):

ludoboludo:
Hey @M00NSH0T, what you could do is paste it in the `quantity_selector` block on `main-product.liquid` for example. Here is a [screenshot](https://screenshot.click/21-10-09cxd-arprr.png). I also changed the form attribute value to be `form="product-form-{{ section.id }}"` instead of `form="{{ product_form_id }}"`

assets/product-form.js
```diff
@@ -23,11 +23,12 @@ if (!customElements.get('product-form')) {
 
       const config = fetchConfig('javascript');
       config.headers['X-Requested-With'] = 'XMLHttpRequest';
-      config.body = JSON.stringify({
-        ...JSON.parse(serializeForm(this.form)),
-        sections: this.cartNotification.getSectionsToRender().map((section) => section.id),
-        sections_url: window.location.pathname
-      });
+      delete config.headers['Content-Type'];
```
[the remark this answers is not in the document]
LucasLacerdaUX: Oh, it's because we're no longer sending a `Content-Type` `application/json` as part of the header.

The function that returns these config headers is still being used in other endpoints across the theme, and they're still sending JSON requests, so that's why I only removed this from the `/cart` request.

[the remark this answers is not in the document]
ludoboludo: Ah great. Thanks for the explanation 🙂

RedPlugDesign:
@LucasLacerdaUX While experimenting with this in Chrome, I found an odd behavior for the file.  Using dawn and your test setup, the files do pass via AJAX, but if you click those same files, the file saves as a WEBP, despite the actual file extension.  I disabled the AJAX, https://www.screencast.com/t/4AtCSRoVewx, and tried again with a similar file, and when going straight to cart, the file will download as it should.  Firefox was not an issue.

While not a big deal, it will confuse the average customer

aresena:
Hey @RedPlugDesign! I think it saves as a WEBP file because of the Shopify CDN: https://cdn.shopify.com/. It automatically detects which file formats are supported on the client-side and sends the best option

aresena:
@RedPlugDesign True, you need to download it to see it, and it's a WEBP. Good point that some merchants could have difficulties opening it!

shexpire:
Hi @LucasLacerdaUX, is there a way to send multiple products changing the `<input name="id" value="xxxx">` for `<input type="hidden" name="items" value="{id:xxxxx,quantity:1},{id:xxxxx,quantity:1},{id:xxxx,quantity:1}">`, tried but not allowing me, got 400 bad request

MELO-06:
Hey, its been a long time since this pull was created

I was wondering if its possible to change this so the client can upload multiple files and they all show as photos separated by commas

Thanks in advance

poblabs:
This is a great option, thank you for this. However, once an image is uploaded to the cdn, how does the shop owner retain control over it?

Issue #739, opened 2021-10-06
CRITICAL BUG - customer page does not work accross domains


gregjotau:
We have several domains.

If a customer log in to **fammestore.dk**, but go into the order, then they are sent to **famme.no** which is the main domain.

**All links must be relative to the current domain**, now the user is logged out.

![Screenshot 2021-10-06 at 10 48 02](https://user-images.githubusercontent.com/13047497/136170510-14eaee0d-c932-4e75-af74-66477480ff7e.png)

ludoboludo:
Ah interesting! Thanks a bunch for sharing this 👌

gregjotau:
We will also merge it in ASAP to test live 🚀

bakura10:
This fix looks a bit hacky. It would be better if it could be fixed at Shopify level. If tomorrow Shopify allows to change the patterns of the URL we’re screwed up. @Thibaut would it be possible to have this customer variable be locale aware ?

tjoyal:
Had a quick look. Laying down some facts:

- Typically, `_url` methods are relative paths, not actual absolute urls. This one seems to be the exception and is why this specific url cause problem.
- That seems to have always been broken when domains didn't redirect to primary.
- The above is a window that is being greatly expended with the addition of intl domains and markets (pushing merchants to not redirect to primary).

I am hesitant to say what should be done here because I lack some sleep to make a good call 😄 . 
- I feel we can't change the url => path behaviour as for now anything liquid is mostly lifetime support.
- For the time being, I think we could change so the domain is set to use the current domain as I do not see "redirect to primary domain" being a feature, it seems to be a simple bug.
- I think other drops like `CustomerDrop` might be affected by this issue (eg.: reset_password_url)

Skudo:
We shipped a fix for the storefront that creates `order.customer_url` with the current domain. That should address the issue OP was facing.

If we've missed something and we need to have another look at this, please don't hesitate to reopen this issue or create a new one! 🙇‍♂️

PR #758, opened 2021-10-10
Fix missing internationalization on account page on order url

gregjotau:
https://github.com/Shopify/dawn/issues/739


gregjotau:
@ludoboludo testet on fammestore.dk / famme.no, you can login with 

gregjotau@gmail.com
T8nrC4BQhBYgPy3

to test and see that it works.

gregjotau:
To check the fix:

1. login to fammestore.dk
2. See that domain url is still fammestore.dk on order number link

This PR fixes this bug.

gregjotau:
I have already signed the CLA so you have a bug here.

ludoboludo:
Hey @gregjotau, since one of our team is looking at a fix for this issue, we won't be shipping this change you have in the PR. I will keep an eye on the issue they have created for it to see when a fix might ship for it 👍

ludoboludo:
Thanks for bringing up this work around in the mean time 🥇

PR #768, opened 2021-10-13
Version 2.3.0

tyleralsbury:
Version bump


gregjotau:
@tyleralsbury appreciate the version bump. IMO you guys should do the versions bumps and releases frequently since it makes it much easier to follow with notifications: 

Screenshot 2021-10-13 at 21 00 15

PR #773, opened 2021-10-14
Logical properties exploration

ludoboludo:

This is an exploration into using logical properties moving forward so that our theme can be used in [different directions](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) (ltr or rtl) and as well with [different writing modes](https://www.smashingmagazine.com/2019/08/writing-modes-layout/). 

Here are some of the logical properties:
|  New property | Old one   |
|---|---|
| block-size | height |
| inline-size | width |
| border-block-start | `border-top` |
| border-block-end | `border-bottom` |
| border-inline-start | `border-left` |
| border-inline-end | `border-right` |
| margin-block-start | `margin-top` |
| margin-block-end | `margin-bottom` |
| margin-inline-start | `margin-left` |
| margin-inline-end | `margin-right` |
| padding-block-start | `padding-top` |
| padding-block-end | `padding-bottom` |
| padding-inline-start | `padding-left` |
| padding-inline-end | `padding-right` |
| inset-block-start | `top` |
| inset-block-end | `bottom` |
| inset-inline-start | `left` |
| inset-inline-end | `right` |
| inline-start | value for `float/clear` |
| inline-end | value for `float/clear` |
| start | value for `text-align: left` |
| end | Value for `text-align: right` |

I focused on getting RTL to work decently and see where updates are/were needed or not. I did not update values for things like `padding-top/bottom`, `margin-top/bottom`, `border`.
I didn't do any JS updates to fix some of the functionalities either (eg: slider). 

**Other considerations**

Here are a few things I noted/noticed while testing this out: 
- [ ] Ellipsis for text is moved to the left while it should stay on the right for english text. Adding the attribute `dir= auto;` fixes it. `Punctuation characters like ( ) . ! and more are non-directional in Unicode` [source](https://engineering.atspotify.com/2019/04/15/right-to-left-the-mirror-world/).
- [ ] Added some CSS specific to RTL for things like arrow icons, float (logical value for this property isn't well supported). 
- [ ] Slider would need some work so buttons could be working with the new direction and slider number show appropriately.

Here is the support at the moment for logical properties. It's pretty good and feel like it should soon be our standard as we have merchant from all over the world: 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126519050262)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126519050262/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -332,7 +332,7 @@ body,
   top: 0;
   inset-inline-end: 0;
   bottom: 0;
-  left: 0;
+  inset-inline-start: 0;
   z-index: 2;
 }
 
@@ -444,12 +444,12 @@ blockquote {
   font-style: italic;
   color: rgba(var(--color-foreground), 0.75);
   border-left: 0.2rem solid rgba(var(--color-foreground), 0.2);
-  padding-left: 1rem;
+  padding-inline-start: 1rem;
 }
 
 @media screen and (min-width: 750px) {
   blockquote {
-    padding-left: 1.5rem;
+    padding-inline-start: 1.5rem;
   }
 }
 
@@ -595,7 +595,7 @@ h6:empty {
 
 .link-with-icon .icon {
   width: 1.5rem;
-  margin-left: 1rem;
+  margin-inline-start: 1rem;
 }
 
 .circle-divider::after {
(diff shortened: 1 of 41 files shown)


gregjotau:
This might just by my opinion, but how large of the market share using the theme actually need RTL support ? And compare that with using dev time on *a lot* of bugs/improvements for a much larger share of merchants 😁

ironnysh:
Hi there, any news about implementing this fix?

ludoboludo:
Made a newer PR with the same approach. Closing this one.

PR #787, opened 2021-10-18
Mobile navigation focus issue

### LINKED ISSUE #76: [Header] Mobile sub-navigation control does not manage focus
## Issue summary

When navigating the mobile menu, sub-navigation controls do not manage focus on behalf of the user. When a sub-navigation item is activated, focus remains on the previous, now hidden `button` element. As a result, no announcement is made informing the user of a change in UI. Some sighted user groups may be confused when the focus indicator disappears as a result.

<details>

Gif of Dawn mobile menu. Focus remains on the Info control while the UI changes to reveal the hidden sub-menu.

</details>

<!-- ### Current code

#### HTML

```html

``` -->

## Steps to reproduce

1. Go to a shop using Dawn with any supported browser
2. Switch to a mobile break point
3. Activate the mobile menu control
4. Activate the "Info" `button`

## Behavior

### Expected

- Keyboard focus to move to the revealed "Info" `button`.

### Actual

- Focus remains on the parent activator.

## Recommendation

When the sub-navigation is revealed, shift focus to the `.menu-drawer__close-button` control. This will help provide context on the UI update and reveal the current focus position.

<!-- ### Recommended code

#### HTML

```html

``` -->

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **Component:** Header
- **WCAG SC:** [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
- **Severity:** Medium
- **Effort:** Low

### PR DESCRIPTION
**Pull request by ludoboludo**

Fixes #76 

The code in placed looked good. What I found is that the issue was related to the transition of the submenu appearing. So I had to move the `trapFocus()` into a `setTimeout()` function as well. 

It seemed to work with `50`ms but I used `100`ms as I thought it gave us a safe gap. I tried with no value mentioned but it wasn't working sill. I tried on slow 3g and it worked fine. 

**Other considerations**

Open to other suggestions. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126531829782)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126531829782/editor)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -323,7 +323,7 @@ class MenuDrawer extends HTMLElement {
 
     if (detailsElement === this.mainDetailsToggle) {
       if(isOpen) event.preventDefault();
-      isOpen ? this.closeMenuDrawer(summaryElement) : this.openMenuDrawer(summaryElement);
+      isOpen ? this.closeMenuDrawer(event, summaryElement) : this.openMenuDrawer(summaryElement);
     } else {
       setTimeout(() => {
         detailsElement.classList.add('menu-opening');

### CODE REVIEW

File: assets/global.js [hunk 1]
```diff
@@ -308,17 +308,19 @@ class MenuDrawer extends HTMLElement {
     const detailsElement = summaryElement.parentNode;
     const isOpen = detailsElement.hasAttribute('open');
 
+    function addTrapFocus(event) {
```
tyleralsbury, line 317:
```suggestion
    function addTrapFocus() {
```

Event isn't being used here.

tyleralsbury (approved):

tyleralsbury (dismissed):

assets/global.js
```diff
@@ -18,6 +18,7 @@ document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
     event.currentTarget.setAttribute('aria-expanded', !event.currentTarget.closest('details').hasAttribute('open'));
   });
 
+  if (summary.closest('header-drawer')) return;
```
ludoboludo, line 21: I'm adding a check here so it doesn't add the event listener on the summary elements from the menu drawer. As it has it's own behaviour that is tackled in the class declaration (onKeyUp(event)). 

But is it good idea to add a check for a specific element within a function that is general like this one. 

I could also remove the IDs on the `<details` elements from the menu drawer and re add an accessibility function in the menu drawer class that would basically do the same as what's happening in this one.

ludoboludo:
Hmm yeah from what I notice, talkback doesn't trigger the animation of the next menu happen to appear. And since we're waiting for the `transitionend` event, then trapFocus() doesn't end up running. I wonder why in VoiceOver it does trigger things properly but not in TalkBack 🤔 
I will try and see what I can find.

assets/global.js
```diff
@@ -306,24 +307,29 @@ class MenuDrawer extends HTMLElement {
...
+    const reducedMotion = window.matchMedia("(prefers-reduced-motion: reduce)");
+
+    function addTrapFocus() {
+      trapFocus(summaryElement.nextElementSibling, detailsElement.querySelector('button'));
+      summaryElement.nextElementSibling.removeEventListener('transitionend', addTrapFocus);
+    }
 
     if (detailsElement === this.mainDetailsToggle) {
       if(isOpen) event.preventDefault();
       isOpen ? this.closeMenuDrawer(summaryElement) : this.openMenuDrawer(summaryElement);
```
tyleralsbury, line 326: Since you changed `closeMenuDrawer` to take an `event` as the first argument, I think you need to add the event here as well:

```suggestion
      isOpen ? this.closeMenuDrawer(event, summaryElement) : this.openMenuDrawer(summaryElement);
```

ludoboludo: 👍

PR #791, opened 2021-10-19
Fix product slider on mobile

### LINKED ISSUE #665: Product Page: Incorrect variant image displaying on mobile

On Android mobile devices, selecting a new variant will display the incorrect image on the image slider.

alt text

Video - https://screenshot.click/21-09-94gqm-qo0eh.gif

The first variant image displays correctly but then it seems it stops updating and starts again but with the previously selected variant's image instead of the current one.

The image slider is expected to display the corresponding variant image when a new variant is selected on mobile.

- Dawn Version: 2.1.0 (replicable on 1.1.0)
- Android devices only. IOS working as expected (Used browserstack to test on multiple Android devices, using the inspect tool on desktop did not produce the issue)

Video - https://screenshot.click/21-09-94gqm-qo0eh.gif

ludoboludo:

Fixes #665 

There is some inconsistency for whatever reason when you're clicking a variant with an attached image on android devices. It's supposed to scroll it into the view, right after the image is pushed to the beginning of the product media list.

So i added another scrolling action: `parent.scrollLeft = 0;`. 

I kept `scrollIntoView` as I still need it on desktop and it's working fine there. My guess is that there is a hiccup when we're doing `parent.prepend(newMedia);` and then trying to scroll that element into the view. Some of the timing might be off and it's messing the functionality of the slider. 

**Other considerations**

I tried to see if using `parent.firstChild` instead of `parent.querySelector(...` would help but it didn't change anything. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126536024086)

**Testing**

In order to test you'll need to check the preview from an android device. The issue wasn't happening using only the inspect tools. 

You can search for the `sandal` product and click the different variant colours to see if there is any issue when it's supposed to scroll you to the variant image associated to the colour. 

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -601,7 +601,7 @@ class VariantSelects extends HTMLElement {
     }
     window.setTimeout(() => {
       parent.scrollLeft = 0;
-      parent.querySelector('li.product__media-item').scrollIntoView({behavior: "smooth"});
+      parent.querySelector('li.product__media-item').scrollIntoView({behavior: 'smooth'});
     });
   }
 


tyleralsbury (dismissed):
Tested on my Google Pixel and it worked nicely.

I tried while on the first image and further down the line, it seems to always work. 🎉

assets/global.js
```diff
@@ -599,7 +599,10 @@ class VariantSelects extends HTMLElement {
     if(this.stickyHeader) {
       this.stickyHeader.dispatchEvent(new Event('preventHeaderReveal'));
     }
-    window.setTimeout(() => { parent.querySelector('li.product__media-item').scrollIntoView({behavior: "smooth"}); });
+    window.setTimeout(() => {
+      parent.scrollLeft = 0;
+      parent.querySelector('li.product__media-item').scrollIntoView({behavior: "smooth"});
```
KaichenWang, line 604: I think removing the `{behavior: "smooth"}` would also fix this issue. I feel this might have to do with limited support of the `smooth` behaviour option. Some browsers may just ignore the option, but others (older versions of Android) may not initiate the scroll altogether.

Screen Shot 2021-10-19 at 4 38 56 PM

KaichenWang (dismissed):

KaichenWang (approved):

assets/global.js
```diff
@@ -599,7 +599,10 @@ class VariantSelects extends HTMLElement {
     if(this.stickyHeader) {
       this.stickyHeader.dispatchEvent(new Event('preventHeaderReveal'));
     }
-    window.setTimeout(() => { parent.querySelector('li.product__media-item').scrollIntoView({behavior: "smooth"}); });
+    window.setTimeout(() => {
+      parent.scrollLeft = 0;
+      parent.querySelector('li.product__media-item').scrollIntoView({behavior: 'smooth'});
```
KaichenWang, line 604: @ludoboludo Sorry for revisiting this, but if we are to keep `parent.scrollLeft = 0;` as a fallback, would it make sense to have it after `scrollIntoView()`? This way on mobile if the `behavior: 'smooth'` option is supported, the smooth scrolling will occur first and if it isn't supported, `parent.scrollLeft = 0` will kick in. Having it the opposite way prevents the `behavior: 'smooth'` from ever being effective?

```suggestion
      parent.querySelector('li.product__media-item').scrollIntoView({behavior: 'smooth'});
      parent.scrollLeft = 0;
```

KaichenWang: Nevermind, in my testing I'm not really seeing a difference 😅

okaykomputer:
Hey folks, when I test this fix on Dawn 2.3.0 (merchant's theme and fresh install in my test store | Chrome) the `scrollIntoView()` method doesn't function consistently on mobile screens (using dev tools on MacBook Pro or mobile device). [Video here](https://screenshot.click/29-41-76th1-3uod0.mp4). 

[Test product here with issue](https://kid-for-today.myshopify.com/products/rossignol-circuit-amptek-snowboard-2016?variant=19557369249814)

Did anyone else experience this happening with the fix above?

okaykomputer:
Hey @ludoboludo , the issue does not replicate in 2.4.0 and after because (I'm presuming) the code for the product image slider was removed and redone using the code for the new slideshow section that was recently added in 2.4.0.

PR #793, opened 2021-10-19
Image banner buttons alignment

### LINKED ISSUE #734: [Image Banner] Buttons aligned left when two linked buttons and text block are added
On Dawn 2.2, when you add a second button, the buttons are aligned correctly. However, when you add a text block, the buttons are aligned left, even though the element still have `justify-content: center` active.

The buttons should stay centered in their container

**Possible solution**
Can fix by removing the `max-width: 45rem` from the buttons' container, or by using `margin: 0 auto` to center the buttons.

I've tested on Dawn 2.0 and 2.1 and the behaviour is expected. Something seems to have changed on Dawn 2.2 that is causing this bug. It should also be noted that this will only happen when a link is added to the second button.

**Expected behaviour**
[Image banner working normally](https://screenshot.click/05-17-b4i4z-9rxy1.png)

**Issue**
[image with banner section with arrows pointing to settings and issue](https://screenshot.click/05-43-dk6vp-tfs0o.png)

@Shopify/themes-fed @Shopify/theme-support

ludoboludo:

Fixes #734 

Added `margin-left: auto;`,`margin-right: auto;` on the parent element of the buttons. 

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126537367574)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126537367574/editor)

**Checklist**

### DIFF
--- a/sections/image-banner.liquid
+++ b/sections/image-banner.liquid
@@ -91,7 +91,7 @@
               <span>{{ block.settings.text | escape }}</span>
             </div>
           {%- when 'buttons' -%}
-            <div class="banner__buttons{% if block.settings.button_label_1 != blank and block.settings.button_link_1 != blank and block.settings.button_label_2 != blank and block.settings.button_link_2 != blank %} banner__buttons--multiple{% endif %}" {{ block.shopify_attributes }}>
+            <div class="banner__buttons{% if block.settings.button_label_1 != blank and block.settings.button_label_2 != blank %} banner__buttons--multiple{% endif %}" {{ block.shopify_attributes }}>
               {%- if block.settings.button_label_1 != blank -%}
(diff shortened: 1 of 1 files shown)


kmeleta (commented):
This definitely fixes the centering issue 👍 , but I suspect we hadn't noticed this ourselves because when these buttons don't have links, they're actually positioned differently and don't create the alignment issue in this case ([see here](https://screenshot.click/20-00-rofr3-kfoq0.mp4))

I think we should change the logic around adding the `banner__buttons--multiple` class so that these buttons display/align the same whether they have links or not. It currently only gets added when both buttons have links. Let me know if you disagree or want to handle in another issue though.

PR #807, opened 2021-10-20
Fix variant spacing on product page

### LINKED ISSUE #677: Product Page: Media slider when only 1 media
When on the product page there seems to be inconsistent padding around the featured media (especially noticeable on mobile). 

This issue is visible when the ‘Hide other variants’ media after selecting a variant’ setting is checked, and there is only one media. [Screenshot](https://screenshot.click/21-09-895ce-bpi4d.png)

Steps to reproduce: 

Create a product with different variants.
Assign images to variants, but don't add any additional image.
In the product info section setting, enable the ‘Hide other variants’ media after selecting a variant’ setting
Observe the inconsistent padding + usage of a slider even though there is only one image surfaced.

sofiamatulis:

Fixes #677 .

Add spacing to the right on mobile. Details on how to reproduce: https://github.com/Shopify/dawn/issues/677

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126542675990/editor)

You can test it with the product `Shoe Test variant product`

Before: https://screenshot.click/21-09-895ce-bpi4d.png
After: https://screenshot.click/20-44-jd5ry-qvyq1.png

**Checklist**

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -463,7 +463,6 @@ a.product__text {
 
 .product__media-item--variant:first-child {
   display: block;
-  padding-right: 1.5rem;
 }
 
 @media screen and (max-width: 749px) {
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -466,6 +466,13 @@ a.product__text {
   padding-right: 1.5rem;
 }
 
+@media screen and (max-width: 749px) {
+  .product__media-item--variant:first-child {
+    padding-right: 1.5rem;
+  }
+
+}
+
 @media screen and (min-width: 750px) and (max-width: 989px) {
   .product__media-list .product__media-item:first-child {
     padding-left: 0;
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -463,6 +463,7 @@ a.product__text {
 
 .product__media-item--variant:first-child {
   display: block;
+  padding-right: 1.5rem;
 }
 
 @media screen and (min-width: 750px) and (max-width: 989px) {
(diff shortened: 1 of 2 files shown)


LucasLacerdaUX:
@sofiamatulis The issue with this approach is that products that doesn't have every media attached to a variant would get the slider disabled.

An example: https://os2-demo.myshopify.com/products/brick-apple
This product has 8 images, but only 2 of them are attached to variants. After enabling the setting, we'd still end up with 7 images, so we'd still want these buttons to be displayed.

If you look at the slider counter, it's already using a logic similar to what I suggested above:
```liquid
            <span class="slider-counter--total">{% if section.settings.hide_variants %}{{ product.media.size | minus: variant_images.size | plus: 1 }}{% else %}{{ product.media.size }}{% endif %}</span>
```

So ideally, we would move this `product.media.size | minus: variant_images.size | plus: 1` up as a variable and use it on the `if product.media.size > 2` instead.

LucasLacerdaUX (approved):
Looks perfect now! :D

Tested on:
- [Granola Box](https://os2-demo.myshopify.com/products/granola-bar?_pos=1&_sid=b1b89266b&_ss=r) (1 image, no variants)
- [Brick](https://os2-demo.myshopify.com/products/brick-apple) (2 variants, not every image associated with variants)
- [Shoe Test variant product](https://os2-demo.myshopify.com/products/test-variant-product) (2 variants, all images associated with variants)
- [Naomi Mini Wallet](https://os2-demo.myshopify.com/products/copy-of-naomi-mini-wallet-st-tropez) (no variants)

They all work as expected with the setting ON and OFF!

PR #808, opened 2021-10-20
Ensure "share" grabs the variant of the product

### LINKED ISSUE #504: Share button doesn't seem to link to a variant

When clicking on the `share` button on the product page, it seems to only grab the product page's url but not the variant specific one. 

Should it be linking to the specific variant you're looking at ? 

- Dawn Version: 1.0.0
- macOS Version 11.3.1

**Possible solution**

Im not sure if it should be added or not as we would need to update the url when we detect a variant change and add some extra logic. 

sofiamatulis:

Fixes #504 .

1) First I ensured that there was a variant and if there was one I should be ensuring to add it to the url with this in liquid:

```
{%- liquid 
  assign share_link = product.url
   if product.selected_variant.url != blank
     assign share_link = product.selected_variant.url
    endif
-%}

```
2) This became an issue with JS as the url updates when the variant updates when you click on the button. Because of that I created a function to update the input

**Other considerations**

This works only with JS and it's for featured product and the product page.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=126544478230)

**Checklist**

### DIFF
--- a/assets/share.js
+++ b/assets/share.js
@@ -42,6 +42,7 @@ if (!customElements.get('share-button')) {
     }
 
     updateUrl(url) {
+      this.urlToShare = url;
       this.elements.urlInput.value = url;
     }
   });


sections/featured-product.liquid
```diff
@@ -139,13 +139,21 @@
...
+                      -%}
                       <input type="text"
                             class="field__input"
                             id="url"
-                            value="{{ shop.url | append: product.url }}"
+                            value="{{ shop.url | append: share_link }}"
                             placeholder="{{ 'general.share.share_url' | t }}"
                             onclick="this.select();"
                             readonly
+                            share-input
```
svinkle, line 155: This generates invalid HTML. I'd suggest applying a unique `id` and use this as the selector.

svinkle:
Just trying this out and I'm not sure if the generated URL is correct? Specifically the `oseid` here:

`https://os2-demo.myshopify.com/?oseid=5GxAMz6K9Yzo5dD7xd9XBuR2/products/test-3d-models-video?variant=39328407453718`

sofiamatulis:
> Just trying this out and I'm not sure if the generated URL is correct? Specifically the `oseid` here:
> 
> `https://os2-demo.myshopify.com/?oseid=5GxAMz6K9Yzo5dD7xd9XBuR2/products/test-3d-models-video?variant=39328407453718`

@svinkle I am not seeing the `oseid` 🤔  Do you mind sharing a video? This is what I see: https://screenshot.click/21-33-u88wy-d63wh.mp4

svinkle:
Dawn product page. User adjusts variants then clicks Share button. URL includes ?osedi parameter in the middle of the URL string.

sections/featured-product.liquid
```diff
@@ -139,13 +139,21 @@
...
+                        assign share_link = product.url
+                        if product.selected_variant.url != blank
+                          assign share_link = product.selected_variant.url
+                        endif
+                      -%}
                       <input type="text"
                             class="field__input"
                             id="url"
-                            value="{{ shop.url | append: product.url }}"
+                            value="{{ shop.url | append: share_link }}"
```
KaichenWang, line 151: I think we can avoid the `assign` above if we use the `default` filter
```suggestion
                            value="{{ product.selected_variant.url | default: product.url | prepend: shop.url }}"
```

assets/global.js
```diff
@@ -615,6 +616,12 @@ class VariantSelects extends HTMLElement {
     window.history.replaceState({ }, '', `${this.dataset.url}?variant=${this.currentVariant.id}`);
   }
 
+  updateShareUrl() {
+    if (!this.currentVariant) return;
+    console.log(document.querySelector("[data-share]").dataset, document.querySelector("[data-share]"))
```
KaichenWang, line 621:
```suggestion
```

assets/global.js
```diff
@@ -615,6 +616,12 @@ class VariantSelects extends HTMLElement {
...
+    document.querySelector("[data-share]").value = `${ document.querySelector("[data-share]").dataset.url}${this.dataset.url}?variant=${this.currentVariant.id}`;
```
KaichenWang, line 622: If there are multiple featured product sections on a page, `document.querySelector("[data-share]")` will return the first node that matches the selector, which is not necessarily the correct one

sofiamatulis:
@svinkle I apologize, I realized I linked the editor instead of the store. You can test it here: https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=126544478230

Also updated the link in the description

assets/global.js
sofiamatulis: Oh yes, good catch! I saw that it only allows for one `share` block, but I forgot about this scenario with the featured product. I will add a section id

PR #810, opened 2021-10-21
Add role to empty links

### LINKED ISSUE #387: [Global] Disabled links are still actionable
## Issue summary

When a link is created without a destination, it's still actionable via mouse and keyboard. Activating the link refreshes the page. This may cause confusion for multiple user groups who may think activation will load a new resource.

<details>

Dawn theme homepage. Disabled link with dimmed text is highlighted.

</details>

#### HTML

```html
<a
  href=""
  class="button button--secondary"
  aria-disabled="true"
  …
>
  Purposefully broken
</a>
```

## Steps to reproduce

1. Go to a Dawn homepage that contains a "disabled" link with any supported browser
2. With the keyboard, mouse, or active screen reader, navigate to and activate the disabled link

### Expected

- "Disabled" link to not be focusable or actionable via mouse/keyboard

### Actual

- "Disabled" link remains focusable and actionable

## Recommendation

In order to create a disabled link which is not actionable yet retains its semantic meaning:

1. Remove the `href` attribute. This will cause the link to not receive focus or be actionable via mouse click/<kbd>Enter</kbd> key.
2. Add the `role` attribute, setting its value to `link`. This will provide semantic meaning for assistive technology.

Read more from the post [Disabling a link](https://www.scottohara.me/blog/2021/05/28/disabled-links.html) from @scottaohara.

#### HTML

```html
<a
  class="button button--secondary"
  aria-disabled="true"
  role="link"
  …
>
  Purposefully broken
</a>
```

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **WCAG SC:** [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
- **Severity:** High
- **Effort:** Low

### PR DESCRIPTION
**Pull request by kmeleta**

Fixes #387 

Disabled/empty links should have `aria-disabled="true"` and `role="link"` instead of a `href`. I found `image-banner`, `image-with-text`, and `rich-text` needed the role. I didn't find any other areas where we render links with no href.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126547558422/editor)

**Checklist**

### DIFF
--- a/sections/collection-list.liquid
+++ b/sections/collection-list.liquid
@@ -29,7 +29,7 @@
 
       {%- for block in section.blocks -%}
         <li class="collection-list__item grid__item{% if section.settings.swipe_on_mobile %} slider__slide{% endif %}" {{ block.shopify_attributes }}>
-          <a{% if block.settings.collection != blank and block.settings.collection.all_products_count > 0 %} href="{{ block.settings.collection.url }}"{% endif %}
+          <a{% if block.settings.collection != blank and block.settings.collection.all_products_count > 0 %} href="{{ block.settings.collection.url }}"{% else %} role="link" aria-disabled="true"{% endif %}
             class="card animate-arrow link{% if block.settings.collection.featured_image != blank %} card--media{% else %}{% if section.settings.image_ratio != 'adapt' %} card--stretch{% endif %}{% endif %}{% unless section.settings.image_padding %} card--light-border{% endunless %}"
           >
             <div class="card--stretch card-colored color-{{ section.settings.color_scheme }}">

### CODE REVIEW

**Comment by svinkle** [comment]
@kmeleta Also check `multicolumn`

**Review by svinkle — APPROVED** [review]

File: sections/image-banner.liquid [hunk 1]
```diff
@@ -93,10 +93,10 @@
           {%- when 'buttons' -%}
             <div class="banner__buttons{% if block.settings.button_label_1 != blank and block.settings.button_label_2 != blank %} banner__buttons--multiple{% endif %}" {{ block.shopify_attributes }}>
               {%- if block.settings.button_label_1 != blank -%}
-                <a{% if block.settings.button_link_1 != blank %} href="{{ block.settings.button_link_1 }}"{% endif %} class="button{% if block.settings.button_style_secondary_1 %} button--secondary{% else %} button--primary{% endif %}"{% if block.settings.button_link_1 == blank %} aria-disabled="true"{% endif %}>{{ block.settings.button_label_1 | escape }}</a>
+                <a{% if block.settings.button_link_1 == blank %} role="link" aria-disabled="true"{% else %} href="{{ block.settings.button_link_1 }}"{% endif %} class="button{% if block.settings.button_style_secondary_1 %} button--secondary{% else %} button--primary{% endif %}">{{ block.settings.button_label_1 | escape }}</a>
```
LucasLacerdaUX, line 96: Just a pattern we can apply here.

Since there's always going to be something being outputted by this condition, we can move the whitespace before the `{% if %}`.

```suggestion
                <a {% if block.settings.button_link_1 == blank %}role="link" aria-disabled="true"{% else %}href="{{ block.settings.button_link_1 }}"{% endif %} class="button{% if block.settings.button_style_secondary_1 %} button--secondary{% else %} button--primary{% endif %}">{{ block.settings.button_label_1 | escape }}</a>
```

This would apply to every change in this PR.

LucasLacerdaUX (approved):
Looks good to me. Just left a code style suggestion if you think it makes sense :P

sections/image-banner.liquid
kmeleta: Ah yeah good observation, I'm not sure I feel too strongly about it though. I think I'd have to either move the class attr in front of the others or put it on a new line or else risk a double space. So I don't feel we're gaining much from it.

PR #811, opened 2021-10-21
Add unit price to cart

### LINKED ISSUE #497: Display line item price in Cart
We recently cleaned up the cart page from the price column because we felt there was a lot of information gymnastics on this page for buyers to parse - and believe the total column provides the price information in most cases. 

For some usecases however, still seeing the price of an individual unit of a line item may be necessary. 

image

sofiamatulis:

Fixes #497 .

Followed what we use for the total column, but used `original_price` and `final_price` instead of `final_line_price` and `original_line_price`

**Other considerations**

Unit was already living under the total, so I kept it there

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=126548672534)

**Checklist**

### DIFF
--- a/sections/main-cart-items.liquid
+++ b/sections/main-cart-items.liquid
@@ -111,7 +111,7 @@
                           {{ 'products.product.price.regular_price' | t }}
                         </dt>
                         <dd>
-                          <s class="cart-item__old-price">
+                          <s class="cart-item__old-price product-option">
                             {{ item.original_price | money }}
                           </s>
                         </dd>
--- a/sections/main-cart-items.liquid
+++ b/sections/main-cart-items.liquid
@@ -70,14 +70,14 @@
                           {{ 'products.product.price.regular_price' | t }}
                         </span>
                         <s class="cart-item__old-price product-option">
-                          {{ item.original_price | money }}
+                          {{- item.original_price | money -}}
                         </s>
                         <span class="visually-hidden">
                           {{ 'products.product.price.sale_price' | t }}
                         </span>
-                        <strong class="product-option">
+                        <span class="product-option">
                           {{ item.final_price | money }}
-                        </strong>
+                        </span>
                       </div>
                     {%- else -%}
                       <div class="product-option">
--- a/sections/main-cart-items.liquid
+++ b/sections/main-cart-items.liquid
@@ -65,26 +65,24 @@
                     <a href="{{ item.url }}" class="cart-item__name h4 break">{{ item.product.title | escape }}</a>
 
                     {%- if item.original_price != item.final_price -%}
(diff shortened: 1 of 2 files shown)


sections/main-cart-items.liquid
```diff
@@ -105,6 +105,28 @@
                         </li>
                       {%- endfor -%}
                     </ul>
+                    {%- if item.original_price != item.final_price -%}
+                      <dl class="cart-item__discounted-prices">
+                        <dt class="visually-hidden">
+                          {{ 'products.product.price.regular_price' | t }}
+                        </dt>
+                        <dd>
+                          <s class="cart-item__old-price product-option">
```
sofiamatulis, line 114: I used the existing class `product-option` to ensure the styles are the same. This is not technically an option so maybe we can rename it or create a new class with the same styles? Any thoughts?

sections/main-cart-items.liquid
```diff
@@ -105,6 +105,28 @@
...
+                        </dt>
+                        <dd class="product-option">
+                          {{ item.final_price | money }}
+                        </dd>
+                      </dl>
+                    {%- else -%}
+                      <span class="product-option">
+                        {{ item.original_price | money }}
+                      </span>
+                    {%- endif -%}
```
ludoboludo, line 129: Shouldn't this be before the product options/variants ?

vs

sections/main-cart-items.liquid
```diff
@@ -64,6 +64,27 @@
...
+                        <span class="visually-hidden">
+                          {{ 'products.product.price.sale_price' | t }}
+                        </span>
+                        <strong class="cart-item__final-price product-option">
+                          {{ item.final_price | money }}
+                        </strong>
+                      </div>
+                    {%- else -%}
+                      <div class="product-option">
+                        {{ item.original_price | money }}
```
ludoboludo, line 84: I wonder if we should show the compare at price as well ? 🤔 And how it would work when there is both a discount and a compare at price.
Show the discount and if there is no discount applied then check if there is a compare at price on line 84.

sofiamatulis: @ludoboludo Oh I see what you mean 🤔 I created an issue and tagged you and Meli: https://github.com/Shopify/dawn/issues/854

Feel free to add any comments

kmeleta (approved):

sections/main-cart-items.liquid
melissaperreault: > And how it would work when there is both a discount and a compare at price.

For that specific reason, we are not showing the original price on cart to avoid confusion. We don't show the original price either at Checkout. I'll try to find more docs related to that topic.

PR #817, opened 2021-10-22
Remove lazyload attribute from feat. product image

ludoboludo:

Looking into the impact of having the featured image of the main-product page 

Just a simple `if` statement running around the attribute but it's not the perfect approach as it impacts the feat. product section as well. 
I'm also using the attribute `decode="async"` when not using `loading="lazy"`. [More info about decode](https://www.smashingmagazine.com/2021/04/humble-img-element-core-web-vitals/#image-decoding).
So if the benefits were deemed valuable enough here, then we need to update the approach to be specific to the main-product image and not the feat. product section. 

We would also need to apply the same logic within the `noscript` elements. 

**Other considerations**

Here are some of the observations I noted: 

| with lazyload  | without lazyload |
| ------------- | ------------- |
| Image loads at: 1s  | Image loads at: 0.7s  |
| Discovered: 0.849s  | Discovered: 0.565s  |
| Time to first bite: 43ms  | Time to first bite: 56ms  |
| Content download: 51ms  | Content download: 85ms  |
| Largest contentful paint: 1s  | Largest contentful paint: 0.87s  |

Here is a screenshot comparison of web page test for a product page on my test store: 

Running the test of the same page a few times is going to give slightly different results. So this is to take with a grain of salt. 

**Demo links**

To run your own test, you need a store that isn't password protected. And add the changes from this PR into the same file within your admin. 

The theme needs to be published, then grab the url of a product page and go test it in [webpagetest.org](webpagetest.org)

**Checklist**

### DIFF
--- a/snippets/product-thumbnail.liquid
+++ b/snippets/product-thumbnail.liquid
@@ -83,9 +83,11 @@
         {{ media.preview_image | img_url: 'master' }} {{ media.preview_image.width }}w"
       src="{{ media | img_url: '1500x' }}"
       sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | times: 0.64 | round }}px, (min-width: 750px) calc((100vw - 11.5rem) / 2), calc(100vw - 4rem)"
-      {% unless position == 'featured' %}
+      {% if position == 'featured' %}
+        decode="async"
+      {% else %}
         loading="lazy"
-      {% endunless %}
+      {% endif %}
       width="576"
       height="{{ 576 | divided_by: media.preview_image.aspect_ratio | ceil }}"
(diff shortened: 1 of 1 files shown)


ludoboludo:
Hey @baldurh this is something we are still interested in exploring/fixing. 

It's hard to assume usually what is going to come first on the page as you can now add sections (everywhere) above your the product page content. Though it still seems rare enough of a pattern that we could get away with it and it would be beneficial to mostly everyone. 

I had closed this as it was more of an exploration but some work is being done on the product page still.

So don't think of this as if we were not going to look back it and try come with a better way to deal with it. 👍

baldurh:
Makes sense. Thanks 😊

PR #831, opened 2021-10-26
Show 4 entries in a single row for a collection list

### LINKED ISSUE #781: Collection List with 4 entries should display them in a single row

The collection list shows the wrong number of columns on tablet/desktop, specifically:
* 1 collection: Correct, 100% width.
* 2 collections: Correct, 50% width.
* 3 collection: Correct, 33.33% width.
* 4 collections: Incorrect, 50% width.
* 5+ collections: Correct, 33.33% width.

When displaying exactly 4 items in a collection list, they should be displayed in a single row on a tablet and desktop. This also matches the logic provided in the template (`collection-list.liquid`), which creates custom classes for up to 4 items (and then returns to 3 items per row):
https://github.com/Shopify/dawn/blob/main/sections/collection-list.liquid#L18

- Dawn Version: 2.3.0 (actually, latest from origin/main)
- Chrome Version 94.0.4606.81
- Windows Version 10

**Possible solution**

This would fix it, happy to create this as pull request:
https://github.com/thoean/dawn/commit/621248b4a819ed1a94ad77d40a3b6ba3d7bcf573

_Screenshots from behavior before:_

3 items are correctly rendered as 3 columns:
Screenshot of 3 items

4 Items are incorrectly reverted to just show 2 columns next to each other:
Screenshot of 4 items

_Screenshots from behavior after suggested code change:_

A collection list with exactly 4 entries is displayed in a single row:
Screenshot of 4 items after code change

thoean:

Fixes #781. The issue contains a very detailed description and screenshots.

Identify the problem that setting grid positions should be done on the parent (grid) level, not on the individual item level. Verified this is aligned with the original intent of the CSS class name.

**Other considerations**

I initially tried to set `.grid__item:nth-child(n)` (to avoid the gaps with `n + 4`, but quickly figured out that's the wrong part to fix this.

**Demo links**

See https://thesmallbatchproject.ch/en/collections (you'll need to use the developer consoles to delete a few <li> items to get exactly 4 items).

**Checklist**

### DIFF
--- a/assets/section-collection-list.css
+++ b/assets/section-collection-list.css
@@ -98,14 +98,16 @@
     box-shadow: none;
   }
 
+  .collection-list.grid--2-col-tablet .grid__item {
+    max-width: 50%;
+  }
+
   .collection-list.grid--3-col-tablet .grid__item {
     max-width: 33.33%;
   }
 
-  .collection-list--4-items .grid__item,
-  .collection-list--7-items .grid__item:nth-child(n + 4),
-  .collection-list--10-items .grid__item:nth-child(n + 7) {
-    width: 50%;
+  .collection-list.grid--4-col-tablet .grid__item {
+    max-width: 25%;
   }
 }
 


ludoboludo:
We've added more settings that let you control how many items per row you can use. I will close this PR and it can be re opened if needed.

PR #837, opened 2021-10-28
Fix counter for swipe on mobile section when section has html in the…

EmersonEmerson:

Within the customizer, for multiple sections (multi-column for example), if a user creates an HTML list within the text field (description field in multi-column) and has the "Enable swipe on mobile" setting enabled, the counter for the slider on mobile will show incorrect current/total slide numbers. 

You can see here in this multi column I have 3 columns added:
Screen Shot 2021-10-27 at 7 50 40 PM

But if I add lists in the body text like so, you'll notice the counter is now reading 17 despite 3 actual slides. The current slide counter eg 1/3, 2/3, 3/3 will also be incorrect but only on the last side.

Screen Shot 2021-10-27 at 7 47 56 PM

I understand inputting html into the text fields isn't optimal to do but it is possible and as such users will use it, hence how I stumbled upon this.

Fixes #0.

I simply added the parent class for the slide element into the logic for the counter so other child `<li>` elements won't be included in the count:

```
// global.js

// before
this.sliderItems = this.querySelectorAll('li');

// after
this.sliderItems = this.querySelectorAll('li.slider__slide');
```

**Other considerations**

The collection-list, featured-blog, featured-collection, and multicolumn sections all seem to be affected by this.

**Checklist**
- [X] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [X] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [X] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [X] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -497,7 +497,7 @@ class SliderComponent extends HTMLElement {
   constructor() {
     super();
     this.slider = this.querySelector('ul');
-    this.sliderItems = this.querySelectorAll('li');
+    this.sliderItems = this.querySelectorAll('li.slider__slide');
     this.pageCount = this.querySelector('.slider-counter--current');
     this.pageTotal = this.querySelector('.slider-counter--total');
     this.prevButton = this.querySelector('button[name="previous"]');


ludoboludo:
Hey @EmersonEmerson 👋 

Thanks a lot for bringing this to our attention and contributing to make the theme better. 

What you're showing makes me wonder if we should use `strip_html` to prevent it from happening 🤔  I'll chat with the team and see what they think.

But either way, I'm working on a slideshow section where I changed the selector, kind of like you did but using IDs instead 🙂 

So it should be fixed in an upcoming release once the slideshow section is merged.

PR #840, opened 2021-11-01
Slideshow section

### LINKED ISSUE #833: New Slideshow section
## Context

Merchants have high expectations in terms of layout flexibility and customization, and this is proven through support tickets. Slideshows weren't a UX pattern we wanted to continue to encourage because the more slides you add to a slideshow, the less likely a buyer is to consume the content, but merchants still want to have this kind of section for various reasons and goals.

In the meantime we do more education, we want to offer a new Slideshow section to meet merchant’s needs for this section type.

[Internal Figma reference](https://www.figma.com/file/XVGy0rXJhYSZoDh031jLNj/Slideshows-Design?node-id=151%3A53130)

### Behaviour details

**General**
- [x] The auto-rotating slides should stop when user interacts with the slider arrows or visual UI controls + Paused on hover.
- [x] Maximum of slides (block) - to determine 5 or 6. See questions at the end.
- [x] Similar settings from Image banner section will live in this new section but we won't display 2 buttons, just 1 per slide instead.
- [x] Support swipe gesture on mobile.
 

**No JS**
- [x] Slides won't auto-rotate when JS is disabled when the settings is set to `true`.

### New settings details

**Section level**

- [ ] Layout
   - [x] Full width (default)
   - [x] Grid
   - [ ] Grid with a peek
- [x] Slide height = Same height options as from `Image banner` section.
   - [x] Adapt to first image (default)
   - [x] Small
   - [x] Medium
   - [x] Large
- [x] Slider visual
   - [x] Counter 
   - [x] Dots
   - [x] Numbers (default)
- [x] Heading font size
   - [x] Medium (default)
   - [x] Large
- [x] Desktop text position
   - [x] Top left
   - [x] Top center 
   - [x] Top right
   - [x] Middle left
   - [x] Middle center (default)
   - [x] Middle right
   - [x] etc...
- [x] Show text box on desktop (`true` by default)
- [x] Image overlay opacity (range from 0% to 100%, default 0%. Increments of 10)
- [x] Auto-rotate slides (`false` by default)
- [x] Change slides every (range from 3s to 9s, default 5s. Increments of 2)
- [x] Mobile layout
   - [x] Show text below on mobile (`true` by default)

**Block settings**
- [x] Image (image picker)
- [x] Heading (Default: Image slide)
- [x] Subheading (Default: Tell your brand's story through images)
- [x] Button label (Default: Shop all)
- [x] Button link (Default: empty) + (Info text below the select: `Leave the label blank to hide the button.`)
- [x] ~Use outline button style (`false` by default)~
- [x] Box color scheme (Info text below the select: `Visible when text box displayed.`)
   - [x] Accent 1 (default)
   - [x] Accent 2
   - [x] Background 1
   - [x] Background 2
   - [x] Inverse

## Additional considerations/questions UX to answer

- [ ] To follow up the need to create Slideshow guidelines [similar to Debut](https://help.shopify.com/en/manual/online-store/themes/os/themes-by-shopify/debut/sections#slideshow)
- [x] Should heading font size have `Small` option in the list like Rich text section?
- [x] Do we want to keep maximum of slides to 5 like Debut or have 6? Checkout [internal reference](https://docs.google.com/document/d/1SkIEDRg5SdkTB0CpR80cRq2cPwnLq7BlDRXe-w_-Pfc) for latest decisions on Slideshow from the past.
- [x] Does `Desktop text box position` setting also drive the text alignment inside the box? On mobile, the text will always be center-aligned
- [ ] Investigating the position of the pause still.
- [x] Do we want to keep `Use outline button style` setting? I believe we have it for image banner that has 2 buttons, but no where else. **_Answer: No!_**

## UX Checklist

 - [x] Scott reviewed the accessibility annotations
 - [ ] Open an issue for Docs update

ludoboludo:

Fixes #833

I reused quite a bit of the styling for the image banner as if you have only one slide the look is very similar. Section settings are also very similar so it seemed to make sense. 

I create a new JS class that extends the functionality of the existing SliderComponent. 

**Other considerations**

- [x] I added an accessibility setting for merchants to give a TLDR about the purpose of the slideshow
- [x] Navigation button size. On mobile they're not set to be 44x44 but I think they probably should. But it would have an impact on the design: [screenshot](https://screenshot.click/21-11-g8ets-rv9h6.png) (play button isn't active on this example). 
- [ ] Grid peek style isn't added at the moment. I thought I would add it as a follow up.
- [x] Tab order on mobile isn't great if we display the controls in between the text image and text box. Though it could be re arranged in the structure, I would have to duplicate some styling as the pattern so far has been to have it after the slider. 

Issue: when editing blocks/slides it won't show you the current slide you're editing in the theme editor.

Potential benefits to add: 
- The slideshow is currently using the `image-banner.css` file as they share quite a bit of styling. But the code in common could be put into its own file as around 40% of the `image-banner.css` isn't necessary to the slideshow. 
- Look into using `content-visibility`. How does it behave if used in browser that don't support the property ? Remove lazy loading of the first slide to load faster when it's the first section and rely on content-visibility to delay the loading when it isn't. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126715396118) 
- [Editor](https://os2-demo.myshopify.com/admin/themes/126715396118/editor) 

**Checklist**
- [x] Request new documentation for the help docs.


svinkle:
@ludoboludo Looking good. I left a few notes.

> Should clicking the number or dots from the slideshow navigation set the focus on the slide ?

No, since they're `button` controls which imply focus remains on the control. The `aria-live` announces the newly revealed content.

> I added an accessibility setting for merchants to give a TLDR about the purpose of the slideshow

Consider switching to "Slideshow description"

> Navigation button size. On mobile they're not set to be 44x44 but I think they probably should.

I agree the button and number control sizing could be increased. I doesn't need to be the `44` by `44` but at least maximize the space available via `padding`.

svinkle:
@ludoboludo I think after making some adjustments, I'll setup a few sessions with Fable to get user feedback. I'll report back with any new notes. If you need to merge before I get this feedback, go for it. Just be ready to jump back in if I get high severity issues reported back.

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+          },
+          "options__9": {
+            "label": "Bottom right"
+          }
+        },
+        "options": {
+          "content": "Options"
+        },
+        "show_text_box": {
+          "label": "Show text box"
```
melissaperreault, line 2067:
```suggestion
          "label": "Show text box on desktop"
```

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+          "label": "Image overlay opacity"
+        },
+        "auto_rotate": {
+          "label": "Auto-rotate slides"
+        },
+        "change_slides_speed": {
+          "label": "Change slides every"
+        },
+        "mobile": {
+          "content": "Mobile"
```
melissaperreault, line 2079:
```suggestion
          "content": "Mobile layout"
```

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+        },
+        "show_text_below": {
+          "label": "Show text below images"
+        },
+        "accessibility": {
+          "content": "Accessibility",
+          "label": "Slideshow description"
```
melissaperreault, line 2086: I am reflecting on this one, it might need some additional helper text for merchants.

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+        "layout": {
+          "label": "Layout",
+          "options__1": {
+            "label": "Full width"
+          },
+          "options__2": {
+            "label": "Grid"
+          },
+          "options__3": {
+            "label": "Grid and peek"
```
melissaperreault, line 1994:
```suggestion
            "label": "Grid with a peek"
```

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+        "slider_visual": {
+          "label": "Slider visual",
+          "options__1": {
+            "label": "Numbers"
+          },
+          "options__2": {
+            "label": "Dots"
+          },
+          "options__3": {
+            "label": "Counter"
```
melissaperreault, line 2021:
```suggestion
          "options__1": {
            "label": "Counter"
          },
          "options__2": {
            "label": "Dots"
          },
          "options__3": {
            "label": "Numbers"
```

Changing it for alphabetical order.

melissaperreault:
Thank you so much Ludo! Great work!

Made a first pass, will do another round later!

### UX Update on the remaining points I needed to get back to you + some feedback:

- [ ] Set the maximum of slide to 5
- [ ] I thought we could navigate the slides with left and right keyboard arrows, is this still possible? [We did it on Debut](https://screenshot.click/21-11-z9kat-5d0mx.mp4).
- [ ] I wonder if the pause on hover should happen after a few seconds instead of instant? That could leave some time to users to just experience the merchant's first intention. Thoughts? cc. @danielvan 
- [ ] Border opacity around the controls should be  `solid rgba(var(--color-foreground),.08);` instead of `solid rgba(var(--color-foreground),.2);` unless there are other evident areas that we use this 20% opacity 🤔
- [ ] We need to pair for the position of the play/pause to test other options quickly, it definitely gets lost to the opposite left. I also want to improve the thickness of the pause strokes.
- [ ] We won't include a small heading font size, I can see the purpose for Rich text but less on this type of section that needs to stand out.

**Slider visual controls**
- [ ] I believe we are missing some hover states on the numbers, they could increase in opacity and receive the underline once clicked or in focus/view? I took a look at the pagination current implementation and it might not be clear actually.
- [ ] The cursor on the inactive arrow doesn't change and convey the arrow is clickable. I suspect we would need to fix it for the product template nav too. 

**Content**
- [ ] I am wondering if `Counter` and `Numbers` are clear for merchants?

### Answers to your questions/considerations

> > Should clicking the number or dots from the slideshow navigation set the focus on the slide ?

> No, since they're button controls which imply focus remains on the control. The aria-live announces the newly revealed content.

> > Navigation button size. On mobile they're not set to be 44x44 but I think they probably should.

> I agree the bullet and number control sizing could be increased. I doesn't need to be the 44 by 44 but at least maximize the space available via padding.

I am not sure I understand well the expectations for the focus on the slide. Buyers can click/tap dots or number stepper but it is not the primary way to navigate the slides. The arrows are the main controls. Those main control should have the 44px safe tap target, but not the secondary pagination. They give us a hint of where we are mainly but could be navigated to get to a specific slide.

### Remaining things that I need to review

- [ ] All the settings combinations
- [ ] No JS
- [ ] Mobile experience
- [ ] Start requesting help doc update

locales/en.default.schema.json
```diff
@@ -1978,6 +1978,162 @@
...
+          "options__8": {
+            "label": "Bottom center"
+          },
+          "options__9": {
+            "label": "Bottom right"
+          }
+        },
+        "options": {
+          "content": "Options"
+        },
```
melissaperreault, line 2065: I think we can get rid of this heading, it doesn't add value in the hierarchy of the settings, not consistent with  other areas.

locales/en.default.schema.json
svinkle: Agreed. For example, video content we have the following hint text:

> "Describe the video to make it accessible for customers using screen readers."

Perhaps we could include something like,

> "A brief description of the slideshow for customers using screen readers."

sections/slideshow.liquid
```diff
@@ -0,0 +1,418 @@
...
+              {%- if block.settings.image.width >= 2000 -%}{{ block.settings.image | img_url: '2000x' }} 2000w,{%- endif -%}
+              {%- if block.settings.image.width >= 3000 -%}{{ block.settings.image | img_url: '3000x' }} 3000w,{%- endif -%}
+              {%- if block.settings.image.width >= 3840 -%}{{ block.settings.image | img_url: '3840x' }} 3840w,{%- endif -%}
+              {{ block.settings.image | img_url: 'master' }} {{ block.settings.image.width }}w"
+              sizes="100vw"
+              src="{{ block.settings.image | img_url: '1500x' }}"
+              loading="lazy"
+              alt="{{ block.settings.image.alt | escape }}"
+              width="{{ block.settings.image.width }}"
+              height="{{ block.settings.image.width | divided_by: block.settings.image.aspect_ratio }}"
```
svinkle, line 70: Could we round the value up here–the HTML validator doesn't like decimal points in the attribute value.

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+
+    this.isPlaying = true;
+    this.sliderAutoplayButton.classList.remove('slideshow__autoplay--paused');
+    this.slider.setAttribute('aria-live', 'off');
+    this.sliderAutoplayButton.setAttribute('aria-label', 'Pause slideshow');
+    this.autoplay = setInterval(this.autoRotateSlides.bind(this), this.autoplaySpeed);
+  }
+
+  pause() {
+    if (this.isPlaying = false) return;
```
LucasLacerdaUX, line 640:
```suggestion
    if (!this.isPlaying) return;
```

Also: do we need this condition at all on both `play()` and `pause()`? Isn't it being controlled by the other conditions outside this function?

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+
+    this.isPlaying = false;
+    this.sliderAutoplayButton.classList.add('slideshow__autoplay--paused');
+    this.slider.setAttribute('aria-live', 'polite');
+    this.sliderAutoplayButton.setAttribute('aria-label', 'Play slideshow');
```
LucasLacerdaUX, line 645: Hardcoded translation.

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+    }
+  }
+
+  play() {
+    if (this.isPlaying) return;
+
+    this.isPlaying = true;
+    this.sliderAutoplayButton.classList.remove('slideshow__autoplay--paused');
+    this.slider.setAttribute('aria-live', 'off');
+    this.sliderAutoplayButton.setAttribute('aria-label', 'Pause slideshow');
```
LucasLacerdaUX, line 635: Same here.

LucasLacerdaUX (commented):
Just some quick comments. I haven't started my review yet but just caught these by looking at the code.

danielvan:
> [ ] I wonder if the pause on hover should happen after a few seconds instead of instant? That could leave some time to users to just experience the merchant's first intention. Thoughts? cc. @danielvan

I think the user intent on hover is to look at the slide, so I would pause immediately on hover, and play again once the mouse is out. For accessibility, I think it is also better to pause immediately – @svinkle can you let me know if this is true?

assets/global.js
```diff
@@ -514,17 +514,31 @@ class SliderComponent extends HTMLElement {
...
     this.currentPage = Math.round(this.slider.scrollLeft / this.sliderLastItem.clientWidth) + 1;
+    this.sliderControlButtons = this.querySelectorAll('.slider-counter__link');
+
+    if (this.sliderControlButtons.length > 0) {
+      this.sliderControlButtons.forEach(link => {
+        link.classList.remove('slider-counter__link--active');
+        link.removeAttribute('aria-current');
+      });
+      this.sliderControlButtons[this.currentPage - 1].classList.add('slider-counter__link--active');
+      this.sliderControlButtons[this.currentPage - 1].setAttribute('aria-current', true);
```
LucasLacerdaUX, line 535: Is this feature exclusive to the `SlideshowComponent`? It sounds like it is. 

If that is the case, I would revert back to the original code and add this to `SlideshowComponent`:
```js
  update() {
      super.update();
      this.sliderControlButtons = this.querySelectorAll('.slider-counter__link');
      if (!this.sliderControlButtons.length) return;

      this.sliderControlButtons.forEach(link => {
        link.classList.remove('slider-counter__link--active');
        link.removeAttribute('aria-current');
      });
      this.sliderControlButtons[this.currentPage - 1].classList.add('slider-counter__link--active');
      this.sliderControlButtons[this.currentPage - 1].setAttribute('aria-current', true);
  }
```

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
+    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
+    if( this.sliderItemsToShow.length > 0) this.currentPage = 1;
+
+    this.sliderControlLinks = this.sliderControlWrapper.querySelectorAll('.slider-counter__link');
+    this.sliderControlLinksArray = Array.from(this.sliderControlLinks);
+    this.sliderAutoplay = this.slider.hasAttribute('data-autoplay');
+    this.mediaQueryMobile = window.matchMedia('(max-width: 749px)');
+    this.isPlaying = false;
+    this.autoPlayEnabled = false;
```
LucasLacerdaUX, line 584: Unused variable.

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
 
 customElements.define('slider-component', SliderComponent);
 
+class SlideshowComponent extends SliderComponent {
+  constructor() {
+    super();
+    this.sliderControlWrapper = this.querySelector('.slider-buttons');
+    this.bannerContents = this.slider.querySelectorAll('.banner__content');
```
LucasLacerdaUX, line 571: Would you mind having a:

```js
this.elements = {
 // all the variables that you are assigning to a querySelector
}
```
?

I think we're using this pattern in a few places, but it really helps with legibility. I'm often unsure what is a numeric / boolean variable and what is an element.

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+
+    if (!this.sliderControlWrapper) return;
+
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
+    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
+    if( this.sliderItemsToShow.length > 0) this.currentPage = 1;
+
+    this.sliderControlLinks = this.sliderControlWrapper.querySelectorAll('.slider-counter__link');
+    this.sliderControlLinksArray = Array.from(this.sliderControlLinks);
```
LucasLacerdaUX, line 580: Would it make sense to combine both variables into a single one?
I get that you need the array to use `indexOf`, but do you really need the `NodeList` one specifically?

locales/en.default.schema.json
ludoboludo: 👍  if I do this layout in a follow up PR I might remove it from the settings for now 🤔

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+  constructor() {
+    super();
+    this.sliderControlWrapper = this.querySelector('.slider-buttons');
+    this.bannerContents = this.slider.querySelectorAll('.banner__content');
+
+    if (!this.sliderControlWrapper) return;
+
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
+    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
+    if( this.sliderItemsToShow.length > 0) this.currentPage = 1;
```
LucasLacerdaUX, line 577: Isn't that `this.currentPage = 1;` automatically inherited from `SliderComponent` initialization? Or are there scenarios where this could be initialized with a different value?

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+class SlideshowComponent extends SliderComponent {
+  constructor() {
+    super();
+    this.sliderControlWrapper = this.querySelector('.slider-buttons');
+    this.bannerContents = this.slider.querySelectorAll('.banner__content');
+
+    if (!this.sliderControlWrapper) return;
+
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
+    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
```
LucasLacerdaUX, line 576: Correct me if I'm wrong, but isn't that being initialized in `initPages` on the `SliderComponent`?

Looks like the same code to me.
```js
    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
```

assets/global.js
```diff
@@ -553,6 +564,131 @@ class SliderComponent extends HTMLElement {
...
+    this.slider.addEventListener('focusin', this.autoplayFocusHandling.bind(this));
+    this.slider.addEventListener('focusout', this.autoplayFocusHandling.bind(this));
+    this.prevButton.addEventListener('focusout', this.autoplayFocusHandling.bind(this));
+    this.prevButton.addEventListener('focusin', this.autoplayFocusHandling.bind(this));
+    this.nextButton.addEventListener('focusin', this.autoplayFocusHandling.bind(this));
+    this.nextButton.addEventListener('focusout', this.autoplayFocusHandling.bind(this));
+    this.prevButton.addEventListener('mouseenter', this.autoplayFocusHandling.bind(this));
+    this.prevButton.addEventListener('mouseleave', this.autoplayFocusHandling.bind(this));
+    this.nextButton.addEventListener('mouseenter', this.autoplayFocusHandling.bind(this));
+    this.nextButton.addEventListener('mouseleave', this.autoplayFocusHandling.bind(this));
```
LucasLacerdaUX, line 608: This should probably be moved to a separate `setAutoplayEvents` function.

There's a lot happening on the `constructor` right now and I find that would help us know what is absolutely required and what isn't.

LucasLacerdaUX (commented):
Just dumping some initial ideas and questions before the weekend. 

Also: you did an amazing work putting up together such a complex section. Thanks again for that . There's still a lot of functionality to test on my end, but from my quick look at the Theme Editor I love what I'm seeing! :)

I'll come back on Monday to finish my review.

assets/component-slider.css
```diff
@@ -118,15 +118,37 @@ slider-component {
...
+.slider-counter__link--numbers {
+  transition: transform 0.2s ease-in-out;
+}
+
+.slider-counter__link--active.slider-counter__link--dots .dot,
+.slider-counter__link--active.slider-counter__link--numbers,
+.slider-counter__link--dots:hover .dot,
+.slider-counter__link--numbers:hover {
+  transform: scale(1.4);
+}
```
ludoboludo, line 140: @melissaperreault here I played with the hover state for the sake of exploration. The test store link has been updated in the PR and here is a video recap: [video](https://screenshot.click/21-11-z06vj-m00hl.mp4)

sections/slideshow.liquid
ludoboludo: Added the `| round` filter 👍 
I don't think that should impact the layout shift much so it should be ok to add.

assets/global.js
ludoboludo: You're right 👍  I removed the `NodeList` 👌

assets/global.js
ludoboludo: I thought I needed it because it comes up as undefined when I console log it on initialization. But seems to work fine without.

assets/global.js
ludoboludo: Maybe we can take a look together/pair. I feel like some values are also used in `SliderComponent` and some are based on conditionals. So curious to see what would make the most sense.

svinkle:
@ludoboludo I just noticed a focus management issue. Ex., when you're on the second slide and you activate the "previous slide" control, the control is set as `disabled` and focus moves to the top of the DOM.

Instead of setting the control as `disabled`, could we wrap and load the last slide? This way focus remains on the `button` as expected. Wrapping may present a better UX and meet user expectations. @melissaperreault @danielvan thoughts?

ludoboludo:
> @ludoboludo Just noticed another unexpected behaviour. When the slideshow is auto-playing and the user interacts with the "Previous" or "Next" `button` controls, either mouse or keyboard, the slideshow play/pause `button` toggles to a "play" state. This might be confusing as the user did not interact with this control explicitly. This control should remain as-is unless interacted with directly.
> 
> Also, it seems like when the mouse is hovering over the `.slider-buttons` container, the slideshow does not pause. It does pause on keyboard focus here, but not the mouse. Ideally this would be the overall container.

Ah ok. My thought was that if you're clicking on a previous/next button, you're meaning to take over the control. So if I click to see the next slide and 3sec later it moves away from it, it might be frustrating. But I guess that's why hovering over the slide pauses it. But I wouldn't think as a user that once I get to the slide I want to see, hovering will pause the auto play. And since there isn't a visual cue that it does, it's something I have to find out by myself. 
But if I add the pausing on hover for the previous/next button it should do without me noticing that it's happening.

svinkle:
The color contrast of the non-active bullet control is `2.37:1`. It needs to be `3:1` to meet [1.4.11
Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html).

Dawn slideshow non-active bullet controls are light grey.

assets/global.js
```diff
@@ -558,6 +562,170 @@ class SliderComponent extends HTMLElement {
...
+    this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
+    this.slider.addEventListener('mouseenter', this.focusInHandling.bind(this));
+    this.slider.addEventListener('mouseleave', this.focusOutHandling.bind(this));
+    this.slider.addEventListener('focusin', this.focusInHandling.bind(this));
+    this.slider.addEventListener('focusout', this.focusOutHandling.bind(this));
+
+    this.sliderControlWrapper.addEventListener('mouseenter', this.focusInHandling.bind(this));
+    this.sliderControlWrapper.addEventListener('mouseleave', this.focusOutHandling.bind(this));
+    this.sliderControlWrapper.addEventListener('focusin', this.focusInHandling.bind(this));
+    this.sliderControlWrapper.addEventListener('focusout', this.focusOutHandling.bind(this));
```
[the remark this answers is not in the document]
kmeleta: Nah if you had a reason to separate them like excluding play button events, feel free to disregard. Just trying to save reduce some lines. Not worth reworking anything for.

melissaperreault (commented):
Thank you Ludo!! This is looking good so far but there are a few adjustments needed:

**Position of controls + Icon**
- [x] We revised the pause and play balance, you can find the new [component here](https://www.figma.com/file/XVGy0rXJhYSZoDh031jLNj/Slideshows-Design?node-id=352%3A58670)
- [x] Medium and up device size, have the control [centered like here](https://screenshot.click/21-11-kmfpi-hi6ga.png) and have a vertical divider line between the navigation and the control
- [ ] Mobile, have the control pause/play aligned to the right and [hide the vertical divider line](https://screenshot.click/21-11-2awuu-8rhfm.png)

**Settings**
- [ ] On mobile the the visual with the control bar between is [not appealing and introduces a break](https://screenshot.click/21-11-yvysz-4x7m2.png) in the reading flow of the section, the [control bar needs to be styled like the text box color scheme with some adjustments](https://screenshot.click/21-11-ul8kw-1a4tt.png).
- [x] At the block level: bring back the `[Use outline button style](https://screenshot.click/21-11-ga477-3qaxu.png)`. This is due to recent conversation fo the benefit of the ability to chose between a fill and an outline button across sections.
- [x] At the section level: Add a `Text alignment` setting that will change the alignment of the text inside the box. This way, the box could be `center-left` aligned but have a `center` text alignment.
   - [x] Where? [Between](https://screenshot.click/21-11-2gn2r-xahro.png) `Heading font size` and `Desktop box position`
   - [x] Select option: Center (default), Left, Right


**Hover animation**
- I'll set up some time to pair and provide more guidance

assets/component-slider.css
Oliviammarcello: I like your suggestion to scale, but I think we can reduce it slightly.

**For the numbers:** [Video example](https://screenshot.click/21-53-t3x8y-wyic4.mp4)
- [ ] Can we change `transform: scale(1.4);` to `transform: scale(1.1);`
- [ ] Also remove the underline on hover? Scale and opacity should be enough and this way the underline is reserved for the active state. 

**For the dots:** We may need to pair on this one to get the exact values, but can we try this?
- [ ] Can the active dots remain the same size as the other dots? [Screenshot](https://screenshot.click/21-11-ezqsq-yp1g4.png)
- [ ] On hover can we change `transform: scale(1.4);` to `transform: scale(1.1);`
- [ ] I can't figure out where the values are in the code, but on hover can we try to increase the stroke of the outline instead of filling in the circle? (This may not look great, but just wanted to try) 

**Arrows**
- [ ] What is the scaling added here? Can we try the same scaling used on the dots and numbers?

**Pagination**
- [ ] Can we update the pagination to match the styling for numbers 


cc @ludoboludo @melissaperreault

svinkle:
@ludoboludo Currently the slideshow does not take into account the user's [reduced motion OS setting](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion). This will be critical for slideshow accessibility as it contains a wide-sweeping animation which can [negatively affect the user experience](https://css-tricks.com/revisiting-prefers-reduced-motion-the-reduced-motion-media-query/#the-people-affected).

I may suggest if this setting is enabled to remove all motion, including hover effects on controls. The slide would just appear and not "slide" in to view.

How we approach animation should be up for discussion. @melissaperreault @danielvan thoughts? Cc @johanstromqvist.

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,160 @@
...
+          "label": "Heading font size",
+          "options__1": {
+            "label": "Medium"
+          },
+          "options__2": {
+            "label": "Large"
+          }
+        },
+        "text_align": {
+          "label": "Text alignment",
```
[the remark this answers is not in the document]
melissaperreault:
```suggestion
          "label": "Desktop content position",
```
Synced with @katycobb and this is aligned with other sections copy.

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,160 @@
...
+        "mobile": {
+          "content": "Mobile layout"
+        },
+        "show_text_below": {
+          "label": "Show text below images"
+        },
+        "accessibility": {
+          "content": "Accessibility",
+          "label": "Slideshow description",
+          "info": "A brief description of the slideshow for customers using screen readers."
```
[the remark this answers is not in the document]
melissaperreault: _Katy and I synced and here is our recommendation:_

Info text is more instructional to the merchants so perhaps removing "make it accessible" is best with the following sentence format:

- Slideshow
   - `Describe the slideshow for customers using screen readers.`
- Video
   - `Describe the video for customers using screen readers.`

I'll open an PR to request the content info text change. 🙌

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,175 @@
...
+          },
+          "options__3": {
+            "label": "Medium"
+          },
+          "options__4": {
+            "label": "Large"
+          }
+        },
+        "slider_visual": {
+          "label": "Slider visual",
```
melissaperreault, line 2018:
```suggestion
          "label": "Pagination style",
```
Synced with Katy and we are adding more clarity about what to expect to find under this setting.

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,175 @@
...
+          "options__1": {
+            "label": "Counter"
```
melissaperreault, line 2020: We've considered `Fraction` but still need some time to confirm if this the best choice. But definitely not going to go with `Counter`.

melissaperreault:
### UX Update

@ludoboludo I wanted to come back to you about the mobile "text alignment" setting you saw for `Image with text` and `Image banner`.

Confirmed we'll need to add it to `Slideshow` as well. Here is the link to the revised hierarchy of settings. Should read like:
- Desktop content alignment
- Mobile content alignment

sections/slideshow.liquid
```diff
@@ -0,0 +1,445 @@
...
+    <div class="slideshow__controls slider-buttons no-js-hidden color-{{ section.settings.color_scheme }}">
+      <button type="button" class="slider-button slider-button--prev" name="previous" aria-label="{{ 'accessibility.previous_slideshow' | t }}" aria-controls="Slider-{{ section.id }}">{% render 'icon-caret' %}</button>
+      <div class="slider-counter{% if section.settings.slider_visual == 'counter' or section.settings.slider_visual == 'numbers' %} caption{% endif %}">
+        {%- if section.settings.slider_visual == 'counter'-%}
+          <span class="slider-counter--current">1</span>
+          <span aria-hidden="true"> / </span>
+          <span class="visually-hidden">{{ 'accessibility.of' | t }}</span>
+          <span class="slider-counter--total">{{ section.blocks.size }}</span>
+        {%- else-%}
+          <div class="slideshow__control-wrapper" role="group" aria-label="{{ 'accessibility.load_slide' | t }}">
```
svinkle, line 109: From Fable session, let's remove `role="group"` and `aria-label="{{ 'accessibility.load_slide' | t }}"`; this did not have the positive affect I assumed.

sections/slideshow.liquid
```diff
@@ -0,0 +1,445 @@
...
+            {%- for block in section.blocks -%}
+              <button class="slider-counter__link slider-counter__link--{{ section.settings.slider_visual }} link" aria-label="{{ forloop.index }} {{ 'accessibility.of' | t }} {{ forloop.length}}" aria-controls="Slider-{{ section.id }}">
```
svinkle, line 111: From Fable session, let's adjust the `aria-label` to be more clear and precise on what these controls do.

For example, an accName of `1 of 3` did not provide enough detail to convey the purpose of the control. Adding more detail, such as `Load slide 1 of 3` is more informative.

ludoboludo: Should I update the copy to say that a slide is active/loaded ? Say if I'm on slide 1 already, should the aria-label say `Loaded slide 1 of 3` or something along those lines ? Though I guess it should announce the new slide when shown into the view. It doesn't seem to do it atm though 🤔

assets/component-slider.css
ludoboludo: Should be good now 👍  If you want to check the [linked demo store in the PR description](https://os2-demo.myshopify.com/admin/themes/126938021910/editor)

sections/slideshow.liquid
```diff
@@ -0,0 +1,465 @@
...
+              <a{% if block.settings.link %} href="{{ block.settings.link }}"{% else %} role="link" aria-disabled="true"{% endif %} class="button {% if block.settings.button_style_secondary %}button--secondary{% else %}button--primary{% endif %}">{{ block.settings.button_label | escape }}</a>
+            {%- endif -%}
+          </div>
+        </div>
+      </div>
+    {%- endfor -%}
+  </div>
+
+  {%- if section.blocks.size > 1 -%}
+    <div class="slideshow__controls slider-buttons no-js-hidden color-{{ section.settings.color_scheme }}">
```
ludoboludo, line 100: @svinkle, the fact that the slideshow controls is below the images makes sense on desktop I believe but is it an issue for mobile ? 

Since some of the content of the slide, depending on the section settings, can be below the control bar you will tab to the slide button that is below the controls first and then get to the controls. 
[video](https://screenshot.click/21-11-tw714-6rtw7.mp4) (I'm fixing the focus ring now)

svinkle: Right I noticed this, too. Are you able to adjust the placement of the controls in the DOM to match the visual order?

ludoboludo: So I'd see two options and curious to hear your thoughts:  

- 1) move the controls above the image and text box container (slide container). Which would work as expected on mobile. But on desktop it'd mean that focus would first go on the controls (below the slide) and then back up onto the button of the slide, if any, and then onto whatever is the next section/focusable element. [Video of tab order](https://screenshot.click/21-11-4vd5y-znn6n.mp4)
- 2) duplicate the controls code and have one for mobile that is before the slides and then one on desktop after the slides 🤔

svinkle: Curious to know, which is worse in terms of perf; duplicate controls or moving the set of controls in the DOM when a breakpoint condition is met?

@melissaperreault, @danielvan curious to understand the rationale behind the slideshow placement on mobile. It's resulting in some friction on the dev side of things. Could we normalize and keep in the same position?

melissaperreault: Like on Debut, the controls are placed in this position to maintain its position regardless of the slide content length. Otherwise we lose the controls as you navigate the slides, moving up or down, sometimes in the viewport, sometimes outside. I can share a quick prototype we've made to base this decision from.

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,187 @@
...
+        "layout": {
+          "label": "Layout",
+          "options__1": {
+            "label": "Full width"
+          },
+          "options__2": {
+            "label": "Grid"
+          },
+          "options__3": {
+            "label": "Grid with a peek"
```
melissaperreault, line 1999: @katycobb Another copy I wanted to get your feedback on.

This is a setting to set the Layout. So it goes to:

- Layout
   - Full width
   - Grid
   - Grid with a peek

assets/component-slider.css
Oliviammarcello: The active dot is still slightly larger than the non-active dots. [Screenshot.](https://screenshot.click/21-30-wvg5l-0a818.png) Otherwise, it's perfect 👍

ludoboludo: Sweet, done 👌

locales/en.default.schema.json
katycobb: @melissaperreault I'd suggest: 

Layout
- Full width
- Grid
- Grid with preview

Let me know if you are aligned!

sections/slideshow.liquid
ludoboludo: Yeah I can see how it's useful on mobile specifically when having the text box below the image. Cause then the content can be quite long vertically. 

In terms of the HTML structure... I think it makes sense to keep the controls after the slideshow because that's where it needs to be on desktop and also on mobile when you set the text box to be over the image and not below it.  So I only need to tweak the positioning of the controls for one situation. 

Could I change the tabbing order specifically in that one situation @svinkle ? I feel like that's something we would ideally avoid but it might make sense here 🤔

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,187 @@
...
+        },
+        "auto_rotate": {
+          "label": "Auto-rotate slides"
+        },
+        "change_slides_speed": {
+          "label": "Change slides every"
+        },
+        "color_scheme": {
+          "label": "Box color scheme",
+          "info": "Visible when text box displayed.",
```
melissaperreault, line 2094:
```suggestion
          "info": "Visible when container displayed.",
```
Aligns with the most recent decision for Image banner. cc @katycobb for visibility

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,187 @@
...
+            "label": "Medium"
+          },
+          "options__4": {
+            "label": "Large"
+          }
+        },
+        "slider_visual": {
+          "label": "Pagination style",
+          "options__1": {
+            "label": "Fraction"
```
melissaperreault, line 2020:
```suggestion
            "label": "Counter"
```
Ludo! We are reverting this version based on feedback. Sorry! cc. @katycobb 
This `1/2` style should be the default style when you add a section.

locales/en.default.schema.json
ludoboludo: `Visible when container **is** displayed.` ?

assets/component-slideshow.css
```diff
@@ -0,0 +1,271 @@
...
+  .slideshow:not(.banner--mobile-bottom) + .slideshow__controls {
+    background-color: rgb(var(--color-base-background-1));
+    color: rgba(var(--color-base-text), 0.75);
+  }
+
+  .slideshow:not(.banner--mobile-bottom) + .slideshow__controls .slider-button,
+  .slideshow:not(.banner--mobile-bottom) + .slideshow__controls .slider-counter__link--active.slider-counter__link--numbers,
+  .slideshow:not(.banner--mobile-bottom) + .slideshow__controls .slider-counter__link--numbers {
+    color: rgba(var(--color-base-text), 0.75);
+  }
```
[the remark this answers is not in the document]
ludoboludo: I need it because they're child elements of `.slideshow`'s sibling. I don't think I have another way to select it 🤔 

Unless I add a class on the `slider-component` which would help reduce the amount of selectors. Ill look into it

assets/component-slideshow.css
```diff
@@ -0,0 +1,271 @@
...
+@media screen and (min-width: 750px) {
+  .slideshow__text-wrapper {
+    height: 100%;
+    padding: 5rem;
+  }
+
+  .slideshow__text-wrapper.slideshow__text-wrapper--top-left {
+    align-items: flex-start;
+    justify-content: flex-start;
+  }
```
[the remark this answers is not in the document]
ludoboludo: You're right, I removed the `.class.class` and we're still good. 

As for the classes, the way it is right now is easy in the sense that I just have to use the setting value at the end of the class. I find it clearer than having to manipulate the value of the setting by doing some liquid string manipulation to get the values from it. 

For `top-left` I'd need to add two classes for `top` and for `left`.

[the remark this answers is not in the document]
tyleralsbury: I think for that, you'd just have the value of the setting be `slideshow__text-wrapper--top  slideshow__text-wrapper--left`.  Then you don't need to do any string manipulation.

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,187 @@
...
+          },
+          "options__2": {
+            "label": "Center"
+          },
+          "options__3": {
+            "label": "Right"
+          }
+        },
+        "show_text_below": {
+          "label": "Show text below images"
```
melissaperreault, line 2127:
```suggestion
          "label": "Show content below images"
```
I missed this one!

assets/global.js
```diff
@@ -558,6 +562,168 @@ class SliderComponent extends HTMLElement {
...
+  }
+
+  focusInHandling(event) {
+    const focusedOnAutoplayButton = event.target === this.sliderAutoplayButton || this.sliderAutoplayButton.contains(event.target);
+    if (focusedOnAutoplayButton && this.isPlaying) {
+      this.play();
+    } else if (this.isPlaying) {
+      this.pause();
+    }
+  }
```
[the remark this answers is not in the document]
ludoboludo: So here I meant to not have it set to true/false. Maybe my naming should be clearer to not create confusion. 

Basically it's used as state management for whether or not the slideshow should be playing or not. So when managing focus and hover states, I want to pause or play but not change the state of what the slideshow should be doing. 
At first I had two states, `this.isPlaying` and `this.wasPlaying` that I was changing here and there but I thought I could have one and set it specifically to match the autoplay button to simplify it. 

Does that make sense ?

assets/global.js
```diff
@@ -558,6 +562,168 @@ class SliderComponent extends HTMLElement {
...
+      const button = item.querySelector('.button');
+      if (index === this.currentPage - 1) {
+        if (button) button.removeAttribute('tabindex');
+        item.setAttribute('aria-hidden', 'false');
+      } else {
+        if (button) button.setAttribute('tabindex', '-1');
+        item.setAttribute('aria-hidden', 'true');
+      }
+    });
+  }
```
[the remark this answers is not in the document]
ludoboludo: I think I can only do it for classes. Seems like at the moment `toggleAttribute` is more for attributes that are either false or true. 

So for `togglePlayButtonState(pauseAutoplay)` I could use `.classlist.toggle` but then I can't for the `aria-label`. So I might as well keep it as is I think 🤔 

And for `setSlideVisibility()` similar thing. I could use `toggleAttribute` for `aria-hidden` but for `tabindex` I need to set it to `-1`. So since I need to keep the if/else I might as well keep it this way. 

Let me know what you think but I don't think I have a way around it.

assets/global.js
```diff
@@ -558,6 +562,168 @@ class SliderComponent extends HTMLElement {
...
+    if (this.slider.classList.contains('banner--mobile-bottom') && this.mediaQueryMobile.matches) {
+      this.sliderControlWrapper.classList.add('slider-buttons--in-between');
+      this.sliderControlWrapper.style.top = this.slider.querySelector('.slideshow__media').offsetHeight + 'px';
+      this.bannerContents.forEach(banner => banner.style.marginTop = this.sliderControlWrapper.offsetHeight + 'px');
+    } else {
+      this.sliderControlWrapper.classList.remove('slider-buttons--in-between');
+      this.sliderControlWrapper.style.top = 0;
+      this.bannerContents.forEach(banner => banner.removeAttribute('style'));
+    }
+  }
```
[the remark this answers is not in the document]
ludoboludo: I think here it's a bit the same situation. I can use `toggle` for the `classlist` but I'd need to keep the `if/else` to set the `marginTop` and `top` values.

sections/slideshow.liquid
```diff
@@ -0,0 +1,463 @@
...
+        {
+          "type": "text",
+          "id": "subheading",
+          "default": "Tell your brand's story through images",
+          "label": "t:sections.slideshow.blocks.slide.settings.subheading.label"
+        },
+        {
+          "type": "text",
+          "id": "button_label",
+          "default": "Shop all",
```
melissaperreault, line 430:
```suggestion
          "default": "Button label",
```
Overlooked, this is to match with other settings where we have a descriptive label instead of an action.

assets/global.js
```diff
@@ -542,21 +549,179 @@ class SliderComponent extends HTMLElement {
     } else {
       this.nextButton.removeAttribute('disabled');
     }
+  }
 
-    this.pageCount.textContent = this.currentPage;
-    this.pageTotal.textContent = this.totalPages;
+  onButtonClick(event) {
+    event.preventDefault();
+    this.slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderLastItem.clientWidth : this.slider.scrollLeft - this.sliderLastItem.clientWidth;
```
ludoboludo, line 556: changed this to use `this.slideScrollPosition` because of an issue in Firefox otherwise. 
When trying to reuse the value as a const in the `slideshowComponent` and `super.onButtonClick` it wasn't grabbing the value. Even though that worked on Chrome and Safari.

assets/component-slideshow.css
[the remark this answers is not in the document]
ludoboludo: Tried this out ^ but that ends up being too many characters. Tried with shorter class name but was still having an error locally. 

One way I could go about it is with CSS selectors like this but it might not be as legible: 

```css
[class*="slideshow__text-wrapper--top"] {
    align-items: flex-start;
  }

.slideshow__text-wrapper[class*="-left"] {
    justify-content: flex-start;
  }
```

assets/global.js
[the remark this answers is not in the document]
tyleralsbury: Gotcha... I see.

Yeah it's not super clear. I wonder if `shouldBePlaying` or something is clearer.

`isPlaying`, to me, means "the video is playing right now".  So having `this.isPlaying` as a condition for `this.play()` and having `this.play()` usually trigger a  change to `isPlaying` is pretty confusing.

`shouldBePlaying` is the first thing that comes to mind but sounds kinda weird.

tyleralsbury (commented):
Looks great overall.  We're getting close here.  I'll wait for another pass by UX before finalizing, but a quick 🎩 and code pass looks great overall.

I think we should try to show the correct slide when the merchant makes a change, because it's pretty confusing to be on slide 2, edit a setting, then be snapped back to slide 1.

These are the available events that the editor emits.  There are events for sections and blocks: 

https://shopify.dev/themes/architecture/sections/integrate-sections-with-the-theme-editor

assets/global.js
```diff
@@ -542,21 +549,164 @@ class SliderComponent extends HTMLElement {
...
+  }
+
+  togglePlayButtonState(pauseAutoplay) {
+    if (pauseAutoplay) {
+      this.sliderAutoplayButton.classList.add('slideshow__autoplay--paused');
+      this.sliderAutoplayButton.setAttribute('aria-label', window.accessibilityStrings.playSlideshow);
+    } else {
+      this.sliderAutoplayButton.classList.remove('slideshow__autoplay--paused');
+      this.sliderAutoplayButton.setAttribute('aria-label', window.accessibilityStrings.pauseSlideshow);
+    }
```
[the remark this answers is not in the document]
ludoboludo: This I removed due to a recent decision in keeping the controls a the bottom :)

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,181 @@
...
+          },
+          "options__2": {
+            "label": "Dots"
+          },
+          "options__3": {
+            "label": "Numbers"
+          }
+        },
+        "heading_size": {
+          "label": "Heading font size",
```
melissaperreault, line 2027:
```suggestion
          "label": "Heading size",
```
Sorry, this changed on all the other sections.

melissaperreault (commented):
Thank you Ludo! Looking good so far! 

I will also ask help from @duygukalaycioglu to do one final exhaustive functionality pass on her own devices, to test the experience of the component and various setting combinations one last time.

I would love to ask @nicklepine to play with the editor before merging as well 🙏 

**Question for Nick**
- [ ] Should the `Heading size` also live at the block level since we have `Content alignment` now under there?
- [ ] I noticed on other final setting sections and that they offer `Small, Medium and Large` options rather than just `Medium, Large` - where do we stand in terms of consistently offering the same sizes?


**General**

- [ ] I agree with Tyler, let's not auto-rotate slide as merchants interact with the block settings. We achieved this on Debut if you need a reference on how to do this. If merchants interacts again with the slideshow, auto-rotate can continue.
- [ ] It's weird [because I see the](https://screenshot.click/21-12-em66e-ejbav.png) `Pagination style`'s options ordered alphabetically on Github but not in the editor. 🤔 Not sure how I should check this here without looking in the editor. (Just looking for tips)

nicklepine:
> Question for Nick
> 
> 1. Should the Heading size also live at the block level since we have Content alignment now under there?

Good catch! Now that all the layout settings live at the slide (block) level, I would also move this one to the block level. 

> 2. I noticed on other final setting sections and that they offer Small, Medium and Large options rather than just Medium, Large - where do we stand in terms of consistently offering the same sizes?

I'm all for consistency. I think we're now offering small heading font size on Image with Text, we should also make sure to update Image Banner / Email Signup (currently being worked on by @sofiamatulis) and Rich Text (when we get to it next quarter). So yes, in favour of adding it to Slideshow too

layout/theme.liquid
```diff
@@ -163,6 +163,8 @@
 
       window.accessibilityStrings = {
         shareSuccess: `{{ 'general.share.success_message' | t }}`,
+        pauseSlideshow: `{{ 'accessibility.pause_slideshow' | t }}`,
+        playSlideshow: `{{ 'accessibility.play_slideshow' | t }}`,
```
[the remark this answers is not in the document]
ludoboludo: Good point 👍

assets/global.js
```diff
@@ -542,21 +549,164 @@ class SliderComponent extends HTMLElement {
...
+    this.enableSliderLooping = true;
+
+    if (!this.sliderControlWrapper) return;
+
+    this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
+    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
+    if( this.sliderItemsToShow.length > 0) this.currentPage = 1;
+
+    this.sliderControlLinksArray = Array.from(this.sliderControlWrapper.querySelectorAll('.slider-counter__link'));
+    this.sliderAutoplay = this.slider.hasAttribute('data-autoplay');
```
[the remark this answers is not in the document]
ludoboludo: Nah you're right 👍  followed your suggestion from the next comment.

assets/global.js
```diff
@@ -542,21 +549,164 @@ class SliderComponent extends HTMLElement {
...
+
+  autoPlayToggle() {
+    this.togglePlayButtonState(this.isPlaying);
+    if (this.isPlaying) {
+      this.pause()
+      this.isPlaying = false;
+    } else {
+      this.play();
+      this.isPlaying = true;
+    }
```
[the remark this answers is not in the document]
ludoboludo: I like it 👍

assets/global.js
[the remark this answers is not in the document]
ludoboludo: Gave this a try but it doesn't seem to like the if statement within the `setAttribute` declaration 🤔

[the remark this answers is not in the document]
LucasLacerdaUX: That's weird 👀 I tested it before commenting and I swear it was working 👀 I'll check again.

sections/slideshow.liquid
```diff
@@ -0,0 +1,465 @@
...
+    </noscript>
+  {%- endif -%}
+
+</slideshow-component>
+
+<script>
+  window.accessibilityStrings = {
+    pauseSlideshow: `{{ 'accessibility.pause_slideshow' | t }}`,
+    playSlideshow: `{{ 'accessibility.play_slideshow' | t }}`,
+  }
```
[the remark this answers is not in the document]
ludoboludo: Ah gotcha, I think the comment was posted twice. I will take a look at it. In the mean I moved this back to the `theme.liquid`.

sections/slideshow.liquid
```diff
@@ -0,0 +1,465 @@
...
+        {%- endif -%}
+      </div>
+      <button type="button" class="slider-button slider-button--next" name="next" aria-label="{{ 'accessibility.next_slideshow' | t }}" aria-controls="Slider-{{ section.id }}">{% render 'icon-caret' %}</button>
+
+      {%- if section.settings.auto_rotate -%}
+        <button type="button" class="slideshow__autoplay slider-button no-js-hidden{% if section.settings.auto_rotate == false %} slideshow__autoplay--paused{% endif %}" aria-label="{{ 'accessibility.pause_slideshow' | t }}">
+          <svg class="pause" width="8" height="12" viewBox="0 0 8 12" fill="none" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true">
+            <path d="M1.2 0.75C0.813401 0.75 0.5 0.985051 0.5 1.275V10.725C0.5 11.0149 0.813401 11.25 1.2 11.25C1.5866 11.25 1.9 11.0149 1.9 10.725V1.275C1.9 0.985051 1.5866 0.75 1.2 0.75Z" fill="currentColor"/>
+            <path d="M6.8 0.75C6.4134 0.75 6.1 0.985051 6.1 1.275V10.725C6.1 11.0149 6.4134 11.25 6.8 11.25C7.1866 11.25 7.5 11.0149 7.5 10.725V1.275C7.5 0.985051 7.1866 0.75 6.8 0.75Z" fill="currentColor"/>
+          </svg>
```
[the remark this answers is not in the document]
ludoboludo: Sounds like a good plan. I added it as a file and I'm rendering both now 👍

assets/global.js
```diff
@@ -542,21 +549,158 @@ class SliderComponent extends HTMLElement {
...
+  autoRotateSlides() {
+    const slideScrollPosition = this.currentPage === this.sliderItems.length ? 0 : this.slider.scrollLeft + this.sliderLastItem.clientWidth;
+    this.slider.scrollTo({
+      left: slideScrollPosition
+    });
+  }
+
+  setSlideVisibility() {
+    this.sliderItemsToShow.forEach((item, index) => {
+      const button = item.querySelector('.button');
```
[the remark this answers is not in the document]
ludoboludo: I see your point. 👍  Agreed.

locales/en.default.json
```diff
@@ -69,6 +69,13 @@
     "of": "of",
     "next_slide": "Slide right",
     "previous_slide": "Slide left",
+    "load_slide": "Load slide",
+    "previous_slideshow": "Previous slide",
+    "next_slideshow": "Next slide",
+    "pause_slideshow": "Pause slideshow",
+    "play_slideshow": "Play slideshow",
+    "carousel": "Carousel",
+    "slide": "Slide",
```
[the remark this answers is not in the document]
ludoboludo: Done in this commit: https://github.com/Shopify/dawn/pull/840/commits/7639f5de40dde3313614b84d81741896cc3d9a31

katycobb:
Hi team, I reviewed the settings against the latest settings content. Couple of things to call out: 

- Under "Mobile Layout", add "on mobile" to the end of the checkbox copy (a bit redundant, but adding for clarity). It should read "Show content below images on mobile"

In the image slide block: 
- “leave the label blank to hide the button” help text should be under the button label field, not the link field. 
- "Link" field label should say "Button link" 
- "Box color scheme" should say "Color scheme" 


And a question for @Oliviammarcello (I think): Does "Image overlay opacity" control both desktop and mobile here?

sections/slideshow.liquid
```diff
@@ -0,0 +1,465 @@
...
+
+          <svg class="play" width="8" height="10" viewBox="0 0 8 10" fill="none" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true">
+            <path fill-rule="evenodd" clip-rule="evenodd" d="M1.49909 0.863414C0.832428 0.479443 0 0.960626 0 1.72996V8.24642C0 9.03867 0.87739 9.51631 1.54278 9.08629L6.87442 5.64063C7.50276 5.23456 7.47903 4.3076 6.83074 3.93421L1.49909 0.863414Z" fill="currentColor"/>
+          </svg>
+        </button>
```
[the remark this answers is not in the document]
ludoboludo: Fixed 👌

duygukalaycioglu (changes requested):
**Character limit for Slide/Headings** – seems like there is no character limit for the copy, would be a good idea to have a limit to avoid something like this. 
Screen Shot 2021-12-03 at 11 17 56 AM
Same for mobile: 
Screen Shot 2021-12-03 at 11 33 22 AM

**Character limit for the button label** – I don't think the button should go to 3 lines, let's limit to 1 and have a character limit. 
Screen Shot 2021-12-03 at 11 39 21 AM

If you need me to identify character limits for the above, let me know. 

**Spacing below the heading** – Let's decrease the spacing between the heading and sub-heading. Increase the spacing between the sub-copy and the button, so that heading and sub-heading looks related. 
Screen Shot 2021-12-03 at 11 42 30 AM

**Sub-heading 2nd line** – The sub-copy doesn't wrap after the 2nd line, started to bleed over the image.
Screen Shot 2021-12-03 at 11 36 25 AM

Everything else is looking great!!

duygukalaycioglu (changes requested):
Fix on the overflow text is looking great Ludo! 

Desktop spacing looks good,
Can we update the tablet + mobile spacing as well?

spacing on tab + mob

Thanks!

duygukalaycioglu (approved):
Spacing looks consistent across, thanks!

tyleralsbury (approved):
Looks great!  Performance appears to be good for a Slideshow.

We will want to come back around and look at how we can improve the awareness of the fold to prioritize the download of large screen images like those in the slideshow as well as the image banner section (remove `loading="lazy"`) to eek out additional performance.  

I'm glad we got rid of the JS moving the controls around on page load, removing the layout shift which means no big script evaluations happening.

tyleralsbury (approved):

tyleralsbury (approved):
Stale approval

duygukalaycioglu (commented):
If possible, can we decrease the spacing between the heading and sub-heading so that they feel like a group. The spacing above the button is looking good, no change needed - thanks!

decrease spacing

katycobb:
Can we add help text in the image slide block? Under `Desktop content position` please add `Position is automatically optimized for mobile.`

thanks!

assets/theme-editor.js
```diff
@@ -0,0 +1,20 @@
+document.addEventListener('shopify:block:select', function(event) {
+  const blockSelectedIsSlide = event.target.classList.contains('slideshow__slide');
```
LucasLacerdaUX, line 2: @tyleralsbury Do you think we should have a more specific naming for this file or have this block of code as part of a named function?

I don't want to worry so much about how we'll use this file & function in the future, but `assets/theme-editor.js` and a generic `shopify:block:select` event doesn't immediately tell me this file is purely to update the selected block on slideshow component.

assets/global.js
```diff
@@ -542,21 +550,155 @@ class SliderComponent extends HTMLElement {
...
+        if (button) button.setAttribute('tabindex', '-1');
+        item.setAttribute('aria-hidden', 'true');
+      }
+    });
+  }
+
+  linkToSlide(event) {
     event.preventDefault();
-    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderLastItem.clientWidth : this.slider.scrollLeft - this.sliderLastItem.clientWidth;
+    const slideScrollPosition = this.slider.scrollLeft + this.sliderFirstItem.clientWidth * (this.sliderControlLinksArray.indexOf(event.currentTarget) + 1 - this.currentPage);
```
LucasLacerdaUX, line 694: image
I kept get an error on this line when clicking the Numbers or Dots. 
https://user-images.githubusercontent.com/37168033/145521736-cc2df010-abc0-4653-9101-a591126989dc.mp4

It seems that this variable was renamed to `this.sliderFirstItemNode` but not here.

assets/global.js
```diff
@@ -542,21 +550,155 @@ class SliderComponent extends HTMLElement {
...
 
   onButtonClick(event) {
+    super.onButtonClick(event);
+    const isFirstSlide = this.currentPage === 1;
+    const isLastSlide = this.currentPage === this.sliderItemsToShow.length;
+
+    if (!isFirstSlide && !isLastSlide) return;
+
+    if (isFirstSlide && event.currentTarget.name === 'previous') {
+      this.slideScrollPosition = this.slider.scrollLeft + this.sliderFirstItem.clientWidth * this.sliderItemsToShow.length;
```
LucasLacerdaUX, line 607: Same here. Looks like this variable was renamed to `this.sliderFirstItemNode` but not here. Please confirm 👀

LucasLacerdaUX (commented):
Looking good! Just found a minor bug with dots and numbers that seems to be caused by a variable that no longer exists.

Tested all settings on multiple devices and didn't have any major issues.

assets/theme-editor.js
tyleralsbury: You're right about this probably being a little too generically named.  I think we should isolate all the theme editor code into a single file, so I think the one file is a good idea, but the naming could be better. Let's revisit this because I'd wager there are a few spots we could improve the editor experience with this type of code.

I noticed a spot in Sofia's collapsible rows PR where it would help if the row whose text you're editing would open when the section reloads, for example.  I'll create an issue for us to audit editor behaviours and look into expanding the amount of editor-only code we have improving that experience and the refactor can happen then.

tyleralsbury (dismissed):
Bugs are fixed. Awaiting UX and a11y approval.

LucasLacerdaUX (dismissed):
Dots/Numbers are now working as expected 👍

svinkle:
@tyleralsbury Is there a new store link I could test these changes with?

tyleralsbury (approved):
A monumental task! ✅

tyleralsbury (approved):
✅

LucasLacerdaUX (approved):
👍

PR #826, opened 2021-10-25
Add border and shadow to button

sofiamatulis:
**Why are these changes introduced?**

Fixes #821 .

**What approach did you take?**

Because we already use `box-shadow` for the `focus`, I decided to use `filter: drop-shadow`. Docs here: https://developer.mozilla.org/en-US/docs/Web/CSS/filter

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126559354902/editor)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #844, opened 2021-11-02
Global settings - buttons, inputs, dividers

tyleralsbury:

Adds the new global settings for buttons, dropdowns, inputs, and divider lines.

Note: Translations (including translation strings) are not included here.  We can handle translation stuff in the `all-global-settings` branch.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126768513046/editor)

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -75,14 +75,19 @@
 .button--secondary {
   --color-button: var(--color-base-outline-button-labels);
   --color-button-text: var(--color-base-outline-button-labels);
-  --alpha-button-background: 0;
+  --color-button: var(--color-background);
+  --alpha-button-background: 1;
 }
 
 .color-background-2 .button--secondary,
-.color-inverse .button--secondary,
 .color-accent-1 .button--secondary,
 .color-accent-2 .button--secondary {
   --color-button: var(--color-foreground);
+  --color-button-text: var(--color-foreground);
+}
+
+.color-inverse .button--secondary {
+  --color-button: var(--color-background);
   --color-button-text: var(--color-foreground);
 }
 
@@ -1196,24 +1201,45 @@ deferred-media {
 .button,
 .shopify-challenge__button,
 .customer button {
+  --shadow-horizontal-offset: var(--buttons-shadow-horizontal-offset);
+  --shadow-vertical-offset: var(--buttons-shadow-vertical-offset);
+  --shadow-blur-radius: var(--buttons-shadow-blur-radius);
+  --shadow-opacity: var(--buttons-shadow-opacity);
   cursor: pointer;
   display: inline-flex;
   justify-content: center;
   align-items: center;
   box-sizing: border-box;
   font: inherit;
-  padding: 0.9rem 3rem 1.1rem;
+  font-size: 1.5rem;
+  padding: var(--buttons-padding) calc(var(--buttons-padding) * 3);
   text-decoration: none;
-  border: 0.1rem solid transparent;
-  border-radius: 0;
+  border: 1px solid transparent;
+  border-radius: var(--buttons-radius);
   background-color: rgba(var(--color-button), var(--alpha-button-background));
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-button), var(--alpha-button-border));
+  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button), var(--buttons-border-opacity));
   color: rgb(var(--color-button-text));
   min-width: 12rem;
   min-height: 4.5rem;
   transition: box-shadow var(--duration-short) ease;
   -webkit-appearance: none;
   appearance: none;
+  filter: drop-shadow(
+    var(--shadow-horizontal-offset)
+    var(--shadow-vertical-offset)
+    var(--shadow-blur-radius)
+    rgba(var(--color-button), var(--shadow-opacity))
+  );
+}
+
+.button--secondary {
+  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button-text), var(--buttons-border-opacity));
+  filter: drop-shadow(
+    var(--shadow-horizontal-offset)
+    var(--shadow-vertical-offset)
+    var(--shadow-blur-radius)
+    rgba(var(--color-button-text), var(--shadow-opacity))
+  );
 }
 
 .button:focus-visible {
@@ -1264,6 +1290,10 @@ deferred-media {
 .shopify-challenge__button:hover,
 .customer button:hover {
   box-shadow: 0 0 0 0.2rem rgba(var(--color-button), var(--alpha-button-border));
+}
+
+.button--secondary:not([disabled]):hover {
+  box-shadow: 0 0 0 0.2rem rgba(var(--color-button-text), var(--alpha-button-border));
 }
 
 /* Button - other */
@@ -1419,16 +1449,22 @@ details[open] > .share-button__fallback {
 .customer select {
   -webkit-appearance: none;
   appearance: none;
-  background-color: transparent;
+  background-color: rgb(var(--color-background));
   border: 0.1rem solid transparent;
-  border-radius: 0;
+  border-radius: var(--inputs-radius);
   color: rgb(var(--color-foreground));
   font-size: 1.6rem;
   width: 100%;
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.55);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
   height: 4.5rem;
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
+  filter: drop-shadow(
+    var(--inputs-shadow-horizontal-offset)
+    var(--inputs-shadow-vertical-offset)
+    var(--inputs-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--inputs-shadow-opacity))
+  );
 }
 
 .select__select {
@@ -1444,7 +1480,7 @@ details[open] > .share-button__fallback {
 .customer .field input:hover,
 .customer select:hover,
 .localization-form__select:hover {
-  box-shadow: 0 0 0 0.2rem rgba(var(--color-foreground), 0.55);
+  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 1px) rgba(var(--color-foreground), var(--inputs-border-opacity));
 }
 
 .field__input:focus,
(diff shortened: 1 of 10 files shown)


config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "step": 5,
+        "unit": "%",
+        "label": "Border opacity",
+        "default": 50
+      },
+      {
+        "type": "range",
+        "id": "buttons_border_width",
+        "min": 0,
+        "max": 10,
```
melissaperreault, line 128:
```suggestion
        "max": 8,
```

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "content": "Borders"
+      },
+      {
+        "type": "range",
+        "id": "buttons_border_opacity",
+        "min": 0,
+        "max": 100,
+        "step": 5,
+        "unit": "%",
+        "label": "Border opacity",
```
melissaperreault, line 121:
```suggestion
        "label": "Opacity",
```
No need to repeat the word since we have a header introducing the category.

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "default": 50
+      },
+      {
+        "type": "range",
+        "id": "buttons_border_width",
+        "min": 0,
+        "max": 10,
+        "step": 1,
+        "unit": "px",
+        "label": "Border width",
```
melissaperreault, line 131:
```suggestion
        "label": "Width",
```
No need to repeat the word since we have a header introducing the category.

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "min": 0,
+        "max": 20,
+        "step": 1,
+        "unit": "px",
+        "label": "Padding",
+        "default": 10
+      },
+      {
+        "type": "header",
+        "content": "Borders"
```
melissaperreault, line 213:
```suggestion
        "content": "Border"
```

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "step": 5,
+        "unit": "%",
+        "label": "Border opacity",
+        "default": 50
+      },
+      {
+        "type": "range",
+        "id": "inputs_border_width",
+        "min": 0,
+        "max": 10,
```
melissaperreault, line 229:
```suggestion
        "max": 8,
```

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "step": 1,
+        "unit": "px",
+        "label": "Border width",
```
melissaperreault, line 232:
```suggestion
        "label": "Width",
```

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "content": "Borders"
+      },
+      {
+        "type": "range",
+        "id": "inputs_border_opacity",
+        "min": 0,
+        "max": 100,
+        "step": 5,
+        "unit": "%",
+        "label": "Border opacity",
```
melissaperreault, line 222:
```suggestion
        "label": "Opacity",
```

melissaperreault (commented):
Thank you Tyler! 🎉 
I'll be sharing my feedback by group of settings just to make it easy to follow and review. After a few investigations here are the latest decisions:

### Buttons
- [ ] Remove the padding setting for now. There is additional thinking to make for desktop and mobile padding logic.
- [ ] I wasn't able to test the `Font size scale`, [nothing changed in my preview.](https://screenshot.click/21-11-4iixy-26s81.mp4)
- [ ] Border radius doesn't apply on the [dynamic checkout button](https://screenshot.click/21-11-3p23w-ov5ug.png)

**Corner radius**
- [ ] Under customer account, [the order number wears the tertiary button style](https://screenshot.click/21-11-7qno8-6rfpf.png), I believe this one should receive corner radius too.

**Border width**
- [ ] The initial hover intention on the buttons is to emphasize the size on hover. When we increase the border width, the over is no longer appealing and we lose the animation subtlety. I would like to chat about other opportunities for us to improve hover state animation
- [ ] Similarly when the buttons receive focus, I would expect the focus to surround the new style of the button, [not use the same default one](https://screenshot.click/21-11-e80aw-p4frr.mp4). The style are lost and looks like a bug. Should we consider to increase the focus border too? Let's chat about what we can do!
- [ ] I find the [border width max value too intense](https://screenshot.click/21-11-zf075-o36w9.mp4). I'll suggest a new value!

**Shadow**
- [ ] The shadow should not receive the border styling. When we offset the shadow + have a border [the outcome is not what is expected](https://screenshot.click/21-11-oqdqz-d6sgp.png). ([Screen recording](https://screenshot.click/21-11-q0pli-ajewo.mp4) of the undesired outcome)
- [ ] We explored a different hover experience for when the button has a shadow, to make it feel like you are pressing the button. Sofia and I made explorations over here https://github.com/Shopify/dawn/pull/826, is this no longer possible?

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+    ]
+  },
+  {
+    "name": "Dividers & lines",
+    "settings": [
+      {
+        "type": "range",
+        "id": "dividers_width",
+        "min": 0,
+        "max": 20,
```
melissaperreault, line 298:
```suggestion
        "max": 10,
```

assets/base.css
```diff
@@ -1170,24 +1175,45 @@ deferred-media {
 .button,
 .shopify-challenge__button,
 .customer button {
```
ludoboludo, line 1177: Some buttons that have a transparent background are looking a little funny: 
- [account page button](https://screenshot.click/21-11-gq1p1-hscbi.png)
- [facet filtering](https://screenshot.click/21-11-bqha2-kz3gn.png)

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+      {
+        "type": "range",
+        "id": "buttons_shadow_vertical_offset",
+        "min": -20,
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Vertical offset",
+        "default": 0
+      },
```
ludoboludo, line 177: Here I think we need to find the right balance and limit a bit more the max value. Otherwise you end up with things that are overlapping a bit too much:
- [search and variants](https://screenshot.click/21-11-qz5e8-m3fhu.png)
- [cart notification buttons](https://screenshot.click/21-11-af9si-fvtu3.png)
- [add to cart button](https://screenshot.click/21-11-ox24x-zkc2o.png)

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Blur radius",
+        "default": 0
+      }
+    ]
+  },
+  {
+    "name": "Dividers & lines",
```
ludoboludo, line 292: Should the settings below impact as well the customer/account page and cart page?
- [cart](https://screenshot.click/21-11-0zv1e-71787.png)
- [account page](https://screenshot.click/21-11-rpmym-7mz49.png)

ludoboludo (commented):
Here are a few things I've noticed/questioned: 
- Border on box shadow like Meli mentioned: [screenshot](https://screenshot.click/21-11-o4b9v-ulrkh.png)
- focus style on sort is receiving the border-radius update. Should other filter summaries received the same treatment: [screenshot](https://screenshot.click/21-11-1u4yp-zx9re.png)
- When choosing a filter it adds a "bubble/button". I feel like its border-radius and border thickness should match the settings for buttons. So as a user you have control and can decide to have rounded things everywhere or none. [screenshot](https://screenshot.click/21-11-3zxjc-gfg6j.png)
- Similar to the above, the close button on modal, I think should only have a border radius when set in the theme settings now: [screenshot](https://screenshot.click/21-11-j3cpt-z5ngr.png). 
- Dynamic checkout buttons not receiving the proper styling (already mentioned by Meli): [screenshot](https://screenshot.click/21-11-jt0ta-8a8kd.png)
- The sort input in the facet drawer is looking a bit odd without a border but a box a shadow: [screenshot](https://screenshot.click/21-11-0mrw7-iw3d2.png)
- I feel like local buttons should match the rest of them. So with a border radius and thicker border when one is set: [screenshot](https://screenshot.click/21-11-t2cj5-t7tj5.png)
- the share button is getting the input styling with the drop shadow which mess with its look: [screenshot](https://screenshot.click/21-11-hi4hx-gpj23.png)
- It mentions drop-downs but which ones is it affecting? I feel like it's confusing, as I would expect the navigation dropdown to be affected but it isn't.

And finally as a general feeling about setting all of those things up: 
- it would be nice to have one setting for border-radius, border thickness, shadow etc; that would affect all of the elements we're touching in this PR. It's more of a platform feature request I think but it would be nice to have. 
I find it tedious to have to edit all those sliders for different elements and a bit overwhelming. Having just 4 sliders for my whole theme and trust the logic behind it to make everything look smooth together would be great. With the optional overwrite on the element level like it is right now.

assets/base.css
tyleralsbury: Good catch - I will fix them in a follow-up.  I changed the way outline buttons and inputs work in some spots but there are a few that I missed, it seems.  They should just get the background colour of their parent element rather than having a transparent background.

config/settings_schema.json
tyleralsbury: We've de-scoped "tables" for now but will be trying to add settings in for those as well.  I believe they should be separate from these settings because there are more controls you might want to add like even/odd row colouring, different thicknesses for below the headings v.s. between cells, etc...  Also we would not want to allow shadows there, probably.

Tables are pretty unique.

tyleralsbury:
Will fix the bugs in a follow-up.  Thanks for the thorough reviews, Ludo + Meli.

ludoboludo (approved):
Some follow ups needed that I mentioned in my previous comment but can be merged in the global branch for the time being 👍

config/settings_schema.json
```diff
@@ -86,6 +86,257 @@
...
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Blur radius",
+        "default": 0
+      }
+    ]
+  },
+  {
+    "name": "Dropdowns & inputs",
```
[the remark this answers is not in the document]
martinamarien: Not sure we should have the inputs applied on the filters dropdown.
Screen Shot 2021-11-16 at 4 06 48 PM

Screen Shot 2021-11-16 at 4 09 18 PM


And sort by in drawer - particularly if it's not being applied on desktop and if the border isn't applied.
Screen Shot 2021-11-16 at 4 07 20 PM

martinamarien (approved):
Found a few inconsistencies. We can address them in the main global branch once Meli has a time to review more completely. Mostly similar comments to Meli and Ludo

PR #847, opened 2021-11-03
Fixed RGB color format on 404 template

### LINKED ISSUE #846: RGB color format not working on 404 template page
RGB color format is not loading/working on the 404 template. 

```
--color-base-accent-1: {{ settings.colors_accent_1.red }}, {{ settings.colors_accent_1.green }}, {{ settings.colors_accent_1.blue }};
```
Looks like a bug in Shopify?

chintankotadia:

Fixes #846.

On the 404 template page, the traditional way of RGB color format is not working so replaced with `color_extract` color filter.

**Other considerations**

**Demo links**

- [Store](https://cloverlane.com/404)

**Checklist**

### DIFF
--- a/layout/password.liquid
+++ b/layout/password.liquid
@@ -49,13 +49,13 @@
         --font-body-scale: {{ settings.body_scale | divided_by: 100.0 }};
         --font-heading-scale: {{ settings.heading_scale | times: 1.0 | divided_by: settings.body_scale }};
 
-        --color-base-text: {{ settings.colors_text.red }}, {{ settings.colors_text.green }}, {{ settings.colors_text.blue }};
-        --color-base-background-1: {{ settings.colors_background_1.red }}, {{ settings.colors_background_1.green }}, {{ settings.colors_background_1.blue }};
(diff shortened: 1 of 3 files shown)


bakura10:
We are also being impacted by this on our theme. While this fix indeed works, it would be nice if it could be fixed at Shopify level. Not being able to use some documented features (color access) reliably because they are not supported on some templates is not a good experience for anyone and should really be fixed at Shopify level imho.

chintankotadia:
@bakura10 : I agree with you. However, it's better to apply this patch and fix this issue using the color filter `color_extract`. This is the best approach and we don't need to rely on color access.

ludoboludo:
@chintankotadia thanks a bunch for addressing my comments/concerns. I commented on another issue, but reiterating here. Let's pause on this for the time being and I'm going to see if we can get the issue resolved on the platform level. @bakura10 made some great points [here](https://github.com/Shopify/dawn/issues/873#issuecomment-967215959).

ludoboludo:
Hey folks, the 404 template is now working as expected, give it a shot and let me know if you still see something funky happen. Cheers!

Issue #548, opened 2021-09-02
Local pickup fails with CORS issue on non-primary domain


cbothner:
### Current behavior

Loading the local pickup section fails with a CORS error because it's always being loaded from the primary domain.

<img width="425" alt="image" src="https://user-images.githubusercontent.com/4642599/131850357-cf9253aa-9745-43ad-b5b0-d32adde5fc98.png">
<img width="497" alt="image" src="https://user-images.githubusercontent.com/4642599/131850969-2f1d94b0-c2ec-459f-9a63-13ab43047104.png">

https://ca.catleaves.com/products/invisible-cat-leaf

### Expected behavior

The local pickup section should be loaded from the current domain so that it is not cross-origin and can load successfully

### Version information
- Dawn Version: 1.0.0
- Firefox Version 91.0.2
- macOS Version 11.5.2

### Possible solution

Currently the `data-base-url` being passed to the `pickup-availability` component is generated using `{{ shop.url }}{{ routes.root_url }}`
https://github.com/Shopify/dawn/blob/4a7cad306d48df70e0e58249df1533d3beec8eba/sections/main-product.liquid#L293-L298

But `shop.url` always returns the primary domain in the primary language.

Instead, `{{ request.host }}{{ routes.root_url }}` should be used. `request.host` always returns the current domain.

PR #848, opened 2021-11-03
Fix pickup-in-store international domains and multi-language support

tjoyal:

Fixes https://github.com/Shopify/dawn/issues/548

As per the issue description, we must always use the current domain to issue an ajax call or otherwise a CORS issue will occur.

Additionally, the concatenation was not valid for multi-language use cases where the resulting url was not valid. 

**Demo links**
URL: https://jp.defaultglobal.com/products/blue-silk-tuxedo
Generated XHR: https://defaultglobal.com/variants/31362180874262/?section_id=pickup-availability
Result: 
```
Access to fetch at 'https://defaultglobal.com/variants/31362180874262/?section_id=pickup-availability' from origin 'https://jp.defaultglobal.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```
Screen Shot 2021-11-03 at 7 01 31 PM

URL: https://defaultglobal.com/en-fr/products/blue-silk-tuxedo
Generated XHR: https://defaultglobal.com/en-fr/en-frvariants/31362180874262/?section_id=pickup-availability
Result: 
```
<div id="shopify-section-pickup-availability" class="shopify-section">
</div>
```

**Checklist**

### DIFF
--- a/assets/pickup-availability.js
+++ b/assets/pickup-availability.js
@@ -11,7 +11,11 @@ if (!customElements.get('pickup-availability')) {
     }
 
     fetchAvailability(variantId) {
-      const variantSectionUrl = `${this.dataset.baseUrl}variants/${variantId}/?section_id=pickup-availability`;
+      let rootUrl = this.dataset.rootUrl;
+      if (!rootUrl.endsWith("/")) {
+        rootUrl = rootUrl + "/";
+      }
+      const variantSectionUrl = `${rootUrl}variants/${variantId}/?section_id=pickup-availability`;
 
       fetch(variantSectionUrl)
         .then(response => response.text())
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -320,7 +320,7 @@
 
               <pickup-availability class="product__pickup-availabilities no-js-hidden"
                 {% if product.selected_or_first_available_variant.available and pick_up_availabilities.size > 0 %} available{% endif %}
-                data-base-url="{{ shop.url }}{{ routes.root_url }}"
+                data-root-url="{{ routes.root_url }}"
                 data-variant-id="{{ product.selected_or_first_available_variant.id }}"
                 data-has-only-default-variant="{{ product.has_only_default_variant }}"
               >


assets/pickup-availability.js
```diff
@@ -11,7 +11,7 @@ if (!customElements.get('pickup-availability')) {
     }
 
     fetchAvailability(variantId) {
-      const variantSectionUrl = `${this.dataset.baseUrl}variants/${variantId}/?section_id=pickup-availability`;
+      const variantSectionUrl = `//${window.location.host}/${this.dataset.rootUrl}/variants/${variantId}/?section_id=pickup-availability`;
```
cbothner, line 14: Does the relative path work by itself?
```suggestion
      const variantSectionUrl = `/${this.dataset.rootUrl}/variants/${variantId}/?section_id=pickup-availability`;
```

cbothner (approved):
Yes, this would be something a dev could use `Shopify.routes.root` for

assets/pickup-availability.js
tjoyal: I'm far from being a JS expert, but from what I've tested locally and read in the doc that ended up producing a call to `https://fr-ca/variants/1234/....`

cbothner: I think you might have tested with two leading slashes, making it a protocol-relative path?

Ah, and that's because the root url always starts with a slash, so this would probably work:

```suggestion
      const variantSectionUrl = `${this.dataset.rootUrl}/variants/${variantId}/?section_id=pickup-availability`;
```

wroy (approved):

assets/pickup-availability.js
ludoboludo: Is there a `/` too many ? 
`window.location.host` returns a url without a forward slash at the end of it but `{{ routes.root_url }}` does have one at the beginning of the string it will output. So is the one you manually added necessary ? It ends up being like that:

tjoyal: @ludoboludo After a quick chat, we decided to go the full js version which is a bit more verbose but ensure there is no console warnings
```
      let rootUrl = this.dataset.rootUrl;
      if (!rootUrl.endsWith("/")) {
        rootUrl = rootUrl + "/";
      }
      const variantSectionUrl = `${rootUrl}variants/${variantId}/?section_id=pickup-availability`;
```

assets/pickup-availability.js
```diff
@@ -11,7 +11,11 @@ if (!customElements.get('pickup-availability')) {
     }
 
     fetchAvailability(variantId) {
-      const variantSectionUrl = `${this.dataset.baseUrl}variants/${variantId}/?section_id=pickup-availability`;
+      let rootUrl = this.dataset.rootUrl;
+      if (!rootUrl.endsWith("/")) {
+        rootUrl = rootUrl + "/";
+      }
+      const variantSectionUrl = `${rootUrl}variants/${variantId}/?section_id=pickup-availability`;
```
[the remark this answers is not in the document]
tjoyal: Good catch

> Let me know if you're ok with adding this fix in this PR as well, otherwise I can create a follow up issue and PR

I think since it is unrelated to the specific of the issue being solved here, I would prefer to have this shipped and iterated over.

I would also appreciate since you are offering to have you take over this specific change. I'm not familiar with the specifics of the JS enhancements in Dawn(in themes in general) to be confident as to the blast radius here.

ludoboludo (approved):

cbothner (approved):

assets/pickup-availability.js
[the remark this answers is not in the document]
ludoboludo: Just for sake of visibility, I've created the follow up PR about this issue here: https://github.com/Shopify/dawn/pull/855

But nothing is needed, I requested reviews from my teammates.

PR #855, opened 2021-11-08
Fix pick up availability 'Unavailable'

ludoboludo:

While testing out [this PR](https://github.com/Shopify/dawn/pull/848) I noticed an issue where the product was unavailable at a pick up location. But it wasn't showing the proper message/html due to an error coming up from `renderPreview()` where the event listener being added to the button was failing since there was no button in this scenario. 

Added a check/early return if no button is present in the HTML being rendered. 

**Other considerations**

**Demo links**
I've been testing on this store/page: https://ca.catleaves.com/products/invisible-cat-leaf

I went into the code in the admin and added the check that I'm adding in this PR to confirmed it fixed the issue.

I will try and recreate the same setup in our test store but I can't atm. 

**Checklist**

### DIFF
--- a/assets/pickup-availability.js
+++ b/assets/pickup-availability.js
@@ -57,8 +57,8 @@ if (!customElements.get('pickup-availability')) {
 
       document.body.appendChild(sectionInnerHTML.querySelector('pickup-availability-drawer'));
 
-      if (!this.querySelector('button')) return;
-      this.querySelector('button').addEventListener('click', (evt) => {
+      const button = this.querySelector('button');
+      if (button) button.addEventListener('click', (evt) => {
         document.querySelector('pickup-availability-drawer').show(evt.target);
       });
     }
--- a/assets/pickup-availability.js
+++ b/assets/pickup-availability.js
@@ -57,6 +57,7 @@ if (!customElements.get('pickup-availability')) {
 
       document.body.appendChild(sectionInnerHTML.querySelector('pickup-availability-drawer'));
 
+      if (!this.querySelector('button')) return;
       this.querySelector('button').addEventListener('click', (evt) => {
         document.querySelector('pickup-availability-drawer').show(evt.target);
       });


assets/pickup-availability.js
```diff
@@ -57,6 +57,7 @@ if (!customElements.get('pickup-availability')) {
 
       document.body.appendChild(sectionInnerHTML.querySelector('pickup-availability-drawer'));
 
+      if (!this.querySelector('button')) return;
```
KaichenWang, line 60: We could use an existing pattern for better consistency:
```
const button = this.querySelector('button');
if (button) button.removeEventListener('click', this.onClickRefreshList);
```
Also, `this.querySelector('button')` is queried multiple times in this JS file, could we take this opportunity to add a `if (button)` condition for each instance?

KaichenWang (approved):

assets/pickup-availability.js
```diff
@@ -57,7 +57,8 @@ if (!customElements.get('pickup-availability')) {
 
       document.body.appendChild(sectionInnerHTML.querySelector('pickup-availability-drawer'));
 
-      this.querySelector('button').addEventListener('click', (evt) => {
+      const button = this.querySelector('button');
+      if (button) button.addEventListener('click', (evt) => {
         document.querySelector('pickup-availability-drawer').show(evt.target);
       });
```
sofiamatulis, line 63: Looks good 👍 I wonder since we are creating the const button twice, if we could create it once in the `fetch` and pass it as params 🤔 

But not sure if it would be readable/confusing

sofiamatulis (approved):
Looks good 👍 Added a small comment

assets/pickup-availability.js
ludoboludo: Since one is in the `catch`, I'm not sure if it'd make sense 🤔 
I'm gonna leave it as is for now if that's ok 👍

PR #857, opened 2021-11-08
Update predictive search image to lazy load

dgitman:
Updated to fix theme check suggestion.

Before

```
shopify theme check
Checking . ...
sections/predictive-search.liquid:17: suggestion: ImgLazyLoading: Add a loading="lazy" attribute to defer loading of images.
	<img class="predictive-search__image"

226 files inspected, 1 offenses detected, 0 offenses auto-correctable
```

After
```
dgitman@Davids-MacBook-Pro-2 dawn % shopify theme check
Checking . ...
226 files inspected, 0 offenses detected, 0 offenses auto-correctable
```


sections/predictive-search.liquid
```diff
@@ -19,6 +19,7 @@
                 alt="{{ product.featured_media.alt }}"
                 width="50"
                 height="{{ 50 | divided_by: product.featured_media.preview_image.aspect_ratio }}"
+                loading="lazy"
```
ludoboludo, line 22: Hey @dgitman 👋 
I'm wondering if that's something we do on purpose. We can assume that the rest of the page content has loaded by the time the predictive search can/will be used. 
And when this image(s) will show we want them immediately. It should only come up when you start typing something in the search bar, so it doesn't create any issue on page load nor should block any rendering. 

I think it's just a case where theme-check can't understand the difference and default to ask for any `img` element to have it. 
Here I'd say we can ignore it 🙂 

But let me know if you feel otherwise.

dgitman: Hi @ludoboludo 👋 

According to PageSpeed Insights, images that are not in the first visible viewport should be lazy loaded. 

Let me know what you think.

ludoboludo: In this specific case, I think it's a bit different as it's not an image that is below the fold or in a drawer on first paint. 
Those images are going to show only after we fetch data to match the terms entered in the search bar. 
Since the images is an important part of the predictive search component, I think it makes sense to load them as soon as possible 🤔 

And it shouldn't block anything else from rendering as the page will be loaded already when you can start performing a search query. 

So I think it's ok as is. 
Might be worth exploring further though 🤷

ludoboludo: I will get in touch as well with the team behind theme-check to talk about it. Since it's not very pleasant to have an error come up when you're running it that isn't really one. 

Thanks again @dgitman

PR #889, opened 2021-11-15
Text boxes global settings setup

tyleralsbury:

affects password section, image banner, multicolumn (including image)

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126853808150/editor)

### DIFF
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -197,8 +197,13 @@
 }
 
 .banner__box {
-  border: 0;
-  padding: 4rem 3.5rem;
+  border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
+  padding: var(--text-boxes-padding);
+  border-radius: var(--text-boxes-radius);
+  box-shadow: var(--text-boxes-shadow-horizontal-offset)
+    var(--text-boxes-shadow-vertical-offset)
+    var(--text-boxes-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--text-boxes-shadow-opacity));
   position: relative;
   height: fit-content;
   align-items: center;
@@ -296,7 +301,6 @@
 
 @media screen and (min-width: 750px) {
   .banner__box {
-    padding: 5rem;
     width: auto;
     max-width: 71rem;
     min-width: 45rem;
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -87,6 +87,107 @@
     ]
   },
   {
+    "name": "Text boxes",
+    "settings": [
+      {
+        "type": "header",
+        "content": "Typography"
+      },
+      {
+        "type": "header",
+        "content": "Spacing"
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_padding",
+        "min": 0,
+        "max": 20,
+        "step": 1,
+        "unit": "px",
+        "label": "Padding",
+        "default": 10
+      },
+      {
+        "type": "header",
+        "content": "Borders"
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_border_opacity",
+        "min": 0,
+        "max": 100,
+        "step": 5,
+        "unit": "%",
+        "label": "Opacity",
+        "default": 50
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_border_width",
+        "min": 0,
+        "max": 10,
+        "step": 1,
+        "unit": "px",
+        "label": "Thickness",
+        "default": 1
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_radius",
+        "min": 0,
+        "max": 30,
+        "step": 3,
+        "unit": "px",
+        "label": "Corner radius",
+        "default": 0
+      },
+      {
+        "type": "header",
+        "content": "Shadow"
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_shadow_opacity",
+        "min": 0,
+        "max": 100,
+        "step": 5,
+        "unit": "%",
+        "label": "Opacity",
+        "default": 0
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_shadow_horizontal_offset",
+        "min": -20,
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Horizontal offset",
+        "default": 0
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_shadow_vertical_offset",
+        "min": -20,
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Vertical offset",
+        "default": 0
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_shadow_blur_radius",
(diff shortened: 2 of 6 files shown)


assets/section-image-banner.css
```diff
@@ -197,8 +197,13 @@
 }
 
 .banner__box {
-  border: 0;
-  padding: 4rem 3.5rem;
+  border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
+  padding: var(--text-boxes-padding);
```
kmeleta, line 201: This is where trying to use a single pixel value rather than magic sizes to control multiple sections is going to be tricky. Or at least require a larger range. With banner_box previously designed at ~4rem, the current max for text boxes being 2rem is going to slim down those designs.

config/settings_schema.json
```diff
@@ -86,6 +86,107 @@
...
+        "default": 50
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_border_width",
+        "min": 0,
+        "max": 10,
+        "step": 1,
+        "unit": "px",
+        "label": "Border width",
```
kmeleta, line 131: Not urgent but do we want to update to "thickness"?

config/settings_schema.json
```diff
@@ -86,6 +86,107 @@
...
+        "default": 1
+      },
+      {
+        "type": "range",
+        "id": "text_boxes_radius",
+        "min": 0,
+        "max": 30,
+        "step": 3,
+        "unit": "px",
+        "label": "Corner radius",
```
kmeleta, line 141: I could see keeping "Corner radius" but if these are organized into a "Borders" heading we probably don't need to prepend "Border" in each setting label.

kmeleta (approved):
Left some non critical notes. Otherwise works as expected.

templates/gift_card.liquid
```diff
@@ -63,6 +63,16 @@
...
+
+        --text-boxes-padding: {{ settings.text_boxes_padding }}px;
+        --text-boxes-border-opacity: {{ settings.text_boxes_border_opacity | divided_by: 100.0 }};
+        --text-boxes-border-width: {{ settings.text_boxes_border_width }}px;
+        --text-boxes-radius: {{ settings.text_boxes_radius }}px;
+        --text-boxes-shadow-opacity: {{ settings.text_boxes_shadow_opacity | divided_by: 100.0 }};
+        --text-boxes-shadow-horizontal-offset: {{ settings.text_boxes_shadow_horizontal_offset }}px;
+        --text-boxes-shadow-vertical-offset: {{ settings.text_boxes_shadow_vertical_offset }}px;
+        --text-boxes-shadow-blur-radius: {{ settings.text_boxes_shadow_blur_radius }}px;
       }
```
sofiamatulis, line 76: We dont have sections on gift card so I dont think this is needed

assets/section-image-banner.css
```diff
@@ -296,7 +301,6 @@
 
 @media screen and (min-width: 750px) {
   .banner__box {
-    padding: 5rem;
     width: auto;
     max-width: 71rem;
     min-width: 45rem;
```
sofiamatulis, line 306: For visibility Id add it to the description of the PR which sections are affected which I believe is only the banner and multi column

sofiamatulis: and email sign up banner (password page)

config/settings_schema.json
```diff
@@ -86,6 +86,107 @@
...
+        "id": "text_boxes_shadow_blur_radius",
+        "min": 0,
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "Blur radius",
+        "default": 0
+      }
+    ]
+  },
```
sofiamatulis, line 189: Should we already move these to the `en` file even if we arent requesting translations yet?

sofiamatulis (commented):
Looks good! Added small comments 👍 I think well just need to remove it from gift card

assets/section-image-banner.css
martinamarien: I agree with Ken. The Multicolumn section has some inner padding right now that doesn't allow the border to ever touch the text, while the banner section doesn't. 

The border is very close to the text and touches the button(s) if they're present. 
I think we'll need at least a little bit of default padding. 

Screen Shot 2021-11-16 at 2 03 25 PM

Same for the password page:
Screen Shot 2021-11-16 at 2 15 57 PM

layout/password.liquid
```diff
@@ -65,6 +65,15 @@
 
         --page-width: {{ settings.page_width | divided_by: 10 }}rem;
         --page-width-margin: {% if settings.page_width == '1600' %}2{% else %}0{% endif %}rem;
+
+        --text-boxes-padding: {{ settings.text_boxes_padding }}px;
```
martinamarien, line 69: I personally think anything padding/margin related should be in relative units for accessibility and better responsiveness. This will allow the page to scale appropriately for users who use zoom, or if the font size changes.

config/settings_schema.json
tyleralsbury: I find it easier to wait until we have the final copy and then do that, otherwise you need to go from one file to another to make the associations rather than just get them all in one place, then move them to their final spot.

sofiamatulis (approved):
Thanks for the changes 🎉

templates/gift_card.liquid
```diff
@@ -63,6 +63,16 @@
...
+        --page-width-margin: {% if settings.page_width == '1600' %}2{% else %}0{% endif %}rem;
+
+        --text-boxes-padding: {{ settings.text_boxes_padding }}px;
+        --text-boxes-border-opacity: {{ settings.text_boxes_border_opacity | divided_by: 100.0 }};
+        --text-boxes-border-width: {{ settings.text_boxes_border_width }}px;
+        --text-boxes-radius: {{ settings.text_boxes_radius }}px;
+        --text-boxes-shadow-opacity: {{ settings.text_boxes_shadow_opacity | divided_by: 100.0 }};
+        --text-boxes-shadow-horizontal-offset: {{ settings.text_boxes_shadow_horizontal_offset }}px;
+        --text-boxes-shadow-vertical-offset: {{ settings.text_boxes_shadow_vertical_offset }}px;
+        --text-boxes-shadow-blur-radius: {{ settings.text_boxes_shadow_blur_radius }}px;
```
martinamarien: What will this be used for on the giftcard template?

martinamarien (approved):
A few comments, but nothing we can't address afterwards

assets/section-image-banner.css
tyleralsbury: An alternative would be for it to be allowed to touch but to add `overflow: hidden` so that a merchant could create a text box with the buttons flush with the border - it could be an effect they're looking for.

templates/gift_card.liquid
tyleralsbury: Removed it as per Sofia's comment.

layout/password.liquid
tyleralsbury: We will figure this out with UX and then make a broad decision. Great call-out Martina.

PR #903, opened 2021-11-17
Update secondary button background color variable

martinamarien:
Fixes secondary background color for certain color schemes. 

Changed background color to use `var(--color-background)` instead of `var(--color-foreground)`.


tyleralsbury (approved):
Tested locally, looks good! 🎩 🚢

PR #902, opened 2021-11-17
Clean up popup drawer

sofiamatulis:
**Why are these changes introduced?**

Fixes https://github.com/Shopify/dawn/pull/865/#discussion_r750469308

**What approach did you take?**

Refer to the variable directly

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126899224598/editor)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #865, opened 2021-11-09
Settings popup and drawer

sofiamatulis:
**Why are these changes introduced?**

Fixes #0.

**What approach did you take?**

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126721327126/editor)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #917, opened 2021-11-18
Split popup and drawer settings

kmeleta:

Splits the popup drawer settings into separate categories. This PR does not add any additional settings or fix any bugs. Nor _should_ it add any new bugs. The only difference between both categories now is `Corner radius` has been excluded from `Drawers` and the associated info note has been removed.

Below is how I've applied the respective settings.

Drawers
- Hamburger menu (mobile)
- Facet filters (mobile)
- Pickup availability

Popups
- Main nav submenu (desktop)
- Facet filters (desktop)
- Localization selects
- Cart notification
- Predictive search results
- Main product popup modal

**Other considerations**

For additional context, if needed..

Original popup drawer settings commit https://github.com/Shopify/dawn/pull/865
Follow up to popup drawer settings https://github.com/Shopify/dawn/pull/902

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126921637910/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2285,15 +2285,15 @@ details[open] > .header__icon--menu .icon-hamburger {
 }
 
 .header__submenu.list-menu--disclosure {
-  border-radius: var(--popup-drawer-corner-radius);
-  border-color: rgba(var(--color-foreground), var(--popup-drawer-border-opacity));
+  border-radius: var(--popup-corner-radius);
+  border-color: rgba(var(--color-foreground), var(--popup-border-opacity));
   border-style: solid;
-  border-width: var(--popup-drawer-border-width);
+  border-width: var(--popup-border-width);
   filter: drop-shadow(
-    var(--popup-drawer-shadow-horizontal-offset)
-    var(--popup-drawer-shadow-vertical-offset)
-    var(--popup-drawer-shadow-blur-radius)
-    rgba(var(--color-foreground), var(--popup-drawer-shadow-opacity))
+    var(--popup-shadow-horizontal-offset)
+    var(--popup-shadow-vertical-offset)
+    var(--popup-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--popup-shadow-opacity))
   );
 }
 
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -723,12 +723,12 @@
       },
       {
         "type": "range",
-        "id": "popup_blur_radius",
+        "id": "popup_shadow_blur_radius",
         "min": 0,
         "max": 20,
         "step": 2,
         "unit": "px",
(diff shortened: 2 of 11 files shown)


config/settings_schema.json
```diff
@@ -718,77 +718,141 @@
...
+        "min": -20,
+        "max": 20,
+        "step": 2,
+        "unit": "px",
+        "label": "t:settings_schema.drawers.settings.drawer_shadow_vertical_offset.label",
+        "default": 0
+      },
+      {
+        "type": "range",
+        "id": "drawer_blur_radius",
```
KaichenWang, line 850: Minor, but would `drawer_shadow_blur_radius` be more consistent here?

config/settings_schema.json
```diff
@@ -718,77 +718,141 @@
...
         "step": 2,
         "unit": "px",
-        "label": "t:settings_schema.popup_drawer.settings.popup_drawer_shadow_vertical_offset.label",
+        "label": "t:settings_schema.popups.settings.popup_shadow_vertical_offset.label",
         "default": 0
       },
       {
         "type": "range",
-        "id": "popup_drawer_blur_radius",
+        "id": "popup_blur_radius",
```
KaichenWang, line 785: Minor, but would `popup_shadow_blur_radius` be more consistent here?

assets/base.css
```diff
@@ -2303,15 +2303,15 @@ details[open] > .header__icon--menu .icon-hamburger {
 }
 
 .header__submenu.list-menu--disclosure {
-  border-radius: var(--popup-drawer-corner-radius);
-  border-color: rgba(var(--color-foreground), var(--popup-drawer-border-opacity));
+  border-radius: var(--popup-corner-radius);
```
KaichenWang, line 2306: Not in scope for this PR, but just wanted to make a note.

We apply essentially the same set of CSS properties in multiple files for all "popups" and "drawers". Is there an opportunity to create (and add to the HTML) a utility class  like `.popup` and `.drawer` on which we apply the CSS? Thoughts on this approach vs. the current? This pattern would apply to other settings as well.

KaichenWang (approved):

melissaperreault (approved):
Thanks! The split looks good! We'll address the full list of settings under its own issue!

assets/base.css
kmeleta: Yes absolutely. I'd much prefer to utilize new global classes wherever we can. In a perfect world, global settings would change styles at the highest level with only overrides/exceptions taking place in lower level section styles.

config/settings_schema.json
kmeleta: Yeah might as well fix it now. Updated.

Issue #909, opened 2021-11-17
Global settings priorities discussion


melissaperreault:
## Context

After prototyping and regrouping we've come to space where we can talk about priorities divide and conquer.
Below I will list what comes to mind, let's combined and divide on multiples issues after and moving forward. Just feels right to have everything at one place to see what we have to do.

### Meeting MVP expectations

1. Remove from the demo store reference (if not done yet) the following settings categories to align with the latest decision
- Sections
- Dividers and lines
2. By removing `Sections` settings, 2 things we need to introduce
- Image with text section, the text part would be styled by the Text box setting moving forward and the media will renamed managed by Media setting
- Bring `Vertical space between sections` setting to `Layout` category (Formerly named `Grid`) ([Preview of the new hierarchy internal file](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=141%3A10944))
3. Split `Popups` and `Drawers` to 2 distinct categories and follow the Figma guidance
4. Discuss if setting combination conditions should be done in a single PR or more granularly  
- Examples:
   - Do not duplicate border for touching borders when horizontal spacing is set to 0
   - Remove the corner radius when horizontal spacing is set to zero or when the component hug another one.

### Follow up tasks
- Buttons review feedback, I can see where ensuring section styles behave as expected will have a greater impact for this week
- Prioritize setting's feedback for each setting categories
- Define specific range of each setting variables
- Content update, this can come in next week once we (UX) have more guidance to give for official settings values and condition rules for all items. I could own and contribute to content update!

kmeleta:
Those priorities seems reasonable to me. Just some random remarks on some of them..

- **Image with text** updates _should_ be a relatively quick win I'd think.

- For **vertical spacing between sections**, moving the settings to layout should be quick enough, but I'd love if we had time to solve this more globally than previously done in prototypes. Remove section-specific margin overrides where possible, make title markup and margins predictable/consistent across sections, etc. I feel a little more time spent to clean this up now could save time/bugs later.

- Not sure if Kai already has thoughts on the **combination conditions** but we definitely need give some thought/explore as to how we make elements "spatially aware". The approach likely may just have to be more granular and specific to each section, but it'd be nice to see if there are any global rules we can establish somehow.

[...]

PR #918, opened 2021-11-19
Image with text section - Apply global textbox and media settings

KaichenWang:

Part of https://github.com/Shopify/dawn/issues/909#issuecomment-973200276

- Use global textbox CSS variables in `component-image-with-text.css`
- Remove `color-{{ section.settings.color_scheme }}` from the container to prevent background color from appearing around image and textbox when corner-radius > 0
- Clean up CSS

**Other considerations**

- ⚠️ Like other sections where there is 0 grip gap, logic needs to be added to ensure collapsing of adjacent borders + elimination of border-radius for adjacent corners. This it not yet implemented.
- ⚠️ Shadow overlapping and z-index will require more refinement as well

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126924357654)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126924357654/editor)

**Checklist**

### DIFF
--- a/assets/component-image-with-text.css
+++ b/assets/component-image-with-text.css
@@ -1,5 +1,6 @@
 .image-with-text {
   margin-top: var(--sections-spacing);
+  margin-bottom: var(--sections-spacing);
 }
 
 @media screen and (min-width: 750px) {
--- a/assets/component-image-with-text.css
+++ b/assets/component-image-with-text.css
@@ -2,10 +2,6 @@
   margin-top: var(--sections-spacing);
 }
 
-.image-with-text:not(.color-scheme-background-1) {
-  margin-bottom: var(--sections-spacing);
-}
-
 @media screen and (min-width: 750px) {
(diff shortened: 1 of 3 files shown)


assets/component-image-with-text.css
```diff
@@ -83,9 +76,13 @@
 }
 
 .image-with-text__content {
+  align-items: flex-start;
+  background-clip: padding-box;
```
kmeleta, line 75: Good call 👍

melissaperreault (approved):
Thank you Kai! The application of the setting works as expected. I'll report UX feedback on another issue for the behaviour and aesthetic needs.

kmeleta (approved):
This works as I'd expect right now. Obviously there are some details to work out here in regards to clashing styles between the text box and media settings, but I don't think it's worth trying to deal with them until we've considered those issues from higher level.

PR #933, opened 2021-11-21
Back top button

renanrcrr:
Integrating the back top button with new dawn.


tyleralsbury:
Hi @renanrcrr .  Thanks for your contribution.  We are not looking to add a back to top button at this time.  We will consider it for a later version of Dawn.

PR #937, opened 2021-11-23
Product Media Layout updates

### LINKED ISSUE #828: Product template media layout update
## Context

We'll be offering some flexibility to the product page media and layout through settings update to enable merchant to have more optionality for that important template.

### New settings details
**Large screens layout**
- [ ] Stacked: Basically the current collage grid built on Dawn
   - Zoom behaviour
      - When you click on the media, it opens the modal where you can zoom with your fingers or keyboard controls
- [ ] Thumbnails: A featured media with thumbnails underneath
   - Zoom behaviour
      - When you tap on the thumbnails, the media is previewed in the featured container.
      - You can click on the featured image, when it is an image, to open the modal and preview the list of all media. You can play 3D models and videos inside the modal.
      - The video or 3D models plays inline (Same behaviour as the other layout except you click on the thumbnail to trigger the play)
   - Media badges
      - They will live on the thumbnails instead of the poster image
   - Featured media height
      - [ ] We'll need to set some `max-height` dependant on the media width to ensure the thumbnails are visible on large screens.

### Media size
Similar to settings we can find in Debut, we want to offer media width variations for merchants to pick from.

- [x] A set of sizes to change the media width
   - [x] Small
      - _Media column width 45%_
   - [x] Medium
      - _Media column width 55%_
   - [x] Large (default)
      - _Media column width 65%_

### Thumbnails on mobile
Thumbnails can become quite tall and push the product information further down. We want to give merchants the flexibility to show or hide thumbnails regardless of their layout choice.

## UX Checklist

 - [x] Scott reviewed the accessibility annotations
 - [ ] Open an issue for Docs update

LucasLacerdaUX:
Adds three new settings to product page:
- Large screens layout
- Media size
- Mobile thumbnails

Closes #828.

- Created a new **MediaGallery** component to manage the GalleryViewer. This component manages two sliders (`gallery` and `thumbnails`) and keeps them in sync. It also serves as an interface for other components, like `VariantSwitch`. Instead of changing the DOM directly, we now make this request via a `setActiveMedia(media, prepend)` function.
- **Thumbnails** are implemented using a `<slider-component>`. This is different to the dots/numbers in the `Slideshow` component that @ludoboludo implemented on #840, because both thumbnails and viewer can be controlled separately. The same HTML is used for both Mobile and Desktop.

**Setting 1: Large screens layout**
- **Stacked:**  if sticky column setting is enabled, info column is sticky.
- **Thumbnails:** if sticky column setting is enabled, media column is sticky. Will limit the height of media.

**Setting 2: Media size**
- **Small:** 45% media / 55% content. 4 thumbnails per row. 20x20 thumbnail badges. 500px maximum height if `Thumbnails` layout is set.
- **Medium:** 55% media / 45% content. 5 thumbnails per row. 30x30 thumbnail badges. 600px maximum height if `Thumbnails` layout is set.
- **Large:** 65% media / 35% content. 6 thumbnails per row. 30x30 thumbnail badges. 600px maximum height if `Thumbnails` layout is set.

**Setting 3: Mobile thumbnails**
- **Hide** will use `< 1/3 >` format for slider.
- **Show** can be enabled independently from the **Large screens layout** setting. Uses the same *Thumbnails* component, but inside a slider. Pressing the `next/prev` button will scroll 3 slides at a time. If user scrolls the main gallery slider, the thumbnail slider will be kept in sync.

**Accessibility:**
- Thumbnails (`button`) have a `aria-controls="GalleryViewer-{{ section.id }}"` and are labelled (via `aria-label`) depending on the media type:
  - Images: `Load image {{ index }} in gallery view`
  - 3D Models: `Load 3D Model {{ index }} in gallery view`
  - Videos: `Play video {{ index }} in gallery view`
- If set to Thumbnails mode, the gallery viewer will have a `<h2>Gallery Viewer</h2>` at the top.
- When you click on a Thumbnail:
  - If it's a video or 3D Model, it will autoplay.
  - If it's an image, a `role="status"` div will receive `Image {{ index }} is now available in gallery view` once the `` is loaded for the first time.

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127010603030)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127010603030/editor)

**Checklist**


locales/en.default.json
```diff
@@ -104,8 +104,12 @@
       "on_sale": "Sale",
       "product_variants": "Product variants",
       "media": {
-        "open_featured_media": "Open featured media in gallery view",
-        "open_media": "Open media {{ index }} in gallery view",
+        "gallery_viewer": "Gallery Viewer",
+        "load_image": "Load image {{ index }} in gallery view",
+        "load_model": "Load 3D Model {{ index }} in gallery view",
+        "load_video": "Play video {{ index }} in gallery view",
+        "image_available": "Image {{ index }} is now available in gallery view",
```
LucasLacerdaUX, line 111: Question to @svinkle: should this be announced every time the user clicks the thumbnail or just on the first time the image finishes loading? 

If it's just for the first time, I'm unsure if we can guarantee this to fire, as browsers may skip the `onload` event if the image is cached. For now, it's implemented as a one-time only. Let me know if we should change to always display this message.

assets/global.js
```diff
@@ -598,25 +613,16 @@ class VariantSelects extends HTMLElement {
   updateMedia() {
     if (!this.currentVariant) return;
     if (!this.currentVariant.featured_media) return;
-    const newMedia = document.querySelector(
-      `[data-media-id="${this.dataset.section}-${this.currentVariant.featured_media.id}"]`
-    );
 
-    if (!newMedia) return;
+    this.stickyHeader = this.stickyHeader || document.querySelector('sticky-header');
+    if (this.stickyHeader) this.stickyHeader.dispatchEvent(new Event('preventHeaderReveal'));
```
ludoboludo, line 618: I think this might need to be placed elsewhere 🤔 
[video](https://screenshot.click/21-11-6z3mn-ijgto.mp4)

Also when the sticky header is showing and you click on a variant image, it should probably hide the sticky header if it's currently sticky I think. You can see it at 16s in the video above.

assets/global.js
```diff
@@ -543,13 +543,28 @@ class SliderComponent extends HTMLElement {
...
+    const lastVisibleSlide = this.slider.clientWidth + this.slider.scrollLeft - 10;
+    const isSlideVisible = element.offsetLeft < lastVisibleSlide && element.offsetLeft > this.slider.scrollLeft;
+    if (!isSlideVisible) this.slider.scrollTo({ left: element.offsetLeft });
+  }
+
   onButtonClick(event) {
     event.preventDefault();
-    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderLastItem.clientWidth : this.slider.scrollLeft - this.sliderLastItem.clientWidth;
+    const step = event.currentTarget.dataset.step || 1;
+    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + (step * this.sliderLastItem.clientWidth) : this.slider.scrollLeft - (step * this.sliderLastItem.clientWidth);
```
ludoboludo, line 567: There is a bit of a delay happening sometimes between the thumbnail sliding to the currently active image: [video](https://screenshot.click/21-11-egsr0-gcq7m.mp4)

assets/global.js
```diff
@@ -543,13 +543,28 @@ class SliderComponent extends HTMLElement {
...
+  scrollIntoView(element) {
+    const lastVisibleSlide = this.slider.clientWidth + this.slider.scrollLeft - 10;
+    const isSlideVisible = element.offsetLeft < lastVisibleSlide && element.offsetLeft > this.slider.scrollLeft;
+    if (!isSlideVisible) this.slider.scrollTo({ left: element.offsetLeft });
+  }
+
   onButtonClick(event) {
     event.preventDefault();
-    const slideScrollPosition = event.currentTarget.name === 'next' ? this.slider.scrollLeft + this.sliderLastItem.clientWidth : this.slider.scrollLeft - this.sliderLastItem.clientWidth;
+    const step = event.currentTarget.dataset.step || 1;
```
ludoboludo, line 566: The steps with also the possibility to drag on mobile can create some situation where you can't click the next/previous button: [video](https://screenshot.click/21-11-flgn8-vmm63.mp4)

ludoboludo: Seems to only do it when scrolling back. Starting at the end to go back to the beginning.

ludoboludo (commented):
Looks really good. Found the code nice to read as well 👌 

Tested quite a few scenarios and left a few notes on what I noticed.

assets/global.js
```diff
@@ -543,13 +543,28 @@ class SliderComponent extends HTMLElement {
...
+    }
+
+    if (!this.pageCount || !this.pageTotal) return;
+
     this.pageCount.textContent = this.currentPage;
     this.pageTotal.textContent = this.totalPages;
   }
 
+  scrollIntoView(element) {
+    const lastVisibleSlide = this.slider.clientWidth + this.slider.scrollLeft - 10;
```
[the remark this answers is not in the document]
tyleralsbury: I would consider creating a named constant for it just to make it clear what the `10` represents.

assets/section-main-product.css
```diff
@@ -816,3 +857,152 @@ a.product__text {
...
+
+.thumbnail {
+  position: absolute;
+  top: 0;
+  left: 0;
+  display: block;
+  height: 100%;
+  width: 100%;
+  padding: 0;
+  color: rgb(var(--color-base-text));
```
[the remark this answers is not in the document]
tyleralsbury: If the merchant uploaded an image with no background, I think we would want it to be transparent, in case they've got a specific effect in mind.

locales/en.default.json
svinkle: Each time, as is the case with a video loading (audio playback indicates a successful load) and a 3D model (built-in live region announces "loaded" state.)

svinkle:
Initial first-pass observations:

1. On first click of a video, focus shifts to the `video` or related `iframe` element. Subsequent clicks do not shift focus. **Focus should remain on the thumbnail `button`.**
2. Each media type should have its own index. This helps to provide accurate control accessible naming.

   Ex., tabbing through thumbnail controls with mixed media (2 images, then a video, then another image, then a 3D model) would ideally announce:
   
   > 🗣 "Load image 1 in gallery view <kbd>Tab</kbd> Load image 2 in gallery view <kbd>Tab</kbd> Play video 1 in gallery view <kbd>Tab</kbd> Load image 3 in gallery view <kbd>Tab</kbd> Load 3D Model 1 in gallery view"
3. The gallery viewer markup in this view needs some adjustment. It doesn't make sense to reuse the list structure for the gallery view since there will ever only be 1 list item presented at a time.

   When the conditions of the **Thumbnail layout and desktop viewport** are true, inside the `slider-component` element, remove semantic meaning by applying `role="presentation"` to the `ul` and each `li`.
4. The heading I suggested in Figma throws off the heading order. Let's remove this but instead, create a landmark for the gallery viewer.

   Add `aria-label="Gallery viewer"` and `role="region"` to the `media-gallery` component. 
5. I see the `role="status"` in the DOM but I don't see any updates being pushed to this element. I believe there should be something pushed for when an image has loaded?

snippets/product-thumbnail.liquid
```diff
@@ -73,99 +73,99 @@
...
+  <div class="product__media-container"{% if limit_height > 0 %} style="max-width: {{ limit_height | divided_by: 10 | times: media.preview_image.aspect_ratio }}rem"{% endif %}>
+    <div class="product__media media media--transparent" style="padding-top: {{ 1 | divided_by: media.preview_image.aspect_ratio | times: 100 }}%;">
+      <img
+        srcset="{% if media.preview_image.width >= 288 %}{{ media.preview_image | img_url: '288x' }} 288w,{% endif %}
+          {% if media.preview_image.width >= 576 %}{{ media.preview_image | img_url: '576x' }} 576w,{% endif %}
+          {% if media.preview_image.width >= 750 %}{{ media.preview_image | img_url: '750x' }} 750w,{% endif %}
+          {% if media.preview_image.width >= 1100 %}{{ media.preview_image | img_url: '1100x' }} 1100w,{% endif %}
+          {% if media.preview_image.width >= 1500 %}{{ media.preview_image | img_url: '1500x' }} 1500w,{% endif %}
+          {{ media.preview_image | img_url: 'master' }} {{ media.preview_image.width }}w"
+        src="{{ media | img_url: '1500x' }}"
```
tyleralsbury, line 85: There will need to be a multiplier on the values in the gallery for optimal desktop image sizes dependent on the width of the gallery.  Ping me if you want details, but for example... if the original width of the container was `50%` and an image size for that is `1000`, and the setting is now `60%`, the image needs to be `1200` when that setting is selected.

I think it will take some `assign`s, unfortuantely.

tyleralsbury (commented):
After a quick pass on the code, this is looking really great.  I only have one comment so far, plus the vid I sent you on Slack.  Great catches by Ludo as well. 👍

svinkle:
Product images are not visible in Windows Hight Contrast mode in Firefox. Adjust the CSS of the `button` to use the `:after` pseudo element. Review the product card CSS as an example.

Dawn product page in Windows Hight Contrast mode in Firefox. Gallery viewer is an empty black area.

duygukalaycioglu (changes requested):
- [ ] Decrease the spacing below the image and the thumbnails on mobile.
spacing on mobile

- [ ] When the thumbnails are hidden, decrease the spacing above the product title on mobile. 
thumbnails hidden - spacing
Thanks!

assets/section-main-product.css
```diff
@@ -816,3 +857,152 @@ a.product__text {
...
+  justify-content: center;
+  height: 2rem;
+  width: 2rem;
+  position: absolute;
+  right: 0.4rem;
+  top: 0.4rem;
+  z-index: 1;
+  transition: color var(--duration-short) ease,
+    opacity var(--duration-short) ease;
+}
```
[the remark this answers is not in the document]
LucasLacerdaUX: Good catch. Just combined them and kept the overwrite values below :)

sections/main-product.liquid
```diff
@@ -19,54 +19,138 @@
...
+      <media-gallery id="MediaGallery-{{ section.id }}"{% if section.settings.enable_sticky_info %} class="product__media-gallery"{% endif %}>
+        <div id="GalleryStatus-{{ section.id }}" class="visually-hidden" role="status"></div>
+        <slider-component id="GalleryViewer-{{ section.id }}" class="slider-mobile-gutter">
+          <a class="skip-to-content-link button visually-hidden" href="#ProductInfo-{{ section.id }}">
+            {{ "accessibility.skip_to_product_info" | t }}
+          </a>
+          {% if section.settings.gallery_layout == 'thumbnail' %}
+            <h2 class="visually-hidden">{{ 'products.product.media.gallery_viewer' | t }}</h2>
+          {% endif %}
+          <ul class="product__media-list grid grid--peek list-unstyled slider slider--mobile" role="list">
```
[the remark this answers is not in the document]
LucasLacerdaUX: I've updated to use ids for `ul`, but I don't think we should be adding `id`s to individual slides (`<li>`) 👀

If it's only to restrict the querying, we can query inside the `ul` for its `<li>` elements.

sections/main-product.liquid
```diff
@@ -444,6 +528,77 @@
   }
 
   customElements.define('product-modal', ProductModal);
```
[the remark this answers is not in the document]
LucasLacerdaUX: Oh, got it. We need to check if the product-modal is already defined. :)

assets/global.js
LucasLacerdaUX: Moved to the MediaGallery component! :)

katycobb:
Hi @LucasLacerdaUX I've been working with @duygukalaycioglu  to make some content updates. They're reflected in the Figma file + listed below. Let me know if you need more info, and D let me know if there's anything I've missed! 

- Large screens layout --> `Desktop layout`
- Add help text below Desktop layout: `Thumbnails is the default mobile layout.`
- Media size --> `Desktop media size `
- Add help text  below Desktop media size: `Media is automatically optimized for mobile. `

LucasLacerdaUX:
@katycobb @duygukalaycioglu I have a question about:
> Thumbnails is the default mobile layout.

Does that mean we are changing `Mobile thumbnails` default from `Hide` to `Show`? Right now, the default is Hide.

image

katycobb:
@LucasLacerdaUX oof, good question. @duygukalaycioglu  let's chat about this!

katycobb:
@LucasLacerdaUX  I connected with @duygukalaycioglu: we'll update the help text to `Thumbnails is the default mobile layout.`

No change to the mobile thumbnail default needed!

assets/section-main-product.css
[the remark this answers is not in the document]
LucasLacerdaUX: Good catch. This should've been a transparent background with a border. I'll update it.

sections/main-product.liquid
```diff
@@ -444,6 +528,77 @@
...
+          activeItem.dataset.mediaPosition
+        );
+      };
+    }
+
+    playActiveMedia(activeItem) {
+      window.pauseAllMedia();
+      const deferredMedia = activeItem.querySelector('.deferred-media');
+      if (deferredMedia) deferredMedia.loadContent();
+    }
```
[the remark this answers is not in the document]
LucasLacerdaUX: This was intentional on mobile. Here's the Figma reference that specifies that it should play when tapping thumbnails on mobile:
image
The user will have a visual feedback that the thumbnail they are clicking is a video via the badge icon. I've implemented it so only the thumbnail click causes the video to autoplay. Swiping on the carousel won't.

As for autoplaying on every click vs on the first click only (current), that's a good point. I don't have a strong opinion on that. I'll check with designers.

sections/main-product.liquid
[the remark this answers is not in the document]
LucasLacerdaUX: I'll fix this in a follow up PR. This isn't related to this PR and I'd like to avoid increasing the scope of this one.

#978

assets/global.js
LucasLacerdaUX: I have no idea how to fix this without a big refactor on the slider component.

I don't think it's blocker, though. Would it be okay to solve this on follow up PRs?

svinkle:
@LucasLacerdaUX 

> Clicking a thumbnail will always announce that the image is loaded.

Confirmed. New issue being this text can be found when navigating via virtual cursor. If you can set a timeout to clear `#GalleryStatus` after a couple seconds that'd be great.

> The `Gallery Viewer` title is now applied to the `media-gallery` component via `aria-label` and `role="region"` instead of being an `<h2>`

Confirmed.

> If thumbnails are enabled and we're using desktop viewport, `role="presentation"` is applied to the `<ul>` and all `<li>` in the Gallery Viewer.

Confirmed on desktop and mobile for the thumbnail view.

> `aria-describedby` was added to thumbnail `buttons`, linking them to the `img` element.

Confirmed.

> Each media type now has its own index.

Confirmed, though image type seems to start at `0`. Let's start at `1`.

> Firefox - Windows High Contrast mode fixed by styling the `::after` pseudoelement instead of the button :)

Confirmed. New issue being the focus ring is placed below and to the left of the larger image in the gallery viewer. Seems to be images only. Please review.

> Unfortunately, this seems to be caused by YouTube's video player.

Thanks for looking into this. It seems to be true for hosted, non YouTube video as well.

> This would limit the current UX as it would no longer allow mobile users to navigate through the media via swipe gestures.

Ah, no worries. I didn't realize the swipe feature existed. With this context in mind, ignore my original comment.

katycobb:
Hi @LucasLacerdaUX: Melissa, D and I connected and aligned on the behavior of this section and have made some content updates we think will better describe the settings. Let me know if this makes sense given your feedback and knowledge of the section! 

- Remove help text under `Desktop layout`
- Rename `Mobile thumbnails` field label to `Mobile layout`
- Update options under Mobile layout to: `Show thumbnails` and `Hide thumbnails `

cc: @duygukalaycioglu and @melissaperreault

svinkle:
@LucasLacerdaUX 

> I've followed the `aria-hidden` => `true` pattern 

Confirmed. 👍 

> It now checks if the component is using a non `stacked` layout

Looks good.

> Is this issue restricted to when we are using VoiceOver

Seems to be Safari only, when using <kbd>Tab</kbd>.

<details>
  <summary>Screenshot</summary>

  Dawn product page. Product image in gallery view features small focus ring in bottom left corner.

</details>

> The focus now remains on the thumbnail regardless of the media type you select.

Confirmed.

assets/section-main-product.css
```diff
@@ -540,8 +569,14 @@ a.product__text {
 }
 
 .product__media-toggle {
-  background-color: transparent;
+  display: flex;
   border: none;
+  background: transparent;
```
ludoboludo, line 574: should we keep `background-color` here ? I think it's better since we're only changing the background color.

sections/main-product.liquid
```diff
@@ -19,54 +19,144 @@
...
-              {% render 'product-thumbnail', media: media, position: 'featured', loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true %}
-            </li>
-          {%- endif -%}
-          {%- for media in product.media -%}
-            {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
-              <li class="product__media-item grid__item slider__slide{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
-                {% render 'product-thumbnail', media: media, position: forloop.index, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true %}
+      <media-gallery id="MediaGallery-{{ section.id }}" role="region" {% if section.settings.enable_sticky_info %} class="product__media-gallery"{% endif %} aria-label="{{ 'products.product.media.gallery_viewer' | t }}" data-desktop-layout="{{ section.settings.gallery_layout }}">
```
ludoboludo, line 24:
```suggestion
      <media-gallery id="MediaGallery-{{ section.id }}" role="region" {% if section.settings.enable_sticky_info %}class="product__media-gallery"{% endif %} aria-label="{{ 'products.product.media.gallery_viewer' | t }}" data-desktop-layout="{{ section.settings.gallery_layout }}">
```

melissaperreault:
>  One thing I wanted to bring up with content folks is the name of the setting that encompass both product columns (medias + infos). The name we use right now is Product information which make it sounds like it will only be for the right side of the product "section". So if I was to look for a place to edit the media layout, it might not be very self explanatory 🤔
Something like Product section might be a bit more general and represents well both columns that you can edit via those settings. Just a thought.
[Screenshot](https://screenshot.click/21-12-7awgh-z5ia1.png)


@ludoboludo Thanks, this is good feedback! Not a blocker to ship but I would like to get @katycobb's opinion on the table  when she has a chance.  🙏

assets/component-slider.css
```diff
@@ -32,6 +32,24 @@ slider-component {
   }
 }
 
+@media screen and (min-width: 749px) {
```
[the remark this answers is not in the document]
LucasLacerdaUX: Oh, good catch. I'll update the breakpoint.

As for the similar class, I'll rebase when you merge yours so I can see if they can be combined 👀

assets/global.js
```diff
@@ -598,25 +619,13 @@ class VariantSelects extends HTMLElement {
...
     modalContent.prepend(newMediaModal);
-    parent.prepend(newMedia);
-    this.stickyHeader = this.stickyHeader || document.querySelector('sticky-header');
-    if(this.stickyHeader) {
-      this.stickyHeader.dispatchEvent(new Event('preventHeaderReveal'));
-    }
-    window.setTimeout(() => {
-      parent.scrollLeft = 0;
-      parent.querySelector('li.product__media-item').scrollIntoView({behavior: 'smooth'});
-    });
```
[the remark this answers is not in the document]
LucasLacerdaUX: We are still scrolling to the variant image, except when using the `Thumbnails` layout. If you set it to Stacked, we'll still scroll.

`main-product.liquid`:
```js
      window.setTimeout(() => {
        if (this.elements.thumbnails) {
          activeMedia.parentElement.scrollTo({ left: activeMedia.offsetLeft });
        }
        if (!this.elements.thumbnails || this.dataset.desktopLayout === 'stacked') {
          activeMedia.scrollIntoView({behavior: 'smooth'});
        }
      });
```

sections/main-product.liquid
```diff
@@ -19,54 +19,144 @@
...
+              assign variant_images = product.images | where: 'attached_to_variant?', true | map: 'src'
+              assign media_count = product.media.size
+              if section.settings.hide_variants
+                assign media_count = media_count | minus: variant_images.size | plus: 1
+              endif
+            -%}
+            {%- if product.selected_or_first_available_variant.featured_media != null -%}
+            {%- assign featured_media = product.selected_or_first_available_variant.featured_media -%}
+              <li id="ProductMedia-{{ section.id }}-{{ featured_media.id }}" class="product__media-item grid__item slider__slide is-active{% if featured_media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains featured_media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ featured_media.id }}">
+                {% render 'product-thumbnail', media: featured_media, position: 0, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true %}
```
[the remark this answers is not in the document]
LucasLacerdaUX: Updated to start at 1. Also changed how it's being handled so it's no longer based on the `forloop.index` but on a variable that is only incremented when the media is rendered. Previously, you could have the index skipping 1 media if the featured product media wasn't the first media in the product.

I think we can consider updating these strings in a follow-up PR to be `Open {{ media_type }} {{ media_index }} in pop up" like we did for thumbnails.

LucasLacerdaUX:
PR is ready for final reviews.

Thumbnails / image `srcset` will be updated later this morning, but I would suggest to review everything else in the meantime :)

cc @tyleralsbury @ludoboludo @duygukalaycioglu

duygukalaycioglu (dismissed):
Looking great!!

assets/component-slider.css
```diff
@@ -32,6 +32,24 @@ slider-component {
...
+    scroll-behavior: smooth;
+    scroll-padding-left: 1rem;
+    -webkit-overflow-scrolling: touch;
+    margin-bottom: 1rem;
+  }
+
+  .slider.slider--tablet-up .slider__slide {
+    margin-bottom: 0;
+    padding-bottom: 0;
+  }
```
ludoboludo, line 50: I think you might need to edit the negative margin applied to the `.grid` class here by using a padding-left value or resetting the margin on the `product__media-list` class ? 
On smaller window size (I was testing 915px) the media image is bigger than the container which makes the thumbnail list look not aligned:

assets/global.js
```diff
@@ -505,7 +506,7 @@ class SliderComponent extends HTMLElement {
     this.sliderItems = this.querySelectorAll('[id^="Slide-"]');
     this.enableSliderLooping = false;
     this.currentPageElement = this.querySelector('.slider-counter--current');
-    this.pageTotal = this.querySelector('.slider-counter--total');
+    this.pageTotalElement = this.querySelector('.slider-counter--total');
```
ludoboludo, line 509: I wonder if pageTotalElement is the best. Since for some uses cases it's not the total amount of elements but pages. As a page/view can have 1-3 slides within it:

assets/global.js
```diff
@@ -529,32 +530,51 @@ class SliderComponent extends HTMLElement {
     this.update();
   }
 
+  resetPages() {
+    this.sliderItems = this.slider.querySelectorAll('li');
```
ludoboludo, line 534: This seem specific to sliders using `li` elements. Should it target the elements differently so it's a bit more generic and could be used in any slider situation?

assets/global.js
```diff
@@ -529,32 +530,51 @@ class SliderComponent extends HTMLElement {
...
       this.prevButton.setAttribute('disabled', 'disabled');
     } else {
       this.prevButton.removeAttribute('disabled');
     }
 
-    if (this.currentPage === this.totalPages) {
+    if (this.isSlideVisible(this.sliderLastItem)) {
       this.nextButton.setAttribute('disabled', 'disabled');
     } else {
       this.nextButton.removeAttribute('disabled');
```
ludoboludo, line 565: There seem to be a hick-up here. When I try the slider on a feat. collection section I can't get to the end of that slider:

assets/section-main-product.css
```diff
@@ -23,7 +23,10 @@
 }
 
 @media screen and (min-width: 750px) {
-  .product__info-container--sticky {
+  .product--thumbnail .product__media-gallery,
+  .product--thumbnail_slider .product__media-gallery,
+  .product--stacked .product__info-container--sticky {
+    display: block;
     position: sticky;
```
ludoboludo, line 30: One thing I notice here with position sticky on the media gallery is that it makes it a bit annoying to get to see the video: [video](https://screenshot.click/21-12-dt0hk-74tj7.mp4)

assets/section-main-product.css
ludoboludo: Since background is also referring to other properties that aren't tweaked here, I think `backrgound-color` is the way to go no?

sections/main-product.liquid
```diff
@@ -19,54 +19,146 @@
...
+                        height="200"
+                        width="200"
+                        loading="lazy"
+                      >
+                    </button>
+                  </li>
+                {%- endunless -%}
+              {%- endfor -%}
+            </ul>
+            <button type="button" class="slider-button slider-button--next{% if media_count <= 3 %} small-hide{% endif %}{% if media_count <= 4 %} medium-hide large-up-hide{% endif %}" name="next" aria-label="{{ 'general.slider.next_slide' | t }}" aria-controls="GalleryThumbnails-{{ section.id }}" data-step="3">{% render 'icon-caret' %}</button>
```
ludoboludo, line 158: I'm wondering if the next and previous buttons should be using `aria-hidden` ?
Because the thumbnail slider is a list, when I'm navigating through it, the only times I land on the previous or next arrow it's dimmed. 🤔  but maybe I'm missing use cases. [video](https://screenshot.click/21-12-mu3o3-0vk1h.mp4)

ludoboludo (commented):
Looking very nice 👌 

Left a few comments of things I noticed. Nothing major though 🤔

assets/global.js
LucasLacerdaUX: I added the `Element` to the variable to make it clear that it's a NodeElement. Isn't that always the case? Would you rather use `Node`?

`this.pageTotal` sounded like an integer variable to me :P

assets/global.js
LucasLacerdaUX: Good catch. Forgot to update after rebase.

sections/main-product.liquid
LucasLacerdaUX: > the only times I land on the previous or next arrow it's dimmed. 

Not necessarily true. Here's an example where I got to the `Next` arrow by using the Skip link and then tabbing back: https://user-images.githubusercontent.com/37168033/146236510-8076e15a-bbeb-48f7-a170-6aaf0bc4d930.mp4

Tabs will also cycle back, so if you tab through every element on the page you'll get back to the top, allowing you to focus on a `Prev` arrow:

https://user-images.githubusercontent.com/37168033/146236996-18a67cfd-a98d-478f-90e8-646f81c3bda4.mp4

In most cases, yes, you'll be able to navigate the thumbnails without using the arrows, but it's important that they are visible whenever you reach them via this non-linear paths.

assets/global.js
LucasLacerdaUX: Fixed! (hopefully)

sections/main-product.liquid
ludoboludo: Nice. Very good 👍

snippets/product-thumbnail.liquid
```diff
@@ -22,17 +22,23 @@
     <span class="product__media-icon motion-reduce">{% render 'icon-play' %}</span>
```
ludoboludo, line 22: In the comments at the start of this file, you can add the new `media_width` "property" that it can accept :)

LucasLacerdaUX: 👍

sections/main-product.liquid
```diff
@@ -444,6 +546,107 @@
...
+      });
+      if (this.dataset.desktopLayout !== 'stacked' && this.mql.matches) this.removeListSemantic();
+    }
+
+    onSlideChanged(event) {
+      const thumbnail = this.elements.thumbnails.querySelector(`[data-target="${ event.detail.currentElement.dataset.mediaId }"]`);
+      this.setActiveThumbnail(thumbnail);
+    }
+
+    setActiveMedia(mediaId, prepend) {
```
ludoboludo, line 573: when I have a feat. product section and change the variant, it should change the image and show the associated image I believe. Atm it's throwing an error about: `Cannot read properties of null (reading 'setActiveMedia')`

I think it's because this code was moved from global.js and the ` updateMedia()` function to this file.

assets/section-main-product.css
```diff
@@ -554,6 +590,28 @@ a.product__text {
...
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: 0.3rem;
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
+
+.product__media-toggle.focused::after {
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: 0.3rem;
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
```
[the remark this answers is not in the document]
LucasLacerdaUX: We can't really combine these. The `.focused` class is only available when `:focus-visible` is not supported by the browser. And if the browser does not recognize part of a selector in CSS, it will ignore the entire group of selectors.

I'll update the spacing after commas and background color, though.

assets/global.js
```diff
@@ -529,32 +530,51 @@ class SliderComponent extends HTMLElement {
...
+    if (this.isSlideVisible(this.sliderItemsToShow[0])) {
       this.prevButton.setAttribute('disabled', 'disabled');
     } else {
       this.prevButton.removeAttribute('disabled');
     }
 
-    if (this.currentPage === this.totalPages) {
+    if (this.isSlideVisible(this.sliderLastItem)) {
       this.nextButton.setAttribute('disabled', 'disabled');
     } else {
```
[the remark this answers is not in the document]
LucasLacerdaUX: I think I've fixed it now. Can you test it again? I accidentally used `<`  and `>` instead of `<=` and `>=`

sections/main-product.liquid
LucasLacerdaUX: Oh no 🤦‍♂️ That's not good news. I'll need to do some refactor and add the `media-gallery` component to the product page tomorrow morning. 

Thanks for flagging this - it's a really important bug that I completely missed. Glad that you found this before we merged ❤️

assets/section-main-product.css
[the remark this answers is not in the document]
ludoboludo: ah yeah right! Forgot about this as a reason 👍

sections/main-product.liquid
LucasLacerdaUX: Just fixed. `media-gallery` is now a shared script between `main-product` and `featured-product`. I've added a `<media-gallery>` component to `featured-product` and made its media list a `GalleryViewer-` element, so the function will now work as expected.

I've also fixed a bug with low resolution images being served on featured product due to missing `media_width` variable.

assets/section-main-product.css
melissaperreault: Hum, this is interesting. A few things comes to mind:
- This is not only impacting videos but any media in a landscape ratio
- This is not something that was concerning with the stacked layout because we had more vertical visuals to view
- Since merchants can deactivate the `Enable sticky header` setting and the other one, I am not too concerned. That behaviour is not meant for all instances and this is the reason why the settings exist.
- Our defaults set merchants for success if they remain with the `Stacked` layout.

assets/media-gallery.js
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      this.preventStickyHeader();
+      window.setTimeout(() => {
+        if (this.elements.thumbnails) {
+          activeMedia.parentElement.scrollTo({ left: activeMedia.offsetLeft });
+        }
+        if (!this.elements.thumbnails || this.dataset.desktopLayout === 'stacked') {
+          activeMedia.scrollIntoView({behavior: 'smooth'});
+        }
```
ludoboludo, line 48: I kinda feel like it's nice to have it scroll back to the top of the media being opened. But I guess it's weird to keep focus on the thumbnail button and potentially have it outside of the viewport 🤔 

This can be a follow up conversation anyway. Not a blocker and quick fix if needed.
[video](https://screenshot.click/21-12-eg9gh-7dj93.mp4)

ccing @melissaperreault for future conversation on this.

assets/component-slider.css
```diff
@@ -32,6 +32,24 @@ slider-component {
...
+@media screen and (min-width: 750px) {
+  .slider.slider--tablet-up {
+    position: relative;
+    flex-wrap: inherit;
+    overflow-x: auto;
+    scroll-snap-type: x mandatory;
+    scroll-behavior: smooth;
+    scroll-padding-left: 1rem;
+    -webkit-overflow-scrolling: touch;
+    margin-bottom: 1rem;
```
ludoboludo, line 44: Due to this margin-bottom, it looks like the slider buttons aren't quite vertically aligned: [video](https://screenshot.click/21-12-wnay2-9vv9z.mp4)

ludoboludo (approved):
Noted maybe one thing that could be changed in the PR. The vertical alignment of the thumbnail slider buttons. Otherwise didn't notice any other issues.

sections/main-product.liquid
```diff
@@ -546,109 +546,12 @@
...
-    removeListSemantic() {
-      this.elements.viewer.slider.setAttribute('role', 'presentation');
-      this.elements.viewer.sliderItems.forEach(slide => slide.setAttribute('role', 'presentation'));
-    }
-  }
-  customElements.define('media-gallery', MediaGallery);
 {% endjavascript %}
 
+{% if product.media.size > 0 %}
+  <script src="{{ 'media-gallery.js' | asset_url }}" defer="defer"></script>
```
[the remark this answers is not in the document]
LucasLacerdaUX: I'll create an issue to revisit this. This is a good point, but like I shared in our DMs, I don't have a better alternative right now 😸

ludoboludo (approved):

assets/media-gallery.js
```diff
@@ -0,0 +1,101 @@
+if (!customElements.get('media-gallery')) {
+  customElements.define('media-gallery', class MediaGallery extends HTMLElement {
```
tyleralsbury, line 2: This is clever.  Using this approach we might even be able to inline some of our smaller custom elements.  Keeping this one in the back of my mind for asset management ideas.

tyleralsbury (approved):
Looks good - I gave it some tests and looked again at the code.  🚢

PR #952, opened 2021-11-26
Fix section vertical spacing

### LINKED ISSUE #925: [Global settings] Layout
### Settings reference
- [Internal Figma link](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=141%3A10944)

### Some considerations

**Vertical spacing between sections**
   - [x] Increase range max - Meli to test
   - Minimum value 0
      - [x] Rich text [Background 1 color scheme should behave](https://screenshot.click/21-11-xg6ow-omnf4.png) like the other scheme and have some top and bottom padding
      - [x] Featured collection View all button to have a bottom spacing to [avoid content hug like this](https://screenshot.click/21-11-23jj5-74pba.png)
      - [x] To consider testing sections when they don't have a heading displayed

**Switching to from PX to REM**
- ~We should make is to have a true representation of what's being changed for anything related to spacing, margins, padding.~ To be discussed on this issue https://github.com/Shopify/dawn/issues/984

**Scaling values across devices**
-  ~Adjust the values applied to be reduced on smaller screens since they won't need the same value to accomplish the same aesthetic.~ _No longer relevant if we have a device specific setting available. - Meli_

KaichenWang:

Fixes #925 

#### Remove:
- "one-off" top and bottom margins set at section level (use approach described in # 2 instead)
- negative section margins (use approach in # 3 instead)

### 2 - Create utility classes for section spacing
| HTML class            | Formerly known as             | Behaviour                                                                                                                                                | Notes                                                                                                                              |
|-----------------------|-------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
| `.section`            | `.spaced-section`             | Sets `margin-top` to `var(--sections-safe-spacing)`. Sets `margin-bottom` to `var(--sections-safe-spacing)` if the element is last `.section` on the page | Updated naming because all `.section` elements should inherently (minus exceptions) have consistent vertical spacing between them. |
| `.section-no-margin`  | `.spaced-section--full-width` | Sets `margin-top` to `var(--sections-safe-spacing)`  only if the element's previous sibling has class `.section`                                         | Updated naming to be less abstract. These sections are typically full-width and/or handle margins within the section itself.       |
| `.section-collapse-grid`     |                               | Described in # 3 |      
| `.section-heading`    |                               | Sets `margin-top` to zero                                                                                                                                | Prevent headings at the top of sections from creating additional vertical space.                                                   |
| `.element-margin-top` | `.element-margin`             | Sets `margin-top` to `var(--sections-safe-spacing)`                                                                                                      |                                                                                                                                    |
| `.element-margin`     |                               | Sets `margin-top` and `margin-bottom` to `var(--sections-safe-spacing)`                                                                                  |                                                                                                                                    |

#### "Special" sections
- Collage
- Collection list

#### Criteria - when a "special" section is:
- placed directly after another "special" section on the page
- has an empty heading

#### Behaviour
The vertical spacing between the sections is set to the value of `settings.grid_desktop_vertical_spacing` (and its mobile equivalent for mobile breakpoints)

## **Other considerations**

Currently, default vertical spacing between sections is set to the value of `settings.settings.sections_spacing`, but with a floor or 15px. I think it makes more sense to simply limit the range of the setting to start at a minimum of `15px`. Overall range limits are still TBD, so this can be done at a later time.

## **Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126954012694)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126954012694/editor)

## **Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -229,14 +229,8 @@ html.no-js .no-js-hidden {
   margin-top: var(--sections-safe-spacing);
 }
 
-.section-collapse-grid + .section-collapse-grid {
-  margin-top: var(--grid-mobile-vertical-spacing);
-}
-
-@media screen and (min-width: 750px) {
-  .section-collapse-grid + .section-collapse-grid {
-    margin-top: var(--grid-desktop-vertical-spacing);
-  }
+.section-collapse + .section-collapse {
+  margin-top: var(--sections-spacing);
 }
 
 .section-heading,
--- a/assets/base.css
+++ b/assets/base.css
@@ -231,36 +231,27 @@ html.no-js .no-js-hidden {
 
 /* Special section spacing rules */
 
-.section-collage + .section-collage {
-  margin-top: var(--grid-mobile-vertical-spacing);
-}
-
-.section-collage + .section-collection-list {
+.section-collage + .section-collage,
+.section-collage + .section-collection-list,
+.section-collection-list + .section-collection-list,
+.section-collection-list + .section-collage {
   margin-top: var(--grid-mobile-vertical-spacing);
 }
 
 @media screen and (min-width: 750px) {
-  .section-collage + .section-collage {
-    margin-top: var(--grid-desktop-vertical-spacing);
-  }
-  
-  .section-collage + .section-collection-list {
+  .section-collage + .section-collage,
+  .section-collage + .section-collection-list,
+  .section-collection-list + .section-collection-list,
+  .section-collection-list + .section-collage {
     margin-top: var(--grid-desktop-vertical-spacing);
   }
 }
 
-.section-heading {
+.section-heading,
+.section-heading.title,
+.section-heading.title-wrapper-with-link,
+.section-heading.title--primary {
   margin-top: 0;
-}
-
-.section {
-  outline: 4px solid red;
-  outline-offset: -4px;
-}
-
-.section-no-margin {
-  outline: 4px solid blue;
-  outline-offset: -4px;
 }
 
 .element-margin {
--- a/assets/base.css
+++ b/assets/base.css
@@ -217,42 +217,59 @@ html.no-js .no-js-hidden {
   }
 }
 
-.element-margin {
-  margin-top: var(--sections-spacing);
+.section {
+  margin-top: var(--sections-safe-spacing);
 }
 
-.spaced-section {
-  margin-top: var(--sections-spacing);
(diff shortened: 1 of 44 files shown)


tyleralsbury (approved):
Looks good and I love the systematic approach.  It really makes it easier to understand, so great job getting rid of all the special cases.

One copy question that I'm wondering about... We've got the following:

> Affects collection, search and blog pages. Also affects collage, collection, collection list, multicolumn and blog sections.

But I wonder if it's better to indicate something a little more systematic. Like, maybe...

> Does not affect sections with "make section full width" enabled

And then we can make that a system-wide rule, and all other sections are affected.  We should be avoiding exceptions, so does this make sense as a general rule? cc @melissaperreault

assets/base.css
```diff
@@ -267,13 +267,14 @@ body,
...
+  padding-top: 4rem;
+  padding-bottom: 5rem;
 }
 
 @media screen and (min-width: 750px) {
   .background-secondary {
-    padding: calc(6rem + var(--page-width-margin)) 0
-      calc(5rem + var(--page-width-margin));
+    padding-top: calc(6rem + var(--page-width-margin));
+    padding-bottom: calc(5rem + var(--page-width-margin));
```
kmeleta, line 277: I feel like the padding used for background-secondary sections should scale a bit along with section spacing. https://screenshot.click/29-58-ddtp3-ewv3w.png 6/5rem seems out of place to me with a small section spacing value set. cc: @melissaperreault

kmeleta (approved):
Nice, I think this is basically what I was hoping for. Much cleaner. Couple notes, but haven't uncovered any issues.

assets/base.css
melissaperreault: Good question! I'll think more about i!

melissaperreault:
> I've now removed this spacing, but I believe it was present on all product grids (including collections, search, product recos, etc.) and may have served a purpose that I'm missing?

I don't think so, it was just so tall! I think it was overlooked. 🤞

KaichenWang:
Updating PR with latest feedback

KaichenWang:
New PR with  updated approach https://github.com/Shopify/dawn/pull/976

PR #954, opened 2021-11-29
Update global setting range values

KaichenWang:

Update setting ranges for:
- Buttons
- Inputs
- Layout
- Media
- Content containers

Use updated values and naming

**Other considerations**

May change in the future

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126958305302)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126958305302/editor)

**Checklist**

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -566,7 +566,7 @@
         "max": 100,
         "step": 5,
         "unit": "px",
-        "label": "Blur radius",
+        "label": "Blur",
         "default": 0
       }
     ]
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -327,7 +327,7 @@
     "settings": [     
       {
         "type": "header",
-        "content": "Borders"
+        "content": "Border"
       },
       {
         "type": "range",
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -480,7 +480,7 @@
         "type": "range",
         "id": "inputs_shadow_blur_radius",
         "min": 0,
-        "max": 40,
+        "max": 20,
         "step": 5,
         "unit": "px",
         "label": "Blur radius",
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -483,7 +483,7 @@
         "max": 20,
         "step": 5,
         "unit": "px",
-        "label": "Blur radius",
+        "label": "Blur",
         "default": 0
       }
     ]
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -324,25 +324,7 @@
   },
   {
     "name": "Buttons",
-    "settings": [
-      {
-        "type": "header",
-        "content": "Typography"
-      },
-      {
-        "type": "header",
-        "content": "Spacing"
-      },
-      {
-        "type": "range",
-        "id": "buttons_padding",
-        "min": 0,
-        "max": 20,
-        "step": 1,
-        "unit": "px",
-        "label": "Padding",
-        "default": 10
-      },
+    "settings": [     
       {
         "type": "header",
         "content": "Borders"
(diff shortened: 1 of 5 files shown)


config/settings_schema.json
```diff
@@ -516,21 +480,17 @@
         "type": "range",
         "id": "inputs_shadow_blur_radius",
         "min": 0,
-        "max": 20,
-        "step": 2,
+        "max": 40,
+        "step": 5,
```
tyleralsbury, line 484: I'm wondering if the steps here are too large for blur radius.  I feel like each `px` you add to a blur radius adds a lot.  Not a strongly held opinion, so I defer to @melissaperreault and @KaichenWang on the decision.

config/settings_schema.json
```diff
@@ -854,11 +814,11 @@
         "type": "range",
         "id": "sections_spacing",
         "min": 0,
-        "max": 40,
+        "max": 120,
```
tyleralsbury, line 817: I like the larger max, I think it will lead to some nice open layouts.

config/settings_schema.json
```diff
@@ -989,8 +949,8 @@
         "type": "range",
         "id": "media_shadow_blur_radius",
         "min": 0,
-        "max": 20,
-        "step": 2,
+        "max": 100,
+        "step": 5,
```
tyleralsbury, line 953: I wonder if 100 is too large of a blur radius.  Maybe not, but it is massive.  I'm okay with it, but I am curious if folks will actually use it.

layout/password.liquid
```diff
@@ -74,7 +74,6 @@
         --text-boxes-shadow-vertical-offset: {{ settings.text_boxes_shadow_vertical_offset }}px;
         --text-boxes-shadow-blur-radius: {{ settings.text_boxes_shadow_blur_radius }}px;
 
-        --buttons-padding: {{ settings.buttons_padding }}px;
```
tyleralsbury, line 77: I like the idea of removing buttons / input padding - it will ensure accessible buttons and inputs.  I'm trying to think if there would be a valid art direction for very narrow buttons or inputs, but I think an opinion on function over form makes sense for controls like this.

tyleralsbury (approved):
A couple of comments.  Nothing blocking merge as far as I can see.

config/settings_schema.json
```diff
@@ -516,21 +480,17 @@
...
-    "settings": [
-      {
-        "type": "header",
-        "content": "Typography"
-      },
+    "name": "Content containers",
+    "settings": [    
       {
         "type": "header",
         "content": "Borders"
```
melissaperreault, line 496:
```suggestion
        "content": "Border"
```
We have the singular form, like for `Shadow`.

config/settings_schema.json
```diff
@@ -324,25 +324,7 @@
...
-        "max": 20,
-        "step": 1,
-        "unit": "px",
-        "label": "Padding",
-        "default": 10
-      },
+    "settings": [     
       {
         "type": "header",
         "content": "Borders"
```
melissaperreault, line 330:
```suggestion
        "content": "Border"
```

config/settings_schema.json
```diff
@@ -415,35 +397,17 @@
...
-        "step": 1,
-        "unit": "px",
-        "label": "Padding",
-        "default": 10
-      },
+    "name": "Inputs",
+    "settings": [     
       {
         "type": "header",
         "content": "Borders"
```
melissaperreault, line 413:
```suggestion
        "content": "Border"
```

melissaperreault (commented):
Thank you Kai! 🎉 

Looking good! I added a few notes feel free to disregard if you feel they should be tackled from another PR. 🙏 

**Buttons, Inputs, Content containers, and Media**
- [x] Have the `Thickness` setting appear before `Opacity`


**Layout**
- [ ] Have the `horizontal` setting display before `vertical` for both Desktop and Mobile grid
- [ ] Let's adjust the Mobile grid values for both axis to match the system grid 

> 

    max": 40,
    "step": 4,
    "default": 4,

config/settings_schema.json
```diff
@@ -415,35 +397,17 @@
         "type": "range",
         "id": "buttons_shadow_blur_radius",
         "min": 0,
-        "max": 20,
-        "step": 2,
+        "max": 40,
```
melissaperreault, line 400:
```suggestion
        "max": 20,
```
For the same reasons of Inputs

layout/password.liquid
melissaperreault: We are confident that the initial button spacing was scalable for most brands. We are insuring to let merchants have a well balanced label over the button background to grab buyers attention. We'll monitor though if that becomes a common request.

There are some tweaks to make to buttons though with the flexibility settings, coming in another issue!

KaichenWang:
Since this PR only changes setting range values and copy, I'm going to merge. I believe it's best to continue adjusting and discussing in a follow up.

Issue #906, opened 2021-11-17
[Content] Accordion versus tabs to align on


melissaperreault:
From @svinkle:

> Re: the current “Collapsible tab” block, or the upcoming “Tabs” component, let’s reconsider how these are named.
The name “Tabs” might be confusing to some folks (devs, designers, perhaps merchants, too.) Why? The **vertical grouping** of show/hide page elements we currently know as a “Collapsible tab” block are actually often referred to as an “[accordion](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion)” pattern ([accordion example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html).) On the other hand, a “[tab](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)” or “tab panel” pattern is usually made up of **horizontal controls** which also shows/hides page content ([tab example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html).) Both patterns are wildly different in terms of conveying semantic meaning to assistive technology and usability. Just something to consider moving forward.

To inform the following issue https://github.com/Shopify/dawn/issues/907

cc. @katycobb @Oliviammarcello @nicklepine

PR #956, opened 2021-12-01
New section - Collapsible content

### LINKED ISSUE #907: New Section - "Tabs"
Add an accordion section (example use case: FAQs). Each collapsible tab in the accordion is a block.

**Describe alternatives you've considered**
Could reuse implementation from Product page's collapsible tabs

n/a

sofiamatulis:

Fixes #907 .

Used the product page accordion as a starting point

**Other considerations**

Please note that the options are: `small`, `large` and `adapt` where the content adapts to the image size. We decided against adapting the image to the content as that would require JS and would cause layout shift. 

Hover state: https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=14437%3A84209
Designs: https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=2043%3A44305

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/126968430614/editor)

**Checklist**


assets/tabs.css
```diff
@@ -0,0 +1,64 @@
...
+    padding-top: 7rem;
+    padding-bottom: 9rem;
+  }
+}
+
+.caption-with-letter-spacing + h2 {
+  margin-top: 1rem;
+}
+
+.tabs .accordion:hover {
```
sofiamatulis, line 58: @Oliviammarcello Should the hover state be applied to the product accordion as well?

locales/en.default.schema.json
```diff
@@ -1983,6 +1983,182 @@
       "presets": {
         "name": "Video"
       }
+    },
+    "tabs": {
```
sofiamatulis, line 1987: @Oliviammarcello Do you mind taking a look at the content and seeing if it looks good? :) If so I believe I can already request translations

sofiamatulis:
> @sofiamatulis Where is this available to 🎩 in the demo store?

@svinkle You can use this link and add a new section called `Tabs` 

Quick video: https://screenshot.click/01-28-3sh8z-0pz7h.mp4

sections/tabs.liquid
```diff
@@ -0,0 +1,335 @@
...
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet tabs__grid{% if section.settings.desktop_layout == 'image_first' %} tabs__grid--reverse{% endif %}">
+        <div class="grid__item">
+          {%- for block in section.blocks -%}
+            <div class="accordion{% if section.settings.secondary_background == 'on_rows' %} color color-{{ section.settings.color_scheme }}{% endif %}" {{ block.shopify_attributes }}>
+              <details id="Details-{{ block.id }}-{{ section.id }}"{% if section.settings.open_first_collapsible_tab and forloop.first %} open{% endif %}>
+                <summary>
+                  <div class="summary__title">
+                    {% render 'icon-accordion', icon: block.settings.icon %}
+                    <h2 class="h4 accordion__title">
```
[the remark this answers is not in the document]
sofiamatulis: In this tabs section structure we have this new title (optional), but in the product page, these are individual collapsible blocks, would it make sense that both of them are `h3`s? Even when there is no title?

Quick video: https://screenshot.click/01-31-b8giv-5q7ox.mp4

sections/tabs.liquid
```diff
@@ -0,0 +1,335 @@
...
+        {% endif %}
+        {%- if section.settings.heading != blank -%}
+          <h2 class="h1">{{ section.settings.heading }}</h2>
+        {% endif %}
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet tabs__grid{% if section.settings.desktop_layout == 'image_first' %} tabs__grid--reverse{% endif %}">
+        <div class="grid__item">
+          {%- for block in section.blocks -%}
+            <div class="accordion{% if section.settings.secondary_background == 'on_rows' %} color color-{{ section.settings.color_scheme }}{% endif %}" {{ block.shopify_attributes }}>
+              <details id="Details-{{ block.id }}-{{ section.id }}"{% if section.settings.open_first_collapsible_tab and forloop.first %} open{% endif %}>
```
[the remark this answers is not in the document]
svinkle: Yes. This JS should execute for any elements matching the `[id^="Details-"] summary` selector.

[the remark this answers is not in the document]
sofiamatulis: I apologize, I shared the editor link but it works on the preview link:

https://screenshot.click/01-34-qoxhk-ykcsb.png

https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=126968430614

sections/tabs.liquid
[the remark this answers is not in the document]
svinkle: Dynamic heading structure can get complex pretty quickly. And we don't want to generate out of order headings.

Thoughts on this approach… for either product accordion or homepage section accordions:

- Always include an `h2` section heading
  - If the merchant supplies a heading title, apply the title within he `h2`
  - If the merchant **does not** supply a title, apply a generic title: `Accordion content` (not `Tab content` as [this will be confusing](https://github.com/Shopify/dawn/issues/906).)
- Accordion item headings will be set as `h3`.

This way the heading structure will be:

1. `h1` page or product title
2. `h2` accordion section title
3.  `h3` accordion item title

cc @katycobb

assets/tabs.css
Oliviammarcello: Yes please!

locales/en.default.schema.json
katycobb: @Oliviammarcello @sofiamatulis 
The main thing I'm not confident about is the name "Tabs" which Scott flagged (thanks!). I will get some feedback tomorrow re: some naming options (Accordion, Collapsible content etc.) and get back to you on it.

sections/tabs.liquid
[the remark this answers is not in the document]
svinkle: > In the case of the product page, we dont have a higher title and not even an option to add it. In that case would wed always add the title `Accordion content` .

I assumed the two templates shared the same code base. If they're separate components, leave the product section as-is.

And yes, I meant to mention the generic `h2` heading would be set as `visually-hidden` (updated my comment above.) Good catch. 👍

sections/tabs.liquid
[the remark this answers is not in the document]
svinkle: Ah, excellent. All good.

Curious why it doesn't work in the Editor? Merchants with disabilities would benefit from this same treatment.

sections/tabs.liquid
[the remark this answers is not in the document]
sofiamatulis: > I assumed the two templates shared the same code base. If they're separate components, leave the product section as-is.

Because some of the code is repetitive, I thought of moving it to a re-usable file but that hasnt been done yet 🤔 cc @tyleralsbury 

Here is my original comment: https://github.com/Shopify/dawn/pull/956#discussion_r760313831

sections/tabs.liquid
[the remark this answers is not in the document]
sofiamatulis: @svinkle I just re-tested it in the editor and it seems to work there too for me🤔 

https://screenshot.click/02-37-r0r5q-s7o3f.png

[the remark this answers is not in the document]
svinkle: Ok strange. I swear it wasn't there before. Cool, let's close this thread.

katycobb:
Hi @sofiamatulis  

Not a dealbreaker, but if possible please add a period at the end of the Desktop layout help text `Image is always first on mobile.` Otherwise everything looks good to me! 

@Oliviammarcello  and I jammed on updated labels for `Secondary background` + options. 
Field label: `Layout`
Dropdown options: `No container`, `Row container`, `Section container `

I've updated Figma to reflect the changes.

Oliviammarcello:
Looking great! Just a few things:
- [ ] The `Row content` isn't working for me for some reason. It should work both when it is the only content and when it is  [Screenshot](https://screenshot.click/21-04-s1hd5-nafj8.png)
- [ ] There is no padding between the section and the footer. [Screenshot](https://screenshot.click/21-07-r3z1v-sfwwx.png)
- [ ] When switching the image layout from text first to image first, the spacing and alignment break. [Video](https://screenshot.click/21-17-upgqm-27lnk.mp4)
- [ ] On tablet, the mobile layout of having the image on top is used instead of it being to the side of the tabs. [Figma](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=14160%3A99796)
- [ ] We now need to add a "right" option to the heading alignment dropdown since it was added to the other sections
- [ ] By default can "Open first collapsible row" be unchecked
- [ ] Can the spacing under the row content match the spacing around the title and icon. [Screenshot ](https://screenshot.click/21-48-a8dpb-hzvxz.png)
- [ ] The left and right spacing for the row content works in some places, but then is too tight in other areas. [Video.](https://screenshot.click/21-55-fth6w-vu2od.mp4) (Content at the bottom is too close to the edges, but content at the top is perfect.)

When global settings are merged is there a way to know that the corners of the tabs when layout is set to row will be rounded, have shadows, borders, etc.?

assets/collapsible-content.css
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+@media screen and (max-width: 749px) {
+  .collapsible-content__grid {
+    flex-direction: column-reverse;
+  }
+}
+
+@media screen and (min-width: 750px) {
```
tyleralsbury, line 25: I would combine this media query with the one below it to save a few lines.

config/settings_data.json
```diff
@@ -1,5 +1,81 @@
 {
-  "current": "Default",
```
tyleralsbury, line 2: Changes to this file will need a revert

templates/index.json
```diff
@@ -2,105 +2,337 @@
   "sections": {
```
tyleralsbury, line 2: Changes here will need a revert as well.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,349 @@
...
+                  <div class="summary__title">
+                    {% render 'icon-accordion', icon: block.settings.icon %}
+                    <h2 class="h4 accordion__title">
+                      {{ block.settings.heading | default: block.settings.page.title }}
+                    </h2>
+                  </div>
+                  {% render 'icon-caret' %}
+                </summary>
+                <div class="accordion__content rte" id="ProductAccordion-{{ block.id }}-{{ section.id }}">
+                  {{ block.settings.tab_content }}
```
tyleralsbury, line 30:
```suggestion
                  {{ block.settings.row_content }}
```

tyleralsbury (changes requested):
There is a bug when using rich text  - `tab_content` is still referenced in the Liquid but it should be `row_content`.

Oliviammarcello:
Update to color system:

- [ ] We will be adding a `Container color scheme` picker under the `Color scheme` picker. This will allow merchants to select the colour of their rows or their section container. [Figma settings](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=15607%3A92564) 

Let me know if you have any questions, thanks!!

sofiamatulis:
> * [x]  On tablet, the mobile layout of having the image on top is used instead of it being to the side of the tabs. [Figma](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=14160%3A99796)

I am seeing the correct layout on tablet 🤔 https://screenshot.click/08-37-cnj3t-1qloa.png What do you see?

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+                <div class="accordion__content rte" id="ProductAccordion-{{ block.id }}-{{ section.id }}">
+                  {{ block.settings.row_content }}
+                  {{ block.settings.page.content }}
+                </div>
+              </details>
+            </div>
+          {%- endfor -%}
+        </div>
+        {%- if section.settings.image != blank -%}
+          <div class="grid__item">
```
svinkle, line 38: Since the image is visually first on mobile, and `Image first` is an option, could we swap the order of the `.grid__item` elements? This will generate a more natural content flow and match the visual order when navigating by screen reader/braille reader.

sofiamatulis:
> I will wait for a cycle of UX review + removal of test store settings commits before finalizing my approval - overall the code looked great.

@tyleralsbury I have removed the store commits :) Do you mind also giving your 2 cents here https://github.com/Shopify/dawn/pull/956#discussion_r761230918?

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+
+<div class="collapsible-content{% if section.settings.layout == 'no_container' %} color color-{{ section.settings.color_scheme }}{% endif %}{% unless section.settings.layout == 'no_container' %} color-{{ section.settings.container_color_scheme }}{% endunless %}{% if section.settings.layout == 'section_container' %} page-width{% endif %}">
+  <div class="collapsible-content-wrapper{% if section.settings.layout == 'section_container' %} color color-{{ section.settings.color_scheme }}{% endif %}">
+    <div class="{% if section.settings.image == blank %} collapsible-content-wrapper--narrow{% else %} page-width{% endif%}">
+      <div style="text-align: {{ section.settings.heading_alignment }}">
+        {%- if section.settings.caption != blank -%}
+          <p class="caption-with-letter-spacing"> {{ section.settings.caption }}</p>
+        {% endif %}
+        {%- if section.settings.heading != blank -%}
+          <h2 class="h1">{{ section.settings.heading }}</h2>
```
svinkle, line 12: Are we able to [add a condition to implement a fallback heading `h2`](https://github.com/Shopify/dawn/pull/956#discussion_r760551926) for when title text is not provided?

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet collapsible-content__grid{% if section.settings.desktop_layout == 'image_first' %} collapsible-content__grid--reverse{% endif %}">
+        <div class="grid__item grid__item--content">
+          {%- for block in section.blocks -%}
+            <div class="accordion{% if section.settings.layout == 'row_container' %} color color-{{ section.settings.color_scheme }}{% endif %}" {{ block.shopify_attributes }}>
+              <details id="Details-{{ block.id }}-{{ section.id }}"{% if section.settings.open_first_collapsible_row and forloop.first %} open{% endif %}>
+                <summary>
+                  <div class="summary__title">
+                    {% render 'icon-accordion', icon: block.settings.icon %}
+                    <h2 class="h4 accordion__title">
```
svinkle, line 23: If we have the `h2` above, we'll need to swap these for `h3`.

Oliviammarcello:
> > * [x]   On tablet, the mobile layout of having the image on top is used instead of it being to the side of the tabs. [Figma](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=14160%3A99796)
> 
> I am seeing the correct layout on tablet 🤔 https://screenshot.click/08-37-cnj3t-1qloa.png What do you see?

I see this for some reason [https://screenshot.click/21-44-b25uq-w0le8.png](https://screenshot.click/21-44-b25uq-w0le8.png )

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+          <h2 class="h1">{{ section.settings.heading }}</h2>
+        {% endif %}
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet collapsible-content__grid{% if section.settings.desktop_layout == 'image_first' %} collapsible-content__grid--reverse{% endif %}">
+        <div class="grid__item grid__item--content">
+          {%- for block in section.blocks -%}
+            <div class="accordion{% if section.settings.layout == 'row_container' %} color color-{{ section.settings.color_scheme }}{% endif %}" {{ block.shopify_attributes }}>
+              <details id="Details-{{ block.id }}-{{ section.id }}"{% if section.settings.open_first_collapsible_row and forloop.first %} open{% endif %}>
+                <summary>
+                  <div class="summary__title">
```
svinkle, line 21: This generates invalid HTML. Now that I inspect and see what it does, I'm not sure if there's any way around this. Even changing to `span` this leads to another invalid issue of a heading not being a valid child of `span`. 🤔

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+                    {% render 'icon-accordion', icon: block.settings.icon %}
+                    <h2 class="h4 accordion__title">
+                      {{ block.settings.heading | default: block.settings.page.title }}
+                    </h2>
+                  </div>
+                  {% render 'icon-caret' %}
+                </summary>
+                <div class="accordion__content rte" id="ProductAccordion-{{ block.id }}-{{ section.id }}">
```
svinkle, line 29: Refer back to the [a11y spec in Figma on adding `aria-labelledby`](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=14722%3A86225) to the container here, referencing the `id` of the `summary` activator.

svinkle (changes requested):

Oliviammarcello:
Only a few issues left: 
- [ ] The spacing was fixed on the right, but not the left of the row container. [Screenshot](https://screenshot.click/21-47-g31zf-0gj4z.png)
- [ ] For the help text under the container color scheme can we capitalize Layout, Row and Container since they are referencing setting names. "Visible when Layout is set to Row or Section container."
- [ ] The spacing above the section when Row container is selected is too much. It should match the spacing used for No container. [Video](https://screenshot.click/21-56-n7qoq-dy7nw.mp4)
- [ ] The color scheme setting is currently changing the container color and the container color is changing the background. This should be vice versa.

sections/tabs.liquid
```diff
@@ -0,0 +1,335 @@
...
+                <summary>
+                  <div class="summary__title">
```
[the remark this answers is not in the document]
sofiamatulis: Removed it 👍

assets/tabs.css
sofiamatulis: Created an issue here to update the block on the product page: https://github.com/Shopify/dawn/issues/1001

sections/collapsible-content.liquid
sofiamatulis: I originally used the `id` in the `details`, but updated to use the `id` of the `summary`:

https://github.com/Shopify/dawn/pull/956/commits/54b2ee74a0e87a834755ea261e6560537817e4d9

Oliviammarcello (dismissed):
🚢 Looks great!!

tyleralsbury (dismissed):
Looks absolutely great.  Code is good, quick tests work well.  A small enhancement, not for now, but for later would be to add some admin only JS that opened a tab for which you just edited the content.  Ludo had something similar on the slideshow and it would be useful for the merchant to be able to see the content they added.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+
+<div class="collapsible-content layout-{{ section.settings.layout | split: '_' | first }}-container color-{{ section.settings.color_scheme }}{% if section.settings.layout == 'section_container' %} page-width{% endif %}">
+  <div class="collapsible-content-wrapper{% if section.settings.layout == 'section_container' %} color-{{ section.settings.container_color_scheme }}{% endif %}">
+    <div class="{% if section.settings.image == blank %} collapsible-content-wrapper--narrow{% else %} page-width{% endif%}">
+      <div style="text-align: {{ section.settings.heading_alignment }}">
+        {%- if section.settings.caption != blank -%}
+          <p class="caption-with-letter-spacing"> {{ section.settings.caption }}</p>
+        {% endif %}
+        {%- if section.settings.heading != blank -%}
+          <h2 class="h1">{{ section.settings.heading }}</h2>
```
ludoboludo, line 12: here we need to add some styling to break the word if it's too long:

ludoboludo, line 12: I would also like to get designer's feedback on the size of this heading. Imo it's too big compared to other sections' heading. 

I can see how it's as big as for example the image with text but I think the layout is different there and it makes sense.
In this section layout it looks like more like the `feat. collection` for example and should match the heading size 🤔

assets/collapsible-content.css
```diff
@@ -0,0 +1,104 @@
...
+  margin-top: 1rem;
+}
+
+@media screen and (min-width: 750px) {
+  .collapsible-content .accordion {
+    margin-top: 0;
+  }
+}
+
+.layout-row-container .accordion {
```
ludoboludo, line 72: I find the class naming a little too generic and not specific enough to the context. 

`layout-row-container` vs `collapsible-row-layout` for example.

assets/collapsible-content.css
```diff
@@ -0,0 +1,104 @@
...
+  border: 0.1rem solid rgba(var(--color-foreground), 0.1);
+  margin-bottom: 1.5rem;
+}
+
+.layout-row-container .accordion summary, .layout-row-container .accordion .accordion__content {
+  padding-left: 1.5rem;
+  padding-right: 1.5rem;
```
ludoboludo, line 79: I think we need to add a `overflow-x: auto;` here maybe. If I am to create a size table and it's wider than the width of this div, then it's hidden but creates a horizontal scroll bar:

assets/collapsible-content.css
```diff
@@ -0,0 +1,104 @@
...
+    padding-left: 5rem;
+    padding-right: 0;
+  }
+}
+
+@media screen and (min-width: 990px) {
+  .collapsible-content__grid:not(.collapsible-content__grid--reverse) .grid__item:last-child,
+  .collapsible-content__grid--reverse .grid__item--image {
+    padding-left: 7rem;
+    padding-right: 0;
```
ludoboludo, line 102: I don't think we need to apply the `padding-right` here as it's being applied in the media query above which will cover `990px` and over display.

assets/collapsible-content.css
```diff
@@ -0,0 +1,104 @@
...
+  padding-bottom: 5rem;
+  padding-top: 5rem;
+}
+
+@media screen and (min-width: 750px) {
+  .layout-section-container {
+    padding-bottom: 7rem;
+    padding-top: 7rem;
+  }
+}
```
tyleralsbury, line 11: I would probably combine some of these same media query blocks into the same block to save on some lines since they are pretty close to each other.  Can be done later.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
+{{ 'component-accordion.css' | asset_url | stylesheet_tag }}
+{{ 'collapsible-content.css' | asset_url | stylesheet_tag }}
+
+<div class="collapsible-content layout-{{ section.settings.layout | split: '_' | first }}-container color-{{ section.settings.color_scheme }}{% if section.settings.layout == 'section_container' %} page-width{% endif %}">
+  <div class="collapsible-content-wrapper{% if section.settings.layout == 'section_container' %} color-{{ section.settings.container_color_scheme }}{% endif %}">
```
tyleralsbury, line 5: The class here should probably be `collapsible-content__wrapper` to follow BEM.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+              {% if section.settings.image_ratio == 'adapt' and section.settings.image != blank %} style="padding-bottom: {{ 1 | divided_by: section.settings.image.aspect_ratio | times: 100 }}%;"{% endif %}
+            >
+              <img
+                srcset="{%- if section.settings.image.width >= 165 -%}{{ section.settings.image | img_url: '165x' }} 165w,{%- endif -%}
+                  {%- if section.settings.image.width >= 360 -%}{{ section.settings.image | img_url: '360x' }} 360w,{%- endif -%}
+                  {%- if section.settings.image.width >= 535 -%}{{ section.settings.image | img_url: '535x' }} 535w,{%- endif -%}
+                  {%- if section.settings.image.width >= 750 -%}{{ section.settings.image | img_url: '750x' }} 750w,{%- endif -%}
+                  {%- if section.settings.image.width >= 1070 -%}{{ section.settings.image | img_url: '1070x' }} 1070w,{%- endif -%}
+                  {%- if section.settings.image.width >= 1500 -%}{{ section.settings.image | img_url: '1500x' }} 1500w,{%- endif -%}
+                  {{ section.settings.image | img_url: 'master' }} {{ section.settings.image.width }}w"
```
tyleralsbury, line 30: I'd add a media size between `1070` and `1500` to close the gap between the two a bit and account for other screen sizes a bit better.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+                src="{{ section.settings.image | img_url: '1500x' }}"
+                sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: 2 }}px, (min-width: 750px) calc((100vw - 130px) / 2), calc((100vw - 50px) / 2)"
+                alt="{{ section.settings.image.alt | escape }}"
+                loading="lazy"
+                width="{{ section.settings.image.width }}"
+                height="{{ section.settings.image.height }}"
+            >
```
ludoboludo, line 37:
```suggestion
              >
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+            </div>
+          </div>
+        {% endif %}
+        <div class="grid__item grid__item--content">
```
ludoboludo, line 41: Not sure if this class is used anywhere: `grid__item--content`. I don't think so.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
...
+          {%- for block in section.blocks -%}
+            <div class="accordion{% if section.settings.layout == 'row_container' %} color-{{ section.settings.container_color_scheme }}{% endif %}" {{ block.shopify_attributes }}>
+              <details id="Details-{{ block.id }}-{{ section.id }}"{% if section.settings.open_first_collapsible_row and forloop.first %} open{% endif %}>
+                <summary id="Summary-{{ block.id }}-{{ section.id }}">
+                  {% render 'icon-accordion', icon: block.settings.icon %}
+                  <h3 class="accordion__title h4">
+                    {{ block.settings.heading | default: block.settings.page.title }}
```
ludoboludo, line 48: Might need to do an `| escape` here 🤔

sections/collapsible-content.liquid
ludoboludo: Might need the `| escape` filter here too

tyleralsbury (approved):
Approving - there are a few small things that I would change, but we can always tackle them as part of the follow-up, they're nitpicks.

I tested...

1. with and without an image
2. different image sizes
3. With content and content from page, and both combined
4. A few of the icons
5. Different colour schemes

The only thing I noticed that seems off is that when you pick the "Image first" layout setting but don't have an image, the content has a left padding applied when it should be full width.

It seems to be this selector:

``` css
.collapsible-content__grid:not(.collapsible-content__grid--reverse) .grid__item:last-child {
    padding-left: 5rem;
    padding-right: 0;
}
```

It probably needs to also check either whether it's the next sibling of the image, or if it's an `:only-child`, and not include that padding if it's the case.

ludoboludo:
@tyleralsbury I would like to tackle most of those things you and I brought up next week. And I can get another review maybe from Ken or Kai to confirm it's all good and shippable. 
There are a couple things on which I'd be curious to hear designer's feedback as well. 

Is that ok with you ?

assets/collapsible-content.css
```diff
@@ -0,0 +1,104 @@
...
+    padding-bottom: 4rem;
+  }  
+}
+
+.collapsible-content-wrapper--narrow {
+  margin: 0 auto;
+  padding-right: 1.5rem;
+  padding-left: 1.5rem;
+  max-width: 73.4rem;
+}
```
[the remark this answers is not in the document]
Oliviammarcello: Good point, I agree! We had a list of a few features that were nice to haves that I can add this to for us to explore in the future.

sections/collapsible-content.liquid
Oliviammarcello: I believe @wiktoriaswiecicka had been working on adding a `Heading size` dropdown to every section. I think it got descoped so we can reduce it to h2 for the time being. 👍

locales/en.default.json
```diff
@@ -73,7 +73,8 @@
     },
     "loading": "Loading...",
     "total_reviews": "total reviews",
-    "star_reviews_info": "{{ rating_value }} out of {{ rating_max }} stars"
+    "star_reviews_info": "{{ rating_value }} out of {{ rating_max }} stars",
+    "collapsible_content_title": "Collapsible content"
```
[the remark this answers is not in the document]
ludoboludo: I won't mess with this so that we don't need to re request translations.

assets/collapsible-content.css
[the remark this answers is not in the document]
ludoboludo: That would great! I won't tackle this potential layout now then. It's better that way anyway 👍

sections/collapsible-content.liquid
ludoboludo: Sweet, now using a h2 style to match other sections heading.

melissaperreault (approved):
Thanks Ludo! I reviewed a bunch of combination, looks good to me!

- [ ] One minor thing I found is the [focus being cut off](https://screenshot.click/21-12-scwhe-d4dx7.png) on the left

assets/component-accordion.css
```diff
@@ -49,6 +49,8 @@
 .accordion__content {
   margin-bottom: 1.5rem;
   word-break: break-word;
+  overflow-x: auto;
+  padding: 0 0.6rem;
```
ludoboludo, line 53: Add to add the padding to account for the focus ring. As adding the `overflow` property was cutting it off as noticed by Meli.

assets/collapsible-content.css
```diff
@@ -0,0 +1,117 @@
...
+    height: 31.4rem;
+  }
+
+  .collapsible-content__media--large {
+    height: 69.5rem;
+  }
+}
+
+.collapsible-content__grid {
+  margin-left: 0;
```
KaichenWang, line 32: This is overriding the negative margin set by `.grid` and can cause a slight mis-alignment on the left. Is that desired?

Kapture 2021-12-23 at 10 06 26

assets/collapsible-content.css
```diff
@@ -0,0 +1,117 @@
...
+  padding-left: 1.5rem;
+  padding-right: 1.5rem;
+}
+
+.collapsible-content summary:hover {
+  background: rgba(var(--color-foreground), 0.04);
+}
+
+.collapsible-content summary:hover .accordion__title {
+  text-decoration: underline;
```
KaichenWang, line 101: We've typically paired this with `text-underline-offset: 0.3rem;`. Could add that here for consistency

assets/collapsible-content.css
```diff
@@ -0,0 +1,117 @@
+.collapsible-section-layout {
```
KaichenWang, line 1: We should review the file naming conventions in a separate PR. I would expect something like `section-collapsible-content.css` for consistency.

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
+{{ 'component-accordion.css' | asset_url | stylesheet_tag }}
+{{ 'collapsible-content.css' | asset_url | stylesheet_tag }}
+
+<div class="collapsible-content collapsible-{{ section.settings.layout | split: '_' | first }}-layout color-{{ section.settings.color_scheme }}{% if section.settings.layout == 'section_container' %} page-width{% endif %}">
```
KaichenWang, line 4: Instead of `image_first` and `text_first` as layout option values, we could just use `image` and `text` in the schema respectively? Then we could simplify to `collapsible-{{ section.settings.layout }}-layout`?

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+  <div class="collapsible-content__wrapper{% if section.settings.layout == 'section_container' %} color-{{ section.settings.container_color_scheme }}{% endif %}">
+    <div class="{% if section.settings.image == blank %} collapsible-content-wrapper--narrow{% else %} page-width{% endif%}">
```
KaichenWang, line 6: `collapsible-content-wrapper` doesn't seem to exist so it doesn't make sense to have `--narrow` modifier
```suggestion
    <div class="{% if section.settings.image == blank %} collapsible-content-wrapper-narrow{% else %} page-width{% endif%}">
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+      <div class="collapsible-content__header" style="text-align: {{ section.settings.heading_alignment }}">
```
KaichenWang, line 7:
```suggestion
      <div class="collapsible-content__header" style="text-align: {{ section.settings.heading_alignment }};">
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+          <h2 class="collapsible-content__heading">{{ section.settings.heading | escape }}</h2>
+        {%- else -%}
+          <h2 class="visually-hidden">{{ 'accessibility.collapsible_content_title' | t }}</h2>
+        {% endif %}
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet collapsible-content__grid{% if section.settings.desktop_layout == 'text_first' %} collapsible-content__grid--reverse{% endif %}">
+        {%- if section.settings.image != blank -%}
+          <div class="grid__item grid__item--image">
+            <div class="collapsible-content__media collapsible-content__media--{{ section.settings.image_ratio }} media"
+              {% if section.settings.image_ratio == 'adapt' and section.settings.image != blank %} style="padding-bottom: {{ 1 | divided_by: section.settings.image.aspect_ratio | times: 100 }}%;"{% endif %}
```
KaichenWang, line 21: This is inside a `{%- if section.settings.image != blank -%}`, so checking for a blank image again would be redundant.
```suggestion
              {% if section.settings.image_ratio == 'adapt' %} style="padding-bottom: {{ 1 | divided_by: section.settings.image.aspect_ratio | times: 100 }}%;"{% endif %}
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+                sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: 2 }}px, (min-width: 750px) calc((100vw - 130px) / 2), calc((100vw - 50px) / 2)"
+                alt="{{ section.settings.image.alt | escape }}"
+                loading="lazy"
+                width="{{ section.settings.image.width }}"
+                height="{{ section.settings.image.height }}"
+              >
+            </div>
+          </div>
+        {% endif %}
+        <div class="grid__item grid__item--content">
```
KaichenWang, line 42: I can't find any instances of `grid__item--content` being used to target this element
```suggestion
        <div class="grid__item">
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+        {% endif %}
+        {%- if section.settings.heading != blank -%}
+          <h2 class="collapsible-content__heading">{{ section.settings.heading | escape }}</h2>
+        {%- else -%}
+          <h2 class="visually-hidden">{{ 'accessibility.collapsible_content_title' | t }}</h2>
+        {% endif %}
+      </div>
+      <div class="grid grid--1-col grid--2-col-tablet collapsible-content__grid{% if section.settings.desktop_layout == 'text_first' %} collapsible-content__grid--reverse{% endif %}">
+        {%- if section.settings.image != blank -%}
+          <div class="grid__item grid__item--image">
```
KaichenWang, line 19: Since the styles applied to `.grid__item--image` are really only valid in the context of the collapsible-content section, we shouldn't use a modifier of `.grid__item`, but rather a selector that's associated with collapsible-content. Something like `collapsible-content__grid-item--image`

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+                </div>
+              </details>
+            </div>
+          {%- endfor -%}
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+
```
KaichenWang, line 65:
```suggestion
```

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,383 @@
...
+          "id": "row_content",
+          "label": "t:sections.collapsible_content.blocks.collapsible_row.settings.row_content.label"
+        },
+        {
+          "type": "page",
+          "id": "page",
+          "label": "t:sections.collapsible_content.blocks.collapsible_row.settings.page.label"
+        },
+        {
+          "type": "select",
```
KaichenWang, line 257: More of a design question, but why is the `icon` setting at the very end, when it's paired on the page with the `heading` setting that's at the top?

KaichenWang (commented):
Looks good overall! Left some comments.

locales/en.default.schema.json
```diff
@@ -2429,6 +2429,204 @@
...
+            "label": "Large"
+          }
+        },
+        "desktop_layout": {
+          "label": "Desktop layout",
+          "options__1": {
+            "label": "Text first"
+          },
+          "options__2": {
+            "label": "Image first"
```
kmeleta, line 2530: I believe we updated image with text to offer "image first" and "image second". We should probably match that schema here.

sections/collapsible-content.liquid
melissaperreault: Good observation, not aware of the discussion around the order but it is following the product `Collapsible row` [Block](https://screenshot.click/21-12-lj315-pjmaj.png) order as well. Something we could revisit with the team, not concerned to keep as is for now, but took a note! 🙏

assets/collapsible-content.css
ludoboludo: Hmm yeah not sure if this was needed for a specific case 🤔  I will remove it for now as I don't when it's needed.

assets/collapsible-content.css
ludoboludo: Nice, good spotting.

sections/collapsible-content.liquid
ludoboludo: Hmm this is using `section.settings.layout`. The values you're mentioning are for `settings.desktop_layout` so I think we can keep it as is.

sections/collapsible-content.liquid
ludoboludo: The eyes of an eagle.

sections/collapsible-content.liquid
ludoboludo: I did `collapsible-content__grid-item` so I don't use a modifier otherwise I'd also need `collapsible-content__grid-item`

sections/collapsible-content.liquid
ludoboludo: @melissaperreault what do you think of that ? I feel like Kai has a point, moving the icon drop down below the heading would make sense: 


I moved it for now. I chatted with Katy about it and also mentioned the option of icon first then heading to follow the visual order but decided to keep it like like suggested in the screenshot. 
>leaning towards heading first, since this would determine what icon you’d want to use

sections/collapsible-content.liquid
KaichenWang: Ah yeah true! I feel like the values can still be simplified to `none`, `row`, and `section` to avoid `split` filter

sections/collapsible-content.liquid
```diff
@@ -0,0 +1,382 @@
+{{ 'component-accordion.css' | asset_url | stylesheet_tag }}
+{{ 'collapsible-content.css' | asset_url | stylesheet_tag }}
+
+<div class="collapsible-content collapsible-{{ section.settings.layout | split: '_' | first }}-layout color-{{ section.settings.color_scheme }}{% if section.settings.layout == 'section_container' %} page-width{% endif %}">
+  <div class="collapsible-content__wrapper{% if section.settings.layout == 'section_container' %} color-{{ section.settings.container_color_scheme }}{% endif %}">
+    <div class="{% if section.settings.image == blank %} collapsible-content-wrapper-narrow{% else %} page-width{% endif %}">
```
KaichenWang, line 6:
```suggestion
    <div class="{% if section.settings.image == blank %}collapsible-content-wrapper-narrow{% else %}page-width{% endif %}">
```

KaichenWang (approved):
Looks great!

sections/collapsible-content.liquid
ludoboludo: Good point 👍  done 👌

PR #969, opened 2021-12-06
Use standar REM units

### LINKED ISSUE #883: Leave REM unit at default
Currently, the down theme [use][1] the known [62.5% trick][2] to have a REM unit of 10px, that allows easier calculations, as we are known to be more confortable with those numbers.

While that might be true for most people, the main problem I see with that on a baseline theme, is that it changes a **very well known and trusted default**, and is almost impossible to change once it is there.

Then, the problems are that, for example: copypasted css is broken when it use rem units, derived themes are forced to inherit the same default and so on.

Seeing the 62.5% trick through the lenses of the main guidelines on the README:

- Web-native in its purest form: It breaks a very well known standard.
- Lean, fast, and reliable: Add a trick with no functionality impact, just developer confort, that makes the theme less reliable.
- Functional, not pixel-perfect: Promotes usage of pixels to think design.

[1]: https://github.com/Shopify/dawn/blob/main/layout/theme.liquid#L86
[2]: https://css-tricks.com/accessible-font-sizing-explained/#method-1-the-62-5-rule

## Proposal

So I propose a big search and replace commit, that change all the REM units (I think there is a script doing that on a gist) and
change all the `divide: 10` to 16 on liquid statements.

We can add some rounding to make the source code look nice, but there will be no functionality impact on the theme itself.

The main issue, is that there are already other themes that are based on this, and they will be broken if they depend on this... so there should be a warning for those cases, but that will be get harder once more people depend on this.

eloyesp:
Remove REM = 10px alias.

Fixes #883 

Replaced all mentions of REM units that depends on the non-standard rem value with an almost equal value that do not depends on the non-standard rem value.

**Other considerations**

It use the following code:

``` ruby
require 'bigdecimal'

def custom_round value
  BigDecimal(value * 2, 2).to_f / 2
end

Dir.glob('**/*.{css,liquid}').each do |file|
  fixed_file_content = File.read(file).gsub(/(?<!\w)([\d\.]+)rem/) do
    value_in_pixels = $1.to_f * 10
    value_in_standard_rems = value_in_pixels.to_f / 16
    "#{ custom_round(value_in_standard_rems) }rem"
  end
  File.write file, fixed_file_content
end
```

**Demo links**

I would really appreciate if someone can add that as I have no idea about how to get that...

**Checklist**
- [X] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [X] Linted with [Theme Check](https://github.com/Shopify/theme-check)

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -174,7 +174,7 @@ html.no-js .no-js-hidden {
 .page-width {
   max-width: var(--page-width);
   margin: 0 auto;
-  padding: 0 1.5rem;
+  padding: 0 0.95rem;
 }
 
 .page-width-desktop {
@@ -184,11 +184,11 @@ html.no-js .no-js-hidden {
 
 @media screen and (min-width: 750px) {
   .page-width {
-    padding: 0 5rem;
+    padding: 0 3.15rem;
   }
 
   .page-width--narrow {
-    padding: 0 9rem;
+    padding: 0 5.5rem;
   }
 
   .page-width-desktop {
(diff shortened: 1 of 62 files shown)


eloyesp:
@blanklob I've pointed on the description that the answer for that is on #883.

Main advantage is that not breaking the default makes reusing code easier in both directions.

PR #976, opened 2021-12-07
Fix section vertical spacing

KaichenWang:

#925 

### 1 - Simplify and remove:

- "one-off" top and bottom margins set at section level
- negative section margins
- concept of "space-spacing" - margin between sections can be set to zero if desired
- margin between header + first rearrangeable section
- margin between last rearrangeable section + footer

### 2-  Set consistent margins between all rearrangeable sections

- Use `.section` class
- Add margin equal to ` {{ settings.spacing_sections_desktop }}` between every `.section` element

### 3 - Add separate setting for mobile

- ` {{ settings.spacing_sections_mobile }}` is the margin value for mobile breakpoints

### What needs to be tackled separately

1. Sections should ensure its inner elements do not contribute to the section's top and bottom margins. As an example, section headings currently contribute to the section's top margin and therefore there's perceived space above the section despite ` {{ settings.spacing_sections_desktop }}` being set to zero.

2. Since "space-spacing" is no longer guaranteed between sections, sections should handle its internal vertical top and bottom spacing with a section-level setting that sets top and bottom padding.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126954012694)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126954012694/editor)

**Checklist**

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -853,7 +853,7 @@
       {
         "type": "range",
         "id": "spacing_grid_vertical_desktop",
-        "min": 0,
+        "min": 4,
         "max": 72,
         "step": 4,
         "default": 8,


melissaperreault (commented):
Thank you Kai! Your PR descriptions are always super clear! 🎉 

Some observations, some might need to be tackled on another PR:

- [x] Will we introduce the concept of REM in another PR?
- [x] It is unclear to me why the `Featured collection` grid [receives this amount of padding](https://screenshot.click/21-12-5u4c8-7oi01.mp4) between the cards and the button, unsure if this is due to the vertical space but you can see where I spotted the observation on the video. IT seems to only happen on `Desktop`.
- [x] Email signup: When the `Make section full width` is false, it's unclear to me [if the margins should be adjusted](https://screenshot.click/21-12-3nw02-5hzva.png) at the section level or considered in this PR. 
   - [ ] To compare with `Rich text` section, similar layout but [the outcome is totally different](https://screenshot.click/21-12-pbzp0-62q6u.png) with the same set of settings.
- [ ] Collection list: [Something is off for the cards on mobile](https://screenshot.click/21-12-pien5-qwxss.png), not sure where this is introduced. They lost their padding and the [vertical grid spacing isn't working](https://screenshot.click/21-12-2zlah-nihxe.png).
- [ ] Any section relationship with the header or footer - `Image with text` is a good example to hope for a section level margins to offer the ability to i.e. [Space the last section from the footer](https://screenshot.click/21-12-gxrux-gukbv.png). Otherwise merchants has no way to make the choice themselves.
   - [ ] cc. @tyleralsbury To discuss! Other section like `Image banner`, `Slideshow` are good examples too to consider.

KaichenWang:
>  Will we introduce the concept of REM in another PR?

Changing the units to `rem` is pretty simple, but I haven't done it in this PR. I know we've had some team discussions on the topic already, but after some thought, I have some hesitation towards changing to `rem` because:

1. `rem` is a relative unit, so without knowing the root font size, the value has no real meaning
2. If we provided the root font size as a note, we still don't have empirical data to support the fact that `rem` units are widely understood by merchants

I personally would keep `px` or remove units.

> It is unclear to me why the Featured collection grid receives this amount of padding between the cards and the button, unsure if this is due to the vertical space but you can see where I spotted the observation on the video. IT seems to only happen on Desktop.

The vertical space is due to the `desktop grid vertical spacing` setting. Padding of `72px` is applied even though there is only 1 row of cards. This definitely needs to be fixed, but I would wait until https://github.com/Shopify/dawn/pull/946 is merged to avoid conflicts. I can add this as a to-do in https://github.com/Shopify/dawn/issues/925

> Email signup: When the Make section full width is false, it's unclear to me if the margins should be adjusted at the section level or considered in this PR.

Thanks for catching this. I've now removed hardcoded margins added within the Email signup section. The section behaves more consistently with Rich text section now.

> Collection list: Something is off for the cards on mobile, not sure where this is introduced. They lost their padding and the vertical grid spacing isn't working.

I accidentally removed a class from the section container. I've now restored it.

> Any section relationship with the header or footer - Image with text is a good example to hope for a section level margins to offer the ability to i.e. Space the last section from the footer. Otherwise merchants has no way to make the choice themselves.

Right now that spacing between `header + first rearrangeable section` and `last rearrangeable section + footer` is zero, but I'm open to letting the vertical spacing setting also control that spacing.

tyleralsbury:
> Any section relationship with the header or footer - Image with text is a good example to hope for a section level margins to offer the ability to i.e. Space the last section from the footer. Otherwise merchants has no way to make the choice themselves.

We could add a top margin setting on the footer itself.  Then they can adjust it regardless of which section is directly before.

tyleralsbury (approved):
Looks good code-wise.  Tested a bit and didn't find any noticeable issues.  Approved. 🚀

melissaperreault:
Thank you Kai!!
Answering some of the remaining points and will do a final review pass!

> I personally would keep px or remove units.

I understand, let's move this conversation over here https://github.com/Shopify/dawn/issues/984 where we can make our final decision. IF you could add you voice that would be best! 🙏 

> I can add this as a to-do in #925

Ok, lets! Thank you!


> Right now that spacing between header + first rearrangeable section and last rearrangeable section + footer is zero, but I'm open to letting the vertical spacing setting also control that spacing.

I think Tyler's suggestion is best this way both could happen and keep the ability to hug the last section with the footer. 👍 
- _We could add a top margin setting on the footer itself. Then they can adjust it regardless of which section is directly before._

@tyleralsbury Not sure where is best to document this need for footer and header. Let me know how I can help!

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
         "type": "range",
-        "id": "grid_mobile_vertical_spacing",
+        "id": "spacing_grid_vertical_mobile",
         "min": 0,
         "max": 40,
         "step": 4,
         "default": 4,
         "unit": "px",
-        "label": "Vertical spacing"
+        "label": "Grid vertical spacing"
```
melissaperreault, line 895:
```suggestion
        "label": "Multicolumn vertical space"
```

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
-        "default": 52
-      },
-      {
-        "type": "paragraph",
-        "content": "Affects collection, search and blog pages. Also affects collage, collection, collection list, multicolumn and blog sections."
-      },
-      {
-        "type": "header",
-        "content": "Desktop grid"
+        "default": 72
```
melissaperreault, line 841:
```suggestion
        "default": 0
```

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
       },
       {
         "type": "range",
-        "id": "grid_desktop_horizontal_spacing",
+        "id": "spacing_grid_horizontal_desktop",
         "min": 0,
```
melissaperreault, line 846:
```suggestion
        "min": 4,
```
We won't allow `0` until we plan more time to fix adjacent borders, etc.

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
         "default": 8,
         "unit": "px",
-        "label": "Horizontal spacing"
+        "label": "Grid horizontal spacing"
       },
       {
         "type": "range",
-        "id": "grid_desktop_vertical_spacing",
+        "id": "spacing_grid_vertical_desktop",
         "min": 0,
```
melissaperreault, line 856:
```suggestion
        "min": 4,
```
We won't allow 0 until we plan more time to fix adjacent borders, etc.

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
+        "step": 4,
+        "unit": "px",
+        "label": "Vertical spacing between sections",
+        "default": 52
       },
       {
         "type": "range",
-        "id": "grid_mobile_horizontal_spacing",
+        "id": "spacing_grid_horizontal_mobile",
         "min": 0,
```
melissaperreault, line 880:
```suggestion
        "min": 4,
```

config/settings_schema.json
```diff
@@ -826,67 +826,73 @@
...
         "default": 4,
         "unit": "px",
-        "label": "Horizontal spacing"
+        "label": "Grid horizontal spacing"
       },
       {
         "type": "range",
-        "id": "grid_mobile_vertical_spacing",
+        "id": "spacing_grid_vertical_mobile",
         "min": 0,
```
melissaperreault, line 890:
```suggestion
        "min": 4,
```

melissaperreault (approved):
@KaichenWang I forgot to request these last change before sending my review! We can merge after this 🙏

Issue #979, opened 2021-12-08
[Global settings] Adjust shadow to display behind components


melissaperreault:
### Observation

Look for any instances, here are examples:

- [ ] Variant pills: Make the shadow [appear behind components](https://screenshot.click/21-12-4sygy-h8xhu.png)
- [ ] Cards: Have the [shadow](https://screenshot.click/21-12-dyw1d-do7u1.png) appear below all elements of a section

kmeleta:
Also an issue with product media

kmeleta:
Image with text shadows will be addressed here https://github.com/Shopify/dawn/pull/1003

kmeleta:
Also need multicolumn text boxes..

kmeleta:
FYI, I'm currently looking at pill shadows along with some possibly related issues in https://github.com/Shopify/dawn/issues/1021. I don't think it's worth pushing a quick fix for just the shadow overlap issue without considering the other related changes or we could end up reworking it twice.

kmeleta:
These areas have been addressed. I'm not currently aware of any others, besides buttons, which are being handled separately.

- [x] Product cards
- [x] Blog cards
- [x] Variants pills
- [x] Product media
- [x] Multicolumn cards
- [x] Image with text

PR #1013, opened 2021-12-12
Media shadow updates

kmeleta:

Refs #979 
Fixes potential shadow overlap in product media and collage media (image/video).

Applied the shadow on a pseudo element within the media list items so their z-index can be controlled. The approach used on product media is exactly the same as on collage but applies to different elements.

If there were more instances of media type items needing these styles I would consider creating some new global classes to abstract away some of the redundancies.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -474,8 +474,22 @@
 .collage-card:not(.collage-product, .collage-collection) {
   border: var(--media-border-width) solid rgba(var(--color-foreground), var(--media-border-opacity));
   border-radius: var(--media-radius);
-  box-shadow: var(--media-shadow-horizontal-offset)
-    var(--media-shadow-vertical-offset)
-    var(--media-shadow-blur-radius)
-    rgba(var(--color-foreground), var(--media-shadow-opacity));
+  background: rgb(var(--color-background));
+}
+
+.collage-card:not(.collage-product, .collage-collection):after {
+  border-radius: var(--media-radius);
+  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-foreground), var(--media-shadow-opacity));
+  content: '';
+  position: absolute;
+  width: calc(var(--media-border-width) * 2 + 100%);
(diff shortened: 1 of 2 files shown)


melissaperreault (approved):
Thank you Ken! 🎉 

Every thing looks good from my first pass!
Adding a note that [the shadow](https://screenshot.click/21-12-87r26-zmc4s.png) overlap when `Section padding` + `Vertical space between sections` is set to 0 will be fixed by https://github.com/Shopify/dawn/pull/1003

assets/collage.css
```diff
@@ -474,8 +474,22 @@
...
+  border-radius: var(--media-radius);
+  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-foreground), var(--media-shadow-opacity));
+  content: '';
+  position: absolute;
+  width: calc(var(--media-border-width) * 2 + 100%);
+  height: calc(var(--media-border-width) * 2 + 100%);
+  top: calc(var(--media-border-width) * -1);
+  left: calc(var(--media-border-width) * -1);
+  z-index: -1;
+}
```
ludoboludo, line 490: I'm wondering if we should treat this as a card instead 🤔  So it matches the styles from the product or collection block.
Otherwise to tweak this section, you would have to touch the `cards` settings as well as the `media` ones. Curious to hear your thoughts on this.

assets/section-main-product.css
```diff
@@ -828,11 +823,26 @@ a.product__text {
   width: 2.2rem;
 }
 
-.product__media {
```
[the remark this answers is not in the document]
kmeleta: Lucas has a big product media PR going that will modifying or adding to the dom structure so I don't want to make expand the scope of my changes here. Will keep it in mind though we we need to merge into those updates though.

assets/collage.css
melissaperreault: We will once we tackle this issue https://github.com/Shopify/dawn/issues/1015

PR #1024, opened 2021-12-15
add border radius to dynamic checkout buttons

### LINKED ISSUE #987: [Global settings] Corner radius bugs
### Observations

- [ ] Apply corner radius on Dynamic checkout buttons
- [ ] Media double border and ...

ludoboludo:

Fixes #987 

There are two options to go about it, I thought: 
- add the styling only in the `section-main-product.css` file. Which implies repeating some of the code declaration done in `base.css`. 
- separate some of the styling applied in `base.css` so that shared styling can be applied once, there. But then we have product/feat. product specific styling in the `base.css` file 🤷 

Curious to hear thoughts about this or a different approach/take on it ^ 

**Other considerations**

There are some little things happening with atc buttons depending on the border thickness and blur values that I don't think should be tackled in this PR:
- when the blur value is 0 and thickness greater than 0, because the button background is black and the border too, you can really tell it's there. [screenshot](https://screenshot.click/21-12-yq52d-vthge.png)
- when the border thickness is at its max, ATC and dynamic checkout kinda overlap each other. [screenshot](https://screenshot.click/21-12-h0092-p7lxo.png)
- and there is some interesting behaviour on hover as well. When the border thickness is set to be `2px` it's the same as the original box shadow so there is not changes on hover. On the other hand, if the border thickness is greater then on hover the border kinda disappears: [video](https://screenshot.click/21-12-tygt3-nogrc.mp4)

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127018303510)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127018303510/editor)

You can test branded options by adding this at the end of the product page url: 
- Amazon Pay - ?shopify-debug=true&show=Amazon
- Apple Pay (Safari only) - ?shopify-debug=true&show=ApplePay
- Google Pay - ?shopify-debug=true&show=Google
- PayPal - ?shopify-debug=true&show=PayPal
- Venmo - ?shopify-debug=true&show=Venmo
- Unbranded - ?shopify-debug=true&show=checkout

**Checklist**


tyleralsbury:
This should probably be handling the ones in the cart, too:

Screen Shot 2021-12-15 at 5 06 21 PM

Aside from that, it looks great.  I'd love to see the shadows, too, but we could also consider doing them in a separate PR for review simplicity.

Were you unable to take an axiomatic approach to cut down on the selectors?  Something like `.product-form__buttons > * { ... }`?

ludoboludo:
I can get some of the buttons on the cart to have the `border-radius` but not the paypal one as it's using an iframe. So I would need to use JS in order to style it. 

As for the shadow on the buttons it seemed to make sense to me not to apply it to those buttons since their colors (purple for shop pay, yellow for paypal and black for google pay and apple pay) wouldn't match the shadow color and I wouldn't do the paypal one unless we decide to use JS 🤔 

Here is how it would look like with shadow: 


> Were you unable to take an axiomatic approach to cut down on the selectors?

If I do it then it overwrites some of the property, for example for `.button--secondary` which has then a lower specificity in its targeting and I'd need to fix that and its hover, focus and focus visible state 🤔

assets/base.css
```diff
@@ -1180,21 +1176,41 @@ deferred-media {
...
+  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button), var(--buttons-border-opacity));
+}
+
+.button,
+.shopify-challenge__button,
+.customer button,
+.shopify-payment-button__button--unbranded,
+.shopify-payment-button [role="button"],
+.cart__dynamic-checkout-buttons div[role='button'],
+.cart__dynamic-checkout-buttons iframe {
```
ludoboludo, line 1201: that worked, targeting the `iframe` @tyleralsbury 👍 

The focus-visible style isn't super great but still noticeable so I think that's ok.

tyleralsbury (approved):
Tested and approved.  It might be nice to see if we can figure out a more axiomatic way of doing this, but it's too much of an overhaul on specificity to do right now.  We should do a specificity audit in general of the theme when we've got time.

assets/base.css
```diff
@@ -1227,19 +1224,64 @@ deferred-media {
...
+
+.button--secondary {
+  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button-text), var(--buttons-border-opacity));
+  filter: drop-shadow(
+    var(--shadow-horizontal-offset)
+    var(--shadow-vertical-offset)
+    var(--shadow-blur-radius)
+    rgba(var(--color-button-text), var(--shadow-opacity))
+  );
+}
```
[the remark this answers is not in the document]
ludoboludo: Ah shoot must have messed it while I was doing the rebase thanks for spotting it 👍

assets/base.css
```diff
@@ -1200,21 +1196,41 @@ deferred-media {
...
+.shopify-payment-button__button--unbranded,
+.shopify-payment-button [role="button"] {
+  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button), var(--buttons-border-opacity));
+}
+
+.button,
+.shopify-challenge__button,
+.customer button,
+.shopify-payment-button__button--unbranded,
+.shopify-payment-button [role="button"],
```
KaichenWang, line 1219: What we're really targeting with this selector are the **branded** buttons? Though I don't think this makes a difference practically.

```suggestion
.shopify-payment-button__button--branded [role="button"],
```

assets/base.css
```diff
@@ -1200,21 +1196,41 @@ deferred-media {
...
+.cart__dynamic-checkout-buttons div[role='button'],
```
KaichenWang, line 1220: Do we need to specify `div` in the selector?
```suggestion
.cart__dynamic-checkout-buttons [role='button'],
```

KaichenWang (approved):
Looks good! Left 2 questions

PR #1034, opened 2021-12-16
Adding colour scheme settings to various sections

### LINKED ISSUE #966: Color scheme picker section update
### Sections that require updates:

- [ ] **Collection list:** In this section, the cards are changing color using the colour scheme picker instead of the section background. The cards will now change in the global settings and the background will change on the local level using the colour picker. Background 1 will no longer be the secondary background color, but now the actual background 1 value.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A6904)  |  [Screens](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=222%3A10322)
- [ ] **Featured product:** Add `Color scheme` picker + `Container color scheme` picker. [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9109)  |  [Screens](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=222%3A10702)
- [ ] **Product recommendations:** Add color scheme picker. The color of the cards can be changed in global settings. [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A6728)  
- [ ] **Featured collections:** Add `Color scheme` picker. The color of the cards can be changed in global settings.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A7085)
- [ ] **Contact us:** Add ` Color scheme` picker.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9109)  
- [ ] **Blog posts:** Add color scheme picker. The color of the cards can be changed in global settings. [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9110)  |  [Screens](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=231%3A13436)
- [ ] **Multicolumn:** Add `Color scheme` picker + `Container color scheme` picker. We will be removing secondary background and replacing it with a `Layout` setting. This will work the same way the `Collapsible rows` section works.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9114)` |  [Screens](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9182)
- [ ] **Collage:** Remove color picker from all blocks and secondary background from the product card. Allow the global setting to change the card color for the blocks (including image and video).  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A6650) |  [Screens ](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=294%3A9034)
- [ ] **Collection banner:** Add color scheme picker. Background 1 will no longer be the secondary background color, but now the actual background 1 value.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=161%3A5594)  |  [Screens](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=222%3A11165)
- [ ] **Header:** Add color scheme picker.  [Settings](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A9270)

On all sections that rely on the global settings to change card colors, add "To change the card color scheme, update your theme settings." as helper text. This should also be added to the product and collection blocks in the collage section.

tyleralsbury:

Fix #966

For the sake of simplicity, we are applying this to each section in mostly the same way - generally by adding a wrapping element that has `color-{{ section.settings.color_scheme }} gradient` and adding it to the schema.

**Other considerations**

There may be more nuanced approaches that could be taken for some sections / other sections that allow for more tweaking to happen on block levels, etc... but this initial implementation is meant to simply add flexibility to the ones that are the quickest and easiest to add the setting to.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=127021219862)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127021219862/editor)

**Checklist**


sections/collage.liquid
```diff
@@ -191,6 +193,34 @@
...
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
+    },
     {
```
ludoboludo, line 224: Here the image block still has a color scheme setting that needs to be removed. I'm doing it in one of my PR making sure the video and image block are using the card styling. So maybe no need to do it here 🤷

sections/featured-product.liquid
```diff
@@ -30,7 +30,7 @@
   <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
 {%- endif -%}
 
-<section{% if section.settings.secondary_background %} class="background-secondary"{% endif %}>
+<section class="{% if section.settings.secondary_background %} background-secondary{% endif %} color-{{ section.settings.color_scheme }} gradient">
```
ludoboludo, line 33: Should be an if/else here as `.gradient` adds a background color that overwrites the one set in `background-secondary`. But still keep `color-{{ section.settings.color_scheme }}` at all time 🤔

sections/featured-product.liquid
```diff
@@ -652,6 +652,34 @@
...
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
+    },
```
ludoboludo, line 682: Based on the issue it looks like we also need to add a container color scheme settings. The figma has this:

sections/main-cart-items.liquid
```diff
@@ -302,6 +304,34 @@
...
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
+    },
```
ludoboludo, line 334: Should there be an option for Subtotal as well ? Otherwise it looks a bit visually bad because subtotal looks disconnected from the main items:

sections/main-collection-product-grid.liquid
```diff
@@ -8,55 +8,57 @@
...
-        {%- endif -%}
-      </div>
-    {%- endif -%}
-  {%- endpaginate -%}
+          {%- if paginate.pages > 1 -%}
+            {% render 'pagination', paginate: paginate, anchor: '' %}
+          {%- endif -%}
+        </div>
+      {%- endif -%}
+    {%- endpaginate -%}
```
ludoboludo, line 60: I wonder if paginate should have a bit more spacing below it 🤔  [screenshot](https://screenshot.click/22-01-x7b5r-rm6gt.png). Not necessary in this PR but just noticed.

sections/multicolumn.liquid
```diff
@@ -219,6 +219,34 @@
...
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
+    },
```
ludoboludo, line 249: Based on the issue I think it needs a container color scheme picker as well that would apply the color to the block 🤔

ludoboludo: [screenshot](https://screenshot.click/22-01-b9fcw-ozuhs.png)

sections/product-recommendations.liquid
```diff
@@ -131,6 +133,34 @@
...
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
+    },
```
ludoboludo, line 163: I think it should be move above and not within product cards:

sections/featured-product.liquid
ludoboludo: Also think it should be moved outside of the media settings:

ludoboludo (commented):
Looking good 👌  Noticed a few small things.

I think it needs to be also [added to the `contact form`](https://screenshot.click/22-01-thr12-port4.png) section. And maybe they announcement bar as well ? [screenshot](https://screenshot.click/22-01-etglx-mz5y6.png)

Should the video section have it as well 🤔  It only would come in handy when using a vimeo video and if the image is a different aspect ratio than the video. [screenshot](https://screenshot.click/22-01-pl5dv-7rocf.png)

Also I feel like the position of the setting can vary a little from section to section. Like header and footer have it as the first option but the rest of the section have it somewhere else.

sections/collage.liquid
tyleralsbury: I can wait until your collage PR merges with that removal just to ensure this is all behaving as expected.

sections/featured-product.liquid
tyleralsbury: Gotcha - good point.

sections/product-recommendations.liquid
melissaperreault: Was there UX guidance for the color settings add-on? I suspect we'll need help from @katycobb to either change the setting label for `Section color scheme` or place it under a header called as `Section` and have `Top padding`, `Bottom padding` live under the same header. cc @Oliviammarcello

sections/main-collection-banner.liquid
```diff
@@ -1,6 +1,6 @@
 {{ 'component-collection-hero.css' | asset_url | stylesheet_tag }}
 
-<div class="collection-hero{% if section.settings.show_collection_image and collection.image %} collection-hero--with-image{% endif %}">
+<div class="collection-hero{% if section.settings.show_collection_image and collection.image %} collection-hero--with-image{% endif %} color-{{ section.settings.color_scheme }} gradient">
```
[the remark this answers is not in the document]
tyleralsbury: Good catch - had to remove the background colour from it.

sections/multicolumn.liquid
tyleralsbury: This also goes beyond what I'd consider to be a quick win.  @Oliviammarcello Can you see if we're okay with the current implementation?

sections/product-recommendations.liquid
tyleralsbury: Good point!

sections/main-collection-product-grid.liquid
```diff
@@ -138,6 +140,34 @@
...
+          "value": "background-2",
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "background-1",
+      "label": "t:sections.all.colors.label"
```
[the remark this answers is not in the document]
tyleralsbury: That is an objectively terrible place for them, good point! XD

tyleralsbury:
> I think it needs to be also added to the contact form section. And maybe they announcement bar as well ? screenshot

Each individual announcement bar block does have a colour scheme setting, so I think we're good for those.

sections/product-recommendations.liquid
Oliviammarcello: Yes @katycobb and I reviewed it together in figma. We opted to not include the word 'Section' in front of the colour scheme because it would get repetitive and we don't use 'section' for other settings that apply to the entire section in most areas. 

There are a few inconsistencies as Ludo pointed out (I will also review all tomorrow). All screens should match what is done on this [page.](https://www.figma.com/file/0bdfenAhtGXTiH93THYx0a/Colour?node-id=214%3A6728) Since the section padding was added before we updated Figma, to clarify, all colour scheme settings should be included above section padding.

sections/featured-product.liquid
```diff
@@ -30,7 +30,7 @@
   <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
 {%- endif -%}
 
-<section class="{% if section.settings.secondary_background %} background-secondary{% endif %} color-{{ section.settings.color_scheme }} gradient">
+<section class="{% if section.settings.secondary_background %}background-secondary{% else %}gradient{% endif %} color-{{ section.settings.color_scheme }}">
```
ludoboludo, line 33: There are still a few things to figure out I think with how `show seconday background` should work with the color scheme chosen 🤔 
[screenshot](https://screenshot.click/22-01-cphvv-ks1yh.png)

sections/main-cart-items.liquid
```diff
@@ -21,275 +21,277 @@
...
-                        height="{{ 150 | divided_by: item.image.aspect_ratio | ceil }}"
-                      >
-                    {% endif %}
-                  </td>
-
-                  <td class="cart-item__details">
-                    {%- if section.settings.show_vendor -%}
-                      <p class="caption-with-letter-spacing light">{{ item.product.vendor }}</p>
-                    {%- endif -%}
+<div class="color-{{ section.settings.color_scheme }} gradient">
```
ludoboludo, line 24: Noticed some issue with the color of the delete icon depending on the color scheme chosen:

ludoboludo (commented):
Looking good! Just noted a couple things

sections/featured-product.liquid
Oliviammarcello: Makes sense! 👍

sections/product-recommendations.liquid
tyleralsbury: One thing - I noticed Slideshow as one of the sections, but that one's colour schemes are set block by block so there is no section setting for it that makes sense.

sections/featured-product.liquid
tyleralsbury: Alright, stuck the class further down the chain - it was overriding the foreground colour with white which was causing the issue. 🤦

tyleralsbury: I see - okay, I think that's the behaviour we had before because it was giving it a white background when the text was white.  I'll do it that way and we can test to ensure that it's what we're looking for.

For now, we can keep the secondary background because adding a second picker adds complexity.

ludoboludo (commented):
Should we also add it to the Video section, Slideshow (for when it's using the grid layout) and maybe Custom liquid as well? 🤔 

Should collection list also get that setting (not the section) ? 🤔  Since on the collection page we can do it for the header and product grid, I'm wondering if that'd make sense to had it there too.
But then that'd mean adding it to the templates like page, product page, etc. So maybe I'm overthinking it and we can look into this later 😅 

Other than this ^ and what I mentioned in my couple comments, things look good 👌

sections/featured-product.liquid
tyleralsbury: Okay, I think I've got it working the way we expect it to.

1. If secondary background is not selected, it gets a full screen background with the same colour as the internal container
2. If it is selected, the text colour is taken (either the actual text colour for background, the background colour for inverse, or the button outline colour for accent)

Should be working like that now - I also fixed the gradient so that it works in both cases.

tyleralsbury:
> It shouldn't be on the product grid because then we would have to add it to all templates like you've specified.

I actually did add it to the product grid because otherwise you can't have a custom colour for the banner that matches the rest of the page.  I also did the same for the cart page so you can have a more branded cart.

We can definitely add it to other templates as well, but I would prefer for us to do a proper pass and handle it as a follow-up severity 2 issue.

sections/featured-product.liquid
```diff
@@ -30,9 +30,9 @@
   <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
 {%- endif -%}
 
-<section class="{% if section.settings.secondary_background %}background-secondary{% else %}gradient{% endif %} color-{{ section.settings.color_scheme }}">
+<section class="{% if section.settings.secondary_background %}background-secondary{% endif %}">
```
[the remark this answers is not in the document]
tyleralsbury: This should be working in a reasonable way right now when it's checked and unchecked based on Olivia's comment.

sections/featured-blog.liquid
```diff
@@ -120,6 +120,34 @@
       "default": 3,
       "label": "t:sections.featured-blog.settings.post_limit.label"
```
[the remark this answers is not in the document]
tyleralsbury: Looks like this is a bug with cards I see you created #1123 for this it, so I will consider this comment resolved.

tyleralsbury:
> There shouldn't be a colour picker on any of the template pages like cart or the product grid. The only exceptions are collection banner and product recommendations.

Should we remove them?  I can, but if we are going to eventually add them then should we keep them and just add to more template sections in the future as iterations?

tyleralsbury:
> Collage: The colour scheme picker should be removed on the image block. All blocks in collage should be affected by the card colour scheme in the global settings.

This is done in https://github.com/Shopify/dawn/pull/1090

sections/product-recommendations.liquid
```diff
@@ -22,24 +22,26 @@
...
-            show_secondary_image: section.settings.show_secondary_image,
-            show_vendor: section.settings.show_vendor,
-            show_rating: section.settings.show_rating
-          %}
-        </li>
-      {% endfor %}
-    </ul>
-  {% endif %}
-</product-recommendations>
+<div class="color-{{ section.settings.color_scheme }} gradient">
```
KaichenWang, line 25: This section may remain visible on the page even if there are no product recommendations returned by the server. This is actually due to the section padding settings, but is now more noticeable when a different color scheme is selected.

Screen Shot 2022-01-10 at 11 59 12 2 AM

sections/multicolumn.liquid
```diff
@@ -16,7 +16,7 @@
   }
 {%- endstyle -%}
 
-<div class="multicolumn{% unless section.settings.background_style == 'none' and settings.text_boxes_border_thickness > 0 or settings.text_boxes_shadow_opacity > 0 %} background-{{ section.settings.background_style }}{% endunless %}{% if section.settings.title == blank %} no-heading{% endif %}">
+<div class="multicolumn{% unless section.settings.background_style == 'none' and settings.text_boxes_border_thickness > 0 or settings.text_boxes_shadow_opacity > 0 %} background-{{ section.settings.background_style }}{% endunless %}{% if section.settings.title == blank %} no-heading{% endif %} color-{{ section.settings.color_scheme }} gradient">
```
KaichenWang, line 19: Switching between `background-none` and `background-secondary` appears to only adjust section padding/margin and has no effect on the section's background color. Would the quick solution to remove one of those setting options?

sections/product-recommendations.liquid
Oliviammarcello: Yup! Because the slideshow is full width we decided not to add it to the section level. Similar to Image banner.

ludoboludo:
> 



> > Collage: The colour scheme picker should be removed on the image block. All blocks in collage should be affected by the card colour scheme in the global settings.
> 
> This is done in #1090
> 
> > Not sure if this happened here, but on the collage product and collection block Add image padding and its helper text is missing.
> 
> I think this is also as expected in #1090 , is that right @ludoboludo

I believe `image padding` isn't part of the product and collection card anymore and was removed during Kai's revamp of it. Now both of them won't crop their image so the setting wouldn't be useful. The padding on those cards is dealt with through the global settings for cards. 
It's mentioned in the PR description here: https://github.com/Shopify/dawn/pull/1014

Oliviammarcello:
> Should we remove them? I can, but if we are going to eventually add them then should we keep them and just add to more template sections in the future as iterations?

I don't believe we will be adding them in the future. I think it needs to be explored once we look at the entire system as a whole, but I would be confident in removing them for consistency. That way only sections will have them.

KaichenWang (approved):
Code looks good. Left comments on some observed visual bugs

tyleralsbury:
Should be good for another round @KaichenWang @ludoboludo @Oliviammarcello

assets/section-product-recommendations.css
```diff
@@ -6,3 +6,14 @@
   margin: 0;
   margin-bottom: 3rem;
 }
+
+.product-recommendations:not(.product-recommendations--loaded) {
+  background: var(--color-base-background-1);
```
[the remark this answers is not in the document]
tyleralsbury: Good point - can be a bit clearer that way.

sections/product-recommendations.liquid
```diff
@@ -56,9 +58,14 @@
             const html = document.createElement('div');
             html.innerHTML = text;
             const recommendations = html.querySelector('product-recommendations');
+
             if (recommendations && recommendations.innerHTML.trim().length) {
               this.innerHTML = recommendations.innerHTML;
             }
+
+            if (html.querySelectorAll('.grid__item').length > 0) {
```
KaichenWang, line 66: I believe we only need to check if there is at least 1 grid item, so we don't need to query all instances

```suggestion
            if (html.querySelector('.grid__item')) {
```

sections/product-recommendations.liquid
```diff
@@ -56,9 +58,14 @@
...
+              document.querySelector('product-recommendations').classList.add('product-recommendations--loaded');
```
KaichenWang, line 67: In the context of this custom HTML element, `this` should be targeting the `product-recommendations` element

```suggestion
              this.classList.add('product-recommendations--loaded');
```

assets/section-product-recommendations.css
```diff
@@ -6,3 +6,14 @@
   margin: 0;
   margin-bottom: 3rem;
 }
+
+.product-recommendations:not(.product-recommendations--loaded) {
```
KaichenWang, line 10: I'm not sure this addresses the background color of the container? Could we simply set:
``` css
.product-recommendations:not(.product-recommendations--loaded) {
  display: none;
}
```

Screen Shot 2022-01-10 at 3 19 01 PM 2

KaichenWang (commented):
Looks good. Just a few comments on the fix for product recos

sections/product-recommendations.liquid
tyleralsbury: True - good point!

assets/section-product-recommendations.css
tyleralsbury: This is a good idea - I didn't think of just doing that 🤦 Much simpler!

Oliviammarcello:
- [ ] Not sure if this happened in this PR, but shouldn't the blog posts have cards? [Screenshot](https://screenshot.click/22-15-l29qm-wwq7b.png)
- [ ] On 'featured product' can the colour picker go above the `show secondary background` 

Otherwise this LGTM 👍

assets/section-product-recommendations.css
```diff
@@ -2,6 +2,10 @@
   display: block;
 }
 
+.product-recommendations:not(.product-recommendations--loaded) {
+  display: none;
```
KaichenWang, line 6: My previous suggestion actually has some unwanted side-effects. `IntersectionObserver` which is used to trigger the network request doesn't play well with `display: none` elements. 

```suggestion
  padding-bottom: 0;
  padding-top: 0;
```

KaichenWang (commented):
Left another comment for product recos. Otherwise looks good

assets/section-product-recommendations.css
tyleralsbury: Oh interesting - is that what was happening... I had some inconsistent behaviour but it looked like it was to do with files not updating right away (the `{% javascript %}` tag has some aggressive caching.

Good catch!

Oliviammarcello:
> > Not sure if this happened in this PR, but shouldn't the blog posts have cards? Screenshot
> 
> There are two styles of cards - standard and card. Looks like in that screenshot you're configured to Standard.

Ah, I see. Is it possible to fix the type alignment when set to standard like we do with product cards? [Screenshot](https://screenshot.click/22-39-l5403-xzwwk.png). I can create a new issue if needed.

tyleralsbury:
Rebase done - should be good to review @ludoboludo @KaichenWang

KaichenWang (dismissed):

KaichenWang (changes requested):
Noted some issues with the latest commit

assets/base.css
```diff
@@ -2453,8 +2453,8 @@ details-disclosure > details {
 }
 
 .gradient {
-  background: var(--color-background);
-  background: var(--gradient-background);
+  background-color: rgb(var(--color-background));
+  background-image: var(--gradient-background);
```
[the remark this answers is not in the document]
tyleralsbury: I can revert this one - it was part of a fix for shadows, but it turned out the `z-index` thing below was what actually fixed it.

assets/collage.css
```diff
@@ -5,6 +5,8 @@
 
 .collage {
   display: grid;
+  position: relative;
+  z-index: 0;
```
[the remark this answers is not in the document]
tyleralsbury: TL;DR - the shadows for cards have a `z-index` of `-1` and when a gradient is applied to the section, the shadows disappear.  They end up being put behind the background - not 100% sure why, but it might be that it's treated as an "image" technically.

I'll look into the modal issue.  Worst case, we might need to remove the setting from this section.

Oliviammarcello:
> > > > Not sure if this happened in this PR, but shouldn't the blog posts have cards? Screenshot
> > > 
> > > 
> > > There are two styles of cards - standard and card. Looks like in that screenshot you're configured to Standard.
> > 
> > 
> > Ah, I see. Is it possible to fix the type alignment when set to standard like we do with product cards? [Screenshot](https://screenshot.click/22-39-l5403-xzwwk.png). I can create a new issue if needed.
> 
> Yeah, let's create a new issue for that one just to keep a proper record of it. Does the behaviour change when there's a border? Because if there was a border there, then that'd put it right up against it, no?


For reference [here is the issue](https://github.com/Shopify/dawn/issues/1167) thanks!

KaichenWang:
The collage modal fix is working well. Will review again when conflicts are resolved.

tyleralsbury:
@KaichenWang @ludoboludo Rebase is complete.

assets/global.js
```diff
@@ -440,6 +440,8 @@ class ModalDialog extends HTMLElement {
         if (event.target.nodeName === 'MODAL-DIALOG') this.hide();
       });
     }
+
+    document.body.appendChild(this);
```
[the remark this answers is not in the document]
tyleralsbury: How about if we use `connectedCallback`?

That only runs when the element is added to the page.

https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks

ludoboludo:
Though maybe if we decide to add it to the video section we can do it as a follow up as we will need to request translation and that might delay when the PR will be merged.

tyleralsbury:
> Though maybe if we decide to add it to the video section we can do it as a follow up as we will need to request translation and that might delay when the PR will be merged.

Nah, I've added a new key for translations that are shared throughout the theme - I didn't adjust the keys for every instance of the colour schemes, but I've got stuff like `"label": "t:sections.all.colors.accent_2.label"`.

tyleralsbury:
@ludoboludo @Oliviammarcello I added it to video.  Was ezpz.

assets/global.js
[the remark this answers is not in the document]
KaichenWang: I think moving to `connectedCallback` is triggering an infinite loop, maybe best to leave in constructor at the moment

Screen Shot 2022-01-13 at 11 53 34 AM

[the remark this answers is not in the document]
tyleralsbury: Oof.  Odd that I didn't notice that.  I added an early return.  Should fix it.

ludoboludo (approved):
🎉

KaichenWang (dismissed):

Issue #1035, opened 2021-12-16
[Global setting] Adjust Cards Standard padding for various text alignment


melissaperreault:
### Expectation

- [ ] Product cards and Blog cards would need to handle different padding depending on the text alignment and Layout Grid Horizontal space to avoid text touching and improve legibility

Issue #929, opened 2021-11-19
[Global settings] Cards


melissaperreault:
### Settings reference
- [Internal Figma link](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=141%3A11236)

### Observations

**Featured collection**
- [x] When the style is set to `Card`, the height needs to be fixed and follow the tallest card to avoid various rectangle sizes. ~Same thing if you add a border or shadow to the `Standard` style.~
   - [ ] Have the quick buy button always snap at the bottom too (Only relevant once we ship this feature.)

**Blog posts section/template**
- [x] They are [not receiving any styles](https://screenshot.click/21-12-9klon-6s4jp.png) yet

**Collection list**
- [ ] Some images in the cards [don’t align with the border](https://screenshot.click/21-12-6wya5-gnefx.png)

**Collage section**
   - [x] Card style `Standard`: Product block and [is not styled like expected](https://screenshot.click/21-12-krpli-aj8va.png). We are left with a sharp edge border all around.
   - [x] Card style `Card`:  Product block has [a double border applied](https://screenshot.click/21-12-5k42g-15oi7.png), to be removed.

 **Very low priority**
- [ ] #1049

### For testing reference purposes

| Settings |  Current location | New location |
| ------------- | ------------- | ------------- |
| Image Padding on Product Cards | All sections that include product cards: | Card: Image Padding |
| | - Featured Collection | |
| | - Product Recommendation | |
| | - Collage | |
| | - Search Result | |
| Show Image Border | All sections that include product cards: | Card: Border settings |
| | - Featured Collection | |
| | - Product Recommendation | |
| | - Collage | |
| | - Search Result | |

Closing this issue to tackle the remaining items in separate issues instead.

nicklepine:
To look into: 
- [ ] If a merchant adds a gradient to one of their color, when picking a card color scheme that includes a gradient, should we add the gradient to the card? My assumption would be yes. [(recording) ](https://screenshot.click/21-16-rc9rw-x9cpq.mp4)

KaichenWang:
> Blog posts section/template
> - [ ] They are not receiving any styles yet

@kmeleta cards for articles should be compatible with the markup in `main-search.liquid` under `{%- when 'article' -%}`

kmeleta:
An update on some items in here..

> - [ ] Some images in the cards don’t align with the border

[...]

PR #1036, opened 2021-12-17
Blog cards update

kmeleta:

Refs #929
Updates blog cards to use the same structure and styles as product cards.

Changes affect..
- Featured blog section
- Search results (with articles)
- Main blog

Markup was updated to match product cards (and the existing blog search result card), but blog cards had a specific semantic structure that I'm not sure is worth replicating. For now I've decided it's best for them to match products exactly, but there may be a11y concerns involved.

There are two ways article card media can be sized via the rendering section, height (small, med, large, adapt) or aspect ratio. Height values are needed for the main blog image height settings (which correspond to specific css rules).

Replaced the inline article card markup in the search section with an article-card render.

**Other considerations**

An existing issue with standard style cards that will be more noticeable on blog cards now is the lack of padding between the card content and adjacent cards. This will be handled in a separate issue https://github.com/Shopify/dawn/issues/1035

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127022006294/editor)

**Checklist**

### DIFF
--- a/assets/component-article-card.css
+++ b/assets/component-article-card.css
@@ -27,32 +27,6 @@
   box-sizing: border-box;
 }
 
-.article-content {
-  width: 100%;
-  height: 100%;
-  display: flex;
-  flex-direction: column;
-  text-decoration: none;
-  color: inherit;
-}
-
-.article-content:hover .article-card__title {
-  text-decoration: underline;
-  text-underline-offset: 0.3rem;
-}
-
-.article-card__image {
-  overflow: hidden;
-}
-
-.article-content img {
-  transition: transform var(--duration-long) ease;
-}
-
-.article-content:hover img {
-  transform: scale(1.07);
-}
-
 .article-card__image-wrapper > a {
   display: block;
 }
@@ -80,10 +54,6 @@
 
 .article-card__link {
   text-underline-offset: 0.3rem;
-}
-
-.article-content:hover .article-card__link {
-  text-decoration-thickness: 0.2rem;
 }
 
 .article-card .card__heading {
--- a/assets/component-article-card.css
+++ b/assets/component-article-card.css
@@ -27,10 +27,6 @@
   box-sizing: border-box;
 }
 
-.article-card.card--card {
-  height: 100%;
-}
-
 .article-content {
   width: 100%;
   height: 100%;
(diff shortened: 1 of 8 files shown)


Oliviammarcello (approved):
Reviewing this for @melissaperreault and it looks great!

I have no notes besides the issue with the spacing between cards that you highlighted above, and the shadow not going past the slider UI or towards the heading which I see is tackled in issue 1037.

tyleralsbury (approved):
This looks great.  Tested on the homepage, in search results, and on the blog page itself.   Desktop + mobile.  Played around with various card settings as well and it's all looking great.

I am 100% aligned with matching the card markup with the other types of cards.  Less CSS is great for performance and maintainability.  I love consolidating them into a more straightforward set of components.

I wonder if we could take it even further and remove the need for an `article-card` component entirely at some point.  Not recommending that right now, but it would be great if we could get to the point where every single card used the same component - this would include those in collage as well.

We would need to ensure that it is actually simpler to do it that way, since it also needs to be flexible enough to fit into different contexts, which might be hard to do while keeping it simple.

kmeleta:
> I wonder if we could take it even further and remove the need for an `article-card` component entirely at some point. Not recommending that right now, but it would be great if we could get to the point where every single card used the same component - this would include those in collage as well.
> 
> We would need to ensure that it is actually simpler to do it that way, since it also needs to be flexible enough to fit into different contexts, which might be hard to do while keeping it simple.

This is definitely where I'd like cards to go when we have the chance to tinker with it. I have thoughts. Just seemed quicker/safer to go this route for now.

tyleralsbury:
> > I wonder if we could take it even further and remove the need for an `article-card` component entirely at some point. Not recommending that right now, but it would be great if we could get to the point where every single card used the same component - this would include those in collage as well.
> > We would need to ensure that it is actually simpler to do it that way, since it also needs to be flexible enough to fit into different contexts, which might be hard to do while keeping it simple.
> 
> This is definitely where I'd like cards to go when we have the chance to tinker with it. I have thoughts. Just seemed quicker/safer to go this route for now.

Let's create an issue to look into it - we will tackle it in the new year when we've got some time.  I'm itching to clean up `.title` as well.  I'm noticing some other opportunities to do this in other places, too, but I agree that for now it's safer to just make the changes we need and then look at this properly once we've got the first iteration of the enhanced global settings finalized.

Issue #986, opened 2021-12-08
[Global settings] Update settings values


melissaperreault:
### Expectations

- [ ] Update all range values based on the final intention (Internal Figma reference)
- [ ] Validate all setting options are available as expected

melissaperreault:
Remaining changes that I couldn't do until some PR gets merged:

- [ ] Popups is missing the intro helper text
- [ ] Drawers is missing the color scheme (To investigate when we'll add it)

PR #1042, opened 2021-12-17
Update global settings range values

melissaperreault:

This is fixing some part of https://github.com/Shopify/dawn/issues/986

I am referring to the final [Internal Figma link](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=364%3A12692) for the final range values. 

Most range values share the same min and max range even if on some components the max value may not be the best outcome. We wanted to offer the same flexibility across and let merchants make the best decision for their brand.

**Other considerations**

- I couldn't update at 100%, I will do another PR to solve the remaining missing pieces from the original issue.
- Layout will receive some changes later from https://github.com/Shopify/dawn/issues/925

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127024070678/editor)

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -232,7 +232,7 @@
       {
         "type": "range",
         "id": "buttons_border_thickness",
-        "min": 1,
+        "min": 0,
         "max": 24,
         "step": 1,
         "unit": "px",
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -233,7 +233,7 @@
         "type": "range",
         "id": "buttons_border_thickness",
         "min": 0,
-        "max": 8,
+        "max": 24,
         "step": 1,
         "unit": "px",
         "label": "t:settings_schema.global.settings.thickness.label",
@@ -253,7 +253,7 @@
         "type": "range",
         "id": "buttons_radius",
         "min": 0,
-        "max": 32,
+        "max": 72,
         "step": 2,
         "unit": "px",
         "label": "t:settings_schema.global.settings.corner_radius.label",
@@ -276,8 +276,8 @@
       {
         "type": "range",
         "id": "buttons_shadow_horizontal_offset",
-        "min": -12,
-        "max": 12,
+        "min": -40,
+        "max": 40,
         "step": 2,
         "unit": "px",
         "label": "t:settings_schema.global.settings.horizontal_offset.label",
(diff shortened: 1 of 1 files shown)


config/settings_schema.json
```diff
@@ -1032,10 +1028,6 @@
   {
     "name": "t:settings_schema.styles.name",
     "settings": [
-      {
```
melissaperreault, line 1035: Removing the header because we don't need it anymore.

config/settings_schema.json
```diff
@@ -474,10 +474,6 @@
   {
     "name": "t:settings_schema.cards.name",
     "settings": [
-      {
```
melissaperreault, line 477: Removing the header because we don't need it anymore.

config/settings_schema.json
martinamarien: Just remember to remove this translation when we've merged 👍🏻 You could probably do it now, but we can update that later too.

martinamarien (approved):
Just two comments. It would be great to update the Button default value, since Dawn, by default, has a border on its buttons

config/settings_schema.json
martinamarien: Same here.

KaichenWang (approved):
Looks good!

I've updated some of the values in the [settings doc](https://docs.google.com/document/d/15GZ4D6J2k2eZ53st6EQUbdi4qpX_RDiRasPVpbfXZ8k/edit) to match the changes here

config/settings_schema.json
melissaperreault: I was wondering what was the best way to do it since I am working on 2 separate branches. I am keeping a note to do another pass after on that on another PR next week once everything is merged, thank you!

config/settings_schema.json
melissaperreault: Taking track of this here 🙏 https://github.com/Shopify/dawn/issues/986#issuecomment-997008098

melissaperreault:
> I've updated some of the values in the settings doc to match the changes here

OHHH! Thank you Kai! This was my next step! 🙏 🙏 🙏

config/settings_schema.json
```diff
@@ -233,7 +233,7 @@
         "type": "range",
         "id": "buttons_border_thickness",
         "min": 0,
-        "max": 8,
+        "max": 24,
```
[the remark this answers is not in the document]
melissaperreault: Speaking with @kmeleta we are revisiting the default values to improve predictability. So thickness would be default at 1 and opacity at 100%. We are investigating tertiary button style impact still.

KaichenWang (approved):
Looks great!

Issue #985, opened 2021-12-08
[Global settings] Final content pass


melissaperreault:
### Expectations

- [x] Order of settings follow the intention when settings are not collapsed 
- [x] Order of settings inside each category are following the latest intention
- [x] All labels are following the final naming

Issue #961, opened 2021-12-03
[Global settings] Break out Overlays into their own category


melissaperreault:
### Expectation

- [ ] Internal [Figma link](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=377%3A13896)
- [ ] Offer an opacity range setting to adjust the overlay opacity
- [ ] Applies to: Navigation drawer, Pickup availability drawer, Filter drawer, Search dropdown

melissaperreault:
More context why this is mostly needed too:

https://github.com/Shopify/dawn/pull/1242#issuecomment-1018898724

> > I have a question about the behaviour when I tap over an area that there is a link under the navigation. ([Video reference](https://screenshot.click/22-01-hq8l9-sqnsb.mp4)) Can we prevent from happening?

> @melissaperreault  adding an overlay would be a good way to address the behaviour you mentioned. Without an overlay, it's difficult to prevent the user from clicking on elements such as links. Could we create a separate issue for adding an overlay and assess its priority? Do you think the current behaviour is an issue that needs to be addressed before merging?

PR #1044, opened 2021-12-17
Settings content naming update

melissaperreault:

Fixes https://github.com/Shopify/dawn/issues/985
Fixes https://github.com/Shopify/dawn/issues/986

- Update the final naming where possible
- Removed the usage of `Width`, it wasn't used anywhere
- Removed unused translations
- It is missing `Overlays`, this will be tackled with https://github.com/Shopify/dawn/issues/961
- It is missing `Interactive media controls`, it will be tackled once this is tackled. https://github.com/Shopify/dawn/issues/1051
- `Layout` will be updated from this comment https://github.com/Shopify/dawn/issues/925#issuecomment-995414537

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127024234518/editor)


katycobb (commented):
Hi Meli, couple of things:

- Under Cards, I don't think the "general" heading is needed. We don't use this elsewhere. 
- Dropdowns and pop-ups is missing its help text
- Under Icons, do we need the "decorative elements" header? I don't think it's necessary but you might have more context re: why it should be there! 


Otherwise the names look accurate.  If the other stuff isn't relevant to this ticket let me know and I can create a different one!

locales/en.default.schema.json
```diff
@@ -322,7 +319,6 @@
           }
         }
       }
-    }
```
martinamarien, line 325: This line needs to be re-added

locales/en.default.schema.json
```diff
@@ -2439,4 +2435,5 @@
       }
     }
   }
-}
+ }
```
martinamarien, line 2438: This line needs to be removed.

melissaperreault:
> Hi Meli, couple of things:

@katycobb Hey! Yes, this is not in this PR that I accounted for removing them because I was touching just the language file, we should see most of them revised under this PR https://github.com/Shopify/dawn/pull/1042 except for: "_Dropdowns and pop-ups is missing its help text_" --->>> This will be done in a follow up PR once one of them is merged. 🙏

ludoboludo (approved):
Looks good. Should we request translations ?

melissaperreault:
> Looks good. Should we request translations ?

@ludoboludo  Not yet, I have additional changes I would like to apply once I merge https://github.com/Shopify/dawn/pull/1042

katycobb (commented):
One nitpicky catch: Under `Media`, the heading says `Borders` but should be `Border` (no 's'). Otherwise all looks good!

melissaperreault:
> One nitpicky catch: Under Media, the heading says Borders but should be Border (no 's'). Otherwise all looks good!

@katycobb Good catch thanks! I updated it! You can re-approve once you feel all is good!

locales/en.default.schema.json
```diff
@@ -205,11 +200,8 @@
       "name": "Drawers"
     },
     "styles": {
-      "name": "Styles",
+      "name": "Icons",
```
ludoboludo, line 203: Almost feels likes if the only thing to edit about `Icons` is the color, it could potentially live in the `Colors` dropdown 🤔 
Just a thought.

ludoboludo (approved):

locales/en.default.schema.json
melissaperreault: Good observation, but there will be additional settings in the future under Icons such as the thickness of icons stroke  for instance.

PR #1045, opened 2021-12-17
Experimenting section padding math

tyleralsbury:
Playing with a smoother scale for mobile padding ratios.

Note: The Liquid here is very verbose for clarity, we would need to cut it down **or** turn it into a snippet cause there's a lot.


tyleralsbury:
Outdated explorations. Closing.

Issue #1039, opened 2021-12-17
Padding settings follow-up


tyleralsbury:
Looks great! 

#### Collection template - Product grid section

![Screen_Shot_2021-12-17_at_2_01_34_PM](https://user-images.githubusercontent.com/3710548/146597734-418c45aa-39e8-4070-b8b8-511bd17aacaf.jpg)

#### Blog template - Blog posts section
![Screen_Shot_2021-12-17_at_2_09_56_PM](https://user-images.githubusercontent.com/3710548/146597824-f9246cae-622e-4cee-a8cd-7abc0ed691d0.jpg)

#### Search template - Search results section

![Screen_Shot_2021-12-17_at_2_09_07_PM](https://user-images.githubusercontent.com/3710548/146597988-0a3a7c5a-dbc3-4a5b-8c37-919c67f9efb9.jpg)

#### Blog post template - Blog post section
![Screen_Shot_2021-12-17_at_2_04_40_PM](https://user-images.githubusercontent.com/3710548/146598112-13840b39-6738-411e-bf91-97537e02687e.jpg)

_Originally posted by @KaichenWang in https://github.com/Shopify/dawn/pull/1006#pullrequestreview-835562708_

PR #1055, opened 2021-12-21
Update layout settings 

### LINKED ISSUE #925: [Global settings] Layout

KaichenWang:

Fixes #925
More specifically https://github.com/Shopify/dawn/issues/925#issuecomment-995414537

## **What approach did you take?**

- Update settings and setting ranges
- Add logic for scaling spacing values on mobile

  - When set to under 24 it will be a 1:1 relationship
  - When set to 24 or over, mobile will be 0.7 times the desktop value

- Add `en` strings for settings

## **Other considerations**

n/a 

## **Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127031705622)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127031705622/editor)

## **Checklist**

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -151,74 +151,44 @@
         ],
         "default": "1600",
         "label": "t:settings_schema.layout.settings.page_width.label"
-      },
-      {
-        "type": "header",
-        "content": "Desktop"
-      },
+        },
       {
         "type": "range",
-        "id": "spacing_sections_desktop",
+        "id": "spacing_sections",
         "min": 0,
         "max": 120,
         "step": 4,
         "unit": "px",
-        "label": "Vertical space between sections",
+        "label": "t:settings_schema.layout.settings.spacing_sections.label",
         "default": 0
       },
       {
-        "type": "range",
-        "id": "spacing_grid_horizontal_desktop",
-        "min": 4,
-        "max": 72,
-        "step": 4,
-        "default": 8,
-        "unit": "px",
-        "label": "Multicolumn horizontal space"
-      },
-      {
-        "type": "range",
-        "id": "spacing_grid_vertical_desktop",
-        "min": 4,
-        "max": 72,
-        "step": 4,
-        "default": 8,
-        "unit": "px",
-        "label": "Multicolumn vertical space"
-      },    
-      {
         "type": "header",
-        "content": "Mobile"
+        "content": "t:settings_schema.layout.settings.header__grid.content"
+      },
+      {
+        "type": "paragraph",
+        "content": "t:settings_schema.layout.settings.paragraph__grid.content"
       },
       {
         "type": "range",
-        "id": "spacing_sections_mobile",
-        "min": 0,
-        "max": 120,
-        "step": 4,
-        "unit": "px",
-        "label": "Vertical space between sections",
-        "default": 52
-      },
-      {
-        "type": "range",
-        "id": "spacing_grid_horizontal_mobile",
+        "id": "spacing_grid_horizontal",
         "min": 4,
-        "max": 40,
+        "max": 72,
         "step": 4,
-        "default": 4,
+        "default": 8,
         "unit": "px",
-        "label": "Multicolumn horizontal space"
+        "label": "t:settings_schema.layout.settings.spacing_grid_horizontal.label"
       },
       {
         "type": "range",
-        "id": "spacing_grid_vertical_mobile",
+        "id": "spacing_grid_vertical",
         "min": 4,
-        "max": 40,
+        "max": 72,
         "step": 4,
-        "default": 4,
+        "default": 8,
         "unit": "px",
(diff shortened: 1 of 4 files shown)


config/settings_schema.json
```diff
@@ -151,74 +151,44 @@
...
+        "type": "paragraph",
+        "content": "t:settings_schema.layout.settings.paragraph__grid.content"
       },
       {
         "type": "range",
-        "id": "spacing_grid_horizontal_mobile",
+        "id": "spacing_grid_horizontal",
         "min": 4,
-        "max": 40,
+        "max": 72,
```
ludoboludo, line 177: this can definitely mess with things: 
- sliders [video](https://screenshot.click/21-12-krso0-zsxn7.mp4)
- collection cards in a collage scenario [screenshot](https://screenshot.click/21-12-l0q6u-mywuu.png)
- collection cards in a collection list section [screenshot](https://screenshot.click/21-12-s9ok4-mzwe0.png)

Not that those should be fixed here but maybe we can limit the max value to something different. Otherwise we will need follow up issue to tackle those scenarios. 

I also noticed a couple places that will also need addressing, noting them here so I can come back to it when creating issues later 😅 : 
- those spacing value mess with the product page because the `grid__item` have padding override: [screenshot](https://screenshot.click/21-12-se48o-fjxlv.png) (in a stacked and thumbnail layout)
- looks like the spacing is gone under the heading on a collection page. So filters look very close to it, [screenshot](https://screenshot.click/21-12-2tgie-c3ds2.png)
- feat. product slider doesn't seem to be using the right spacing values on a tablet sized viewport: [screenshot](https://screenshot.click/21-12-zht23-z119o.png)

ludoboludo (approved):
The change look good. 

I have some concerns with offering such a high max value. But code wise it the value/setting update looks good 👍

config/settings_schema.json
```diff
@@ -151,74 +151,44 @@
...
-        "type": "range",
-        "id": "spacing_sections_mobile",
-        "min": 0,
-        "max": 120,
-        "step": 4,
-        "unit": "px",
-        "label": "Vertical space between sections",
-        "default": 52
+        "type": "paragraph",
+        "content": "t:settings_schema.layout.settings.paragraph__grid.content"
```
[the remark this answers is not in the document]
melissaperreault: We might reconsider it from a help doc point of view but we didn't add any helper text elsewhere when we had this systematic ratio change based on device size. I would love to  take a  closer  look  in the next  year to  improve how we  communicate those logic with the help of  @KimliW and  @katycobb. cc.  @nicklepine

config/settings_schema.json
melissaperreault: Thanks Ludo! There is a lot of great point we can discuss:

>  Not that those should be fixed here but maybe we can limit the max value to something different

- Do we have an idea of what could be a max value that could not mess with the described findings? We can decrease the  max and take more time to increase it later, but I am curious if regardless we would need to adjust all the instances. 🤔

> those spacing value mess with the product page because the grid__item have padding override

- [ ]  I  would  expect this being a quick  follow up as it means  the setting is not applied consistently

> looks like the spacing is gone under the heading on a collection page. So filters look very close to it

- This should be solved by this section padding follow up issue https://github.com/Shopify/dawn/issues/1039

> feat. product slider doesn't seem to be using the right spacing values on a tablet sized viewport

-  I believe this was fixed already, can't find when 🤔 but otherwise let's open an issue please! 🙌

katycobb (approved):

kmeleta (approved):
Looks good to me. We definitely have some slider grid spacing stuff to work out, but that's not specific to this update.

melissaperreault (approved):
Thanks Kai! 🚢 

I'll be tackling range values for safe space here https://github.com/Shopify/dawn/issues/1066, from there we'll create issues where we'll need to come back and adjust.

PR #1059, opened 2021-12-21
Button refactor and fixes

### LINKED ISSUE #1021: [Global setting] Improve Button experience
### Expectations

There is a few adjustment to make to buttons, including if we are not offering the motion setting (https://github.com/Shopify/dawn/issues/994) sooner. This also implies experience on `Inputs` and `Variant pills`.

- Border and opacity [internal Figma reference](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=967%3A20018)

**View in your space**
- [x] When buttons have shadow [we see through the button and view the label](https://screenshot.click/21-12-68mk8-zda3x.png). To adjust the shadow application on that button

**Hover animation**
- [x] [Only the button and or border should enlarge](https://screenshot.click/21-12-s0okt-o1qhw.mp4), not the shadow.

**Shadow**
- [x] Shadow [should not pick the border thickness opacity](https://screenshot.click/21-12-h7g6j-rbor5.png)

**Focus** 
- [x] Focus should not pick the border thickness opacity

**Secondary button style**
- [x] Secondary button [not styled properly](https://screenshot.click/21-12-2aebt-jy2az.png). It should always have a border of 1px at 100% if opacity is zero. Once opacity increase [we could leverage the value to match](https://screenshot.click/21-12-7kclm-nwg6l.png) the primary button without a fill.
   - Note, `Variant pills` is okay because the selected variant style is [expected to not have a border](https://screenshot.click/21-12-8s6lu-5ccsn.png).

**Border thickness**
- [x] Button border thickness doesn't space out its sibling buttons - [Image banner](https://screenshot.click/21-12-p5shl-dk2ur.png) will suffer from it. Consider the new maximum range for button will be 24px too. - To investigate for `Inputs` too. 
   - [ ] Perhaps these new settings will help https://github.com/Shopify/dawn/issues/983
   - Note from Ken 
   > Ideally we don't solve this by just adding more margin from the #983 settings, but I believe the work that will need to be done for that will potentially involve a refactor of how borders and shadows are applied to buttons and could inherently fix this.

kmeleta:

Updates include..

- Updated hover animations to increase border width by 1px
- Prevent drop shadow from being affected by hover and focus styles
- Prevent potential shadow content issue
- Prevent shadow from overlapping adjacent buttons (see notes below)

To ensure button **shadows don't overlap** adjacent elements without a particular z-index set, the button's z-index must not be set either. Otherwise a new stacking context is created, because css. This usually isn't a problem, but I observed areas where this configuration can actually cause the shadow to be lost behind a parent element because the shadow's effective z-index is -1 ([see here](https://screenshot.click/21-26-n2c1b-vhpib.png)). This fix for this would be to give the button's parent element a z-index needed where this issue occurs. 

The approach I've taken is to give buttons a z-index: 1 by default and reset it in areas where their shadows could _potentially_ overlap other elements. The former approach may be more inline with other shadow settings, but it seems like this approach would require fewer section fixes, but I could be wrong and I don't feel very strongly about one vs the other.

**Other considerations**

I noticed an issue with background color artifacts bleeding through rounded corner buttons. This is likely happening because  ~~css is weird and browsers are terrible~~  I'm adding an opaque background directly behind transparent borders so we don't get weird opacity overlaps with shadows ([see here](https://screenshot.click/21-42-n8pl6-j0rkj.mp4)). I noticed in all browsers I could reduce this effect by offsetting the potentially transparent border from the underlying opaque background color by a fraction of a pixel ([see here](https://screenshot.click/21-48-yncnh-ny5l9.png)).

This issue is only visible along a border radius and the opaque background hack is really only needed when buttons have a drop shadow. So we can probably optimize a bit by avoiding this weird arbitrary fix when these two conditions aren't met.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127031640086/editor)

**Checklist**


melissaperreault (commented):
Thanks Ken! 🎉 

I am not sure what is ideal for the technical side but I'll comment from an experience and UI point of view. I tested all buttons, also with Disabled JS.

- [ ] We fixed the corner radius with this PR https://github.com/Shopify/dawn/pull/1024 but I see we don't have this styles on the demo. To validate this isn't lost.
   - [Product template reference](https://screenshot.click/21-12-fxpe8-0ztwu.png)
   - [Cart template reference](https://screenshot.click/21-12-41nl1-9y28w.png)
- [ ] If we are capable of fixing the [visible outline outside the border](https://screenshot.click/21-12-f0y54-3xw4t.mp4) like described, this is great! We see it even more with the `max` range value.
- [ ] There is an issue with the [delete icon](https://screenshot.click/21-12-xkmin-7grrt.png) in the cart template, the border grows inside instead of outside. This happens to all tertiary button style. [Screenshot of the footer select](https://screenshot.click/21-12-88qu4-xma84.png)

kmeleta:
Updated PR here https://github.com/Shopify/dawn/pull/1064

Issue #930, opened 2021-11-19
[Global settings] Dropdowns and pop-ups


melissaperreault:
### Settings reference
- [Internal Figma link](https://www.figma.com/file/9FOEO0DFKgNDjiqxMakiLB/Design-Settings-Components?node-id=141%3A11142)

### Observations

**Content** 
- [x] Popups = Dropdowns and pop-ups
- [x] Opacity label is named wrong under Shadow

**Search dropdown**
- [ ] Desktop: Unsure if it would be better to have the border [outside instead of inside](https://screenshot.click/21-12-x93tw-jszdl.png)
- [ ] Small screen when the component touches the edges, align with the cart notification popup [where we only apply the settings at the bottom](https://screenshot.click/21-12-wipc6-atub0.png)
   - [ ] [We wouldn't see the shadow](https://screenshot.click/21-12-t9570-w7xgj.png) at the top here

tyleralsbury:
@LucasLacerdaUX Meli will do the 2 content things.  Your focus will be the other 2 points.

melissaperreault:
**Country/region and language inputs**

- [ ] The dropdown [doesn't work occasionally](https://screenshot.click/21-12-bg612-8xfy2.png) (video)

PR #1064, opened 2021-12-22
Refactor button styles

### LINKED ISSUE #1021: [Global setting] Improve Button experience

kmeleta:

Fixes #1021 

Updates include..

- Updated hover animations to increase border width by 1px
- Prevent drop shadow from being affected by hover and focus styles
- Prevent potential shadow content issue
- Prevent shadow from overlapping adjacent buttons (see notes below)

To ensure button shadows don't overlap adjacent elements without a particular z-index set, the button's z-index must not be set either. Otherwise a new stacking context is created, because css. This usually isn't a problem, but I observed areas where this configuration can actually cause the shadow to be lost behind a parent element's background color because the shadow's effective z-index is -1 ([see here](https://screenshot.click/21-26-n2c1b-vhpib.png)). This fix for this is to give the button's parent element a z-index (and positioning). If you see something (media, container, button, etc) that _should_ have a shadow but _doesn't_, this is the likely cause.

I've identified the following sections need this fix..
- Image banner
- Image with text
- Rich text
- Featured product

**Other considerations**

I noticed an issue with background color artifacts bleeding through rounded corner buttons. This is likely happening because  ~~css is weird and browsers are terrible~~  I'm adding an opaque background directly behind transparent borders so we don't get weird opacity overlaps with shadows ([see here](https://screenshot.click/21-42-n8pl6-j0rkj.mp4)). I noticed in all browsers (on my machine at least) I could reduce this effect by offsetting the potentially transparent border from the underlying opaque background color by a fraction of a pixel ([see here](https://screenshot.click/21-48-yncnh-ny5l9.png)). This issue is only visible along a border radius and the opaque background hack is really only needed when buttons have a drop shadow. So we can probably optimize a bit by avoiding this weird arbitrary fix when these two conditions aren't met.

There's also an issue with shadow stacking on the Payment checkout button ([see here](https://screenshot.click/22-34-ikpxg-dhwku.png)). It should only affect the button _above_. If the shadow offset is positive (below) then it shouldn't be noticeable. This button is rendered as an iframe which doesn't accept our pseudo element approach to keeping shadows below stuff. We can revisit in the future and see if we can further hack it.

**Testing notes**
Check the following..

- Primary, secondary, tertiary buttons
- Branded/unbranded checkout buttons
- Buttons in various sections
- Hover and focus states

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127034490902/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1257,7 +1257,6 @@ button.shopify-payment-button__button--unbranded {
   align-items: center;
   border: 0;
   padding: 0 3rem;
-  border-radius: calc(var(--buttons-border-width) + var(--buttons-radius));
   cursor: pointer;
   font: inherit;
   font-size: 1.5rem;
@@ -1323,9 +1322,9 @@ button.shopify-payment-button__button--unbranded {
     0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
 }
 
-.button:focus:not(:focus-visible, .focused),
-.shopify-payment-button__button--unbranded:focus:not(:focus-visible, .focused),
-.shopify-payment-button [role="button"]:focus:not(:focus-visible, .focused) {
+.button:focus:not(:focus-visible):not(.focused),
+.shopify-payment-button__button--unbranded:focus:not(:focus-visible):not(.focused),
+.shopify-payment-button [role="button"]:focus:not(:focus-visible):not(.focused) {
   box-shadow: inherit;
 }
 
--- a/assets/base.css
+++ b/assets/base.css
@@ -1292,7 +1292,7 @@ button.shopify-payment-button__button--unbranded {
   z-index: 1;
   border-radius: var(--buttons-radius);
   box-shadow: 0 0 0 calc(var(--buttons-border-width) + var(--border-offset)) rgba(var(--color-button-text), var(--border-opacity)),
-    0 0 0 var(--buttons-border-width) rgba(var(--color-button), 1);
+    0 0 0 var(--buttons-border-width) rgba(var(--color-button), var(--alpha-button-background));
   transition: box-shadow var(--duration-short) ease;
 }
 
@@ -1302,7 +1302,7 @@ button.shopify-payment-button__button--unbranded {
 .shopify-payment-button__button--unbranded:hover::after {
   --border-offset: 1.3px;
   box-shadow: 0 0 0 calc(var(--buttons-border-width) + var(--border-offset)) rgba(var(--color-button-text), var(--border-opacity)),
-    0 0 0 calc(var(--buttons-border-width) + 1px) rgba(var(--color-button), 1);
+    0 0 0 calc(var(--buttons-border-width) + 1px) rgba(var(--color-button), var(--alpha-button-background));
 }
 
 .button--secondary:after {
@@ -1346,8 +1346,8 @@ button.shopify-payment-button__button--unbranded {
 .button--tertiary {
   font-size: 1.2rem;
   padding: 1rem 1.5rem;
-  min-width: 9rem;
-  min-height: 3.5rem;
+  min-width: calc(9rem + var(--buttons-border-width) * 2);
+  min-height: calc(3.5rem + var(--buttons-border-width) * 2);
 }
 
 .button--small {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1231,6 +1231,12 @@ button.shopify-payment-button__button--unbranded,
   --border-opacity: calc(1 - var(--buttons-border-opacity));
   border-radius: var(--buttons-radius);
   position: relative;
+}
+
+.button,
+.shopify-challenge__button,
+.customer button,
+button.shopify-payment-button__button--unbranded {
   min-width: calc(12rem + var(--buttons-border-width) * 2);
   min-height: calc(4.5rem + var(--buttons-border-width) * 2);
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1218,37 +1218,8 @@ deferred-media {
 
 .button,
 .shopify-challenge__button,
-.customer button {
-  cursor: pointer;
-  display: inline-flex;
-  justify-content: center;
-  align-items: center;
-  box-sizing: border-box;
-  font: inherit;
-  font-size: 1.5rem;
-  padding: 0.9rem 3rem 1.1rem;
-  text-decoration: none;
-  border: 1px solid transparent;
-  background-color: rgba(var(--color-button), var(--alpha-button-background));
-  color: rgb(var(--color-button-text));
-  min-width: 12rem;
-  min-height: 4.5rem;
-  -webkit-appearance: none;
-  appearance: none;
-}
-
-.button,
-.shopify-challenge__button,
 .customer button,
-.shopify-payment-button__button--unbranded,
-.shopify-payment-button [role="button"] {
-  box-shadow: 0 0 0 var(--buttons-border-width) rgba(var(--color-button), var(--buttons-border-opacity));
-}
-
-.button,
-.shopify-challenge__button,
-.customer button,
-.shopify-payment-button__button--unbranded,
+button.shopify-payment-button__button--unbranded,
 .shopify-payment-button [role="button"],
 .cart__dynamic-checkout-buttons [role='button'],
 .cart__dynamic-checkout-buttons iframe {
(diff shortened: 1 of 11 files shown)


kmeleta:
Feedback from Melissa

> Thanks Ken! 🎉
> 
> I am not sure what is ideal for the technical side but I'll comment from an experience and UI point of view. I tested all buttons, also with Disabled JS.
> 
> * [ ]  We fixed the corner radius with this PR #1024 but I see we don't have this styles on the demo. To validate this isn't lost.
>   
>   * [Product template reference](https://screenshot.click/21-12-fxpe8-0ztwu.png)
>   * [Cart template reference](https://screenshot.click/21-12-41nl1-9y28w.png)
> * [ ]  If we are capable of fixing the [visible outline outside the border](https://screenshot.click/21-12-f0y54-3xw4t.mp4) like described, this is great! We see it even more with the `max` range value.
> * [ ]  There is an issue with the [delete icon](https://screenshot.click/21-12-xkmin-7grrt.png) in the cart template, the border grows inside instead of outside. This happens to all tertiary button style. [Screenshot of the footer select](https://screenshot.click/21-12-88qu4-xma84.png)

melissaperreault (commented):
Thank you Ken! 🎉 

Noting some observations while testing:

- [ ] Dynamic checkout buttons; I believe only [Paypal's shadow will overlap on top of the other buttons](https://screenshot.click/21-12-c35k5-88kat.png), not sure what we can do about that one.
- [ ] Unsure why but the [footer select don't work on this demo](https://screenshot.click/21-12-8hkwy-p1s7w.mp4), I tested on other ones and it works. 
- [ ] The `Delete` icon focus [doesn't not display anymore](https://screenshot.click/21-12-cpkox-0zg1n.mp4)
- [ ] The tertiary button border should either have the same border opacity of the primary and secondary OR have a logic where it is always lighter like it was originally on the build, if possible to create that relationship; we could also this later as a polish phase. [We don't want the border darker like this example](https://screenshot.click/21-12-1oa4x-rc8fj.png)

Happy to see `View in your space` button is looking good! 🎉

assets/base.css
```diff
@@ -1216,86 +1216,111 @@ deferred-media {
...
-  --shadow-vertical-offset: var(--buttons-shadow-vertical-offset);
-  --shadow-blur-radius: var(--buttons-shadow-blur-radius);
-  --shadow-opacity: var(--buttons-shadow-opacity);
+.button:after,
+.shopify-challenge__button:after,
+.customer button:after,
+.shopify-payment-button__button--unbranded:after {
+  content: '';
+  position: absolute;
+  inset: var(--buttons-border-width);
```
ludoboludo, line 1286: This seem to mess with higher value. Because we're applying a radius on something that isn't the same size as the actual element, it ends up having a different "curve" to it. So when you go on higher values like `24px` then you notice it even more: 
[not super high value](https://screenshot.click/21-12-eydor-c5dg7.png)
[higher value with not so high border-radius](https://screenshot.click/21-12-ppyh6-8ewox.png)

assets/base.css
```diff
@@ -1216,86 +1216,111 @@ deferred-media {
...
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),
+    0 0 0 0.5rem rgba(var(--color-foreground), 0.5),
     0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
 }
 
-.button:focus:not(:focus-visible),
-.shopify-payment-button__button--unbranded:focus:not(:focus-visible),
-.shopify-payment-button [role="button"]:focus:not(:focus-visible) {
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-button), var(--alpha-button-border));
+.button:focus:not(:focus-visible, .focused),
```
ludoboludo, line 1320: I think here we need to do `button:focus:not(:focus-visible):not(.focused)` instead as I remember we had to tweak it in the past as it was messing something. By the looks of it, IE. Not sure if that was the only reason but we wanted to not break atc buttons on IE even though it's not really supported.

ludoboludo (commented):
Definitely looks tricky to try and implement and have it work with all different settings/possibilities. 

Here are a few things I noted on top of the comments.

- [ ] When using a high value for border thickness, it mess the look of the button on the cart page. Dynamic checkout and delete button don't look good: [screenshot](https://screenshot.click/21-12-fot4l-eos7d.png)
- [ ] When looking at filters, they're pretty tight and might need a tiny bit more padding. Or padding based on the border thickness value 🤷  : [when thickness is at its highest](https://screenshot.click/21-12-5dlyq-k5nxo.png) and [a bit of a smaller value](https://screenshot.click/21-12-vola6-qwure.png)
- [ ] Secondary style button on the address part of the account page are overlapping each other: [screenshot](https://screenshot.click/21-12-67x4j-uty04.png)
- [ ] Should local "buttons" be impacted ? [screenshot](https://screenshot.click/21-12-bvxfd-algk0.png)

KaichenWang (approved):
Code looks good. 

Not sure if `button--tertiary` is supposed to be accounted for in this PR, but when border thickness is increased, the button size doesn't scale well to accommodate that. For instance in the cart

Screen_Shot_2021-12-22_at_5_51_16_PM

kmeleta:
Thanks all. I believe all feedback to this point has been addressed **if not otherwise noted below**

> * [ ]  Dynamic checkout buttons; I believe only Paypal's shadow will overlap on top of the other buttons

Won't address in this commit. As discussed and noted in PR description there are limitations here.

> * [ ]  Unsure why but the [footer select don't work on this demo](https://screenshot.click/21-12-8hkwy-p1s7w.mp4), I tested on other ones and it works.

As discussed, this is an existing issue we hoped was resolved. It's been noted [here](https://github.com/Shopify/dawn/issues/930#issuecomment-999918051)

> * [ ] The tertiary button border should either have the same border opacity of the primary and secondary OR have a logic where it is always lighter like it was originally on the build, if possible to create that relationship; we could also this later as a polish phase. We don't want the border darker like this example

I'm currently working on tertiary styles, but I don't want them to hold up review. The tertiary cart delete button has been restored to its borderless/shadowless style though. And its focus outline has been fixed as well.

> * [ ]  When looking at filters, they're pretty tight and might need a tiny bit more padding. Or padding based on the border thickness value 🤷  : [when thickness is at its highest](https://screenshot.click/21-12-5dlyq-k5nxo.png) and [a bit of a smaller value](https://screenshot.click/21-12-vola6-qwure.png)

These actually should have been ignoring new button styles. Updated.

> * [ ]  Should local "buttons" be impacted ? [screenshot](https://screenshot.click/21-12-bvxfd-algk0.png)

Nope, I think there's a issue open to give them input styles

> I think here we need to do button:focus:not(:focus-visible):not(.focused) instead as I remember we had to tweak it in the past as it was messing something. By the looks of it, IE. Not sure if that was the only reason but we wanted to not break atc buttons on IE even though it's not really supported.

I may end up doing a follow up for focus styles to make sure everything is in order in IE and also safari (which doesn't do focus-visible)

assets/base.css
kmeleta: Good find. This was happening at ranges I wasn't necessarily testing with much. It's a case of needing to add the radius and the border width. Should be fixed but let me know if you can still break it.

assets/base.css
```diff
@@ -1216,86 +1216,116 @@ deferred-media {
...
-    var(--shadow-horizontal-offset)
-    var(--shadow-vertical-offset)
-    var(--shadow-blur-radius)
-    rgba(var(--color-button-text), var(--shadow-opacity))
-  );
+.button:not([disabled]):hover::after,
+.shopify-challenge__button:hover::after,
+.customer button:hover::after,
+.shopify-payment-button__button--unbranded:hover::after {
+  --border-offset: 1.3px;
```
ludoboludo, line 1303: What's the 1.3px based of ?

ludoboludo (approved):
Looks good 👍

kmeleta:
Thanks @melissaperreault 

> * [ ]  Tertiary button border opacity looks good, [just missing growing the border outside](https://screenshot.click/21-12-498th-qe2op.png)

Updated the min-height/width to account for the borders. I wish we didn't have to do that. Lmk if that looks better now though.

> * [ ]  The cart notification [don't display](https://screenshot.click/21-12-kthar-wfxyz.mp4) upon adding to cart

This is definitely a result of that issue that's blocking the locale selects as well. Must be a js error somewhere that happens under some combination of conditions that's blocking a variety of things. Nothing specific to this commit, but should be resolved along with the locales issue.

Issue #1066, opened 2021-12-22
[Global setting] Lower some range values for MVP


melissaperreault:
### Expectations

- Avoid outcome that feels broken for now until we plan for higher max value
   - [ ] Adjust Layout `Grid`
      - [ ] Sliders [video](https://screenshot.click/21-12-krso0-zsxn7.mp4)
      - [ ] Collection cards in a collage scenario [screenshot](https://screenshot.click/21-12-l0q6u-mywuu.png)
      - [ ] Collection cards in a collection list section [screenshot](https://screenshot.click/21-12-s9ok4-mzwe0.png)
   - [ ] `Corner radius` where necessary. When rounding up corners of an element, the text content within the element should remain legible and should not be cut off.
      - [ ] Avoid text being [cut off like this](https://screenshot.click/21-12-oqsu9-z4eni.png) 
   - [ ] `Shadow blur` and `offset` where necessary
      - [ ] [A result like this is expected](https://screenshot.click/22-41-mij03-hst8q.png), with the expectation that it is always behind all elements. But for sliders, it is pushing the content far away if too high for instance.

PR #1069, opened 2021-12-22
Revise range values

melissaperreault:

Fixes https://github.com/Shopify/dawn/issues/1066

Adjusting the values to ensure nothing is breaking.

- I harmonized the `Vertical space between sections` to match the max value from `Section padding`.
- `Corner radius` was set to 40 instead of 72
- Shadow Blur for `Buttons` and `Inputs` was reduced to 30 instead of 100
- Shadow Blur for `Cards`, `Content containers`, `Drawers`, `Popups` and `Media` was reduced to 40 instead of 100

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127035211798/editor)


config/settings_schema.json
```diff
@@ -156,7 +156,7 @@
         "type": "range",
         "id": "spacing_sections",
         "min": 0,
-        "max": 120,
+        "max": 100,
```
melissaperreault, line 159: Matching the Section padding `max` value.

config/settings_schema.json
```diff
@@ -590,7 +590,7 @@
         "type": "range",
         "id": "card_shadow_blur",
         "min": 0,
-        "max": 100,
+        "max": 40,
```
ludoboludo, line 593: I think this should be 30 a to match the others. Or the others should be 40 as well 😅

config/settings_schema.json
```diff
@@ -673,7 +673,7 @@
         "type": "range",
         "id": "text_boxes_shadow_blur",
         "min": 0,
-        "max": 100,
+        "max": 40,
```
ludoboludo, line 676: I think this should be 30 to match the others.

config/settings_schema.json
```diff
@@ -756,7 +756,7 @@
         "type": "range",
         "id": "media_shadow_blur",
         "min": 0,
-        "max": 100,
+        "max": 40,
```
ludoboludo, line 759: Same here, we keep it a 40 ? Or match some of the others with 30 ?

config/settings_schema.json
```diff
@@ -843,7 +843,7 @@
         "type": "range",
         "id": "popup_shadow_blur",
         "min": 0,
-        "max": 100,
+        "max": 40,
```
ludoboludo, line 846: Same here

config/settings_schema.json
```diff
@@ -916,7 +916,7 @@
         "type": "range",
         "id": "drawer_shadow_blur",
         "min": 0,
-        "max": 100,
+        "max": 40,
```
ludoboludo, line 919: Same here :)

ludoboludo (commented):
Should variant pills match the rest of the blur values and have 30 instead of 20 as well ? Or because it's a small element we keep it lower ?

config/settings_schema.json
melissaperreault: I made a scale change, `Variant pills` at 20 because they are so tiny, `Buttons` and `Inputs` at 30 and larger components like `Cards` etc at 40. Very subjective! 😅

melissaperreault: cc. @kmeleta How do you both feel about this curated difference for different components for now?

kmeleta: I'm good with that. The same values across the board would make me feel warm and fuzzy, but if it doesn't artistically make sense to do that I think it's sensible to assign a small, med, and large element max.

ludoboludo: Sounds good! I will resolve all those comments that are about the same thing then 👍

ludoboludo (approved):

PR #1084, opened 2022-01-04
Added global media settings to collection banner

### LINKED ISSUE #1058: [Global Settings] Collection Banner image should inherit Media global style settings
The collection banner section's image should inherit the styles from Media global settings. 

image

ludoboludo:

Fixes #1058 

Added the general media settings to the image container. I didn't use the pseudo element approach as it did not seem necessary in this scenario. 

**Testing guidelines**

Go to a collection page and confirm the shadow and borders are applied properly to the media in the collection banner. 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127055593494)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127055593494/editor)

**Checklist**

### DIFF
--- a/assets/component-collection-hero.css
+++ b/assets/component-collection-hero.css
@@ -16,6 +16,7 @@
   .collection-hero.collection-hero--with-image {
     padding: calc(4rem + var(--page-width-margin)) 0
       calc(4rem + var(--page-width-margin));
+    overflow: hidden;
   }
 
   .collection-hero--with-image .collection-hero__inner {
@@ -70,6 +71,15 @@
 
 .collection-hero--with-image .collection-hero__text-wrapper {
   padding: 5rem 0 4rem;
+}
+
+.collection-hero__image-container {
+  border: var(--media-border-width) solid rgba(var(--color-foreground), var(--media-border-opacity));
+  border-radius: var(--media-radius);
+  box-shadow: var(--media-shadow-horizontal-offset)
+    var(--media-shadow-vertical-offset)
+    var(--media-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--media-shadow-opacity));
 }
 
 @media screen and (max-width: 749px) {
--- a/sections/main-collection-banner.liquid
+++ b/sections/main-collection-banner.liquid
@@ -1,5 +1,15 @@
 {{ 'component-collection-hero.css' | asset_url | stylesheet_tag }}
 
+{%- if settings.media_shadow_vertical_offset > 0 -%}
+  {%- style -%}
+    @media screen and (max-width: 749px) {
+      .collection-hero--with-image .collection-hero__inner {
+        padding-bottom: calc({{ settings.media_shadow_vertical_offset }}px + 2rem);
+      }
+    }
+  {%- endstyle -%}
+{%- endif -%}
+
 <div class="collection-hero{% if section.settings.show_collection_image and collection.image %} collection-hero--with-image{% endif %}">
   <div class="collection-hero__inner page-width">
     <div class="collection-hero__text-wrapper">
--- a/sections/main-collection-banner.liquid
+++ b/sections/main-collection-banner.liquid
@@ -1,14 +1,12 @@
 {{ 'component-collection-hero.css' | asset_url | stylesheet_tag }}
 
-{%- if settings.media_shadow_vertical_offset > 0 -%}
-  {%- style -%}
-    @media screen and (max-width: 749px) {
-      .collection-hero--with-image .collection-hero__inner {
-        padding-bottom: calc({{ settings.media_shadow_vertical_offset }}px + 2rem);
-      }
+{%- style -%}
+  @media screen and (max-width: 749px) {
+    .collection-hero--with-image .collection-hero__inner {
+      padding-bottom: calc({{ settings.media_shadow_vertical_offset | at_least: 0 }}px + 2rem);
     }
-  {%- endstyle -%}
-{%- endif -%}
+  }
+{%- endstyle -%}
 
 <div class="collection-hero{% if section.settings.show_collection_image and collection.image %} collection-hero--with-image{% endif %}">
   <div class="collection-hero__inner page-width">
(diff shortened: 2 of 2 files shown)


assets/component-collection-hero.css
```diff
@@ -16,6 +16,7 @@
   .collection-hero.collection-hero--with-image {
     padding: calc(4rem + var(--page-width-margin)) 0
       calc(4rem + var(--page-width-margin));
+    overflow: hidden;
```
ludoboludo, line 19: This I need to make sure it doesn't "spill" on the next elements: [screenshot](https://screenshot.click/22-01-0h55h-vjs8g.png).

melissaperreault (commented):
Thank you Ludo! 💪 

2 things I observed:

- `Image with text` lost their implementation it seems, unsure when/how this is happening. [Video explanation.](https://screenshot.click/22-01-txe7q-3x6gk.mp4)
- Do we know why the border opacity [doesn't look the same](https://screenshot.click/22-01-cnq43-fpi2n.png) on the `Collection banner` and other media?

sections/main-collection-banner.liquid
```diff
@@ -14,7 +24,7 @@
     </div>
 
     {%- if section.settings.show_collection_image and collection.image -%}
-      <div class="collection-hero__image-container media">
+      <div class="collection-hero__image-container media gradient">
```
ludoboludo, line 27: Good catches @melissaperreault 🙂  thank you. 
One of the issue isn't related to this fix, so we can create a different issue to tackle it. 

As for the border image, I added the `gradient` class it was missing and now the border is coming up with the right color matching what we have for the image with text section. 👍 

Which made me test a scenario with a gradient for the background and realized the collage section is missing that class too:


But that can be created as an issue or actually, I will add this tidbit in the existing issue for Collage.

melissaperreault: Thanks Ludo!

melissaperreault (approved):
Thanks Ludo, looks good now! 🙏

assets/component-collection-hero.css
```diff
@@ -72,6 +73,15 @@
...
 
+.collection-hero__image-container {
+  border: var(--media-border-width) solid rgba(var(--color-foreground), var(--media-border-opacity));
+  border-radius: var(--media-radius);
+  box-shadow: var(--media-shadow-horizontal-offset)
+    var(--media-shadow-vertical-offset)
+    var(--media-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--media-shadow-opacity));
+}
+
```
tyleralsbury, line 84: I'm starting to wonder if it would make sense for us to, as a refactor, add some classes that represent each of these things - `media` for things that take media settings, `content-container` for things that take those settings, etc...

**Some benefits**

* Reduction of CSS
* Clearer relationship in the code between the settings and what they affect

The downside would be that some situations do have special cases, but maybe we can make that work anyway.

Something we can think about!

sections/main-collection-banner.liquid
```diff
@@ -1,5 +1,15 @@
...
 
+{%- if settings.media_shadow_vertical_offset > 0 -%}
+  {%- style -%}
+    @media screen and (max-width: 749px) {
+      .collection-hero--with-image .collection-hero__inner {
+        padding-bottom: calc({{ settings.media_shadow_vertical_offset }}px + 2rem);
+      }
+    }
+  {%- endstyle -%}
+{%- endif -%}
```
tyleralsbury, line 11: I'm wondering if we need the `if` here.  If we still want there to be a minimum padding of `2rem`, then it would just add `0 + 2rem` and still end up as `2rem`.

tyleralsbury (commented):
Question regarding an `if` statement, but overall it looks great.

assets/component-collection-hero.css
martinamarien: I am actually already trying to do this with the media settings in my current PR. I have a global-media-settings class. And we added a `content-containers` class as well.

```
.global-media-settings {
   ... 
}

.global-media-settings:after {
   ...
}

.global-media-settings--full-width,
.content-containers--full-width {
   ... 
}
``` 

I'm going through it now to make sure it applies everywhere well

sections/main-collection-banner.liquid
ludoboludo: But since the vertical offset can be negative then I think we end up with a negative padding value 🤔

ludoboludo: But because of negative values of the vertical offset, it would mess with the padding and end up being `0`. Seems like padding doesn't work with negative value so it will default to `0` 🤔

tyleralsbury: Good point - we could use `at_least: 0`.

https://shopify.dev/api/liquid/filters/math-filters

ludoboludo: Ah yes, that's great 👌  I made the update.

ludoboludo: Ah smart idea! I made the change 👌

melissaperreault (approved):

sections/main-collection-banner.liquid
```diff
@@ -1,5 +1,13 @@
 {{ 'component-collection-hero.css' | asset_url | stylesheet_tag }}
 
+{%- style -%}
+  @media screen and (max-width: 749px) {
+    .collection-hero--with-image .collection-hero__inner {
+      padding-bottom: calc({{ settings.media_shadow_vertical_offset | at_least: 0 }}px + 2rem);
+    }
+  }
+{%- endstyle -%}
```
martinamarien, line 9: Screen Shot 2022-01-12 at 11 51 35 AM

I wonder if we should be trying to do something similar when the offset is negative. It can overlap the text a little bit.

ludoboludo: With both vertical offset set to the negative max and blur to the max it didn't mess with it so much 🤔  So I thought it should be fine.

assets/component-collection-hero.css
martinamarien: Another option could be adding the box-shadow-vertical-offset as padding, like we're doing with sliders and you are doing on mobile, but something we can discuss in the future.

PR #1086, opened 2022-01-04
Change shadow color source

### LINKED ISSUE #1026: [Global setting] Change shadow color source
### Expectation

- [ ] Let's change all shadow color instances to be styled with the `var(--color-base-text)` instead of relying on `var(—color-foreground)`

ludoboludo:

Fixes #1026 

Changed the variables used for both `box-shadow` and `filter` property where the rest of the shadow settings were applied. 

In some instances it doesn't seem necessary to make the change as there isn't a way to change the color scheme on the container but I figured it'd be better for it to use the value everywhere. 

**Other considerations**

I haven't tweaked the values for the button shadows.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127056805910)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127056805910/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1522,7 +1522,7 @@ details[open] > .share-button__fallback {
     var(--inputs-shadow-horizontal-offset)
     var(--inputs-shadow-vertical-offset)
     var(--inputs-shadow-blur-radius)
-    rgba(var(--color-base-text), var(--inputs-shadow-opacity))
+    rgba(var(--color-shadow), var(--inputs-shadow-opacity))
   );
 }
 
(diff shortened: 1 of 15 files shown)


kmeleta (commented):
Probably not necessary, but it might be helpful to alias the `--color-base-text` as `--color-shadow` or something so we can override it in certain contexts a little more easily if we have that need. What do you think?

ludoboludo:
Yes! great idea and it can be leveraged later if we add section overrides. 👍  Thanks @kmeleta

kmeleta (approved):
Looks good. I think we're just missing an instance of --color-shadow in the `button:before` block in base.css that was already using --color-base-text. Shouldn't hurt anything currently but would be good to update so we don't forget.

`box-shadow: var(--shadow-horizontal-offset) var(--shadow-vertical-offset) var(--shadow-blur-radius) rgba(var(--color-base-text), var(--shadow-opacity));`

kmeleta (approved):

tyleralsbury (approved):
Looks good - checked code and a quick smoke test.

PR #1091, opened 2022-01-05
Update template CSS variables

KaichenWang:

A few CSS variables were added/modified in `theme.liquid` in order to add global settings. `password.liquid` and `gift_card.liquid` must be updated to match the CSS variables in `theme.liquid`.

- `password.liquid`: Copy all CSS variables from `theme.liquid`
- `gift_card.liquid`: Update existing CSS variables to match those in `theme.liquid` (not all theme CSS variables are needed on the gift card template)

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127061000214)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127061000214/editor)

**Checklist**


tyleralsbury (approved):
Looks good - I tested by comparing the variables between the theme and password pages and it looks like they all match.  The gift card one has fewer, as mentioned in the description, so I just cross-referenced it with the classes being used on that page and it looks good.

layout/password.liquid
```diff
@@ -63,9 +63,53 @@
...
+        
+        --popup-border-width: {{ settings.popup_border_thickness }}px;
+        --popup-border-opacity: {{ settings.popup_border_opacity | divided_by: 100.0 }};
+        --popup-corner-radius: {{ settings.popup_corner_radius }}px;
+        --popup-shadow-opacity: {{ settings.popup_shadow_opacity | divided_by: 100.0 }};
+        --popup-shadow-horizontal-offset: {{ settings.popup_shadow_horizontal_offset }}px;
+        --popup-shadow-vertical-offset: {{ settings.popup_shadow_vertical_offset }}px;
+        --popup-shadow-blur-radius: {{ settings.popup_shadow_blur }}px;
+
+        --drawer-border-width: {{ settings.drawer_border_thickness }}px;
```
sofiamatulis, line 98: Drawers don't exist in the password page since they are only for the filter, cart, pick up availability and menu (password page has a different menu). Do we need it?

layout/password.liquid
```diff
@@ -63,9 +63,53 @@
...
+        --card-border-width: {{ settings.card_border_thickness | divided_by: 10.0 }}rem;
+        --card-border-opacity: {{ settings.card_border_opacity | divided_by: 100.0 }};
+        --card-shadow-opacity: {{ settings.card_shadow_opacity | divided_by: 100.0 }};
+        --card-shadow-horizontal-offset: {{ settings.card_shadow_horizontal_offset | divided_by: 10.0 }}rem;
+        --card-shadow-vertical-offset: {{ settings.card_shadow_vertical_offset | divided_by: 10.0 }}rem;
+        --card-shadow-blur-radius: {{ settings.card_shadow_blur | divided_by: 10.0 }}rem;
+
+        --badge-corner-radius: {{ settings.badge_corner_radius | divided_by: 10.0 }}rem;
+        
+        --popup-border-width: {{ settings.popup_border_thickness }}px;
```
sofiamatulis, line 90: Same with popup. Do we need it? It seems to be only used in the cart, search, filter, footer, product page and header which dont exist in the password page

layout/password.liquid
```diff
@@ -90,13 +135,13 @@
...
         --inputs-shadow-blur-radius: {{ settings.inputs_shadow_blur }}px;
 
-        --spacing-sections-desktop: {{ settings.spacing_sections }}px;
-        --spacing-sections-mobile: {% if settings.spacing_sections < 24 %}{{ settings.spacing_sections }}{% else %}{{ settings.spacing_sections | times: 0.7 | round | at_least: 20 }}{% endif %}px;
-
-        --grid-desktop-vertical-spacing: {{ settings.spacing_grid_vertical }}px;
-        --grid-desktop-horizontal-spacing: {{ settings.spacing_grid_horizontal }}px;
-        --grid-mobile-vertical-spacing: {{ settings.spacing_grid_vertical | divided_by: 2 }}px;
-        --grid-mobile-horizontal-spacing: {{ settings.spacing_grid_horizontal | divided_by: 2 }}px;
+        --variant-pills-radius: {{ settings.variant_pills_radius }}px;
```
sofiamatulis, line 138: Same with variants. They seem to only exist in the product page.

sofiamatulis (commented):
Looks great 🎉 Just added a few small comments

layout/password.liquid
KaichenWang: I believe variant pills can exist in a featured-product section on the password page as well

layout/password.liquid
KaichenWang: The featured product section should in theory trigger a cart notification, which is a popup, but that functionality is currently not working on password page. Opening an issue https://github.com/Shopify/dawn/issues/1144

layout/password.liquid
KaichenWang: Good point! I'm ok with removing `--drawer-` variables from password page. One argument for keeping them is to ensure the same CSS variables are available across then theme to be used in the Custom Liquid section, but that is a very limited use-case.

layout/password.liquid
sofiamatulis: I believe the issue it doesn't work is because technically if a site is password protected, it means we cant access the other pages. So even though we allow merchants to add any section to the password page, it's not encouraged. So I am not sure if I would consider that an issue

cc @melissaperreault @Oliviammarcello

melissaperreault: Hum, @nicklepine might have more context on that one, I don't remember. 🤔

KaichenWang: Removed the `-popup` variables from password page https://github.com/Shopify/dawn/pull/1091/commits/f04b9cd577d16bc610e78ef0211ec2cc633051fb

tyleralsbury (approved):

PR #1097, opened 2022-01-05
Global styling: Inputs styling not applied consistently

### LINKED ISSUE #993: [Global settings] Inputs styling not applied consistently
### Observation

- [x] Country/region and language selectors [don't receive the Inputs styling yet](https://screenshot.click/21-12-wyyxw-vxfph.png)
- [x] Stroke applied inside [instead of outside on Quantity input](https://screenshot.click/21-12-m4rwj-gy37a.png)
- [x] [Share block](https://screenshot.click/21-12-1dox2-a7oep.png) on Product template [isn't styled properly](https://screenshot.click/21-12-05o3r-h88q6.mp4)
- [x] Enable cart note, the drag corner get's lost [when corner radius is on](https://screenshot.click/21-12-1mrvl-cxqnj.png)
    - [x] Let's remove the corner radius from the bottom-right

sofiamatulis:

Fixes #993 .

Added missing styles to country/language selector, quantity, cart notification and share block

**Other considerations**

Because we removed the corner radius on the bottom right from the cart notification, it can sometimes look like this: https://screenshot.click/05-50-wqgpv-vn6wh.png

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127061721110/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1552,8 +1552,8 @@ details[open] > .share-button__fallback {
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
   border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
-  min-height: calc(4rem + (var(--inputs-border-width)) * 2);
-  min-width: calc(7rem + (var(--inputs-border-width)) * 2);
+  min-height: calc(4rem + (var(--inputs-border-width) * 2));
+  min-width: calc(7rem + (var(--inputs-border-width) * 2));
   position: relative;
   border: 0;
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1905,6 +1905,7 @@ input[type='checkbox'] {
   flex-grow: 1;
   -webkit-appearance: none;
   appearance: none;
+  z-index: 2;
 }
 
 .quantity__button {
@@ -1919,18 +1920,7 @@ input[type='checkbox'] {
   justify-content: center;
   color: rgb(var(--color-foreground));
   padding: 0;
-}
-
-.quantity__button:focus-visible,
-.quantity__input:focus-visible {
-  outline-offset: 0;
-  box-shadow: 0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
-}
-
-.quantity__button:focus,
-.quantity__input:focus {
-  outline-offset: 0;
-  box-shadow: 0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+  z-index: 2;
 }
 
 .quantity__button:first-child {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1615,7 +1615,7 @@ details[open] > .share-button__fallback {
 
 .text-area,
 .select {
-  display: inline-block;
+  display: flex;
   position: relative;
   width: 100%;
 }
@@ -1635,7 +1635,9 @@ details[open] > .share-button__fallback {
 .customer select {
   cursor: pointer;
   line-height: calc(1 + 0.6 / var(--font-body-scale));
-  padding: 0 4rem 0 calc(1.5rem + var(--inputs-border-width));
+  padding: 1.5rem;
+  margin: var(--inputs-border-width);
+  min-height: 4rem;
 }
 
 /* Field */
@@ -1648,7 +1650,7 @@ details[open] > .share-button__fallback {
 }
 
 .customer .field {
-  display: block;
+  display: flex;
 }
 
 .field--with-error {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1921,6 +1921,18 @@ input[type='checkbox'] {
   padding: 0;
 }
 
+.quantity__button:focus-visible,
+.quantity__input:focus-visible {
+  outline-offset: 0;
+  box-shadow: 0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
+
+.quantity__button:focus,
+.quantity__input:focus {
+  outline-offset: 0;
+  box-shadow: 0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
+
 .quantity__button:first-child {
   margin-left: calc(var(--inputs-border-width));
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1531,9 +1531,9 @@ details[open] > .share-button__fallback {
   width: 100%;
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
-  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
-  min-height: calc(4rem + var(--inputs-border-width) * 2);
-  min-width: calc(7rem + var(--inputs-border-width) * 2);
+  border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
+  min-height: calc(4rem + (var(--inputs-border-width)) * 2);
+  min-width: calc(7rem + (var(--inputs-border-width)) * 2);
   position: relative;
   border: 0;
 }
@@ -1546,7 +1546,7 @@ details[open] > .share-button__fallback {
   content: '';
   position: absolute;
   inset: 0;
-  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
+  border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
   box-shadow: var(--inputs-shadow-horizontal-offset) var(--inputs-shadow-vertical-offset) var(--inputs-shadow-blur-radius) rgba(var(--color-base-text), var(--inputs-shadow-opacity));
 }
 
@@ -1563,10 +1563,6 @@ details[open] > .share-button__fallback {
   box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
 }
 
-.field-text-area:after {
-  border-bottom-right-radius: 0;
-}
-
 .select__select {
   font-family: var(--font-body-family);
   font-style: var(--font-body-style);
@@ -1581,6 +1577,7 @@ details[open] > .share-button__fallback {
 .customer select:hover,
 .localization-form__select:hover {
   box-shadow: 0 0 0 0.3rem rgba(var(--color-foreground), var(--inputs-border-opacity));
+  border-radius: var(--inputs-radius);
 }
 
 .field__input:focus,
@@ -1677,9 +1674,9 @@ details[open] > .share-button__fallback {
   padding: 2.5rem calc(var(--inputs-border-width) + 1.5rem) 0.8rem;
 }
 
-.field__input.text-area:focus,
-.field__input.text-area:not(:placeholder-shown),
-.field__input.text-area:-webkit-autofill {
+.field:not(.cart__note) .field__input.text-area:focus,
+.field:not(.cart__note) .field__input.text-area:not(:placeholder-shown),
+.field:not(.cart__note) .field__input.text-area:-webkit-autofill {
   padding-top: 4rem;
 }
 
--- a/assets/base.css
+++ b/assets/base.css
@@ -1548,7 +1548,7 @@ details[open] > .share-button__fallback {
   background-color: rgb(var(--color-background));
   color: rgb(var(--color-foreground));
   font-size: 1.6rem;
-  width: -webkit-fill-available;
+  width: 100%;
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
   border-radius: var(--inputs-radius-outset);
--- a/assets/base.css
+++ b/assets/base.css
@@ -1552,8 +1552,8 @@ details[open] > .share-button__fallback {
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
   border-radius: var(--inputs-radius);
-  min-height: calc(var(--inputs-border-width) * 2);
   height: 4.5rem;
+  min-height: calc(var(--inputs-border-width) * 2);
   min-width: calc(7rem + (var(--inputs-border-width) * 2));
   position: relative;
   border: 0;
@@ -1643,9 +1643,9 @@ details[open] > .share-button__fallback {
 .customer select {
   cursor: pointer;
   line-height: calc(1 + 0.6 / var(--font-body-scale));
-  padding: 1.5rem;
+  padding: 0 1.5rem;
   margin: var(--inputs-border-width);
-  min-height: 4rem;
+  min-height: calc(var(--inputs-border-width) * 2);
 }
 
 /* Field */
--- a/assets/base.css
+++ b/assets/base.css
@@ -1695,7 +1695,7 @@ details[open] > .share-button__fallback {
 .customer .field input:focus,
 .customer .field input:not(:placeholder-shown),
 .customer .field input:-webkit-autofill {
-  padding: calc(var(--inputs-border-width) + 1.5rem);
+  padding: calc(var(--inputs-border-width) + 2.2rem) 1.5rem calc(var(--inputs-border-width) + 0.8rem);
 }
 
 .field:not(.cart__note) .field__input.text-area:focus,
--- a/assets/base.css
+++ b/assets/base.css
@@ -1671,7 +1671,13 @@ details[open] > .share-button__fallback {
 .customer .field input:focus,
 .customer .field input:not(:placeholder-shown),
 .customer .field input:-webkit-autofill {
-  padding:  2.5rem 1.5rem 0.8rem;
+  padding: 2.5rem calc(var(--inputs-border-width) + 1.5rem) 0.8rem;
+}
+
+.field__input.text-area:focus,
+.field__input.text-area:not(:placeholder-shown),
+.field__input.text-area:-webkit-autofill {
+  padding-top: 4rem;
 }
 
 .field__input::-webkit-search-cancel-button,
--- a/assets/base.css
+++ b/assets/base.css
@@ -1921,23 +1921,6 @@ input[type='checkbox'] {
   padding: 0;
 }
 
-.quantity__button:focus-visible,
-.quantity__input:focus-visible {
-  background-color: rgb(var(--color-background));
-  z-index: 2;
-}
-
-.quantity__input:focus-visible :after  {
-  background-color: rgb(var(--color-background));
-}
-
-.quantity__button:focus,
-.quantity__input:focus {
-  background-color: rgb(var(--color-background));
-  z-index: 2;
-}
-
-
 .quantity__button:first-child {
   margin-left: calc(var(--inputs-border-width));
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1548,11 +1548,11 @@ details[open] > .share-button__fallback {
   background-color: rgb(var(--color-background));
   color: rgb(var(--color-foreground));
   font-size: 1.6rem;
-  width: 100%;
+  width: -webkit-fill-available;
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
   border-radius: var(--inputs-radius-outset);
-  min-height: calc(4rem + (var(--inputs-border-width) * 2));
+  min-height: calc(var(--inputs-border-width) * 2);
   min-width: calc(7rem + (var(--inputs-border-width) * 2));
   position: relative;
   border: 0;
@@ -1659,7 +1659,8 @@ details[open] > .share-button__fallback {
 .customer .field input {
   flex-grow: 1;
   text-align: left;
-  padding: calc(var(--inputs-border-width) + 1.5rem);
+  padding: 1.5rem;
+  margin: var(--inputs-border-width);
   transition: box-shadow var(--duration-short) ease;
 }
 
--- a/assets/base.css
+++ b/assets/base.css
@@ -1574,7 +1574,8 @@ details[open] > .share-button__fallback {
 }
 
 .field:hover.field:after,
-.select__select:hover.select:after,
+.select:hover.select:after,
+.select__select:hover.select__select:after,
 .customer .field:hover.field:after,
 .customer select:hover.select:after,
 .localization-form__select:hover.localization-form__select:after {
@@ -1614,7 +1615,7 @@ details[open] > .share-button__fallback {
 .customer select {
   cursor: pointer;
   line-height: calc(1 + 0.6 / var(--font-body-scale));
-  padding: 0 4rem 0 1.5rem;
+  padding: 0 4rem 0 calc(1.5rem + var(--inputs-border-width));
 }
 
 /* Field */
--- a/assets/base.css
+++ b/assets/base.css
@@ -2309,6 +2309,7 @@ details[open] .modal-overlay::after {
 .search-modal {
   opacity: 0;
   border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.08);
+  min-height: calc(100% + var(--inputs-margin-offset) + (2 * var(--inputs-border-width)));
   height: 100%;
 }
 
@@ -2320,6 +2321,15 @@ details[open] .modal-overlay::after {
   height: 100%;
   padding: 0 5rem 0 1rem;
   line-height: calc(1 + 0.8 / var(--font-body-scale));
+  position: relative;
+}
+
+.search-modal__content-bottom {
+  bottom: calc((var(--inputs-margin-offset) / 2) );
+}
+
+.search-modal__content-top {
+  top: calc((var(--inputs-margin-offset) / 2) );
 }
 
 .search-modal__form {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1553,7 +1553,8 @@ details[open] > .share-button__fallback {
 .field:after,
 .select:after,
 .customer .field:after,
-.customer select:after {
+.customer select:after,
+.localization-form__select:after {
   pointer-events: none;
   content: '';
   position: absolute;
@@ -1561,6 +1562,7 @@ details[open] > .share-button__fallback {
   border: 0.1rem solid transparent;
   border-radius: var(--inputs-radius);
   box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  transition: box-shadow var(--duration-short) ease;
 }
 
 .select__select {
@@ -1571,12 +1573,12 @@ details[open] > .share-button__fallback {
   color: rgba(var(--color-foreground), 0.75);
 }
 
-.field:hover,
-.select__select:hover,
-.customer .field:hover,
-.customer select:hover,
-.localization-form__select:hover {
-  box-shadow: 0 0 0 0.3rem rgba(var(--color-foreground), var(--inputs-border-opacity));
+.field:hover.field:after,
+.select__select:hover.select:after,
+.customer .field:hover.field:after,
+.customer select:hover.select:after,
+.localization-form__select:hover.localization-form__select:after {
+  box-shadow: 0 0 0 calc(0.3rem + var(--inputs-border-width)) rgba(var(--color-foreground),var(--inputs-border-opacity));
   border-radius: var(--inputs-radius);
 }
 
@@ -1637,6 +1639,7 @@ details[open] > .share-button__fallback {
   flex-grow: 1;
   text-align: left;
   padding: 1.5rem;
+  transition: box-shadow var(--duration-short) ease;
 }
 
 .field__label,
--- a/assets/base.css
+++ b/assets/base.css
@@ -1936,6 +1936,16 @@ input[type='checkbox'] {
   pointer-events: none;
 }
 
+.quantity__button:focus-visible,
+.quantity__input:focus-visible {
+  background-color: rgb(var(--color-background));
+}
+
+.quantity__button:focus,
+.quantity__input:focus {
+  background-color: rgb(var(--color-background));
+}
+
 .quantity__input:-webkit-autofill,
 .quantity__input:-webkit-autofill:hover,
 .quantity__input:-webkit-autofill:active {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1575,12 +1575,12 @@ details[open] > .share-button__fallback {
   color: rgba(var(--color-foreground), 0.75);
 }
 
-.field__input:hover,
+.field:hover,
 .select__select:hover,
-.customer .field input:hover,
+.customer .field:hover,
 .customer select:hover,
 .localization-form__select:hover {
-  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 1px) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  box-shadow: 0 0 0 0.3rem rgba(var(--color-foreground), var(--inputs-border-opacity));
 }
 
 .field__input:focus,
@@ -1588,8 +1588,9 @@ details[open] > .share-button__fallback {
 .customer .field input:focus,
 .customer select:focus,
 .localization-form__select:focus {
-  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 2px) rgba(var(--color-foreground), var(--inputs-border-opacity));
-  outline: transparent;
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: 0.3rem;
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
 }
 
 .text-area,
@@ -1623,6 +1624,7 @@ details[open] > .share-button__fallback {
   position: relative;
   width: 100%;
   display: flex;
+  transition: box-shadow var(--duration-short) ease;
 }
 
 .customer .field {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1526,26 +1526,43 @@ details[open] > .share-button__fallback {
   -webkit-appearance: none;
   appearance: none;
   background-color: rgb(var(--color-background));
-  border: 0.1rem solid transparent;
-  border-radius: var(--inputs-radius);
   color: rgb(var(--color-foreground));
   font-size: 1.6rem;
   width: 100%;
-  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
-  height: 4.5rem;
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
-  filter: drop-shadow(
-    var(--inputs-shadow-horizontal-offset)
-    var(--inputs-shadow-vertical-offset)
-    var(--inputs-shadow-blur-radius)
-    rgba(var(--color-shadow), var(--inputs-shadow-opacity))
-  );
-  /* fix drop-shadow issue in Safari */
-  transform: translateZ(0);
+  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
+  min-height: calc(4rem + var(--inputs-border-width) * 2);
+  min-width: calc(7rem + var(--inputs-border-width) * 2);
+  position: relative;
 }
 
-.field__input.text-area {
+.field:before,
+.select__select:before,
+.customer .field:before,
+.customer select:before {
+  pointer-events: none;
+  content: '';
+  position: absolute;
+  inset: 0;
+  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
+  box-shadow: var(--inputs-shadow-horizontal-offset) var(--inputs-shadow-vertical-offset) var(--inputs-shadow-blur-radius) rgba(var(--color-base-text), var(--inputs-shadow-opacity));
+}
+
+.field:after,
+.select__select:after,
+.customer .field:after,
+.customer select:after {
+  pointer-events: none;
+  content: '';
+  position: absolute;
+  inset: var(--inputs-border-width);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+}
+
+.field-text-area:after {
   border-bottom-right-radius: 0;
 }
 
@@ -1625,8 +1642,8 @@ details[open] > .share-button__fallback {
 .field__label,
 .customer .field label {
   font-size: 1.6rem;
-  left: 1.5rem;
-  top: 1rem;
+  left: calc(var(--inputs-border-width) + 1.5rem);
+  top: calc(1rem + var(--inputs-border-width));
   margin-bottom: 0;
   pointer-events: none;
   position: absolute;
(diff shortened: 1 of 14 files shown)


assets/base.css
```diff
@@ -1445,7 +1445,15 @@ details[open] > .share-button__fallback {
   z-index: 3;
```
sofiamatulis, line 1445: Because we removed the corner radius on the bottom right from the cart notification, it can sometimes look like this: https://screenshot.click/05-50-wqgpv-vn6wh.png cc @melissaperreault 

I will let you test and see what you think of it with the different settings 🤔

melissaperreault: Yes, this was expected to keep visible and usable the drag handle UI. Thanks!

melissaperreault (commented):
Thank you Sofia! 🎉 

Some small observations:

- [ ] Let's check with @kmeleta , I recall we had a fix to space out siblings when [border thickness is increased](https://screenshot.click/22-01-9z4un-979i4.png). Might have to do with this merged PR https://github.com/Shopify/dawn/pull/1064. Might need to do some more investigation for spacing out labels for border thickness. I know we intend to have a setting to space out siblings for improving the shadow offset sometimes, but unsure if that could become a systematic change. There are some strong legibility issues otherwise. cc. @nicklepine 
- [ ] Share block: Let's pair, I am not sure what would be the ideal solution [to simplify the focus behaviour here.](https://screenshot.click/22-01-lyan8-qv2eq.mp4)
- [x] Quantity input: As we are scaling the border thickness, [we are losing the input number](https://screenshot.click/22-01-l078m-brnjt.mp4) and the delete icon is no longer centered vertically with the component.

assets/base.css
```diff
@@ -1761,8 +1778,7 @@ input[type='checkbox'] {
   color: rgba(var(--color-foreground));
   border-radius: var(--inputs-radius);
   position: relative;
-  height: 4.5rem;
-  width: calc(14rem / var(--font-body-scale));
+  width: calc(16rem / var(--font-body-scale));
```
sofiamatulis, line 1781: @melissaperreault To avoid this: https://screenshot.click/22-01-l078m-brnjt.mp4 , I updated the calculation to use 16 instead of 14. I see this was updated in @KaichenWang 's PR here: https://github.com/Shopify/dawn/commit/fe3de560f4c3fef7e35e18e0d9fc1515cdd99d53#diff-00b03bd62b155daa89082ba52094c182e46422a9e28cd3ec601b3a2291b52718R1646

Kai, since you worked on this change do you mind confirming?

assets/section-footer.css
```diff
@@ -271,7 +271,7 @@
   .localization-form:only-child {
     justify-content: start;
     width: auto;
-    margin: 0 0 0 -1rem;
+    margin: 0 0 0 var(--inputs-border-width);
```
sofiamatulis, line 274: My idea for the borders overlapping would be to use the variable and add it as margin, but like @kmeleta  mentioned, maybe this can be tackled on a separate PR

https://github.com/Shopify/dawn/pull/1097/commits/e8a5780c64f9f9349f7c4a5c5fd4a211a8f97897

On an unrelated note, @melissaperreault I saw that the range the minimum for thickness is 0 where I am thinking it should be 1 so we can always have a border 🤔 

https://screenshot.click/06-15-hebi2-a3jwl.mp4

kmeleta: Solving the border issue with margin may be ok for inputs considering the limited number of places where the issue may exist. For buttons we'd have to add that variable to existing margin everywhere buttons are or may be used and I didn't like that.

The other thing to consider though is that if we want to prevent the animation from affecting the shadow on inputs (which I assume we do) we'll need to change our approach to input border anyway.

assets/base.css
```diff
@@ -1445,7 +1445,15 @@ details[open] > .share-button__fallback {
...
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.55);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  filter: drop-shadow(
+    var(--inputs-shadow-horizontal-offset)
+    var(--inputs-shadow-vertical-offset)
+    var(--inputs-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--inputs-shadow-opacity))
+  );
```
ludoboludo, line 1456: I wonder if we should be adding some kinda margin-top based on the border width value here so that the overlap isn't really an issue 🤔 
We could potentially take in consideration the vertical offset of the shadow too.

ludoboludo (commented):
Noted a few things but not sure if you're in the middle of changing things or not. There was a changed pushed while I was reviewing. 

Here is what I noted: 
- focus styles are a bit off atm: [on locales](https://screenshot.click/22-01-03kt1-iuf9o.png) and [other inputs](https://screenshot.click/22-01-vw8o9-9qsvf.png)
- cart note is now looking a bit funky: [screenshot](https://screenshot.click/22-01-ubrvt-3avuu.png)
- locales are centered but it looks off due to the caret icon. Maybe we can edit the `padding-left` value: [screenshot](https://screenshot.click/22-01-2vsne-l85tm.png)
- There can be overlaps on variants when their layout is a dropdown. [Screenshot](https://screenshot.click/22-01-ldqmj-orw7d.png). Tho with the latest changes it seems like it's changing their size and not applying borders properly: [screenshot](https://screenshot.click/22-01-1x8e8-bivx6.png)
- The contact form section seem to miss the bottom right corner radius. It doesn't like it needs to be applied there 🤔

melissaperreault:
> The contact form section seem to miss the bottom right corner radius. It doesn't like it needs to be applied there

True it should only apply to `Cart note`, not `Comment` component.

sofiamatulis:
@melissaperreault 

Do you mind taking a look at the hover states/focus states and confirming the behaviour when we have thicker borders? 🙏

assets/base.css
```diff
@@ -1534,21 +1567,23 @@ details[open] > .share-button__fallback {
...
+.field:hover,
 .select__select:hover,
-.customer .field input:hover,
+.customer .field:hover,
 .customer select:hover,
 .localization-form__select:hover {
-  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 1px) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  box-shadow: 0 0 0 0.3rem rgba(var(--color-foreground), var(--inputs-border-opacity));
+  border-radius: var(--inputs-radius);
 }
```
ludoboludo, line 1577: I'm finding the hover to look a bit off compared to the border of the element: [video](https://screenshot.click/22-01-xqr3i-wm4fs.mp4)

Also there is a transition I can see to apply the border radius to it: [video](https://screenshot.click/22-01-axajl-h8azh.mp4)

assets/base.css
```diff
@@ -1509,21 +1522,41 @@ details[open] > .share-button__fallback {
...
+.select:before,
+.customer .field:before,
+.customer select:before {
+  pointer-events: none;
+  content: '';
+  position: absolute;
+  inset: 0;
+  border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
+  box-shadow: var(--inputs-shadow-horizontal-offset) var(--inputs-shadow-vertical-offset) var(--inputs-shadow-blur-radius) rgba(var(--color-base-text), var(--inputs-shadow-opacity));
+}
```
ludoboludo, line 1547: This is being applied on the sort input for facet filter but i don't think it should: 



So it probably need an overwrite. I had done one but since things have a changed a bit it's not doing anything anymore I believe.

assets/component-cart.css
```diff
@@ -87,19 +87,26 @@ cart-items {
...
 
+.cart__note:after {
+  border-bottom-right-radius: 0;
+}
+
 .cart__note .field__input {
-  padding: 1rem;
+  padding: calc((var(--inputs-border-width)) + 0.5rem);
+  height: 100%;
 }
```
ludoboludo, line 110: The border now seem to cover the bottom right icon.

assets/component-cart.css
```diff
@@ -87,19 +87,26 @@ cart-items {
 
 .cart__note {
   display: block;
+  height: 100%;
 }
```
ludoboludo, line 91: When trying to type in a message in the cart note the styles are a bit off:

assets/section-main-product.css
```diff
@@ -240,6 +240,7 @@ fieldset.product-form__input .form__label {
 
 .product-form__input .select {
   max-width: 25rem;
+  padding-left: var(--inputs-border-width);
 }
```
ludoboludo, line 244: On product selects, the styling on hover seems off: [video](https://screenshot.click/22-01-9m3ew-q0xju.mp4)

ludoboludo (commented):
Noticed a few oddities. 

Here are the settings I had while testing: [screenshot](https://screenshot.click/22-01-2t0qf-aqaed.png)

assets/base.css
sofiamatulis: I will be pairing with Melissa on the hover/focus states to see what makes the most sense 👍

assets/base.css
melissaperreault: We decided not to style those select because of the relationship with the filters not being styled.

assets/base.css
```diff
@@ -1509,21 +1542,43 @@ details[open] > .share-button__fallback {
...
   transition: box-shadow var(--duration-short) ease;
-  filter: drop-shadow(
-    var(--inputs-shadow-horizontal-offset)
-    var(--inputs-shadow-vertical-offset)
-    var(--inputs-shadow-blur-radius)
-    rgba(var(--color-shadow), var(--inputs-shadow-opacity))
-  );
+  border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
+  min-height: calc(4rem + (var(--inputs-border-width)) * 2);
+  min-width: calc(7rem + (var(--inputs-border-width)) * 2);
```
kmeleta, line 1552: Parens in the wrong place but I think we can exclude them anyway.

```suggestion
  min-height: calc(4rem + (var(--inputs-border-width) * 2));
  min-width: calc(7rem + (var(--inputs-border-width) * 2));
```

assets/base.css
```diff
@@ -1509,21 +1542,43 @@ details[open] > .share-button__fallback {
...
+
+.field:before,
+.select:before,
+.customer .field:before,
+.customer select:before {
+  pointer-events: none;
+  content: '';
+  position: absolute;
+  inset: 0;
+  border-radius: calc((var(--inputs-border-width)) + (var(--inputs-radius)));
```
kmeleta, line 1565:
```suggestion
  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
```

assets/section-footer.css
```diff
@@ -302,12 +301,34 @@
...
+
+.disclosure__button.localization-form__select {
+  padding: calc(1.5rem + var(--inputs-border-width))
+}
+
+.localization-form__select:before {
+  content: '';
+  position: absolute;
+  inset: 0;
+  border-radius: calc(var(--inputs-border-width) + var(--inputs-radius));
```
kmeleta, line 321: I just noticed that this technique can lead to unnecessary radius even when it's 0 (because of the border value).
  
 
I know this isn't going to be unique to inputs though so we don't need to deal with it here, but we'll have to decide how we want to handle that. Something contrived like `calc(var(--inputs-radius) * var(--inputs-radius) + var(--inputs-border-width) + var(--inputs-radius))` could work inline, but we may consider adding additional global vars like `--inputs-radius-outset: {% if settings.radius > 0 %} radius + border_width {% else %) 0 {%endif%}`

sofiamatulis: Good catch!

ludoboludo (commented):
Looking good, added a few comments. 

Also on the contact form section, the comment part changes in heigh when I click on it to edit it: [video](https://screenshot.click/22-01-mkp96-4asfl.mp4)

assets/base.css
```diff
@@ -1509,21 +1542,43 @@ details[open] > .share-button__fallback {
...
+.localization-form__select:after {
+  pointer-events: none;
+  content: '';
+  position: absolute;
+  inset: var(--inputs-border-width);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  transition: box-shadow var(--duration-short) ease;
 }
```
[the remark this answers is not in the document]
tyleralsbury: Do these inputs have a transparent background?  If so, I think giving them the same `background-color` as their parent container might solve this.  If not...

@kmeleta this overlap with the shadows looks similar to the stuff you were dealing with.  Can you check and if so, help @sofiamatulis implement something similar to what you did for all of the grids?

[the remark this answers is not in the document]
tyleralsbury: For the spacing, I assume it should take the horizontal grid spacing property, is that right @melissaperreault ?

assets/section-footer.css
```diff
@@ -302,12 +301,34 @@
...
+  box-shadow: var(--inputs-shadow-horizontal-offset) var(--inputs-shadow-vertical-offset) var(--inputs-shadow-blur-radius) rgba(var(--color-base-text), var(--inputs-shadow-opacity));
+}
+
+.localization-form__select:after {
+  content: '';
+  position: absolute;
+  inset: var(--inputs-border-width);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground),var(--inputs-border-opacity));
```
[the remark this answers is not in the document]
tyleralsbury: I would consider this to be somewhat low severity since it's specifically on focus and is just a small visual bug.  Can we create a Severity 2 or 3 issue to deal with this? @melissaperreault

assets/component-cart.css
```diff
@@ -86,20 +86,26 @@ cart-items {
...
 
 .cart__note {
-  display: block;
+  height: fit-content;
```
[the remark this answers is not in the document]
tyleralsbury: Maybe even like `0.3rem` extra just so it's not stuck to the label, I wouldn't add much..

assets/base.css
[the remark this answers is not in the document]
kmeleta: I think just giving the shadow pseudo element (`.field:before` I believe) a `z-index: -1` will fix the shadow overlap. Just make sure the customer selectors get the same style.

assets/base.css
```diff
@@ -1631,7 +1691,13 @@ details[open] > .share-button__fallback {
 .customer .field input:focus,
 .customer .field input:not(:placeholder-shown),
 .customer .field input:-webkit-autofill {
-  padding: 2.2rem 1.5rem 0.8rem;
+  padding: calc(var(--inputs-border-width) + 2.2rem) 1.5rem calc(var(--inputs-border-width) + 0.8rem);
```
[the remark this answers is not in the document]
sofiamatulis: I have added some margin depending on the width of the border 👍

assets/base.css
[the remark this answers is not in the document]
sofiamatulis: Because of the position relative and before and after, padding wasn't working anymore to separate the elements, so I used margin instead

assets/base.css
```diff
@@ -1534,26 +1590,33 @@ details[open] > .share-button__fallback {
...
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: calc(0.3em + var(--inputs-border-width));
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
+
+.localization-form__select:focus {
+  outline-offset: 0.3em;
 }
 
 .text-area,
```
ludoboludo, line 1617: On the contact form it seems like the "comment" part is a little too close to the border compared to the other fields.

assets/section-footer.css
ludoboludo, line 331: When using the max value for horizontal offset (negative value), the shadow is coming up over the previous summary. 

Not a big issue, just noting it.

ludoboludo (commented):
Feels like it's almost good everywhere! 

On the password page something isn't quite right with the shadows: [screenshot](https://screenshot.click/22-01-6b0dy-ghcns.png), [screenshot 2](https://screenshot.click/22-01-zbk92-njb2j.png).

tyleralsbury:
> Feels like it's almost good everywhere!
> 
> On the password page something isn't quite right with the shadows: [screenshot](https://screenshot.click/22-01-6b0dy-ghcns.png), [screenshot 2](https://screenshot.click/22-01-zbk92-njb2j.png).

I think that's fixed here: https://github.com/Shopify/dawn/pull/1091

assets/base.css
```diff
@@ -1509,21 +1542,44 @@ details[open] > .share-button__fallback {
...
   box-sizing: border-box;
   transition: box-shadow var(--duration-short) ease;
-  filter: drop-shadow(
-    var(--inputs-shadow-horizontal-offset)
-    var(--inputs-shadow-vertical-offset)
-    var(--inputs-shadow-blur-radius)
-    rgba(var(--color-shadow), var(--inputs-shadow-opacity))
-  );
+  border-radius: var(--inputs-radius);
+  min-height: calc(var(--inputs-border-width) * 2);
```
kmeleta, line 1551: I still think the input height isn't quite right. Currently on main, inputs and buttons are 45px tall by default with no additional border. With these changes we're sitting at 54px. This is likely why we needed to adjust the vertical alignment of the search icon. We may just need to reduce the 1.5rem padding a bit. I'd expect the min-height here to be calculated to 45px.

assets/base.css
```diff
@@ -1534,26 +1590,33 @@ details[open] > .share-button__fallback {
...
 .customer .field input:focus,
 .customer select:focus,
 .localization-form__select:focus {
-  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 2px) rgba(var(--color-foreground), var(--inputs-border-opacity));
-  outline: transparent;
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: calc(0.3em + var(--inputs-border-width));
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
+
```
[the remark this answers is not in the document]
ludoboludo: actually disregard this. As Scott mentioned elsewhere: 
>"For any area where it's assumed the user is going to provide keyboard input, the browser will draw a focus indicator."

sofiamatulis:
> ## UX Update
> * [ ]  [Providing the most recent guidance](https://screenshot.click/22-01-e257w-4zn6b.png) for hover and focus experience with the global settings in mind.
>   
>   * Hover: Increase the border thickness
>   * Focus: Increase border thickness + opacity is at 100%

@melissaperreault This means that hover and focus would have the same styles because the merchant could set the border opacity to already be 100%, so if we change it to 100%, it would look the same on focus. In other words, when the user hovers or focuses the only change would be the border thickness.

This is the current experience for visibility/context: 

11-02-8439s-oflnm

assets/base.css
```diff
@@ -1534,26 +1590,33 @@ details[open] > .share-button__fallback {
...
 .select__select:focus,
 .customer .field input:focus,
 .customer select:focus,
 .localization-form__select:focus {
-  box-shadow: 0 0 0 calc(var(--inputs-border-width) + 2px) rgba(var(--color-foreground), var(--inputs-border-opacity));
-  outline: transparent;
+  outline: 0.2rem solid rgba(var(--color-foreground),.5);
+  outline-offset: calc(0.3em + var(--inputs-border-width));
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground),.3);
+}
```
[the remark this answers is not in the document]
sofiamatulis: This is also an issue when we have the shadow be at the top. @melissaperreault What do you think?
https://screenshot.click/11-05-4599u-gs9bi.png

I will make the update once the focus/hover states are finalized. Context: https://github.com/Shopify/dawn/pull/1097#issuecomment-1010271537

assets/base.css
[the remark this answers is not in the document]
sofiamatulis: @ludoboludo Do you mind re-sharing the video with me? 🤔 For some reason it wont load for me

assets/base.css
sofiamatulis: The height is bigger because now we possibly have thicker borders. If I keep the height always the same, because the borders grow "inwards", you end up with less space for the content which isnt desired.

Screen Shot 2022-01-11 at 2 11 20 PM

Thinner borders just means we have less margin:
Screen Shot 2022-01-11 at 2 11 20 PM

But youll notice the height and padding remain the same

kmeleta: Yeah we can't reduce the inner height based on the border, but with only 1px of border the inputs height is still 54px now, as opposed to 45. Reducing this "base" height back down to where it was may also give [this issue](https://camo.githubusercontent.com/e673e7770868e3fbce2be1a2cf30b18a2fc291218c5f7fa51a2f179d8d23fc2e/68747470733a2f2f73637265656e73686f742e636c69636b2f32322d30312d6b757a31692d79717878332e706e67) more room

kmeleta: This is what I mean

melissaperreault:
> This means that hover and focus would have the same styles because the merchant could set the border

The difference in that case is the label inside the input that changes, on hover the label don't move at the top as oppose to when it receives focus. This was one of the tradeoff we decided @nicklepine and I.

assets/base.css
sofiamatulis: Good catch, Ken! This is now fixed

sofiamatulis:
> * Focus

@melissaperreault 
Sounds good 👍 Updated! Note that the label doesnt get smaller for select dropdowns like localization selector. You can see what it looks like on my store

assets/base.css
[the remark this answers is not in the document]
sofiamatulis: This is now fixed, but I noticed the shadow issue above 🤔

assets/section-footer.css
sofiamatulis: I didnt have z index in the `after` and `before` and that his was done for the button PR. I added that and it worked great for all inputs except for the newsletter one (I had to explicitly set the z-index to 0 for the button) and the overlapping is still happening for the localization selectors 🤔 . 

https://github.com/Shopify/dawn/pull/1097/commits/226ef8f74e49019a8c6e664133ac8c9b68c1d24b 

11-34-mqkl6-xz1ex

kmeleta: This is tricky so let me know if you want to pair.. Setting that z-index: 0 on the button does break the relative z context and will prevent this approach from working, but obviously, if you do that, the shadows disappear behind the footer bg. If you were to add a `z-index: 1` and `position relative` to some div that is a child of `.footer.color-background-1` but still a parent of the localization forms, it should fix that. Though the buttons seem to have `background: transparent` too so the shadow will still appear show through. So to fix _that_ we can change that transparent to `var(--color-background)` or whatever the footer gets.

melissaperreault (commented):
Thanks Sofia!
Here are some observations:

- [ ] Have the [shadow appear underneath](https://screenshot.click/22-01-9tasq-jnaae.png) like we did for Buttons with @kmeleta 
- [ ] We can see through the hover/focus [border increase here](https://screenshot.click/22-01-3pu4m-rv7v2.png)
- [ ] Something is off on the email sign up input when you [compare it with the Quantity input focus](https://screenshot.click/22-01-tr5ty-5trib.mp4).
   - [ ] Related to that, when we add corner radius, the outcome is not cute  for the quantity input and others https://screenshot.click/22-01-cy1um-hp62x.png. We see double lines.
   - [ ] Also related to that same screenshot, we can [see the focus on the shadow behind](https://screenshot.click/22-01-cy1um-hp62x.png)
- [ ] Can we have the [shadow not cut and appear on top of the line](https://screenshot.click/22-01-b5tvw-ffjdz.png) on the search input header? There might be close relationship with the text but this is already the case on other areas. 
   - [ ] Related to that, [could we add padding](https://screenshot.click/22-01-8pwyy-l8980.png) as we scale the border to always have some safe space top and bottom the input? Similar output [like this](https://screenshot.click/22-01-zjeui-jwl7o.png)?

melissaperreault:
Adding other observations:

- Share block
   - [ ] Let's make sure label position are maintained like before ([good example](https://screenshot.click/22-01-d6ajl-k9grh.png)), what I see [on the demo test store](https://screenshot.click/22-01-bg7on-9iq48.png).
   - [ ] Let's apply a consistent focus behavior, we can pair to see what works best. I was wondering if we could apply the style from the icon focus to the input as an exception. ([Video reference](https://screenshot.click/22-01-883vs-37jkb.mp4)) 
- [ ] Their seems to be an additional unwanted grey border on top of the black border, [example here](https://screenshot.click/22-01-t2mmy-o407d.png)

kmeleta:
> > * [ ]   We can see through the hover/focus [border increase here](https://screenshot.click/22-01-3pu4m-rv7v2.png)
> 
> @melissaperreault I believe this is expected 🤔 Because if you update the border opacity to be less, it means you can see through it and see the shadow. But youll see that if I set the opacity to be 100% then you wouldnt see it anymore. Does that make sense? Quick video: https://screenshot.click/12-51-cip0p-aym3c.mp4
> 
> Video of the current behaviour on our `main` branch: https://screenshot.click/12-04-berlz-08qbq.mp4

So I did solve for the growing border opacity issue with buttons, though it was annoying ([see here](https://screenshot.click/12-15-naxey-vyv6j.mp4)). I do think the overlap is expected from a technical pov, but I don't think that's ideally what we wanted creatively. If we decide the opacity overlap is acceptable I'd want to make sure we update buttons to behave the same. But basically I overlaid the partially transparent shadow color on top of a fully opaque background that matches the section background to prevent the shadow from showing through the border.

cc @sofiamatulis @melissaperreault

tyleralsbury:
> 

I'm wondering if this is something we can consider properly in a separate, Severity 2 issue.  To give ourselves proper time to evaluate the desired effect and consider the complexity of the implementation.  We can also extract the behaviour out to a class or something else to put all of that complexity into one place instead of strewn about the theme.

sofiamatulis:
> * [x]  Something is off on the email sign up input when you [compare it with the Quantity input focus](https://screenshot.click/22-01-tr5ty-5trib.mp4).


@melissaperreault I have update the `focus` to be `0.1rem` instead of `0.3rem` as per your last comment. Here is the commit: https://github.com/Shopify/dawn/pull/1097/commits/0654b27a3cdb4e688204a4df07fbc185b8055352

Let me know if that makes sense

sofiamatulis:
> * [ ]  Let's also remove the focus visible on click on the Sort select form the Collection template please ([Video recording](https://screenshot.click/22-01-gz688-h1a2m.mp4))

@melissaperreault I believe this is actually expected 🤔 You wont see this in the Filter because it's a `<details>` element, but you will see it in the `sort` and in the localization selector because they are `<select>` elements and when you click on them, they are in focus.

Video: https://screenshot.click/12-48-yk5p4-clmrf.mp4

cc @svinkle

assets/component-cart.css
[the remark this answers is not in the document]
ludoboludo: I think this could maybe be tackled here still ?

assets/base.css
```diff
@@ -1540,26 +1612,35 @@ details[open] > .share-button__fallback {
...
 
 .field__input:focus,
 .select__select:focus,
 .customer .field input:focus,
 .customer select:focus,
+.localization-form__select:focus.localization-form__select:after {
+  box-shadow: 0 0 0 calc(0.1rem + var(--inputs-border-width)) rgba(var(--color-foreground));
+  outline: 0;
+  border-radius: var(--inputs-radius);
+}
```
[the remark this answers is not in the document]
sofiamatulis: cc @melissaperreault @svinkle

assets/component-cart.css
```diff
@@ -110,11 +110,13 @@ cart-items {
...
+.cart__note:after, .cart__note:hover.cart__note:after,
+.cart__note:before, .cart__note:hover.cart__note:before {
   border-bottom-right-radius: 0;
 }
 
-.cart__note .field__input:focus {
+.cart__note .field__input:focus,
+.cart__note .field__input {
   border-bottom-right-radius: 0;
 }
```
ludoboludo, line 121: looks like this could be all grouped together

assets/component-cart.css
```diff
@@ -86,20 +86,35 @@ cart-items {
...
+  top: -2.3rem;
   color: rgba(var(--color-foreground), 0.75);
 }
 
 .cart__note .field__input {
-  padding: 1rem;
+  height: 100%;
+  position: relative;
+  border-radius: var(--inputs-radius);
+}
```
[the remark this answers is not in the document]
melissaperreault: I recall @martinamarien had a suggestion how to solve the preview of the scrollbar where it was hidden inside the container. I'll let her provide thoughts when she has the time.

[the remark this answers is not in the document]
kmeleta: She can correct me if I'm wrong but I assume Martina's suggestion would be something along the lines of adding a new parent element around the `<textarea>` and applying the border radius stuff and an `overflow: hidden` to that instead. I have no problem recording this as an issue and handling as a follow up if everyone agrees this is probably more sev 2.

sofiamatulis:
**Updates on this PR and what to test:**

- Update of overall focus to be 0.1rem instead of 0.3rem. Here is the commit: https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&key=426fb1f420e906418fa2448926c74e347c9832a999315511ee1be9880525803a
- Overall styles for inputs: newsletter input, footer newsletter input, email sign up banner on password page, search, cart note, customer pages, blog comment, contact form, quantity on product page, share on product page, localization selector
- Focus and hover states for inputs: https://screenshot.click/22-01-e257w-4zn6b.png (note that the filter selectors are not included)
- Code: refactored for inputs and using `after` and `before` instead of `filter` (same approach as button)
- Cart note has no border on the bottom right even when border radius is applied 
- Share block to have the same focus as the button next to it. Video reference: https://screenshot.click/22-01-883vs-37jkb.mp4

assets/base.css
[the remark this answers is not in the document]
svinkle: I understand the concern, and agree; in this instance the focus state may difficult to discern. However, this is still technically accessible as the focus ring contrasts highly against its contrasting area (black on white) an is at least `2px`.

I also understand the "box within a box" concern as `:focus-visible` does not apply to `input` elements, therefore sighted mouse users will see the focus ring, causing the effect.

Since the locale/currency are native `summary`/`button` controls, we should be able to apply the `:focus-visible` styling, removing the "box within a box" affect for mouse/touch users. Native `input`s are the exception as focus styles will always display in these instances.

The inconsistent focus styling seems to be seen most prominently on Sense. The focus styles here are particularly difficult to discern due to low contrast. Perhaps there's an opportunity to revisit focus ring styles?

<details>
  <summary>Sense Demo</summary>

<video src="https://user-images.githubusercontent.com/1392632/149366481-da226a9b-e458-4f21-870c-bb82758c71fe.mp4" aria-label="Sense form showcasing low contrast focus rings." controls>



</details>

assets/component-cart.css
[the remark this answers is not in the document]
sofiamatulis: I am aligned to make this a follow up

assets/component-newsletter.css
```diff
@@ -41,6 +45,16 @@
...
   width: 4.4rem;
   margin: 0;
+  right: var(--inputs-border-width);
+  top: 0;
+  height: 100%;
+}
+
+.newsletter-form__button:focus-visible,
+.newsletter-form__button:focus {
+  background: rgb(var(--color-background));
```
kmeleta, line 55: I don't think safari is getting the same background fix

assets/base.css
```diff
@@ -1449,7 +1449,15 @@ details[open] > .share-button__fallback {
...
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.55);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  filter: drop-shadow(
+    var(--inputs-shadow-horizontal-offset)
+    var(--inputs-shadow-vertical-offset)
+    var(--inputs-shadow-blur-radius)
+    rgba(var(--color-foreground), var(--inputs-shadow-opacity))
+  );
```
kmeleta, line 1460: The shareback button focus gets partially hidden. Adding a z-index on focus could help.

assets/component-newsletter.css
```diff
@@ -50,6 +50,12 @@
   height: 100%;
 }
 
+.newsletter-form__button:focus-visible,
+.newsletter-form__button:focus {
```
ludoboludo, line 54: One thing, and I haven't tested it here but those two selectors need to be separated as Safari won't read the styling for `:focus` because it doesn't support `:focus-visible`.

assets/base.css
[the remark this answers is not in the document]
ludoboludo: Yeah I agree that in some situations, based on the settings it's just hard to tell where the focus is: [video](https://screenshot.click/22-01-o9snn-lqzfh.mp4)

Maybe the box-shadow can be slightly bigger than `0.1rem` 🤔

assets/base.css
ludoboludo: Also, the input could receive an overwrite not to have a radius to the right:

assets/base.css
```diff
@@ -1762,20 +1844,34 @@ input[type='checkbox'] {
...
-    var(--inputs-shadow-horizontal-offset)
-    var(--inputs-shadow-vertical-offset)
-    var(--inputs-shadow-blur-radius)
-    rgba(var(--color-shadow), var(--inputs-shadow-opacity))
-  );
+  border-radius: var(--inputs-radius);
+  min-height: calc((var(--inputs-border-width) * 2) + 4.5rem);
+}
+
+.quantity:after {
```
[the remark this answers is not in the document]
ludoboludo: I was thinking the same:

assets/component-facets.css
```diff
@@ -75,6 +75,21 @@
...
+}
+
+.facet-filters__field .select__select,
+.mobile-facets__sort .select__select {
+  border-radius: 0;
+  min-width: auto;
+  min-height: auto;
+  transition: none;
+}
+
```
ludoboludo, line 92: Should we leave the focus style on it when we click it?
[video](https://screenshot.click/22-01-c6lrm-rpg47.mp4)

ludoboludo (commented):
Feels so close to merge! 

Add a few small comments

assets/component-newsletter.css
sofiamatulis: This has been updated, but I noticed the focus doesnt seem to work on main for the newsletter button either 🤔 https://github.com/Shopify/dawn/pull/1097/commits/23d48c6a5a5404efc097234ff61509bc9c746227

assets/base.css
sofiamatulis: @ludoboludo I believe it makes sense to keep it rounded and affected by the corner radius for consistency 🤔 @melissaperreault what do you think?

assets/base.css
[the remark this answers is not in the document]
sofiamatulis: Both should be fixed now 👍 https://screenshot.click/13-08-4o75n-jilaq.png

But something I noticed is that by adding this fix, we get the focus on click as well since the element is in focus: https://screenshot.click/13-14-e2lms-dn808.mp4 🤔

assets/component-facets.css
sofiamatulis: Yes, this is expected!

assets/base.css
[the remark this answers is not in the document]
kmeleta: You may need something like this as well
```
.newsletter-form__button:not(:focus-visible) {
  box-shadow: inherit;
}
```

assets/base.css
melissaperreault: If we are though, let's also have roundness to the copy icon focus https://screenshot.click/22-01-hn3iy-g6cz5.gif

assets/base.css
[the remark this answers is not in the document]
ludoboludo: For the input to have the same radius, it might be tricky for some elements I think 🤔  but it can be something to look into later.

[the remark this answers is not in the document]
kmeleta: We need to shift our focus outline approach to fully use `box-shadow` instead of `outline` everywhere to fully get radius outlines working in safari regardless (safari's outline doesn't respect radius). So we probably don't want to spend too much time worrying about that kind of detail.

melissaperreault (dismissed):
🎉 Thanks Sofia for all the hard work! I think we can say it is a wrap, let's follow up with polishing task. I'll look around and try to map out a plan for some suggestions that I made  for future behavior improvement in general.

Last observation:

-  [ ] When corner radius is applied, only the search icon focus will display the border inside  the focus as oppose to the others. ([Video example](https://screenshot.click/22-01-xxf5k-nyyh1.mp4))

assets/base.css
```diff
@@ -1449,7 +1449,15 @@ details[open] > .share-button__fallback {
   z-index: 3;
   width: 100%;
   min-width: max-content;
-  box-shadow: 0 0 0 0.1rem rgba(var(--color-foreground), 0.55);
+  border: 0.1rem solid transparent;
+  border-radius: var(--inputs-radius);
+  box-shadow: 0 0 0 var(--inputs-border-width) rgba(var(--color-foreground), var(--inputs-border-opacity));
+  filter: drop-shadow(
```
kmeleta, line 1455: Was there a reason I missed that we didn't update this shadow? If we add it as box-shadow instead of the filter we can probably avoid the weird effects. Though after making that change we may have to move this to a pseudo element so we can properly scale the shadow dimensions to account for thicker borders.
Screen Shot 2022-01-14 at 10 03 03 AM

kmeleta (commented):
One more note 😬

assets/base.css
sofiamatulis: Good catch! Refactored to use the `after` and `before` as well: https://github.com/Shopify/dawn/pull/1097/commits/6a223875d14de4daaf4c978f9fa1a67240b0539f

kmeleta (approved):
Thanks for all the updates Sofia! I think this looks good to me. Let's just make sure all follow ups are accounted for in issues. In the future I'd also like to look at utilizing a shared `.settings-inputs` class to reduce some redundancy for many of these styles.

sofiamatulis:
> Thanks for all the updates Sofia! I think this looks good to me. Let's just make sure all follow ups are accounted for in issues. In the future I'd also like to look at utilizing a shared `.settings-inputs` class to reduce some redundancy for many of these styles.

@kmeleta I agree 👍 I thought of the same thing since some of the styles are shared. I will be handling this in a follow up

ludoboludo (approved):
The only thing I noticed was this: [screenshot](https://screenshot.click/22-01-9aa2e-qkzds.png)

The rest looked great. 

Thanks for all the adjustments along the way 🎉

sofiamatulis:
> screenshot

Created an issue here :) https://github.com/Shopify/dawn/issues/1193

Thank you all for the great and detailed reviews 👍

melissaperreault (approved):

PR #1005, opened 2021-12-10
Add variant pills settings

martinamarien:
**Why are these changes introduced?**

Fixes #982 .

**What approach did you take?**

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #1112, opened 2022-01-06
Content - update button to pills

martinamarien:
This PR updates Variant picker content from Button to Pills. 
This means that product content now matches global setting content.

https://github.com/Shopify/dawn/pull/1005#issuecomment-1006839343

cc. @katycobb


ludoboludo (approved):
Ahah agreed. Pills in french is `comprimés` which kinda sounds odd but also makes sense I think. Most people know the shape of it.

PR #1127, opened 2022-01-07
Adjust grid space to use gap values instead of padding

### LINKED ISSUE #1054: Grid - Vertical spacing after last row

Grid `vertical space` global setting allows for up to 72px padding between grid rows. This bottom padding is applied to items in the last row of the grid.

The Grid `vertical space` setting value should be applied as padding between grid rows, but not on the last row.

- Dawn branch: `all-global-settings`
- Chrome Version 95.0.4638.69
- macOS Version  12.0.1

**Possible solution**

Add negative bottom margin to the grid container. Since this affects multiple sections, this requires thorough testing.

``` css
.grid {
  margin-bottom: calc(0px - var(--grid-mobile-vertical-spacing));
}

  .grid {
    margin-bottom: calc(0px - var(--grid-desktop-vertical-spacing));
  }
}
```

The key is ensure consistent behavior that makes sense in multiple scenarios

#### Featured collection
Screen_Shot_2021-12-21_at_10_03_28_AM

#### Collection with pagination
Screen_Shot_2021-12-21_at_10_03_15_AM

#### Collection without pagination
Screen_Shot_2021-12-21_at_10_03_41_AM

ludoboludo:

Fixes #1054 

Using `column-gap` and `row-gap` instead of dealing with the spacing using `padding`. 

**Other considerations**

Safari supports those properties in its current version 15 and previous version 14. But not prior to that.

**Testing**

Check where the horizontal and vertical space have an effect: 
- Collage,
- collection page, 
- feat. collection section,
- multicolumn section, 
- search page,
- collection list, 
- sliders (slideshow, blog posts, multicolumn, feat. collection, etc)

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127066767382)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127066767382/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -971,14 +971,18 @@ summary::-webkit-details-marker {
   display: flex;
   flex-wrap: wrap;
   margin-bottom: 2rem;
-  margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));;
+  /* margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing)); */
   padding: 0;
   list-style: none;
+  column-gap: var(--grid-mobile-horizontal-spacing);
+  row-gap: var(--grid-mobile-vertical-spacing);
 }
 
 @media screen and (min-width: 750px) {
   .grid {
-    margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing));
+    /* margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing)); */
+    column-gap: var(--grid-desktop-horizontal-spacing);
+    row-gap: var(--grid-desktop-vertical-spacing);
   }
 }
 
@@ -987,8 +991,8 @@ summary::-webkit-details-marker {
 }
 
 .grid__item {
-  padding-left: var(--grid-mobile-horizontal-spacing);
-  padding-bottom: var(--grid-mobile-vertical-spacing);
+  /* padding-left: var(--grid-mobile-horizontal-spacing);
+  padding-bottom: var(--grid-mobile-vertical-spacing); */
   width: calc(25% - 0.5rem * 3 / 4);
   max-width: 50%;
   flex-grow: 1;
@@ -997,16 +1001,16 @@ summary::-webkit-details-marker {
 
 @media screen and (min-width: 750px) {
   .grid__item {
-    padding-left: var(--grid-desktop-horizontal-spacing);
-    padding-bottom: var(--grid-desktop-vertical-spacing);
+    /* padding-left: var(--grid-desktop-horizontal-spacing);
+    padding-bottom: var(--grid-desktop-vertical-spacing); */
     width: calc(25% - var(--grid-desktop-horizontal-spacing) * 3 / 4);
     max-width: 50%;
   }
 }
 
-.grid--gapless .grid__item {
-  padding-left: 0;
-  padding-bottom: 0;
+.grid--gapless.grid {
+  column-gap: 0;
+  row-gap: 0;
 }
 
 @media screen and (max-width: 749px) {
--- a/assets/base.css
+++ b/assets/base.css
@@ -976,15 +976,19 @@ summary::-webkit-details-marker {
   list-style: none;
 }
 
+.grid:last-child {
+  margin-bottom: calc(-1 * var(--grid-mobile-vertical-spacing));
+}
+
 @media screen and (min-width: 750px) {
   .grid {
     margin-bottom: calc(2rem - var(--grid-desktop-vertical-spacing));
     margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing));
   }
-}
 
-.grid:last-child {
-  margin-bottom: 0;
+  .grid:last-child {
+    margin-bottom: calc(-1 * var(--grid-desktop-vertical-spacing));
+  }
 }
 
 .grid__item {
@@ -2427,6 +2431,16 @@ details[open] .modal-overlay::after {
   .search-modal__close-button {
     position: initial;
     margin-left: 0.5rem;
+  }
+}
+
+.template-search__results .grid {
+  margin-bottom: calc(-1 * var(--grid-mobile-vertical-spacing));
+}
+
+@media screen and (min-width: 750px) {
+  .template-search__results .grid {
+    margin-bottom: calc(-1 * var(--grid-desktop-vertical-spacing));
   }
 }
 
--- a/assets/base.css
+++ b/assets/base.css
@@ -1077,15 +1077,6 @@ summary::-webkit-details-marker {
   }
 }
 
-.grid__item--vertical-align {
-  align-self: center;
-}
-
-.grid__item--full-width {
-  flex: 0 0 100%;
-  max-width: 100%;
-}
-
 @media screen and (max-width: 749px) {
   .grid--peek.slider--mobile {
     margin: 0;
--- a/assets/base.css
+++ b/assets/base.css
@@ -990,7 +990,7 @@ summary::-webkit-details-marker {
 .grid__item {
   padding-left: var(--grid-mobile-horizontal-spacing);
   padding-bottom: var(--grid-mobile-vertical-spacing);
-  width: calc(25% - var(--grid-mobile-horizontal-spacing) * 3 / 4);
+  width: calc(25% - 0.5rem * 3 / 4);
   max-width: 50%;
   flex-grow: 1;
   flex-shrink: 0;
@@ -1102,7 +1102,7 @@ summary::-webkit-details-marker {
 
   .grid--peek .grid__item {
     padding-left: var(--grid-mobile-horizontal-spacing);
-    width: calc(50% - var(--grid-mobile-horizontal-spacing) / 2);
+    width: calc(50% - 3.75rem / 2);
   }
 
   .grid--peek .grid__item:first-of-type {
@@ -1116,15 +1116,15 @@ summary::-webkit-details-marker {
 
 @media screen and (min-width: 750px) and (max-width: 989px) {
   .slider--tablet.grid--peek .grid__item {
-    width: calc(25% - var(--grid-desktop-horizontal-spacing) * 3 / 4);
+    width: calc(25% - 4rem * 3 / 4);
   }
 
   .slider--tablet.grid--peek.grid--3-col-tablet .grid__item {
-    width: calc(33.33% - var(--grid-desktop-horizontal-spacing) * 2 / 3);
+    width: calc(33.33% - 4rem * 2 / 3);
   }
 
   .slider--tablet.grid--peek.grid--2-col-tablet .grid__item {
-    width: calc(50% - var(--grid-desktop-horizontal-spacing) / 2);
+    width: calc(50% - 4rem / 2);
   }
 
   .slider--tablet.grid--peek .grid__item:first-of-type {
--- a/assets/base.css
+++ b/assets/base.css
@@ -2735,3 +2735,46 @@ details-disclosure > details {
   border-left: none;
   border-right: none;
 }
+
+@supports not (inset: 10px) {
+  .grid {
+    margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));
+  }
+
+  .grid__item {
+    padding-left: var(--grid-mobile-horizontal-spacing);
+    padding-bottom: var(--grid-mobile-vertical-spacing);
+  }
+
+  @media screen and (min-width: 750px) {
+    .grid {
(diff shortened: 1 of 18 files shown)


tyleralsbury:
I wonder if we should add `@supports` - https://developer.mozilla.org/en-US/docs/Web/CSS/@supports - with a fallback on the previous approach.  A total of ~3.5% of global users are still using versions of Safari that don't support flex gap.

ludoboludo:
I wonder 🤔  technically we're supposed to support the latest 2 version of a browse right ?
I like the idea of only using flex gap if it's supported and keep the previous approach if not. But then we're adding stuff to the theme that needs to be removed later most likely. So I'm divided. I feel like it doesn't quite match the theme's principle.

tyleralsbury:
> Some ways around this are through JS. I saw some mention as well of using @supports of other properties that share the same support (starting on version 14.1 of Safari and not supported before) but not sure it's a great work around.

It's better than JS, I think.  We could add a comment - it's kinda silly, but it's still better than using JS.

tyleralsbury (approved):
Looks good - it's unfortunate to need to use `@supports` but I think it is principled to do so.  I'd say it falls under both web native and progressive enhancement.  I would love to exclude it but 3.5% of users is just too many something as core as grid spacing - we can't have that many users see collapsed grids.

melissaperreault (approved):
Thanks Ludo! Works well! 🚢

assets/base.css
```diff
@@ -971,20 +971,20 @@ summary::-webkit-details-marker {
...
   list-style: none;
+  column-gap: var(--grid-mobile-horizontal-spacing);
+  row-gap: var(--grid-mobile-vertical-spacing);
 }
 
 @media screen and (min-width: 750px) {
   .grid {
-    margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing));
+    column-gap: var(--grid-desktop-horizontal-spacing);
+    row-gap: var(--grid-desktop-vertical-spacing);
```
LucasLacerdaUX, line 983: Two questions about the Product Media grid, specifically:
1) Mobile peekaboo is gone if we increase the horizontal spacing. I understand that the original goal was to improve discoverability of other media in the carousel, but wanted to confirm if this is how it should work with these new settings.

image

2) This is probably a bug: if we increase the horizontal spacing, the product media / info layout will break. I believe the calculations for the `product__info-wrapper` and `product__media-wrapper` sizes are still based on the old `1rem` gap. We would need to update `section-main-product` to use the variable there too.

image

image

ludoboludo: Im wondering for `the product__info-wrapper` if I should remove any `padding-left` value it's currently receiving so that the spacing is only coming from the general setting 🤔 

I've made an update in this commit, let me know what you think: https://github.com/Shopify/dawn/pull/1127/commits/34f0638799fa1254ce7e786ce50360e8ce9135d8

ludoboludo: @melissaperreault , i've made it so sliders that have 1 item per page aren't applying any extra spacing to keep the peekaboo effect, is that ok ? I thought it made sense as there isn't anything really to gain from adding more spacing between those slides: [video](https://screenshot.click/22-01-idiet-7cky7.mp4)

ludoboludo: K so I tweaked some of the sliders work a bit better (scroll-padding-left update to match padding applied on the first slide). 

Made it so that horizontal space from the global settings will apply to sliders but it can cause this "issue": [video](https://screenshot.click/22-01-xpvxi-766mv.mp4)

assets/section-main-product.css
```diff
@@ -38,25 +38,25 @@
...
     max-width: 55%;
-    width: calc(55% - 1rem / 2);
+    width: calc(55% - var(--grid-desktop-horizontal-spacing) / 2);
   }
 
   .product--medium:not(.product--no-media) .product__info-wrapper,
   .product--small:not(.product--no-media) .product__media-wrapper {
     max-width: 45%;
-    width: calc(45% - 1rem / 2);
+    width: calc(45% - var(--grid-desktop-horizontal-spacing) / 2);
```
LucasLacerdaUX, line 59: I was thinking more about this and wondering if we really want the grid padding on these 2 columns to be affected by the setting. It currently inherits the gaps from `.grid`, but maybe we should have a fixed spacing.

If we set the Horizontal Spacing to the maximum value (`40px`) and the Product Media size to `Large`, there'll be A LOT of whitespace between the two columns and little available space for product information:
image

I've noticed that other two-column layouts, such as `Image with Text` aren't affected by this setting, so it would also make sense to have this same behaviour on the product page.

assets/base.css
```diff
@@ -1099,10 +1097,6 @@ summary::-webkit-details-marker {
     width: calc(50% - 3.75rem / 2);
```
LucasLacerdaUX, line 1097: To avoid the weird bug you mentioned (https://screenshot.click/22-01-xpvxi-766mv.mp4), can we set the `grid--peek .grid__item` size to be based on the horizontal spacing?

I would probably suggest `calc(50% - var(--grid-mobile-horizontal-spacing) * 2)`.

```suggestion
    width: calc(50% - var(--grid-mobile-horizontal-spacing) * 2);
```

assets/section-main-product.css
LucasLacerdaUX: cc @melissaperreault. Thoughts on this? Should the gap between these two columns be controlled by the same `Layout > Horizontal spacing` setting?

melissaperreault: @LucasLacerdaUX  Do you suggest we reduce the value when the image is set to large?

assets/base.css
```diff
@@ -2518,3 +2518,46 @@ details-disclosure > details {
   border-left: none;
   border-right: none;
 }
+
+@supports not (inset: 10px) {
```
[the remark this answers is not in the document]
KaichenWang: The issue is that we do use `inset` in our codebase. By adding this, we're in a way contradicting ourselves with regards to what browser support we are offering? We should either remove instances of `inset` in the theme or use a different approach here.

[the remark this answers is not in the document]
ludoboludo: I think here because it breaks the whole layout it's a bit more problematic maybe. I can see that it's then opening a door to more questioning about other supported/unsupported things we use and when to add support and when not to worry about it. 
People can still use the theme and go on about their purchase, so maybe we shouldn't worry about it too much 🤷 

cc: @tyleralsbury

assets/base.css
```diff
@@ -970,22 +970,23 @@ summary::-webkit-details-marker {
...
-  margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));;
+  margin-bottom: calc(2rem - var(--grid-mobile-vertical-spacing));
+  margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));
   padding: 0;
   list-style: none;
 }
 
 @media screen and (min-width: 750px) {
   .grid {
+    margin-bottom: calc(2rem - var(--grid-desktop-vertical-spacing));
```
ludoboludo, line 981: this can create a bit of an overlap but didn't seem to prevent you from clicking the `view all` button: [screenshot](https://screenshot.click/22-01-21ryi-zqhfn.png)

ludoboludo:
I've change the approach and focused on fixing the issue mentioned. So adding a calculation to make sure there isn't too much space at the bottom of a grid that is set to have a big `vertical spacing`. 

There is an issue that needs to be created and looked into though about the spacing between items on sliders.

It was at first tackled in here because using the `gap` property made it easy to deal with and work well. But since we're not using it due to browser support, it should be looked into on its own.

LucasLacerdaUX (commented):
The issues from the product page are back now that you reverted these changes. The rest looks good to me.

Do you plan to address these in a separate PR? I think the peekaboo one is related to the slider, so it's fine to scope it there, but the main grid one (`media` / `info` columns) can probably be fixed here.

assets/base.css
```diff
@@ -970,14 +970,15 @@ summary::-webkit-details-marker {
 .grid {
   display: flex;
   flex-wrap: wrap;
-  margin-bottom: 2rem;
-  margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));;
+  margin-bottom: calc(2rem - var(--grid-mobile-vertical-spacing));
```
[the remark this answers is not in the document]
ludoboludo: As for this, it's not technically part of the issue the PR was initially created for.

There is an overall not so great experience on sliders in general when using the horizontal grid spacing. It was easier to deal with when using `gap`. 
Ill create an issue for it unless there is an existing one and we can chat with Meli to see what might be our best option/approach for it 🤔 
Wether it's applying a fixed spacing regardless of the horizontal grid value or something else.

LucasLacerdaUX (approved):
💪 ✅ 👍 
   🦵🦵

tyleralsbury:
For Kai's Point number 2 - I added this to remove unexpected margins between sections to ensure that if a merchant has zero vertical spacing + non-background-1 section level colour schemes it will not have any background-1 peeking through.

Please make sure if you are making any changes to it that those behaviours are maintained.

assets/base.css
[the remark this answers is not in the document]
ludoboludo: Made some edits, let me know if I understood properly 👍  https://github.com/Shopify/dawn/pull/1127/commits/6e325f28d791318082763299e5e1ea6bdba434a4

ludoboludo:
Reverted back to using `gap` and `@support` as using a negative margin proved to be tricky with other settings like `section spacing` and `section padding`. 
Sorry for the back and forth but it was worth exploring.

KaichenWang (commented):
Looks good! Left a comment on grid item width

assets/base.css
```diff
@@ -987,8 +989,6 @@ summary::-webkit-details-marker {
 }
 
 .grid__item {
-  padding-left: var(--grid-mobile-horizontal-spacing);
-  padding-bottom: var(--grid-mobile-vertical-spacing);
   width: calc(25% - 0.5rem * 3 / 4);
   max-width: 50%;
   flex-grow: 1;
```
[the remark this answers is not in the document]
ludoboludo: Good catch 👌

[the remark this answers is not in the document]
KaichenWang: @ludoboludo I think we missed another one 

Screen Shot 2022-01-19 at 3 19 04 PM2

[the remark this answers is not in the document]
ludoboludo: Done 👌

KaichenWang (approved):
Awesome!

LucasLacerdaUX (dismissed):
🐕  🎉

ludoboludo:
Hey folks, @KaichenWang & @LucasLacerdaUX, I had to discard your approvals as after chatting with designers it seems like we dont want the gap value to impact the space between the product media and product info wrapper. So I reverted some of the changes for it so that it uses the hard coded value we already had 👍

melissaperreault (commented):
Thank you Ludo!

Found some edge cases:

- [ ] Search blog cards [don't follow the vertical space](https://screenshot.click/22-01-7nar0-48929.png) _(Might not be related so can be a new issue on its own.)_
- [ ] When `Collection list` has more than 3 cards, the vertical space [is not applied](https://screenshot.click/22-01-we7bj-7smi1.png) - Only obvious when Cards style is set to `Card` as oppose to `Standard`.
- [ ] For a next new issue because that one would need more time, I wonder if the product media thumbnails counts as a slider or if they would remain an exception to the rule. ([Video reference](https://screenshot.click/22-01-qr7ax-qzvqo.mp4))

ludoboludo:
> Thank you Ludo!
> 
> Found some edge cases:
> 
> * [ ]  Search blog cards [don't follow the vertical space](https://screenshot.click/22-01-7nar0-48929.png) _(Might not be related so can be a new issue on its own.)_
> * [ ]  When `Collection list` has more than 3 cards, the vertical space [is not applied](https://screenshot.click/22-01-we7bj-7smi1.png) - Only obvious when Cards style is set to `Card` as oppose to `Standard`.
> * [ ]  For a next new issue because that one would need more time, I wonder if the product media thumbnails counts as a slider or if they would remain an exception to the rule. ([Video reference](https://screenshot.click/22-01-qr7ax-qzvqo.mp4))

Hmm weird, I'm not seeing the issue on my side where the space isn't there where set to cards 🤔  I did notice something that needed to be changed as well but it was more for the width of the elements in the collection list section but not about the spacing between items. Would you mind trying it again maybe ?

LucasLacerdaUX (approved):
Looks good! :DD

KaichenWang (approved):

PR #1128, opened 2022-01-07
Adding new icon options to the .json file

danielvan:
Adding new icon names and alternatives we are providing in a future release.
Approach was to keep the icon options in alphabetical order.

We want to add flexibility to merchants.

[Demo store Editor](https://os2-demo.myshopify.com/admin/themes/127115526166/editor)

Kept alphabetical order in the .json files, which means some icons have new placements in the array.


tyleralsbury (approved):
Thank you for providing each of them directly in a collapsible tab section!  Very helpful 👍 

Looks good to me, visually.  Code looks good, at least as much as I can actually read of SVGs.

I used inspect element to check the viewbox and the positions of the icons within each viewbox looks good, and I also tested to ensure the slightly larger size didn't throw the collapsible content off, visually.  Everything looks good.

LucasLacerdaUX (approved):
They all look good to me! Tested every icon on both sections with different color palettes and they're all ✅ :DD

PR #1140, opened 2022-01-10
Fix border card hover safari

### LINKED ISSUE #1121: [Safari] Second image on hover doesn't follow corner radius

The hover state is not working properly, the corners of the images [are not rounded when you hover](https://screenshot.click/22-01-et525-8rw79.png) on product images on a collection page. [Video example](https://screenshot.click/22-01-htdpj-d0ty7.mp4)

Maintain the corner roundness.

- Safari Version 15.2 (17612.3.6.1.6)
- macOS Version 12.1

sofiamatulis:

Fixes #1121 .

Updated the z-index for the card and the media. The reason this was happening is because when we apply `transform` or `opacity` (which we are for the animation), the order doesn't stay the same even with `position: relative`. So the solution is to set the index. More here: https://coder-coder.com/z-index-isnt-working/

**Other considerations**

**Demo links**

Test on safari: 
- [Editor](https://os2-demo.myshopify.com/admin/themes/127073845270/editor)

**Checklist**

### DIFF
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -8,6 +8,7 @@
 .card {
   text-align: var(--card-text-alignment);
   text-decoration: none;
+  position: relative;
 }
 
 .card--card {
@@ -37,6 +38,7 @@
 
 .card .card__inner .card__media {
   overflow: hidden;
+  -webkit-mask-image: -webkit-radial-gradient(white, black);
   border-radius: calc(var(--card-corner-radius) - var(--card-border-width) - var(--card-image-padding));
 }
 
@@ -175,14 +177,6 @@
   .card:hover .media.media--hover-effect > img:first-child:only-child,
   .card-wrapper:hover .media.media--hover-effect > img:first-child:only-child {
     transform: scale(1.03);
-  }
-
-  .card .card__inner .card__media {
-    z-index: 1;
-  }
-
-  .card .card__inner  {
-    z-index: 2;
   }
 
   .card-wrapper:hover
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -37,6 +37,7 @@
 
 .card .card__inner .card__media {
   overflow: hidden;
+  /* Fix for Safari border bug on hover */
   z-index: 0;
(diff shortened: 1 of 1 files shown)


assets/component-card.css
```diff
@@ -177,6 +177,14 @@
     transform: scale(1.03);
   }
 
+  .card .card__inner .card__media {
+    z-index: 1;
+  }
+
+  .card .card__inner  {
+    z-index: 2;
```
kmeleta, line 185: This seems safe for card style cards, but breaks the z order we're relying on for standard card shadows. Shadows need to remain behind adjacent cards.

kmeleta (commented):
Good find! I think we just need to fiddle with it a little more to ensure we don't cause shadow overlaps. Hopefully it's straightforward adjustment.

assets/component-card.css
sofiamatulis: Good catch Ken! Updated the z-index

sofiamatulis:
I have opted to fix this with `z-index` since it's a cleaner/more intuitive solution, but I found a thread about this same bug here: https://discourse.webflow.com/t/safari-not-hiding-overflow-on-rounded-corner-divs/55060

Another solution for Safari seems to be to apply:

```
 -webkit-mask-image: -webkit-radial-gradient(white, black);
````

File: assets/component-card.css [hunk 2]
```diff
@@ -37,6 +37,7 @@
 
 .card .card__inner .card__media {
   overflow: hidden;
+  z-index: 0;
```
kmeleta, line 40: I know we try to avoid comments in code, but I'm tempted to include a short one on this line that makes it known it's needed for a safari bug. I could see it appearing as though it's not doing anything and getting removed in the future. Not sure how everyone feels about that..

kmeleta (approved):
Nice! glad the z-index fix could still work out. Tested the fix in safari and looked for shadow/z-related regressions and didn't notice any issues.

assets/component-card.css
tyleralsbury: It might be a good idea - CSS isn't very self-explanatory.

tyleralsbury (approved):
Approved - Maybe add a comment like Ken mentioned.

Issue #1099, opened 2022-01-05
Some section shadows not inheriting content container border corner radius


katycobb:
**Describe the current behavior**
Observed on these sections: Rich text, image with text. 
When the content container corner radius is set to greater than zero, and a shadow is applied, the shadow does not inherit the corner radius of the content container. 

Side note: it appears like there is a gap in the border on the top right corner too. If this is an unrelated issue, let me know and I can create a separate ticket. 

**Describe the expected behavior**
I THINK the shadow should inherit the corner radius of the content container (as seen in multicolumn section.) 

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 2.5.0
- Chrome Version 97.0.4692.71
- macOS Version 12.1 

**Possible solution**
Align with the treatment used in multicolumn section. 

**Additional context/screenshots**

-content container border corner radius is set to 40px
-content container shadow horizontal and vertical offset are both set to 40px

<img width="1506" alt="05-45-h06e7-w3qan" src="https://user-images.githubusercontent.com/93404168/148301541-64e27b85-9155-435b-a643-a1be608cd570.png">
<img width="1387" alt="05-51-tdgah-yuawc" src="https://user-images.githubusercontent.com/93404168/148301557-c4cc0b80-02d4-46ef-a3aa-06bbbbe6e8e4.png">

Expected behavior: 
<img width="1007" alt="05-48-hjwm6-fpx8z" src="https://user-images.githubusercontent.com/93404168/148301552-b3fa820b-4cc3-4f46-8f2b-27302b6e3192.png">

nicklepine:
Video to support the case on Image with Text reported by @katycobb  https://screenshot.click/06-05-e92kg-aj34a.mp4

nicklepine:
Note: this doesn't seem to happen for media 
![image](https://user-images.githubusercontent.com/8127145/148407372-1ea5880f-1881-44a3-9606-3aeafad2d90e.png)

melissaperreault:
- [ ] Noticed you can have a vertical offset but still [see an horizontal gap over here](https://screenshot.click/06-59-id2qz-2ltwt.png)

kmeleta:
The image with text portion of this will be addressed as part of Martina's PR https://github.com/Shopify/dawn/pull/1108 so we're not making conflicting updates.

The rich text/newsletter issue I'll have in its own PR.

PR #1142, opened 2022-01-10
Fix fullwidth content container shadows

kmeleta:

Refs #1099 

Fixes the weird shadow offset noticeable on fullwidth content container sections like newsletter and rich text.

Removed the left/right positioning offset for the shadow pseudo element when the content container is full width.

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127074271254/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2491,10 +2491,7 @@ details-disclosure > details {
 .content-container:after {
   content: '';
   position: absolute;
-  width: calc(var(--text-boxes-border-width) * 2 + 100%);
-  height: calc(var(--text-boxes-border-width) * 2 + 100%);
-  top: calc(var(--text-boxes-border-width) * -1);
-  left: calc(var(--text-boxes-border-width) * -1);
+  inset: calc(var(--text-boxes-border-width) * -1);
   z-index: -1;
   border-radius: var(--text-boxes-radius);
   box-shadow: var(--text-boxes-shadow-horizontal-offset)
@@ -2504,7 +2501,8 @@ details-disclosure > details {
 }
 
 .content-container--full-width:after {
-  width: 100%;
+  left: 0;
+  right: 0;
   border-radius: 0;
 }
 


katycobb (approved):
Thanks Ken, this looks good to me, but I'd feel more confident with Meli and/or Nick also taking a look at this one!

tyleralsbury (approved):

melissaperreault (approved):
Thanks! I took a look and looks good for both specified section. I'll check on the other PR for the fix. 🙏

assets/base.css
```diff
@@ -2491,10 +2491,7 @@ details-disclosure > details {
 .content-container:after {
   content: '';
   position: absolute;
-  width: calc(var(--text-boxes-border-width) * 2 + 100%);
-  height: calc(var(--text-boxes-border-width) * 2 + 100%);
-  top: calc(var(--text-boxes-border-width) * -1);
-  left: calc(var(--text-boxes-border-width) * -1);
+  inset: calc(var(--text-boxes-border-width) * -1);
```
ludoboludo, line 2494: Weren't we a bit worried about using `inset` when it's no supported on 14 for Safari desktop and mobile 🤔 
Mainly mobile. I was trying to find the percentage uses of each version but not finding anything so far.

ludoboludo (commented):
Looks good! Just wondering about `inset` as I remember you bringing it up in one of Kai's PR I think.

assets/base.css
tyleralsbury: https://caniuse.com/?search=inset

You can check here for usage stats for unsupported versions by hovering.  It's a total of 3.5%.

If this is not the sort of thing that will totally break the site on those versions, we can allow them to exist as a progressive enhancement for browsers that do support them.

tyleralsbury: I prefer to go by the `caniuse` stats because they're accounting for actual usage in the wild - on the web.

kmeleta: Also if I interpret those correctly that 3.5% is mobile and desktop safari combined and any versions between 3.2 and 14/14.4, meaning some of that percentage could be well outside the last 2 major versions anyway.

PR #1148, opened 2022-01-11
Collage video block focus style update

ludoboludo:

The focus style on the video block is updated to use the same kinda logic (focus/focus-visible) as the cards.

Removed `focus-within` and used the `focus, focus-visible` approach.

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127083380758)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127083380758/editor)

**Checklist**

### DIFF
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -150,8 +150,13 @@
   border: 0;
 }
 
-.collage-card .deferred-media__poster:focus-visible,
 .collage-card .deferred-media__poster:focus {
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
+  outline-offset: 0.3rem;
+  outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
+}
+
+.collage-card .deferred-media__poster:focus-visible {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
   outline-offset: 0.3rem;
   outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -108,12 +108,6 @@
   }
 }
 
-.collage__item--video:focus-within {
-  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
-  outline-offset: 0.3rem;
-  outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
-}
-
 .collage-card {
   background: rgb(var(--color-background));
   border-radius: var(--card-corner-radius);
@@ -156,8 +150,15 @@
   border: 0;
 }
 
-.collage-card .deferred-media__poster:focus,
-.collage-card .collage-card__link:focus {
+.collage-card .deferred-media__poster:focus-visible,
+.collage-card .deferred-media__poster:focus {
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
+  outline-offset: 0.3rem;
+  outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
+}
+
+.collage-card .deferred-media__poster:focus:not(:focus-visible),
+.collage-card .collage-card__link:focus:not(:focus-visible) {
   outline-offset: 0.3rem;
   outline: none;
   box-shadow: none;
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -159,10 +159,10 @@
   right: calc(var(--card-border-width) * -1);
   top: calc(var(--card-border-width) * -1);
   outline-offset: 0.3rem;
+  border-radius: var(--card-corner-radius);
 }
 
 .collage-card .deferred-media__poster:focus:after {
-  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
   outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
 }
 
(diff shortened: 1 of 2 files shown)


tyleralsbury (approved):
Looks good - Tested and the focus style only appears when focusing manually or when the modal sets the focus to the video element on close.  It does not appear when clicked, which is the goal 👍

assets/collage.css
```diff
@@ -154,8 +148,15 @@
...
-.collage-card .collage-card__link:focus {
+.collage-card .deferred-media__poster:focus-visible,
+.collage-card .deferred-media__poster:focus {
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
+  outline-offset: 0.3rem;
+  outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
+}
+
+.collage-card .deferred-media__poster:focus:not(:focus-visible),
+.collage-card .collage-card__link:focus:not(:focus-visible) {
```
martinamarien, line 159: In safari, the focus looks different for the selected video vs. the onboarding state. If you click outside of the safari window, the grey border is displayed, but not when tabbing. The focus state looks right for the regular video section.

https://user-images.githubusercontent.com/26778637/148977678-67f95c60-6cff-42e1-97aa-ab4c066adadc.mov


Screen Shot 2022-01-11 at 11 06 00 AM

Screen Shot 2022-01-11 at 11 06 16 AM

martinamarien (changes requested):
The border in Safari doesn't seem right to me. It doesn't seem to have the more prominent/contrasted focus state.

assets/collage.css
ludoboludo: Ah I know why. I grouped selectors for `:focus` and `:focus-within` which I should not because `:focus-within` isn't supported in safari so the styling for `:focus` is not applied. I separated them and it should now be good 🙂

melissaperreault: Related to Safari, I noticed the focus is not following the UI shape here, I can open an issue to not busy up this PR https://screenshot.click/22-01-mtov6-l1wzh.png

tyleralsbury: It may be that Safari is being more strict about the actual CSS spec with regards to outlines (which are not really for decorative use, technically).

melissaperreault (commented):
Thanks Ludo! Noticed this detail where the outline is not applied the same way as on other cards ([Video](https://screenshot.click/22-01-fbzuu-u9eh3.mp4))

tyleralsbury (dismissed):
Tested on desktop and mobile, with different settings for borders and shadows for cards. Different colour schemes as well.  Everything looks good to me.  👍

assets/collage.css
martinamarien: > From what I can tell most of the focus outlines/shadows, I think all of them actually, aren't receiving the border radius like in other browsers. It's specific to Safari.

Yeah. I talked to Ken and we need to convert them to box-shadow because `outline` doesn't inherit the border-radius.

martinamarien:
@melissaperreault My Product media PR will fix the focus state for video. You can see it fixed on this [homepage](https://os2-demo.myshopify.com/admin/themes/127066177558/editor). 

One thing to note is that we will need to add an override for border radius on focus. But we can fix that easily. I've also updated model focus to behave the same way.

assets/collage.css
```diff
@@ -154,9 +148,35 @@
...
+}
+
+.collage-card .deferred-media__poster:focus-visible:after {
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
+  outline: 0.2rem solid rgba(var(--color-foreground), 0.5);
+}
+
+.collage-card .deferred-media__poster:focus:not(:focus-visible),
+.collage-card .deferred-media__poster:focus:not(:focus-visible):after,
+.collage-card .collage-card__link:focus:not(:focus-visible) {
```
martinamarien, line 174: I don't think you don't need this class here.
This class is only used if there is no-js and we have styling in our global css file that should handle this: 

```
.no-js :focus:not(:focus-visible) {
    outline: 0;
    box-shadow: none;
}
```

assets/collage.css
```diff
@@ -154,9 +148,35 @@
...
+.collage-card .deferred-media__poster:after {
+  content: "";
+  position: absolute;
+  z-index: 1;
+  bottom: calc(var(--card-border-width) * -1);
+  left: calc(var(--card-border-width) * -1);
+  right: calc(var(--card-border-width) * -1);
+  top: calc(var(--card-border-width) * -1);
   outline-offset: 0.3rem;
+}
```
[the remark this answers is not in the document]
martinamarien: I would update for video, and if there is a separate selector for card then, there as well if it's not too much effort. It's what we are doing for the pills selectors already. So it would be good to continue to do that with the outline approach. Even though it doesn't benefit safari users, the overall UX/UI feels better when the outline follows the border-radius of the card.

martinamarien (dismissed):
Other than my one comment about box-shadow. Looks good. I'll defer to Meli for the decision on whether to remove it or not. If she gives the 👍🏻  then we can ship with a another dev approval. If not, ping me for another review 🙂

tyleralsbury (dismissed):
Approved pending Meli's feedback on the most recent comment.

martinamarien (approved):

tyleralsbury (approved):

melissaperreault (approved):
Thanks Ludo! 🎉 
We can ship 🚢 🚢 

One tiny detail I noticed that we can tackle later is about the change in focus color depending on setting combination ([Video](https://screenshot.click/22-01-v5e2a-u2evb.mp4))

tyleralsbury (approved):
Tested a few styles and it all feels good to me.

I noticed that the cards don't have gradients when a gradient colour scheme selected.  Is this intended or should we also incorporate the gradient colour schemes into the cards?  Not for this PR, but to maybe spin off another issue for later.

cc @melissaperreault

martinamarien (approved):
Looks good. 
Let's create an issue to look at focus states for video in the collage, if we haven't already

ludoboludo:
Issue created: https://github.com/Shopify/dawn/issues/1202

Issue #1018, opened 2021-12-14
[Global setting] Focus to follow corner radius setting


melissaperreault:
### Expectations

- [ ] Let's have the focus [keep the same roundness/corner radius](https://screenshot.click/21-12-5dp11-3u9ew.png) of the element it is surrounding.
- [ ] Sort by gets roundness on focus [but Filter doesn’t](https://screenshot.click/21-12-35csg-pr2up.png).

katycobb:
Hi @melissaperreault  - would this issue cover what I'm observing here? Not sure if it needs a new bug. This appears when clicking the "Share" icon on the PDP.
<img width="549" alt="06-34-gi6t9-xwuok" src="https://user-images.githubusercontent.com/93404168/148433494-fc8fec87-3534-4c4a-9b23-d74a6ff9a3f1.png">

melissaperreault:
> would this issue cover what I'm observing here

@katycobb Yes, this issue will fix! I'll be pairing with Sofia to make this more appealing! 😊

PR #1154, opened 2022-01-11
Prevent unintended focus outline radius

kmeleta:

Fixes an issue I noticed on #1097 that also applies to buttons, where adding the border-radius value to the border-width value of an element can create unnecessary radius when there is no radius setting applied (0px radius + 2px border width = 2px radius).

Basically we only want to apply this sum value if border radius is above 0. I think I found a contrived way this could be done inline in the CSS..
`border-radius: calc(var(--inputs-radius) * var(--inputs-radius) + var(--inputs-border-width) + var(--inputs-radius))`
But I believe it's probably clearer to handle the logic in liquid and just assign a new variable that's 0 when necessary.

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127085772822/editor)

**Checklist**


ludoboludo (approved):
Looks good 👍 

I noticed `.cart__dynamic-checkout-buttons div[role='button']` is always applying a `border-radius`, shouldn't it be using `--button-radius` only ? 🤔

ludoboludo (approved):

kmeleta:
> Looks good 👍
> 
> I noticed `.cart__dynamic-checkout-buttons div[role='button']` is always applying a `border-radius`, shouldn't it be using `--button-radius` only ? 🤔

Well the paypal button applies its own minor border radius regardless, but the other checkout buttons did also needed this fix, so thanks for pointing that out. I did about that rule in the cart css https://github.com/Shopify/dawn/pull/1154/commits/5a4ce5c608b7bd0f4488d3e6d1b6c63a39f5ab92 @ludoboludo

melissaperreault:
I wonder if this is relevant to this issue https://github.com/Shopify/dawn/issues/1018 as well? 🤔

melissaperreault (approved):
Thanks Ken! Tested various scenarios and LGTM! 🚢

PR #1179, opened 2022-01-13
Improve lazy loading strategy

tyleralsbury:

I was doing some performance testing with all of our most recent changes and noticed a drop in performance for both the collection page and the product page.  Looking at the Lighthouse score, a major part of the issue was the fact that we were lazy loading the above-the-fold images.

Originally, our thought was that because of Sections Everywhere, we can't have any guarantees that certain images are above the fold.  However, I think we can make a reasonable assumption about a vast majority of merchants - for places like the product page or collection page, they want to get to the meat of the page as quickly as possible.

On the collection page, this will mean the collection banner and/or the first product images.  On the product page, this will mean the first piece of media in the gallery.

**Other considerations**

* Other pages, like the article page, would benefit from treatment like this as well.  For now, I kept it to the critical path and only dealt with the product pages, but I'd like to do a follow-up to consider the other pages as well.
* We can't do this with non-template sections because we can't know which ones are on the page first.  I'd love a solution to this so that we can do things like not lazy load the first slideshow image if it's the first section on the page, etc...

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -61,7 +61,7 @@
               {%- assign featured_media = product.selected_or_first_available_variant.featured_media -%}
               <li id="Slide-{{ section.id }}-{{ featured_media.id }}" class="product__media-item grid__item slider__slide is-active{% if featured_media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains featured_media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ featured_media.id }}">
                 {%- assign media_position = 1 -%}
-                {% render 'product-thumbnail', media: featured_media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width %}
+                {% render 'product-thumbnail', media: featured_media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width, lazy_load: false %}
               </li>
             {%- endif -%}
             {%- for media in product.media -%}
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -68,7 +68,7 @@
               {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
                 <li id="Slide-{{ section.id }}-{{ media.id }}" class="product__media-item grid__item slider__slide{% if product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} is-active{% endif %}{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
                   {%- assign media_position = media_position | default: 0 | plus: 1 -%}
-                  {% render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width %}
+                  {% render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width, lazy_load: true %}
                 </li>
               {%- endunless -%}
             {%- endfor -%}
--- a/snippets/article-card.liquid
+++ b/snippets/article-card.liquid
@@ -10,6 +10,7 @@
     - show_date: {String} The setting either show the article date or not. If it's not included it will not show the image by default
     - show_author: {String} The setting either show the article author or not. If it's not included it will not show the author by default
     - show_badge: {String} The setting either show the blog badge or not.
+    - lazy_load: {Boolean} Image should be lazy loaded. Default: true (optional)
 
     Usage:
     {% render 'article-card' blog: blog, article: article, show_image: section.settings.show_image %}
@@ -47,10 +48,10 @@
                 src="{{ article.image.src | img_url: '533x' }}"
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: 2 }}px, (min-width: 750px) calc((100vw - 130px) / 2), calc((100vw - 50px) / 2)"
                 alt="{{ article.image.src.alt | escape }}"
+                class="motion-reduce"
+                {% unless lazy_load == false %}loading="lazy"{% endunless %}
                 width="{{ article.image.width }}"
                 height="{{ article.image.height }}"
-                loading="lazy"
-                class="motion-reduce"
               >
             </div>
           </div>
--- a/snippets/article-card.liquid
+++ b/snippets/article-card.liquid
@@ -49,7 +49,7 @@
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: 2 }}px, (min-width: 750px) calc((100vw - 130px) / 2), calc((100vw - 50px) / 2)"
                 alt="{{ article.image.src.alt | escape }}"
                 class="motion-reduce"
-                {% unless lazy_load == false %}loading="lazy"{% endunless %}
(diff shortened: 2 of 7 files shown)


KaichenWang (dismissed):
Looks good. In local testing, seeing an average 5 point increase in performance score on collection and product pages

KaichenWang (dismissed):

snippets/article-card.liquid
```diff
@@ -47,10 +48,10 @@
                 src="{{ article.image.src | img_url: '533x' }}"
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: 2 }}px, (min-width: 750px) calc((100vw - 130px) / 2), calc((100vw - 50px) / 2)"
                 alt="{{ article.image.src.alt | escape }}"
+                class="motion-reduce"
+                {% unless lazy_load == false %}loading="lazy"{% endunless %}
```
ludoboludo, line 52: What about:
```suggestion
                {% if lazy_load %}loading="lazy"{% endif %}
```
```suggestion
                {% unless lazy_load == false %}loading="lazy"{% endunless %}
```

snippets/card-product.liquid
```diff
@@ -51,8 +52,8 @@
                 src="{{ card_product.featured_media | img_url: '533x' }}"
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 130 | divided_by: 4 }}px, (min-width: 990px) calc((100vw - 130px) / 4), (min-width: 750px) calc((100vw - 120px) / 3), calc((100vw - 35px) / 2)"
                 alt="{{ card_product.featured_media.alt | escape }}"
-                loading="lazy"
                 class="motion-reduce"
+                {% unless lazy_load == false %}loading="lazy"{% endunless %}
```
ludoboludo, line 56: Same here

snippets/product-thumbnail.liquid
```diff
@@ -63,7 +64,7 @@
           {{ media.preview_image | img_url: 'master' }} {{ media.preview_image.width }}w"
         src="{{ media | img_url: '1946x' }}"
         sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | times: media_width | round }}px, (min-width: 990px) calc({{ media_width | times: 100 }}vw - 10rem), (min-width: 750px) calc((100vw - 11.5rem) / 2), calc(100vw - 4rem)"
-        loading="lazy"
+        {% unless lazy_load == false %}loading="lazy"{% endunless %}
```
ludoboludo, line 67: And here

snippets/product-thumbnail.liquid
```diff
@@ -102,7 +103,7 @@
         {{ media.preview_image | img_url: 'master' }} {{ media.preview_image.width }}w"
       src="{{ media | img_url: '1946x' }}"
       sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | times: media_width | round }}px, (min-width: 990px) calc({{ media_width | times: 100 }}vw - 10rem), (min-width: 750px) calc((100vw - 11.5rem) / 2), calc(100vw - 4rem)"
-      loading="lazy"
+      {% unless lazy_load == false %}loading="lazy"{% endunless %}
```
ludoboludo, line 106: Here

snippets/product-thumbnail.liquid
```diff
@@ -146,7 +147,7 @@
...
```
ludoboludo, line 150: And here

snippets/product-thumbnail.liquid
```diff
@@ -37,7 +38,7 @@
           {{ media.preview_image | img_url: 'master' }} {{ media.preview_image.width }}w"
         src="{{ media | img_url: '1946x' }}"
         sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | times: media_width | round }}px, (min-width: 990px) calc({{ media_width | times: 100 }}vw - 10rem), (min-width: 750px) calc((100vw - 11.5rem) / 2), calc(100vw - 4rem)"
-        loading="lazy"
+        {% unless lazy_load == false %}loading="lazy"{% endunless %}
```
ludoboludo, line 41: And here

ludoboludo (commented):
Looks good! Just a question about the `unless`. Couldn't we use an `if` instead ? I remember discussions saying that `if` are easier to read and be our go to.

tyleralsbury:
> Looks good! Just a question about the unless. Couldn't we use an if instead ? I remember discussions saying that if are easier to read and be our go to.

Good point. The "unless" was easier to wrap my head around while implementing it, but I think you're right that an `if` makes more sense.

snippets/article-card.liquid
```diff
@@ -10,6 +10,7 @@
     - show_date: {String} The setting either show the article date or not. If it's not included it will not show the image by default
     - show_author: {String} The setting either show the article author or not. If it's not included it will not show the author by default
     - show_badge: {String} The setting either show the blog badge or not.
+    - lazy_load: {Boolean} Image should be lazy loaded. Default: true (optional)
```
ludoboludo, line 13: Since article card is called in `featured-blogs` and `main-blog` as well, do they we need to pass that variable there ? I think we do 🤔

sections/main-product.liquid
```diff
@@ -61,14 +61,14 @@
...
               </li>
             {%- endif -%}
             {%- for media in product.media -%}
               {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
                 <li id="Slide-{{ section.id }}-{{ media.id }}" class="product__media-item grid__item slider__slide{% if product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} is-active{% endif %}{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}" data-media-id="{{ section.id }}-{{ media.id }}">
                   {%- assign media_position = media_position | default: 0 | plus: 1 -%}
-                  {% render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width %}
+                  {% render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width, lazy_load: true %}
```
ludoboludo, line 71: I think we need to add this to `featured-product.liquid` as well. 
And that's actually due to the change I mentioned with an `if` instead of `unless` 😅 

So maybe unless was better that way.

snippets/card-product.liquid
```diff
@@ -7,7 +7,8 @@
     - show_secondary_image: {Boolean} Show the secondary image on hover. Default: false (optional)
     - show_vendor: {Boolean} Show the product vendor. Default: false
     - show_rating: {Boolean} Show the product rating. Default: false
-    - extend_height: {Boolean} Card height extends to available container space. Default: false (optional)
+    - extend_height: {Boolean} Card height extends to available container space. Default: true (optional)
+    - lazy_load: {Boolean} Image should be lazy loaded. Default: true (optional)
```
ludoboludo, line 11: Same here if we keep the if approach we would need to add `lazy_load` for `collage.liquid` and a few other places.

ludoboludo (commented):
My bad with the if/unless switch. Might have been more straight forward of an approach with `unless` after all.

sections/main-product.liquid
KaichenWang: `lazy_load: true` is not necessary since the that's the default
```suggestion
                  {% render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width %}
```

KaichenWang (approved):

tyleralsbury:
Going to keep it for now just to not dismiss the reviews again.  I'll create an issue to revisit the overall lazy loading - I'd like to do something similar on the blog page as well, for example, and at that point I can remove the unnecessary stuff since it doesn't cause issues.

PR #1204, opened 2022-01-18
fix password input box shadow and password modal input focus

martinamarien:
1. After the latest input update, the styles were not added to the password page for input and button outset/offset. 
2. The password input field `:focus` state was also not behaving correctly. 

password page

password page modal

1. I added the three css variables to `password.liquid` 
```
--buttons-radius-outset
--inputs-margin-offset
--inputs-radius-outset
```

2. The password modal input was wrapped in a `div` whose `display` was set to `block`. I removed the `display` property so that it remains `flex` as that's what the `field` class is. 

**Testing**
- [ ] Test Mobile, Tablet, Desktop
- [ ] Password modal input `:focus`
- [ ] Password modal button styles are properly inherited 
- [ ] Password modal input looks correct - no straight border bottom 
- [ ] Password page email banner input looks correct - no straight border bottom 
- [ ] Password page email banner input button inherits correct styles 
- [ ] Safari 
- [ ] Chrome 
- [ ] Firefox  

**Demo links**
- [Editor](https://os2-demo.myshopify.com/admin/themes/127118213142/editor?previewPath=%2Fpassword)

**Checklist**

### DIFF
--- a/assets/section-password.css
+++ b/assets/section-password.css
@@ -107,7 +107,6 @@ body {
 }
 
 .password-field.field {
-  display: block;
   flex: 1 20rem;
 }
 
--- a/layout/password.liquid
+++ b/layout/password.liquid
@@ -86,7 +86,7 @@
         --card-shadow-vertical-offset: {{ settings.card_shadow_vertical_offset | divided_by: 10.0 }}rem;
         --card-shadow-blur-radius: {{ settings.card_shadow_blur | divided_by: 10.0 }}rem;
 
-        --badge-corner-radius: {{ settings.badge_corner_radius | divided_by: 10.0 }}rem;  
+        --badge-corner-radius: {{ settings.badge_corner_radius | divided_by: 10.0 }}rem;
 
         --spacing-sections-desktop: {{ settings.spacing_sections }}px;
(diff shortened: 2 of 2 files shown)


assets/section-password.css
```diff
@@ -107,7 +107,6 @@ body {
 }
 
 .password-field.field {
-  display: block;
   flex: 1 20rem;
 }
```
sofiamatulis, line 111: This should be fixed on Tyler's PR, but it would be good to check: 

https://screenshot.click/18-41-ykvw3-xp5wy.png

sofiamatulis (approved):
Looks good 👍

assets/section-password.css
martinamarien: Yep! I didn't change the contact-form in my PR. The reason I noticed the issue is because I added the section to the password page. 👍🏻

tyleralsbury (approved):
Looks good - contact form issue was merged so if we did want to test that, we can rebase and check it.  Should be okay, though, since we know what caused it and if I replicate the fix in inspect element it does work.

martinamarien:
I think we should be good to merge. The fix you made is pretty isolated from the ones I made.

PR #1205, opened 2022-01-18
Fixed contact form shadows

tyleralsbury:
**Why are these changes introduced?**

Fixes shadows on the contact form section

**What approach did you take?**

Similar to fix for other sections' shadows, added a relative position and z-index of 0 on the parent container.

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=127118639126)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127118639126/editor)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #1209, opened 2022-01-18
Replace zindex with isolate

### LINKED ISSUE #1206: Refactor z-index to isolate
## Purpose

We are using `z-index: 0` and `position: relative` to fix a few situations where shadows end up behind containers.  Sofia brought up that we could use CSS Isolation instead:

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

Let's refactor this - it would be much more readable than a bunch of `z-index`.

Thank you Sofia!

sofiamatulis:
I am replacing `z-index: 0` and `position: relative` with `isolate`. Where to test (add shadow to the buttons and inputs in global settings):

- Collage (button)
- Newsletter (input)
- Collection list (button)
- Contact form (input)
- Featured blog (button)
- Featured product (button and input)
- Footer (localization dropdown)
- Multicolumn (button)
- Product recommendation (button)
- Collection (button)
- Cards (card shadows)

Example: https://screenshot.click/18-18-chunj-13oy1.mp4

Fixes #1206 .

Original comment: https://github.com/Shopify/dawn/pull/1205#discussion_r786878616

By doing this we are removing 40+ lines of code! More on `isolate` here: https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127453888534/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2771,3 +2771,7 @@ details-disclosure > details {
     }
   }
 }
+
+.isolate {
+  isolation: isolate;
+}
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -22,7 +22,7 @@
 <div class="color-{{ section.settings.color_scheme }} gradient">
   <div class="page-width{% if section.settings.heading == blank %} no-heading{% endif %} section-{{ section.id }}-padding">
     <h2 class="collage-wrapper-title">{{ section.settings.heading | escape }}</h2>
-    <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %} isolate">
+    <div class="collage isolate{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %}">
       {%- for block in section.blocks -%}
         <div class="collage__item collage__item--{{ block.type }} collage__item--{{ section.settings.desktop_layout }}" {{ block.shopify_attributes }}>
           {%- case block.type -%}
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -22,7 +22,7 @@
 <div class="color-{{ section.settings.color_scheme }} gradient">
   <div class="page-width{% if section.settings.heading == blank %} no-heading{% endif %} section-{{ section.id }}-padding">
     <h2 class="collage-wrapper-title">{{ section.settings.heading | escape }}</h2>
-    <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %}">
+    <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %} isolate">
       {%- for block in section.blocks -%}
         <div class="collage__item collage__item--{{ block.type }} collage__item--{{ section.settings.desktop_layout }}" {{ block.shopify_attributes }}>
           {%- case block.type -%}
--- a/sections/collection-list.liquid
+++ b/sections/collection-list.liquid
@@ -19,7 +19,7 @@
 {%- endstyle -%}
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
-  <div class="collection-list-wrapper page-width{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %} section-{{ section.id }}-padding isolate">
+  <div class="collection-list-wrapper page-width section-{{ section.id }}-padding isolate{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %}">
     {% unless section.settings.title == blank %}
       <div class="title-wrapper-with-link{% if section.settings.swipe_on_mobile == true %} title-wrapper--self-padded-tablet-down{% else %} title-wrapper--self-padded-mobile{% endif %}{% if section.settings.title == blank %} title-wrapper-with-link--no-heading{% endif %} title-wrapper--no-top-margin">
         <h2 class="collection-list-title">{{ section.settings.title | escape }}</h2>
--- a/sections/collection-list.liquid
+++ b/sections/collection-list.liquid
@@ -19,7 +19,7 @@
 {%- endstyle -%}
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
-  <div class="collection-list-wrapper page-width{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %} section-{{ section.id }}-padding">
+  <div class="collection-list-wrapper page-width{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %} section-{{ section.id }}-padding isolate">
     {% unless section.settings.title == blank %}
       <div class="title-wrapper-with-link{% if section.settings.swipe_on_mobile == true %} title-wrapper--self-padded-tablet-down{% else %} title-wrapper--self-padded-mobile{% endif %}{% if section.settings.title == blank %} title-wrapper-with-link--no-heading{% endif %} title-wrapper--no-top-margin">
         <h2 class="collection-list-title">{{ section.settings.title | escape }}</h2>
--- a/sections/featured-collection.liquid
+++ b/sections/featured-collection.liquid
@@ -31,7 +31,7 @@
 -%}
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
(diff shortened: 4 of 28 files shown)


sections/collage.liquid
```diff
@@ -22,7 +22,7 @@
 <div class="color-{{ section.settings.color_scheme }} gradient">
   <div class="page-width{% if section.settings.heading == blank %} no-heading{% endif %} section-{{ section.id }}-padding">
     <h2 class="collage-wrapper-title">{{ section.settings.heading | escape }}</h2>
-    <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %}">
+    <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %} isolate">
```
ludoboludo, line 25: Usually good to have the class that we know are always there before the conditional statements.
```suggestion
    <div class="collage isolate{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %}">
```

ludoboludo (commented):
Looking good I didn't notice any issues while checking it (just the class position). 

Noticed this but not specific to this PR:

sofiamatulis:
> Looking good I didn't notice any issues while checking it (just the class position).
> 
> Noticed this but not specific to this PR: 

@ludoboludo Good catch! Created an issue here: https://github.com/Shopify/dawn/issues/1214

sections/collection-list.liquid
```diff
@@ -19,7 +19,7 @@
 {%- endstyle -%}
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
-  <div class="collection-list-wrapper page-width{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %} section-{{ section.id }}-padding">
+  <div class="collection-list-wrapper page-width{% if section.settings.swipe_on_mobile == true %} page-width-desktop{% endif %}{% if section.settings.title == blank %} no-heading{% endif %}{% if section.settings.show_view_all == false or section.blocks.size > collections.size %} no-mobile-link{% endif %} section-{{ section.id }}-padding isolate">
```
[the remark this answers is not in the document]
sofiamatulis: Here we have `section-{{ section.id }}-padding` at the end so I think it would make sense to keep it at the end 🤔 Unless we move both. What do you think?

sections/featured-collection.liquid
```diff
@@ -32,7 +32,7 @@
 -%}
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
-  <div class="collection page-width{% if section.settings.swipe_on_mobile == true and section.settings.collection.all_products_count > 2 and section.settings.products_to_show > 2 %} page-width-desktop{% endif %} section-{{ section.id }}-padding">
+  <div class="collection page-width{% if section.settings.swipe_on_mobile == true and section.settings.collection.all_products_count > 2 and section.settings.products_to_show > 2 %} page-width-desktop{% endif %} section-{{ section.id }}-padding isolate">
```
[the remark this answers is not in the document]
sofiamatulis: Same as here 🤔 https://github.com/Shopify/dawn/pull/1209#discussion_r787823131

assets/base.css
```diff
@@ -2741,3 +2741,7 @@ details-disclosure > details {
   border-left: none;
   border-right: none;
 }
+
+.isolate {
+  isolation: isolate;
+}
```
kmeleta, line 2747: So I guess what I didn't realize was how many places this fix (isolate or zindex/position) was needed now with the color scheme changes on most sections. This isolation fix is a great replacement for what we're currently doing but I don't think either strategy is going to cut it in the long run. Reseting the z context in anyway ruins the relationship between peer elements. So this technique we're using fixes elements _within_ the new context, but breaks the relationship between sections.


One area where I know we need this to work properly is image with text, which fortunately doesn't need this fix or have this problem, but with the ability to reduce section spacing on any section I think we need to fix this another way (right?). Really I think our problem is with the background color. We may need to look at relocating this maybe also be on it's own pseudo element with a lower z-index or something 😅 

Does this make sense? Am I missing something
cc: @tyleralsbury @ludoboludo

kmeleta: > Maybe in this case we can just keep it for the buttons/inputs
This might be safe, but yeah I think we'll need to come up with something else for any more significant elements.

sofiamatulis: I can revert it for `.card .card__inner .card__media` and keep it everywhere else

kmeleta: Well so it's not just cards. If you put the isolate class (or z-index/position) on a multicolumn parent, it's changing the context there too for the content container shadows. I believe (if im not confusing myself) if we fix the section background color dilemma, we won't need any of z context changes like this on buttons or otherwise.

sofiamatulis: In that case, we would only apply `isolate` to product recommendations, featured product, contact form, localization selectors and newsletter since theyre the only ones without cards

kmeleta: > Or perhaps we can add isolate to a higher level to create a new stacking context between the sections and avoid what you have in the screenshot.

Feel free to try it, I definitely could be missing something obvious here, but I think the problem is we want shadows to be below everything anywhere on the page except background color, but z-index: -1 puts it behind _everything_ including the elements with a background color. So those 2 things are at odds.

I'm curious what Tyler or Ludo thinks, but I guess I'm not against merging this as is just because it shouldn't be breaking anything that isn't already "broken" on main.

ludoboludo: So that fix was originally needed and added because of when we were using a gradient as the background. When it's set to background color it was fine but when using a gradient, the shadow of the elements within the section would disappear. 
If `isolate` is not changing the current behaviour, it's maybe a good thing to update to reduce the amount of CSS but doesn't feel like a high priority as it's a refactor vs a fix. 

About the relationship between section and specifically when the shadow would potentially overlap it. Were we maybe thinking of adding some spacing dynamically based on the vertical offset value for the shadow ? Cause if we did then it that would kinda solve our issue no ? So if the value is negative then apply some spacing to the top and to the bottom when positive. Tho it can't be perfect either due to the `blur` effect/value 🤔

tyleralsbury: They can use a combination of vertical spacing between sections and padding settings to adjust things.  I don't know if it's possible for us to account for all of the most extreme circumstances in a way that doesn't create wacky spaghetti code all over the place.

We do need to weigh technical debt against perfection, and if there are already settings that can alleviate the majority of expected cases I think we're okay with some of the more out there things not being 100% perfect.  We can create an issue to explore it, but it's the sort of thing that puts up red flags in my head as being maybe _**possible**_ but not _**reasonable**_ to try and fix.

kmeleta: I was probably thinking of more z-indexception 😬 .. background color styles go on a pseudo element with an even lower z-index than the shadows more so that adding padding based on shadow offset. But as Tyler pointed out, you're never forced into the issue as you can always manually add section padding to avoid it. So let's ignore that for now.

melissaperreault:
> Looking good I didn't notice any issues while checking it (just the class position).
   Noticed this but not specific to this PR:

@ludoboludo what were you settings combination? I can't replicate, checked on main as well. 🤔

melissaperreault (approved):
Thanks Sofia! LGTM! 🚢 🤞

ludoboludo (commented):
I was checking it again and I think we most likely need to apply this class to each section to make sure that the overlapping of shadows is consistent across the board. 

So for example I see some odd things with [email signup](https://screenshot.click/22-01-16g56-ikrfg.png) and [rich text](https://screenshot.click/22-01-4k9rr-ewo2y.png) in terms of overlapping shadows. 

Obviously it could be fixed with section padding and spacing but ideally the behaviour is the same across the board 👌 

Other things I noticed while testing, which aren't related to this PR but I want to log them:
- Wrong order of color scheme options for Image with text: [screenshot](https://screenshot.click/22-01-mqeda-yward.png)
- Video section box-shadow is using the wrong color. Should be using --color-shadow over --color-foreground: [screenshot](https://screenshot.click/22-01-l0oaf-3pirl.png)

kmeleta:
> Obviously it could be fixed with section padding and spacing but ideally the behaviour is the same across the board 👌

Image with text is also susceptible to having its shadow hidden by adjacent sections if we wanted to be consistent. Which I guess I agree with (and still don't like), but we should have @melissaperreault do another pass of this after to make sure this change in shadow overlap isn't going to noticeably change any of the current AD.

ludoboludo:
Actually it's her comment on one of my PR that made me check all sections: https://github.com/Shopify/dawn/pull/1232#pullrequestreview-857416301

kmeleta:
> Actually it's her comment on one of my PR that made me check all sections

Ah yes I see. With the current implementation and with the updated approach in this PR, the result is the same wherever applied. It basically resets the z-index stack to how it worked originally. Overlaps won't occur on elements _after_ other elements because of the way css is. Which kinddd of eliminates the reason we started rendering shadows on the pseudo elements in the first place 😅 . Not fully, but kind of. It's still necessary for cards in collections not overlapping other cards for example, but it does negate the section to section benefit of that approach.

tyleralsbury:
One potential option for this would be to only render the `color-scheme` class on a section when it is different from the global setting - so that otherwise, its background would be transparent.  It feels really clunky to me, though, and would be something that a true setting override can solve - where the setting only renders if it's been detached from its global equivalent.  This may just be a limitation of having colour schemes for each section.

PR #1210, opened 2022-01-18
Add word-break to Image with Text

### LINKED ISSUE #1092: Image with text: long words cut off
_Note: this is my first bug report. Please let me know if I missed anything!_ 

In Image with text section: when a word is too long for a single line, it's cut off, making the remainder of the word invisble.  

Words too long for one line would be truncated with a - and continue on the next line.

- Dawn Version: 2.5.0
- Chrome Version 97.0.4692.71
- macOS Version 12.1

**Possible solution**
Establish max character count that can display on a single line, implement a line break to spread the word across multiple lines when needed. 

- Using desktop browser 
- Font size scale is set to 150%
- Font selected is Bodoni Poster Black 

Image-with-text

LucasLacerdaUX:

Fixes #1092

Added `word-break` property.

**Demo links**

- [Store](url)
- [Editor](url)

### DIFF
--- a/assets/component-image-with-text.css
+++ b/assets/component-image-with-text.css
@@ -311,6 +311,7 @@
 }
 
 .image-with-text__heading {
+  word-break: break-word;
   margin-bottom: 0;
 }
 


ludoboludo (dismissed):
👌

melissaperreault:
I  am not  sure this  is the right  fix  for  that issue. Why? Merchants don't have control how line can break and we might cause some very wrong word break like the [Y here being alone on its line](https://screenshot.click/22-01-kl0s3-zfuhu.png)  without an hyphen to indicate the continuity of the word.

It  doesn't seem like this is localization  proof when I  read "_Word breaks should not be used for Chinese/Japanese/Korean (CJK) text."  [from this resource](https://www.w3schools.com/cssref/css3_pr_word-break.asp).

cc. @katycobb If you have any resources to help us understand the English standard rules, that would help  🙏

ludoboludo:
> I am not sure this is the right fix for that issue. Why? Merchants don't have control how line can break and we might cause some very wrong word break like the [Y here being alone on its line](https://screenshot.click/22-01-kl0s3-zfuhu.png) without an hyphen to indicate the continuity of the word.
> 
> It doesn't seem like this is localization proof when I read "_Word breaks should not be used for Chinese/Japanese/Korean (CJK) text._" [from this resource](https://www.w3schools.com/cssref/css3_pr_word-break.asp).
> 
> cc. @katycobb If you have any resources to help us understand the English standard rules, that would help 🙏

I think this is then something we could look into across the theme. As it's a technique used in many places to prevent the content from overflowing. We could use it along with the `hypen` property as well. 

But if we want a different approach that doesn't cut words we will need more exploration I think and whatever the fix might be it will need to be applied in many places across the theme. I can see 14 instances of `word-break: break-word` not counting the one being added in this PR.

sofiamatulis (approved):
Nice! Maybe we can add it to the caption too?
https://screenshot.click/19-06-ar4iy-umeea.png

tyleralsbury:
> > I am not sure this is the right fix for that issue. Why? Merchants don't have control how line can break and we might cause some very wrong word break like the [Y here being alone on its line](https://screenshot.click/22-01-kl0s3-zfuhu.png) without an hyphen to indicate the continuity of the word.
> > It doesn't seem like this is localization proof when I read "_Word breaks should not be used for Chinese/Japanese/Korean (CJK) text._" [from this resource](https://www.w3schools.com/cssref/css3_pr_word-break.asp).
> > cc. @katycobb If you have any resources to help us understand the English standard rules, that would help 🙏
> 
> I think this is then something we could look into across the theme. As it's a technique used in many places to prevent the content from overflowing. We could use it along with the `hypen` property as well.
> 
> But if we want a different approach that doesn't cut words we will need more exploration I think and whatever the fix might be it will need to be applied in many places across the theme. I can see 14 instances of `word-break: break-word` not counting the one being added in this PR.

Agreed - this is the sort of thing we can look at with fresh eyes and work on iteratively when we figure out what all of our options are.  We can audit the theme and make changes based on that.

We also need to take some time to look at internationalization efforts for the theme - Right to Left languages, etc... are definitely not 100% properly supported.  @ludoboludo had explored the use of logical properties to alleviate this with regards to things like paddings, etc... so I think that improving the overall behaviour of word breaks more broadly (especially considering other languages) falls into that work.

What do you think @melissaperreault ?

assets/component-image-with-text.css
```diff
@@ -311,10 +311,14 @@
 }
 
 .image-with-text__heading {
+  word-break: break-word;
+  hyphens: auto;
```
ludoboludo, line 315: The one thing I'd say is that the `hyphens` property isn't used across the theme so it might feel a bit different here compared to other parts/sections of the theme.

LucasLacerdaUX:
That's a really good point, Meli. I agree that this probably isn't the best solution for this issue. I've added `hyphens: auto` to the issue. It does not work with every font available nor in every browser/OS, but it's definitely better than having no separator at all.

I'm also aligned with what Ludo and Tyler that the best solution isn't as easy as adding a CSS property and will require more exploration, so it can be applied over the entire theme. But for the time being, I think `word-break` is better than having long words breaking the layout and hiding content. What do you think?

If we agree that we should move forward with this approach, I can create a separate issue so we can revisit this later - for the entire theme.

assets/component-image-with-text.css
ludoboludo: Yeah I think it gives a better visual cue that it's the same word. I know in french it makes to see/use it. Not sure for other languages tho 🤔

katycobb: Thanks for the tag @melissaperreault! Agreed adding a hyphen is an improvement for now. I'd love to be involved when we pick this up later - seems like there is a lot of intricacies based on the language being used.

ludoboludo: but then maybe we add `hyphens` in the 14 other places we use `word-break` was my thought 🤔  So that we don't have an hyphen on image with text but not on say image banner 🤔 
Would be a quick thing to add. Im not married to the idea but I think having the same behaviour everywhere would be a plus

katycobb: @ludoboludo  is there a way for me to see how the hyphenation works? I don't currently see it in the editor link above.

tyleralsbury: https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

This is the documentation for hyphens in CSS. We would be using `hyphens: auto` in this solution

tyleralsbury: > Hyphenation rules are language-specific. In HTML, the language is determined by the lang attribute, and browsers will hyphenate only if this attribute is present and the appropriate hyphenation dictionary is available. In XML, the xml:lang attribute must be used.

Looks like the CSS property accounts for language based hyphenation, so since our `html` tag does use the `lang` attribute, we might get baked in i18n support.

sofiamatulis (approved):
Thanks for the changes 🚢

assets/component-image-with-text.css
ludoboludo: Looks like support is a bit [hit or miss depending on the language](https://screenshot.click/22-01-u3sn9-p0auh.png).

Here is where I see it working as expected @katycobb: [screenshot](https://screenshot.click/22-01-qinj9-mopuo.png). I had to resize my browser window a bit to have it come up.

Though it does not come up on the title for some reasons. 🤔  I tried changing the word too but no hyphen

melissaperreault: > Though it does not come up on the title for some reasons. 🤔 I tried changing the word too but no hyphen

It doesn't because some font don't _support_ it ([Video reference](https://screenshot.click/22-01-t5nrj-15hww.mp4))

Issue #1157, opened 2022-01-11
Focus outline on slider items is cut out at the top and bottom


ludoboludo:
**Describe the current behavior**

Focus outlines are cut out a bit at the top and bottom:
![](https://screenshot.click/22-01-p00oi-zthd1.png)

**Describe the expected behavior**

We should be able to see the focus ring properly.

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 2.5.0

Issue #1217, opened 2022-01-18
Slider Component: Box shadow on elements within a slider can get cut off


nicklepine:
Box shadow on elements within a slider can get cut off

![image](https://user-images.githubusercontent.com/8127145/150031727-64853f54-6db5-46cf-9825-7af1bf5c4dc4.png)

melissaperreault:
This is interesting because @kmeleta and I did have a fix for this not so long ago. I also noticed another odd thing:

- [ ] Featured collection: When Cards style are set to `Standard`  + have positive vertical shadow offset with blur, the space  below the cards is increased and it is not as appealing like it can be for the style `Cards`.  Looks like a bug and that it has  too much negative space. ([Video reference](https://screenshot.click/22-01-ij450-ww0tc.mp4))

kmeleta:
> * [ ]  Featured collection: When Cards style are set to `Standard`  + have positive vertical shadow offset with blur, the space  below the cards is increased and it is not as appealing like it can be for the style `Cards`.  Looks like a bug and that it has  too much negative space. ([Video reference](https://screenshot.click/22-01-ij450-ww0tc.mp4))

@melissaperreault would you consider this card specific issue to be sev 1? Or can it wait until we have more time to improve the shadow padding logic in general?

nicklepine:
@melissaperreault @kmeleta a similar issue is happening on the checkout button on the cart template. Is it possible to address on the same PR? 
![image](https://user-images.githubusercontent.com/8127145/150266924-bb86a851-19c7-4faf-a72f-1a6de2fa02e7.png)

kmeleta:
Padding could certainly help if @melissaperreault wants to advise on a value. Otherwise if we didn't want to mess with padding, what you're seeing I think is more related to the PR that @sofiamatulis has up https://github.com/Shopify/dawn/pull/1209. The background of the section below is covering the shadow of the checkout button because of the shadow's -1 z-index. We could probably add it to that as well. Thoughts?

kmeleta:
Fix for the original issue has been resolved in https://github.com/Shopify/dawn/pull/1231. The standard card exception Meli mentioned is being handled in a follow up PR.

PR #1231, opened 2022-01-19
Re-apply slider shadow padding

kmeleta:

Refs #1217 
Fixes https://github.com/Shopify/dawn/issues/1157

- Fixes slider shadow padding on non-product media sliders
- Adds some additional padding to the same logic for focus outlines

Reverted a change from a previous commit that limited the scope of the shadow padding styles.
Also noticed that the blogs slider may have been prevented from receiving the proper shadow offset values.

**Other considerations**

There are other improvements we can make to this logic, but this commit just aims to get it back to working for all sliders in the theme.

Testing notes:
- Collection list (tablet/mobile)
- Featured collection (tablet/mobile)
- Multicolumn (mobile)
- Featured blog (tablet/mobile)
- Product media gallery (mobile)

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127460999190/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -157,6 +157,7 @@
 
 .product-grid,
 .collection-list,
+.blog__posts,
 .card {
   --border-radius: var(--card-corner-radius);
   --border-width: var(--card-border-width);
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -24,9 +24,6 @@ slider-component {
     scroll-padding-left: 1rem;
     -webkit-overflow-scrolling: touch;
     margin-bottom: 1rem;
-  }
-
-  .slider--mobile.product__media-list {
     padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
     padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
   }
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -24,8 +24,8 @@ slider-component {
     scroll-padding-left: 1rem;
     -webkit-overflow-scrolling: touch;
     margin-bottom: 1rem;
-    padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
-    padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
+    padding-top: max(0.5rem, calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius)));
+    padding-bottom: max(0.5rem, calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius)));
   }
 
   .slider.slider--mobile .slider__slide {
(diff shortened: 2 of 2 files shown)


assets/base.css
```diff
@@ -157,6 +157,7 @@
 
 .product-grid,
 .collection-list,
+.blog__posts,
 .card {
```
kmeleta, line 161: We can update these to use shared classes in the future.

assets/component-slider.css
```diff
@@ -24,9 +24,6 @@ slider-component {
     scroll-padding-left: 1rem;
     -webkit-overflow-scrolling: touch;
     margin-bottom: 1rem;
-  }
-
-  .slider--mobile.product__media-list {
     padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
     padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
```
ludoboludo, line 28: I think we might always need to have a bit of padding top and bottom whatever the vertical offset value is 🤔  maybe using a `max(0.5rem, calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius)))` to account for focus styles:

kmeleta: Edit: Actually, using max() may straighten that up nicely. Good call. Will update if all looks well 👍

kmeleta: Donezo

martinamarien (changes requested):
Let me know if you think this should be in a different PR.

ludoboludo:
It fixes this too: https://github.com/Shopify/dawn/issues/1157

assets/component-slider.css
```diff
@@ -64,8 +61,8 @@ slider-component {
     scroll-padding-left: 1rem;
     -webkit-overflow-scrolling: touch;
     margin-bottom: 1rem;
-    padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
-    padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
+    padding-top: max(0.5rem, calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius)));
+    padding-bottom: max(0.5rem, calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius)));
```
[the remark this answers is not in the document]
ludoboludo: good point, so should be conditional to a class applied based on the card/standard layout then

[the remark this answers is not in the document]
kmeleta: That's definitely valid and I believe Meli mentioned it in the ticket too. If the card discrepancy is also a sev 1 concern that we want done asap I can play with it before merging this. Otherwise, the current fix in this PR is definitely sev 1 and should be merged separately so it doesn't potentially get held up by reviews/discussions for the card aspect.

kmeleta:
> It fixes this too: #1157 if we add the `max(...` for focus style.

Thanks Ludo I'll add it to the PR description

Issue #1043, opened 2021-12-17
Collapsible content section: Apply global settings


ludoboludo:
**Describe the change needed**

Collapsible content needs an update to have it's content tweaked accordingly to the general settings added: 

- media border/shadow
- general padding
- content container border/shadow

This is a follow up to #956 

![](https://screenshot.click/21-12-3eago-untsv.png)

tyleralsbury:
I would split this work in two.

The media stuff is very easy - it can be done really quickly by adding a single class @martinamarien 

The content stuff has considerations based on where the content background is applied.

ludoboludo:
For the media stuff whoever ends up looking into it should test with a png with a transparent background and some media shadow: https://github.com/Shopify/dawn/pull/1220#discussion_r788061036

PR #1232, opened 2022-01-19
Collapsible content media with global settings

ludoboludo:

Tackles part of #1043 

Added the `global-media-settings` class on the media wrapper and applied on the section, this styling: 
```css
position: relative;
z-index: 0;
```
So that the shadow could come up. Same fix is applied on other sections and there is a PR looking into refactoring that approach: https://github.com/Shopify/dawn/pull/1209

**To test**

Go to the editor and play around with the global media settings

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127461195798)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127461195798/editor)

**Checklist**

### DIFF
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -10,8 +10,9 @@
   }
 }
 
-.collapsible-content__media {
-  background-color: transparent;
+.collapsible-content {
+  position: relative;
+  z-index: 0;
 }
 
 .collapsible-content__media--small {
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -10,11 +10,6 @@
   }
 }
 
-.collapsible-content {
-  position: relative;
-  z-index: 0;
-}
-
 .collapsible-content__media--small {
   height: 19.4rem;
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -252,6 +252,11 @@ html.no-js .no-js-hidden {
   }
 }
 
+.isolate {
+  position: relative;
+  z-index: 0;
+}
+
 .section + .section {
   margin-top: var(--spacing-sections-mobile);
 }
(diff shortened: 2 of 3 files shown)


assets/collapsible-content.css
```diff
@@ -10,8 +10,9 @@
   }
 }
 
-.collapsible-content__media {
-  background-color: transparent;
+.collapsible-content {
+  position: relative;
+  z-index: 0;
```
tyleralsbury, line 15: Should this be using `isolate` cc @sofiamatulis ?

ludoboludo: It could be I figured i'd use what is currently being used on the theme at this moment.

sofiamatulis: We can re-use the class `.isolate` that I have added on my PR https://github.com/Shopify/dawn/pull/1209

melissaperreault (commented):
Thank you Ludo!

**Observation**

- [ ] Shadow vertical offset negative would [overlap other media](https://screenshot.click/22-01-7yozk-1ufgw.mp4), but not when the value is positive and not if it is a [section](https://screenshot.click/22-01-5clfc-64lzs.png)
- [ ] Just a note for me to check where we landed for the [shadow potentially hiding titles](https://screenshot.click/22-01-67g0e-lx5xr.png), labels, etc. cc. @nicklepine 

Otherwise looks good!

ludoboludo:
>Shadow vertical offset negative would overlap other media, but not when the value is positive and not if it is a section

@melissaperreault I believe it's known and technically logic for the shadow of an element to be hover what precedes them. It's the natural "stacking" context. Think of it as a stair case. The first step from the bottom being the first section, then the second is above it and so on. 

But it can be manually fixed. We're already adding this kind of styling to quite a few sections: 
```css
position: relative;
z-index: 0;
```

Which changes the stacking context. Maybe it should be applied to each section 🤔  And when we're moving on with Sofia's PR it could mean adding here `isolate` class on each section. 

Since it's not something that we're introducing in this PR, I'd say maybe we can add this the to do list in her refactor PR: https://github.com/Shopify/dawn/pull/1209

cc: @tyleralsbury

PR #1237, opened 2022-01-20
Change page width setting to use range

tyleralsbury:

Attempting to use a range setting rather than a dropdown for page width.

There are now 6 distinct sizes the merchant can use - 1000 to 1600 in increments of 100.  We can adjust this.

We need to test a number of things to validate these new ranges

* Different settings - things like gap, section specific settings, the media settings on the product page, etc...
* Images - are they rendering at the correct size in each section?  The way we implemented it, they should, be we need to test.  There may be some sizes missing.
* This should have no changes to mobile

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=127464472598)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127464472598/editor)

**Checklist**

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -34,6 +34,10 @@
     padding-left: 5rem;
   }
 
+  .product__info-wrapper--extra-padding {
+    padding-left: 8rem;
+  }
+
   .product__media-container .slider-buttons {
     display: none;
   }
@@ -343,6 +347,10 @@ a.product__text {
 }
 
 @media screen and (min-width: 750px) {
+  .product__info-container {
+    max-width: 60rem;
+  }
+  
   .product__info-container .price--on-sale .price-item--regular {
     font-size: 1.6rem;
   }
@@ -416,6 +424,10 @@ a.product__text {
 .product--no-media fieldset.product-form__input {
   flex-wrap: wrap;
   margin: 0 auto 1.2rem auto;
+}
+
+.product-form__buttons {
+  max-width: 44rem;
 }
 
 .product--no-media .product__info-container > modal-opener {
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -137,21 +137,15 @@
     "name": "t:settings_schema.layout.name",
     "settings": [
       {
-        "type": "select",
+        "type": "range",
         "id": "page_width",
-        "options": [
-          {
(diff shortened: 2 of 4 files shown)


assets/section-main-product.css
```diff
@@ -34,6 +34,10 @@
     padding-left: 5rem;
   }
 
+  .product__info-wrapper--extra-padding {
+    padding-left: 8rem;
```
ludoboludo, line 38: I don't really feel like it needs this extra padding 🤷

assets/section-main-product.css
```diff
@@ -343,6 +347,10 @@ a.product__text {
 }
 
 @media screen and (min-width: 750px) {
+  .product__info-container {
+    max-width: 60rem;
```
ludoboludo, line 351: this can make it feel like the content is left aligned and not taking the rest of space that other sections use: [video](https://screenshot.click/22-01-opirh-79o3e.mp4)

I don't think it's an issue to have the content take the available space 🤔  plus you can restrict it if you want to by reduce the layout width or change the amount of space the media take.

config/settings_schema.json
```diff
@@ -137,21 +137,15 @@
...
-            "value": "1200",
-            "label": "t:settings_schema.layout.settings.page_width.options__1.label"
-          },
-          {
-            "value": "1600",
-            "label": "t:settings_schema.layout.settings.page_width.options__2.label"
-          }
-        ],
-        "default": "1600",
+        "min": 1000,
```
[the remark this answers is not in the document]
ludoboludo: I like that idea ^ I think it looks better to have 3 per rows in that situation. Can be a follow up though 👍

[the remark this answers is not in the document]
melissaperreault: Yes, one of my observations and suggestions. Aligned if we can include or a fast follow!

assets/section-main-product.css
ludoboludo: ah smart! Sounds good.

assets/section-main-product.css
ludoboludo: Alrighty, that makes sense. Thanks for the resources Meli 👍

ludoboludo (approved):

PR #1240, opened 2022-01-20
Fix misc issues

### LINKED ISSUE #1238: Color scheme order for image with text

The order of color scheme is the same everywhere except on Image with text:

### LINKED ISSUE #1239: Video section not using the proper box shadow color

All shadows are using the new custom property `--color-shadow` but not the video section. It's using `--color-foreground` so when using a color scheme with a white font, the shadow is white instead of black.

ludoboludo:

Fixes #1238 
Fixes #1239 

Updated the order and color variable used for box shadow. 

**Testing**

Check the image with text section and make sure the order of the color scheme options is the same as for other sections. 

Then check the media shadow for the video section, it should be using the text color set for the theme and not `--color-foreground` which changes based on the color scheme. 
Here is the before and after: [screenshot](https://screenshot.click/22-01-e57ju-x7ff8.png)

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127464603670)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127464603670/editor)

**Checklist**

### DIFF
--- a/assets/video-section.css
+++ b/assets/video-section.css
@@ -4,7 +4,7 @@
 }
 
 .video-section__media.deferred-media {
-  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-foreground), var(--media-shadow-opacity));
+  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-shadow), var(--media-shadow-opacity));
 }
 
 .video-section__media.deferred-media:after {
--- a/sections/image-with-text.liquid
+++ b/sections/image-with-text.liquid
@@ -195,6 +195,14 @@
       "id": "color_scheme",
       "options": [
         {
+          "value": "accent-1",
+          "label": "t:sections.image-with-text.settings.color_scheme.options__4.label"
+        },
+        {
+          "value": "accent-2",
+          "label": "t:sections.image-with-text.settings.color_scheme.options__5.label"
+        },
+        {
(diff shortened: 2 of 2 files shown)


assets/video-section.css
```diff
@@ -4,7 +4,7 @@
 }
 
 .video-section__media.deferred-media {
-  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-foreground), var(--media-shadow-opacity));
+  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-shadow), var(--media-shadow-opacity));
 }
```
sofiamatulis, line 8: The shadow doesn't seem to work on `accent-1` and `inverse`

https://screenshot.click/20-25-98oj1-cds2z.png

sofiamatulis (commented):
Looks great! Just noticed a small thing

assets/video-section.css
ludoboludo: It's expected. The shadow is using the general text color set. So if you're setting one of those background to be the same color than the text, you won't see it. 

Here is the original issue to change the color source: https://github.com/Shopify/dawn/issues/1026

melissaperreault: Yeah exactly where I was going to answer, I think this was overlooked, for `Inverse` style, we could have an exception to pick `rgba(var(--color-foreground)`. But I need to think further about it.

melissaperreault: That would be just in the scenario of a component that can receive shadow is on top of another container that can receive color scheme though.

PR #1292, opened 2022-01-31
Update slider shadow padding implementation

### LINKED ISSUE #1217: Slider Component: Box shadow on elements within a slider can get cut off

kmeleta:

Fixes #1217 

We previously added additional padding above and below sliders to compensate for shadow offset. This change makes it possible to remove the additional bottom padding specifically for sliders that contain standard style cards.

Previously, slider shadow padding was applied to the `<ul class="slider">` element. I've changed the approach to put this padding on the `<li class="slider__slide">` elements instead. This was mostly to accommodate a case in collection list where empty collection cards still need the bottom padding regardless of which card style is used ([see here](https://screenshot.click/31-54-tk052-4ij2t.png))

In order to better scope generic shadow variables and account for exception cases I've created new classes to describe the type of children a slider/grid contains. I think these could be utilized more to further simplify global setting variable inheritance, but for now I'm using only where relevant to this issue.

Using the same method, I also added an exception that prevents the previously added .5rem for focus outlines on content-container sliders, which don't receive focus.

**Other considerations**

Probably a UX decision, but instead of just removing bottom padding we could add it between the card image and the title. I don't think the further separation of these two elements looks particularly good, but maybe for preserving contrast we should consider it. Perhaps this is better solved by providing a title spacing setting or something.

Struggled naming the new contains classes. Other ideas welcome.

This PR doesn't modify the actual shadow offset/blur calculations we're using for the padding. We can probably still do more to improve this but should be considered separately from this PR.

Testing notes:

- Check for padding/margin related regressions in all slider sections
- Check with positive and negative vertical shadow offsets

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127526731798/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -155,9 +155,8 @@
   --color-icon: rgb(var(--color-base-outline-button-labels));
 }
 
-.product-grid,
-.collection-list,
-.blog__posts,
+.contains-card--standard,
+.contains-card--card,
 .card {
   --border-radius: var(--card-corner-radius);
   --border-width: var(--card-border-width);
@@ -168,8 +167,8 @@
   --shadow-opacity: var(--card-shadow-opacity);
 }
 
-.multicolumn-list,
-.multicolumn-card {
+.contains-content-container,
+.content-container {
   --border-radius: var(--text-boxes-radius);
   --border-width: var(--text-boxes-border-width);
   --border-opacity: var(--text-boxes-border-opacity);
--- a/assets/base.css
+++ b/assets/base.css
@@ -155,6 +155,7 @@
   --color-icon: rgb(var(--color-base-outline-button-labels));
 }
 
+.contains-card,
 .contains-card--standard,
 .contains-card--card,
 .card {
--- a/assets/base.css
+++ b/assets/base.css
@@ -156,8 +156,6 @@
 }
 
 .contains-card,
-.contains-card--standard,
-.contains-card--card,
 .card {
   --border-radius: var(--card-corner-radius);
   --border-width: var(--card-border-width);
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -10,8 +10,12 @@ slider-component {
 }
 
 .slider__slide {
+  --focus-outline-padding: 0.5rem;
+  --shadow-padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
+  --shadow-padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
   scroll-snap-align: start;
   flex-shrink: 0;
+  padding-bottom: 0;
 }
 
 @media screen and (max-width: 749px) {
@@ -24,13 +28,20 @@ slider-component {
     scroll-padding-left: 1.5rem;
     -webkit-overflow-scrolling: touch;
     margin-bottom: 1rem;
-    padding-top: max(0.5rem, calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius)));
-    padding-bottom: max(0.5rem, calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius)));
   }
 
   .slider.slider--mobile .slider__slide {
     margin-bottom: 0;
-    padding-bottom: 0;
+    padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
+    padding-bottom: max(var(--focus-outline-padding), var(--shadow-padding-bottom));
+  }
+
+  .slider.slider--mobile.contains-card--standard .slider__slide:not(.collection-list__item--no-media) {
+    padding-bottom: var(--focus-outline-padding);
+  }
+
+  .slider.slider--mobile.contains-content-container .slider__slide {
+    --focus-outline-padding: 0rem;
   }
 }
 
@@ -47,7 +58,6 @@ slider-component {
 
   .slider.slider--tablet-up .slider__slide {
     margin-bottom: 0;
-    padding-bottom: 0;
   }
 }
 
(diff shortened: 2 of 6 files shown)


assets/base.css
```diff
@@ -155,9 +155,8 @@
   --color-icon: rgb(var(--color-base-outline-button-labels));
 }
 
-.product-grid,
-.collection-list,
-.blog__posts,
+.contains-card--standard,
+.contains-card--card,
```
ludoboludo, line 159: Since you only have modifiers atm, I would suggest applying those variables to `.contains-card` and add that class where you've added `.contains-card--style`.

assets/component-slider.css
```diff
@@ -24,13 +28,20 @@ slider-component {
...
     margin-bottom: 1rem;
-    padding-top: max(0.5rem, calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius)));
-    padding-bottom: max(0.5rem, calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius)));
   }
 
   .slider.slider--mobile .slider__slide {
     margin-bottom: 0;
-    padding-bottom: 0;
+    padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
+    padding-bottom: max(var(--focus-outline-padding), var(--shadow-padding-bottom));
```
ludoboludo, line 36: Should we also have a padding-top/bottom value when it's not a slider, either on the `ul` or `li:last-child`/`li:first-child`

assets/component-slider.css
```diff
@@ -61,13 +71,20 @@ slider-component {
...
   }
 
   .slider.slider--tablet .slider__slide {
     margin-bottom: 0;
-    padding-bottom: 0;
+    padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
+    padding-bottom: max(var(--focus-outline-padding), var(--shadow-padding-bottom));
+  }
+
+  .slider.slider--tablet.contains-card--standard .slider__slide:not(.collection-list__item--no-media) {
```
ludoboludo, line 82: Just a very general thought. I wonder if we should look into, at some point, at a refactor for class specificity so that we don't rely on `.class.class.class` as much but instead `:where(.class.class.class) .slider__slide:not...`
🤔 
`:where()` has a specificity of zero at all time. So basically it can help a bit with specificity of styling in general. Maybe Ill explore in some free time if it would be helpful/has a positive impact.

ludoboludo (commented):
Looking good, left a few comments.

assets/base.css
```diff
@@ -168,8 +167,8 @@
   --shadow-opacity: var(--card-shadow-opacity);
 }
 
-.multicolumn-list,
-.multicolumn-card {
+.contains-content-container,
```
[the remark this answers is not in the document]
kmeleta: `li.slider-slide.grid-item` is a parent of `.content-container.multicolumn-card` so the grid/slider elements are all currently outside that scope.

assets/component-slider.css
```diff
@@ -24,13 +28,20 @@ slider-component {
...
 
   .slider.slider--mobile .slider__slide {
     margin-bottom: 0;
-    padding-bottom: 0;
+    padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
+    padding-bottom: max(var(--focus-outline-padding), var(--shadow-padding-bottom));
+  }
+
+  .slider.slider--mobile.contains-card--standard .slider__slide:not(.collection-list__item--no-media) {
+    padding-bottom: var(--focus-outline-padding);
```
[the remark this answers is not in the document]
kmeleta: If UX likes the idea of just using the same automatic padding here, I'll add it, but if it's something to be discussed more, it should be handled separately. cc @melissaperreault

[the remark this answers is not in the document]
melissaperreault: Since merchants have the ability to set the opacity to maintain good contrast, I would handle separately please because we need to explore, define and observe how this is being used. We'll consider how it affects section headings as well when the vertical offset is negative.

There might be more opportunity to scale systematically dependent of the shadow opacity, offset, and blur for instance or potentially add new settings.

assets/component-slider.css
melissaperreault: Curious to learn the settings combination as I wasn't able to get it on [Collection cards](https://screenshot.click/22-01-2uhb0-yqpp6.png), but mainly on [Multicolumn](https://screenshot.click/22-01-bjbwa-zfd8n.png).

melissaperreault (approved):
Thank you Ken!
Replied where needed and approving this PR as it is currently fixing the issue described. I would solve any follow up items in another PR/issue. Happy to pair and align on what needs alignment!

assets/base.css
kmeleta: I went back and forth so much on these. I ultimately didn't add it because it wouldn't be useful, as used currently. But it theoretically makes sense. I may opt to just wait til we're making more use of these contexts (grid gap changes and make sure I'm still going to need them the way I think.

ludoboludo: I think you could have `.contains-card` and `.contains-card--standard` 🤔  I think it'd be nice to follow BEM

assets/component-slider.css
kmeleta: I just pushed a change for that. I don't want to go down a rabbit hole in this PR, but if that looks better you, we'll roll with it.

assets/base.css
ludoboludo: Oh what I meant is maybe `contains-card` could be the default using the current styles from `contains-card--card`. And then only have one "modifier" with `--standard`. What do you think 🤔

kmeleta: Ahh, I think I like that actually. Does this make sense? https://github.com/Shopify/dawn/pull/1292/commits/d7f333e89e2157cbf8d01a218e0f2cc45d281777

ludoboludo: Yeah I think so 👍

ludoboludo (approved):
Looking good 👍

kmeleta:
> This all looks good to me. I'm wondering if there would be any value in turning the calculation into a reusable class with a name that tells us what it's there for - it's a bit complex to get your head around unless you've got the context.

Thanks @tyleralsbury I did try to break it up a bit into vars with somewhat descriptive labels like `--shadow-padding-top`. Not actually sure how better to describe the purpose without leaving a comment. I didn't make it a class yet just because its only use is in sliders. However, I think we may want to further tweak the calculation anyway so we can consider it more closely then.

tyleralsbury:
> Thanks @tyleralsbury[](https://github.com/kmeleta) I did try to break it up a bit into vars with somewhat descriptive labels like --shadow-padding-top. Not actually sure how better to describe the purpose without leaving a comment. I didn't make it a class yet just because its only use is in sliders. However, I think we may want to further tweak the calculation anyway so we can consider it more closely then.

It's probably okay as-is if it's only used for sliders.  My concern would be this calculation popping up in different places throughout the theme, making it harder to maintain.  If this isn't something we need to worry about, then I think the named variable is good enough.

PR #1293, opened 2022-01-31
Prevent scrollbar overflow in footer disclosures

### LINKED ISSUE #997: [Global setting] Fix scrollbar and corner radius relationship
### Observation

**Country/region and language inputs**
- [ ] ~**Severity 1** - The [dropdown doesn't work](https://screenshot.click/21-12-bg612-8xfy2.png)~ _Edit: Dec 13, can't replicate anymore!_
- [ ] From an old screenshot, we need to try to have a fix [for this scenario of the scroll bar and corner radius](https://screenshot.click/21-12-dr2bi-edqti.png)

kmeleta:

Fixes #997 

When the footer locale popups have corner radius, the scroll bar can overflow outside the popup. This is because the disclosure list is using `overflow-y: auto`. To keep the list scrollable but contain the content within the border radius I've wrapped the list in a new `<div class="disclosure__list-wrapper">` element. This requires some minor changes to both the css and the js. `disclosure__list-wrapper` should now be the element concerned with the global setting styles and the visibility/positioning of the dialog on the page.

New pattern:
```
<button aria-expanded="false" aria-controls="MyList" aria-describedby="MyLabel">
<div class="disclosure__list-wrapper" hidden>
  <ul id="MyList" role="list"></ul>
</div>
```

**Other considerations**

I believe the changes to markup are safe and didn't notice any changes in the way VO announces the modified pattern, but flagging here in case this does affect expectations from an a11y perspective in some way.

Tip: Disabling whitespace during review will improve readability of these diffs.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127527452694/editor)

**Checklist**

### DIFF
--- a/assets/disclosure.css
+++ b/assets/disclosure.css
@@ -32,6 +32,7 @@
   font-size: 1.4rem;
   padding-bottom: 0.5rem;
   padding-top: 0.5rem;
+  scroll-padding: 0.5rem 0;
   min-height: 8.2rem;
   max-height: 19rem;
   max-width: 22rem;


melissaperreault (commented):
Thanks Ken! Made some observation that are not due to your PR I believe. Let me know if you  would like me to create a new issue.  🙏 

Otherwise, it looks good!
Tested:
- [x] Different settings range, 
- [x] Different Color scheme
- [x] Safari, Chrome
- [x] No JS experience

**No JS**
- The tertiary button style seems they have lost their styling ([no border visible](https://screenshot.click/22-02-i2mfj-jaqmg.png)), even if border is set to 1  and 100%  opacity.  _(Issue opened https://github.com/Shopify/dawn/issues/1300)_

**Mobile** _(Issue opened https://github.com/Shopify/dawn/issues/1301)_
- Noticed on small screen the space between the select [is very narrow](https://screenshot.click/22-02-lhvod-8z9ax.png). They could wrap earlier. Tested on my Android Samsung A70
   - Seems to [happen in the editor](https://screenshot.click/22-02-4dl9u-3c6dg.png) with shorter select option. 
- Related to the above, the [dropdown is off screen](https://screenshot.click/22-02-fuycz-0uma9.png)

kmeleta:
Thanks @melissaperreault I agree, definitely things to address. But maybe let's go ahead and make an issue to clean up all that stuff so it doesn't hold up this update.

kmeleta:
> The same happens to the cart note I believe but since it's going to need a different approach it most likely should be a different issue/PR. It's also kind of an edge case since it only comes up if you write a lot of info in the cart note: [screenshot](https://screenshot.click/01-59-lghjm-rbycz.png)

Hmm yeah I remember that, guess we never logged that in the same issue. But I just did a quick audit and it's also going to be an issue with the textarea on blogs and contact. I was gonna throw that in, but I'm not sure what approach to use for it, given an overflow: hidden would also hide the focus border animation 🤔  Let's plan to handle separately for now. cc @ludoboludo

assets/disclosure.css
```diff
@@ -35,6 +26,19 @@
   box-shadow: var(--popup-shadow-horizontal-offset) var(--popup-shadow-vertical-offset) var(--popup-shadow-blur-radius) rgba(var(--color-shadow), var(--popup-shadow-opacity));
 }
 
+.disclosure__list {
+  position: relative;
+  overflow-y: auto;
+  font-size: 1.4rem;
+  padding-bottom: 0.5rem;
+  padding-top: 0.5rem;
```
ludoboludo, line 34: I wonder if we should have a bit more spacing to allow the focus state to come up fully when borders/border radius is used 🤔 
What do you think ?

ludoboludo (approved):
Looks good to me when testing and voice over was clear when navigating it. 

Just the one thing is maybe a bit more spacing to fully display the focus ring.

assets/disclosure.css
kmeleta: Looks like the full outline is always visible, minus the corner of course, which theoretically we could probably fixed with more advanced focus styles. Or do you just mean there should potentially be some white space between the outline and the border? I feel like the scrolling that happens on tab just results in that positioning sometimes so I'm not sure if we can control that with margin and padding. We could probably add a bit of `scroll-padding-top/bottom` to the list though maybe?

ludoboludo: Hmm good point, looks like adding `scroll-padding: 0.5rem 0;` seem to help with it 👍

kmeleta: Donezo

assets/disclosure.css
```diff
@@ -35,6 +26,20 @@
...
```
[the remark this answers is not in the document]
kmeleta: I'm considering the text area issue separately because it's more complex.

> > The same happens to the cart note I believe but since it's going to need a different approach it most likely should be a different issue/PR. It's also kind of an edge case since it only comes up if you write a lot of info in the cart note: [screenshot](https://screenshot.click/01-59-lghjm-rbycz.png)
> 
> Hmm yeah I remember that, guess we never logged that in the same issue. But I just did a quick audit and it's also going to be an issue with the textarea on blogs and contact. I was gonna throw that in, but I'm not sure what approach to use for it, given an overflow: hidden would also hide the focus border animation 🤔 Let's plan to handle separately for now. cc ludoboludo

assets/disclosure.css
```diff
@@ -35,6 +26,20 @@
...
+  min-height: 8.2rem;
+  max-height: 19rem;
+  max-width: 22rem;
+  min-width: 12rem;
+  width: max-content;
+}
+
 .disclosure__item {
   position: relative;
 }
```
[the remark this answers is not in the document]
melissaperreault: I am aligned that we can live with the tradeoff of cutting it off. Merchant would have other ways to go around this like reducing the border thickness and roundness if they have long list of currency. Very edge case!

melissaperreault (approved):
🎉 🚢

sofiamatulis (approved):
Nice 🎉 🚢

PR #1294, opened 2022-01-31
Use `og:type=website` for collection

michenly:
`<meta property="og:type" content="product.group" />` is not actually a valid value. Collection page should default to `<meta property="og:type" content="website" />` instead.

cc: @jackson-lo

Fixes #0.

**Other considerations**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/snippets/meta-tags.liquid
+++ b/snippets/meta-tags.liquid
@@ -8,8 +8,6 @@
     assign og_type = 'product'
   elsif request.page_type == 'article'
     assign og_type = 'article'
-  elsif request.page_type == 'collection'
-    assign og_type = 'product.group'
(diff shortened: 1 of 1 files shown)


ludoboludo (commented):
Thanks for bringing this up. 
If this needs to be changed, I think we will have to follow up to get it updated here in our docs as well: https://shopify.github.io/liquid-code-examples/example/open-graph-tags

michenly:
@ludoboludo would you be able to do that change?

ludoboludo:
> @ludoboludo would you be able to do that change?

No I think I would need to find which team owns that. I'd start asking in `#docs`.

PR #1297, opened 2022-02-01
Update slider scroll paging method

### LINKED ISSUE #834: Slider numbers are inaccurate on small mobile screens

The pagination in the slider option used in the Feature Collection and Collection List sections breaks on smaller mobile screens (below 362px in width). With larger number of products/collections, at least one page number skips, and causes subsequent page numbers to be inaccurate.

The JS that controls this uses the ratio between the total width of the section, which is responsive to screen size, and the current scroll value. `Math.round` is used to convert this ratio into a page number.

image

Page numbers should progress sequentially.

- Dawn Version: 2.3.0
- Chrome Version 94.0.4606.81
- macOS Version 11.6

**Possible solution**
Change `Math.round` to `Math.ceil` to ensure the ratio is always rounded up. This makes the page numbers work properly on both small and regular size mobile screens, in both sections.

image

[Video showing the problem](https://screenshot.click/27-39-9h72j-q5j9m.mp4)

[Video showing issue resolved after changing to Math.ceil](https://screenshot.click/27-55-vljbm-b63y7.mp4)

kmeleta:
Why are these changes introduced?

Fixes #834

Fixes inconsistencies in the scroll paging/counter logic. Also fixes [this small issue](https://screenshot.click/04-44-8ybpv-mgvi1.mp4).

**tl;dr** Mostly updated the slider js to use slide offsets instead of slide widths

I believe the issue stems from 2 different problems. In both cases, the issue is most noticeable with the featured collection slider due to its ability to show multiple slides per "page", even on the mobile breakpoint.

1. In the `grid--peek` styles, we calculate slide width using ` width: calc(50% - var(--grid-space) - 3rem)` At a small enough viewport, this would lead to 3 slides being totally visible (rather than 2), which would mess with paging calculations. The easiest fix for this is to just add min-width with a static % value to protect against the worst case scenarios (tiny viewport, maximum grid spacing).

2. Originally sliders were updating the slide counter based on the width of the slides. With recent changes to grid/slider (column-gap changes, grid spacing setting), relying on the slide width has become too fragile. To fix this, I used the offset distance between 2 slides as the value to increment pages.

**Other considerations**

I did originally look at the suggested fix of changing `math.round` to `math.ceil`. This appeared to work in some cases, but broke in others, particularly when increasing the grid gap. For context, Math.round was chosen so the current page count label would ideally change half way between slides. The type of rounding method used here shouldn't affect the paging count in any other way. If it does, there's a miscalculation somewhere else.

**Testing steps/scenarios**

- [ ] Featured collection (mobile/tablet)
- [ ] Collection list (mobile/tablet)
- [ ] Multicolumn (mobile)
- [ ] Featured blog (mobile/tablet)
- [ ] Product media (mobile)
- [ ] Slideshow
- [ ] Minimal horizontal grid spacing
- [ ] Maximum horizontal grid spacing
- [ ] Large number of slides
- [ ] Very small viewport sizes (< 360px)
- [ ] Resizing viewport

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127529648150/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1100,12 +1100,12 @@ summary::-webkit-details-marker {
     min-width: 35%;
   }
 
-  .grid--peek .grid__item:first-of-type {
-    padding-left: 1.5rem;
+  .grid--peek.slider .grid__item:first-of-type {
+    margin-left: 1.5rem;
   }
 
-  .grid--peek .grid__item:last-of-type {
-    padding-right: 1.5rem;
+  .grid--peek.slider .grid__item:last-of-type {
+    margin-right: 1.5rem;
   }
 }
 
@@ -1123,11 +1123,11 @@ summary::-webkit-details-marker {
   }
 
   .slider--tablet.grid--peek .grid__item:first-of-type {
-    padding-left: 1.5rem;
+    margin-left: 1.5rem;
   }
 
   .slider--tablet.grid--peek .grid__item:last-of-type {
-    padding-right: 1.5rem;
+    margin-right: 1.5rem;
   }
 }
 
--- a/assets/base.css
+++ b/assets/base.css
@@ -1097,6 +1097,7 @@ summary::-webkit-details-marker {
 
   .grid--peek .grid__item {
     width: calc(50% - var(--grid-mobile-horizontal-spacing) - 3rem);
+    min-width: 35%;
   }
 
   .grid--peek .grid__item:first-of-type {
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -130,7 +130,8 @@ slider-component {
 .slider-counter {
   margin: 0 1.2rem;
   display: flex;
-  min-width: 2rem;
+  justify-content: center;
+  width: 2rem;
 }
 
 @media screen and (max-width: 749px) {
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -128,16 +128,14 @@ slider-component {
 }
 
 .slider-counter {
-  margin: 0 1.2rem;
   display: flex;
   justify-content: center;
-  width: 2rem;
+  min-width: 4.4rem;
 }
 
-@media screen and (max-width: 749px) {
-  .slider-counter--dots,
-  .slider-counter--numbers {
-    margin: 0;
+@media screen and (min-width: 750px) {
+  .slider-counter--dots {
+    margin: 0 1.2rem;
   }
 }
 
--- a/assets/global.js
+++ b/assets/global.js
@@ -531,35 +531,12 @@ class SliderComponent extends HTMLElement {
     this.nextButton.addEventListener('click', this.onButtonClick.bind(this));
   }
 
-  getWaypoints() {
-    const waypoints = [];
-    for (let i = 0; i < this.totalPages; i++) {
-      if (i < this.totalPages - 1) {
-        waypoints.push(this.sliderItemsToShow[i].offsetLeft);
-      } else {
-        waypoints.push(this.slider.scrollWidth - this.slider.clientWidth);
-      }
-    };
-    return waypoints;
-  }
-
-  setNearestPage(x) {
-    const buffer = this.sliderLastItem.clientWidth / 2;
-    for (let i = 0; i < this.waypoints.length; i++) {
-      if (x < this.waypoints[i] + buffer && x > this.waypoints[i] - buffer) {
-        this.currentPage = i + 1;
-        break;
-      }
-    }
-  }
-
   initPages() {
     this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
-    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
-    if (this.sliderItemsToShow.length === 0) return;
-    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItemsToShow[0].clientWidth);
+    if (this.sliderItemsToShow.length < 2) return;
+    this.sliderItemOffset = this.sliderItemsToShow[1].offsetLeft - this.sliderItemsToShow[0].offsetLeft;
+    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItemOffset);
     this.totalPages = this.sliderItemsToShow.length - this.slidesPerPage + 1;
-    this.waypoints = this.getWaypoints();
     this.update();
   }
 
@@ -570,7 +547,7 @@ class SliderComponent extends HTMLElement {
 
   update() {
     const previousPage = this.currentPage;
-    this.setNearestPage(this.slider.scrollLeft);
+    this.currentPage = Math.round(this.slider.scrollLeft / this.sliderItemOffset) + 1;
 
     if (this.currentPageElement && this.pageTotalElement) {
       this.currentPageElement.textContent = this.currentPage;
@@ -592,7 +569,7 @@ class SliderComponent extends HTMLElement {
       this.prevButton.removeAttribute('disabled');
     }
 
-    if (this.isSlideVisible(this.sliderLastItem)) {
+    if (this.isSlideVisible(this.sliderItemsToShow[this.sliderItemsToShow.length - 1])) {
       this.nextButton.setAttribute('disabled', 'disabled');
     } else {
       this.nextButton.removeAttribute('disabled');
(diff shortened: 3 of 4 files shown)


assets/base.css
```diff
@@ -1097,14 +1097,15 @@ summary::-webkit-details-marker {
 
   .grid--peek .grid__item {
     width: calc(50% - var(--grid-mobile-horizontal-spacing) - 3rem);
+    min-width: 35%;
   }
 
-  .grid--peek .grid__item:first-of-type {
-    padding-left: 1.5rem;
+  .grid--peek.slider .grid__item:first-of-type {
+    margin-left: 1.5rem;
```
kmeleta, line 1104: The added `.slider` class here is needed for specificity, but should probably be here anyway.

Additionally, I needed to change these to use margin instead of padding so that `slide.offsetLeft` would return the actual left edge of the slide.

assets/base.css
```diff
@@ -1097,14 +1097,15 @@ summary::-webkit-details-marker {
 
   .grid--peek .grid__item {
     width: calc(50% - var(--grid-mobile-horizontal-spacing) - 3rem);
+    min-width: 35%;
```
ludoboludo, line 1100: Hmm I wonder if should think already about how this might be impacted by Sofia's PR which add a setting to choose the amount of products per row, which is also impacting the amount of cards/slide per page of the slider I think.

assets/global.js
```diff
@@ -547,7 +547,7 @@ class SliderComponent extends HTMLElement {
 
   update() {
     const previousPage = this.currentPage;
-    this.currentPage = Math.round(this.slider.scrollLeft / this.sliderLastItem.clientWidth) + 1;
+    this.currentPage = Math.round(this.slider.scrollLeft / this.sliderItemOffset) + 1;
```
ludoboludo, line 550: Something I noticed here because I might change the amount of slides with scroll by. Right now it's 1. Though we have `step` that is created and used in the`onButtonClick(event)` function. But when adding `data-step="2"` to test, the page count isn't changing properly.
So I'm thinking we could make `step` a variable in the constructor and maybe the attribute can be on the slider wrapper rather than each button. 
And then we use `step` in this calculation as well. So that it looks something like: 
```javascript
this.currentPage = Math.round(this.slider.scrollLeft / (this.sliderItemOffset * step)) + 1;
```

I realize it isn't quite the scope of this PR but thought I'd mention it. It doesn't seem like too big of a thing to fix. Which I could do in my `feat. collection` PR since that's where I most likely will need to use `data-step`. 

and I guess the amount of total page should be change accordingly 🤔

assets/base.css
```diff
@@ -1122,11 +1123,11 @@ summary::-webkit-details-marker {
...
   .slider--tablet.grid--peek .grid__item:first-of-type {
-    padding-left: 1.5rem;
+    margin-left: 1.5rem;
   }
 
   .slider--tablet.grid--peek .grid__item:last-of-type {
-    padding-right: 1.5rem;
+    margin-right: 1.5rem;
   }
 }
```
ludoboludo, line 1133: Noticed that the counter will change it's width as you go through it when you go from single digit number to two digit: [video](https://screenshot.click/04-44-8ybpv-mgvi1.mp4)

Not related to the issue this PR is trying to solve but could be a quick easy win added to the PR description 😬

ludoboludo (approved):
Left a few comments that where not directly related to your fix but the slider in general. 

Otherwise looked good to me 👍

assets/base.css
kmeleta: Worth being aware so we can re-test, but if the mobile settings are only going to allow for 1 or 2 per row this should be fine. If we allow 3, then we'll need to revisit.

assets/base.css
kmeleta: See any problem with this approach? https://github.com/Shopify/dawn/pull/1297/commits/c72df71dd41eb775c60432db73ee6d2c2cedd9e6

ludoboludo: Hmm I wonder about font scale and different font families: [video](https://screenshot.click/07-59-vehqb-4lp1s.mp4)

kmeleta: Seemed relatively safe to me, but the arrows could get closer of course. I don't have real strong feelings on it, the width changing at double digits didn't even bother me that much. If we want to consider it more or implement a more advanced solution I'd rather open an issue for it. Happy to revert this change or not.

kmeleta: Yeah true. No one likes that. Will leave as is for now.

assets/component-slider.css
```diff
@@ -130,7 +130,8 @@ slider-component {
 .slider-counter {
   margin: 0 1.2rem;
   display: flex;
-  min-width: 2rem;
+  justify-content: center;
+  width: 2rem;
```
ludoboludo, line 134: What about making the `min-width` value a bit higher instead of setting a fix value 🤔  Like `min-width: 3rem;`

ludoboludo (dismissed):

assets/component-slider.css
ludoboludo: Nice 👌

ludoboludo (approved):

melissaperreault (commented):
Thank you Ken! 🎉 Everything looks good except one with notes!

**Testing steps/scenarios**

- [x] Featured collection (mobile/tablet)
- [x] Collection list (mobile/tablet)
- [x] Multicolumn (mobile)
- [x] Featured blog (mobile/tablet)
- [ ] Product media (mobile) 
   - [ ] Something is wrong with the left space on that template, regardless of the global settings, [it's offset to the right here](https://screenshot.click/22-02-yoe7d-iaalh.png)
   - [ ] Not sure if relevant, the thumbnails arrow [are not vertically centered](https://screenshot.click/22-02-qa2ro-paj32.png) 
- [x] Slideshow
- [x] Minimal horizontal grid spacing
- [x] Maximum horizontal grid spacing
- [x] Large number of slides
- [x] Very small viewport sizes (< 360px)
- [x] Resizing viewport

melissaperreault (approved):
LGTM!

assets/global.js
```diff
@@ -533,9 +533,9 @@ class SliderComponent extends HTMLElement {
 
   initPages() {
     this.sliderItemsToShow = Array.from(this.sliderItems).filter(element => element.clientWidth > 0);
-    this.sliderLastItem = this.sliderItemsToShow[this.sliderItemsToShow.length - 1];
-    if (this.sliderItemsToShow.length === 0) return;
-    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItemsToShow[0].clientWidth);
+    if (this.sliderItemsToShow.length < 2) return;
+    this.sliderItemOffset = this.sliderItemsToShow[1].offsetLeft - this.sliderItemsToShow[0].offsetLeft;
+    this.slidesPerPage = Math.floor(this.slider.clientWidth / this.sliderItemOffset);
```
ludoboludo, line 538: I'm actually now thinking that the total amount can be managed with some liquid logic 🤔 
Sofia added a setting to set amount of products per row for desktop layout and is about to add it on mobile as well. So I'm thinking that we can most likely do something like this: 
```liquid
{{ numberOfItems | minus: amountOfRows | plus: 1 }}
```

And have to spans, one `totalDesktop` and one `totalMobile` so that we can manage visibility via CSS and media queries 🤔  What do you think ?

assets/base.css
tyleralsbury: cc @sofiamatulis so she's aware.

assets/base.css
tyleralsbury: I'm glad you commented this cause I would have asked. 😄

assets/base.css
kmeleta: Ludo and I are reviewing those PRs too, so it shouldn't get overlooked in 3 column mobile ends up happening there 😄

tyleralsbury (approved):
Looks good - tested + code.

Math is always tricky to get right - great solution, Ken.  You were right to look deeper into it than to just change the rounding strategy, finding the actual source of the miscalculation.

assets/base.css
kmeleta: haha I know. Related to this.. I wouldn't say this case warrants a comment or anything (even if they were more commonplace in our practices), but I do worry that retro specificity reduction for specificity reduction's sake (as we tend to do) could/does lead to small things like this being broken in the future.

tyleralsbury: Yeah, I struggle with this sometimes. CSS is really not an easy thing to just read and understand like other code - we have to assume everything is there for a reason, and that the reason may not be entirely obvious, but that also makes it really hard to refactor CSS as well.  It's a fundamental challenge with CSS, and I don't have a good solution to it.

PR #1315, opened 2022-02-04
Fix empty/disabled links

### LINKED ISSUE #864: Featured collection links missing role
Featured collection empty links are just <a>

Should have aria-disabled="true" and role="link"

- Dawn Version: 1.0.0
- macOS Version 11.3.1

**Possible solution**

Other sections received this update in #810

andrewetchen:

Fixes #864 

Disabled/empty links should have `role="link" aria-disabled="true"` instead of an empty `href` or `href="#"` so that they're not focusable or actionable with a mouse/keyboard. Currently, when a user clicks on these links, the page will reload/refresh.

- Apply similar condition from #810 for empty/disabled `<a>` product and collection links.
- The Multicolumn section has the [desired result](https://user-images.githubusercontent.com/28404165/152587028-2d8a2faf-6dfc-4e15-86cf-794d0cd0296c.png) for disabled links.
- For CSS, I found the following discussion really helpful in terms of specificity, consistency, and to ensure these changes won't also apply to buttons: https://github.com/Shopify/dawn/pull/730/files#r732066483 

**Other considerations**

**Accessibility**: In terms of accessibility, when a link is created without a destination, it's still actionable via mouse and keyboard. Activating the link refreshes the page. This may cause confusion for multiple user groups who may think activation will load a new resource. [More context here](https://github.com/Shopify/dawn/issues/864#issuecomment-1027318490).

**Testing steps/scenarios**

Add the following sections either the homepage or product page and check for empty product or collection links:
- [ ] **Collage**: Collection and Product blocks.
- [ ] **Collection list**: Collection blocks.
- [ ] **Featured collection**: Don't assign a collection and notice the empty product links.
- [ ] **Featured product**: Don't assign a product and notice the [View full details](https://user-images.githubusercontent.com/28404165/152587899-b5f835ef-213e-4f9a-98c0-6e44bc849317.png) link.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127548456982)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127548456982/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -622,7 +622,6 @@ h6:empty {
   margin-left: 1rem;
 }
 
-.link[role="link"]:not([href]),
 a:not([href]) {
   cursor: not-allowed;
 }


sections/featured-product.liquid
```diff
@@ -750,4 +750,4 @@
     }
   ]
 }
-{% endschema %}
+{% endschema %}
```
tyleralsbury, line 753: To avoid these types of diffs in PRs, add an empty line at the end of each file.  You can enable this as a setting in VS Code to not need to think about it

Screen Shot 2022-02-04 at 2 37 47 PM

```suggestion
{% endschema %}

```

snippets/card-product.liquid
```diff
@@ -82,7 +82,7 @@
         <div class="card__content">
           <div class="card__information">
             <h3 class="card__heading">
-              <a href="{{ card_product.url | default: '#' }}" class="full-unstyled-link">
+              <a href="{{ card_product.url }}" class="full-unstyled-link">
```
tyleralsbury, line 85: Has it turned out that it's impossible for there to not be a URL in the case of a product that exists?  I had suspected it, but have you confirmed it?

assets/base.css
```diff
@@ -622,7 +622,8 @@ h6:empty {
   margin-left: 1rem;
 }
 
-.link[role="link"]:not([href]) {
+.link[role="link"]:not([href]),
+.full-unstyled-link[role="link"]:not([href]) {
   cursor: not-allowed;
 }
```
tyleralsbury, line 628: Could we do something like `a:not([href])` or would that not account for every case?

tyleralsbury (commented):
Looking really good overall - some questions.

assets/base.css
andrewetchen: Hi @bacounts, adding `pointer-events: none;` below would actually overwrite `cursor: not-allowed;`.

Using `cursor: not-allowed;` provides a visual indication that a link is disabled and in this particular case, a placeholder link. We also wanted to follow the cursor behaviour for disabled buttons seen on some of our other sections like the Multicolumn section:

image

You can read more about those changes in #730

assets/base.css
```diff
@@ -622,7 +622,8 @@ h6:empty {
   margin-left: 1rem;
 }
 
-.link[role="link"]:not([href]) {
+.link[role="link"]:not([href]),
+a:not([href]) {
```
tyleralsbury, line 626:
```suggestion
a:not([href]) {
```

I'm wondering if we even need `.link[role="link"]:not([href]),` if we've got `a:not([href])`.

andrewetchen: Since `a:not([href])` already takes care of all "disabled" anchor links, I'm good with removing `.link[role="link"]:not([href]),` entirely. I just finished testing without `.link[role="link"]:not([href]),` and everything checks out.

kmeleta: 😆 the amount of discussion this seemingly benign selector has warranted over the span of ~3 PRs. So funny.

As discussed, I feel `a:not([href])` is a little broad to change the cursor. Including the `[role="link"]` makes it a little more intentional and also preserves the original intent of creating a visual cue that might alert us to empty links that are _missing_ the needed `role`. I don't see where the slight increase in specificity would matter for this selector. 

I also cannot remember the context behind using `.link` but it's likely it was an interim "safe scoping" decision at some point so I'm fine removing that if we keep the attr selector.

andrewetchen: **Update**: One thing I noticed when using:

```css
a[role="link"]:not([href]) {
  cursor: not-allowed;
}
```

The buttons from the following sections are using the above CSS rule as it overrides `.button[aria-disabled='true']`:

- [Rich text section > button](https://user-images.githubusercontent.com/28404165/153301443-8cb58be5-9c0e-4983-8c44-297d499c7029.png)
- [Image banner > buttons](https://user-images.githubusercontent.com/28404165/153301634-111d79c2-2e7e-4cb4-91e8-eb27232b47f4.png)
- [Slideshow > buttons](https://user-images.githubusercontent.com/28404165/153301744-9b80e7d4-4016-4189-b169-78056f5ccbc7.png)

And here's what it looks like when using the `a:not([href])` selector:

image

tyleralsbury: I didn't realize this was such a can of worms!  Thanks for all the context @kmeleta .  Indeed, this benign selector has a lot of sway in PR conversations, it seems. 🤣

tyleralsbury (approved):
This feels good to me.  I made a comment above regarding the lack of parity between collection and product cards in their onboarding states, but that's for a separate issue.  I'd like Melissa to weigh in on that, but it should not block merging for this PR.

tyleralsbury:
> I was kind of making a case against that. But it's not much impact/risk either way and I didn't notice any side effects from testing, so lets 🚢

Either way feels good to me - I tend to aim towards the less specific selector if it works and doesn't cause side-effects.  If the two selectors represent the same scenario, then I'd opt for the one with less code.

Issue #1181, opened 2022-01-13
[Section Headings] Introduce new flexible settings


nicklepine:
Impacts all sections with headings: 
- [ ] Featured Collection
- [ ] Collection List
- [ ] Multi Column
- [ ] Video
- [ ] Collage
- [ ] Blog Posts
- [ ] Page
- [ ] Tabs

[Internal Figma Link](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=15199%3A87022)

**WHAT'S NEW**

- [ ] Heading Size
- [ ] Heading Alignment
- [ ] Heading Horizontal Offset
- [ ] Heading Vertical Offset

![image](https://user-images.githubusercontent.com/8127145/149382876-f990c3c2-a2fe-4ff5-951e-eea4e9148749.png)
![image](https://user-images.githubusercontent.com/8127145/149383063-e95fbbd2-4a54-4146-8242-7661020096a1.png)

martinamarien:
I've decided to separate this into separate PRs. I think that these can be tackled and shipped separately. Heading size and alignment provide its own value to merchants, separate from the heading offset feature. 
- [ ] Heading size and alignment
- [ ] Heading offset
  * I'd like to spend some more time exploring this one and possibly refactor how we currently approach heading spacing.

Stephanie-Chou:
We are closing this issue. This issue was reported 2 years ago on an old version of Dawn. If this is still a high priority issue today, please re open with updated reproduction steps


also looks like some work on this was done. Plus with flex sections coming up- we may have new and better solutions

PR #1340, opened 2022-02-10
Add section heading size settings

martinamarien:
**PR Summary**: _A new heading size setting was introduced in all sections on the theme. You can pick between small, medium, or large (defaults to medium). This update may cause your theme's headings to be bigger. To revert this change, change the setting to small_.

We are adding a Heading size setting to all sections with a heading. Currently, section heading blocks already have this setting, and we are now applying it to section headings as well.

Related to: #1181 

I added a select element as the setting and applied the following values: 

|Label|Value|
|---|---|
|Small|h2|
|Medium|h1|
|Large|h0|

Classes were used as the value so that we can simply output the setting value as a class, without needing to check any conditions. This is the same approach that was used for blocks in other places in the Dawn.

**Other considerations**

While making this PR, I noticed that there was a class (`title-wrapper-with-link--no-heading`) that was no longer needed. In most places, this class was only being applied if the section's title/heading was blank. There was a condition on the element itself, but the element was also wrapper in another condition that evaluated to the same thing. I opted to clean this up and remove it. 

This: 
```
    {%- unless section.settings.title == blank -%}
      <div class="title-wrapper-with-link{% if section.settings.title == blank %} title-wrapper-with-link--no-heading{% endif %}">
         ... 
      </div>
    {%- endunless -%}
```
Becomes:    
```
    {%- unless section.settings.title == blank -%}
      <div class="title-wrapper-with-link">
         ... 
      </div>
    {%- endunless -%}
```

**Testing steps/scenarios**
- [x] _List all the testing tasks that applies to your fix and help peers to review your work._

Ensure heading size was applied to these sections and works as expected.
Heading size at section level: 
 - [x] Featured Collection
 - [x] Collection List
 - [x] Multi Column
 - [x] Video
 - [x] Collage
 - [x] Blog Posts
 - [x] Page
 - [x] Collapsible content
 - [x] Contact form
 
 Heading size at block level: 
 Already done previously: 
 - [x] Rich text
 - [x] Email signup
 - [x] Image banner
 - [x] Slideshow
 - [x] Image with text

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127578079254/editor)

**Checklist**


sections/collage.liquid
```diff
@@ -161,6 +161,26 @@
...
+        {
+          "value": "h1",
+          "label": "t:sections.all.heading_size.options__2.label"
+        },
+        {
+          "value": "h0",
+          "label": "t:sections.all.heading_size.options__3.label"
+        }
+      ],
+      "default": "h1",
```
ludoboludo, line 181: Is the default in general supposed to be `h1` ? We used to have `h2` as a default before which is why I ask if the change is intentional or not.

sections/video.liquid
```diff
@@ -20,7 +20,7 @@
     <div{% if section.settings.full_width %} class="page-width"{% endif %}>
       {% unless section.settings.heading == "" %}
         <div class="title-wrapper title-wrapper--no-top-margin">
-          <h2 class="title">{{ section.settings.heading }}</h2>
+          <h2 class="title {{ section.settings.heading_size }}">{{ section.settings.heading }}</h2>
```
ludoboludo, line 23: here when the heading is empty we're still getting some margin from the element even though it's empty:

sections/featured-collection.liquid
```diff
@@ -32,15 +32,15 @@
...
   
         {%- if section.settings.show_view_all and section.settings.swipe_on_mobile and more_in_collection -%}
           <a href="{{ section.settings.collection.url }}" class="link underlined-link large-up-hide">{{ 'sections.featured_collection.view_all' | t }}</a>
         {%- endif -%}
       </div>
-    {% endunless %}
+    {%- endunless -%}
   
     <slider-component class="slider-mobile-gutter">
       <ul id="Slider-{{ section.id }}" class="grid grid--2-col product-grid contains-card{% if settings.card_style == 'standard' %} contains-card--standard{% endif %}{% if products_to_display == 4 or section.settings.collection == blank %} grid--2-col-tablet grid--4-col-desktop{% else %} grid--3-col-tablet{% endif %}{% if products_to_display > 5 %} grid--one-third-max grid--4-col-desktop grid--quarter-max{% endif %}{% if section.settings.collection.all_products_count > 2 and section.settings.swipe_on_mobile and section.settings.products_to_show > 2 %} slider slider--tablet grid--peek{% endif %}" role="list">
```
ludoboludo, line 46: I noticed that here we're getting a margin-top by default of 1rem which makes it use more space than expected when there is no heading.

ludoboludo: I could add this in my own PR by the way 🤔

sections/page.liquid
```diff
@@ -20,11 +20,11 @@
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
   <div class="page-width page-width--narrow section-{{ section.id }}-padding">
-    <h2 class="page-title">
+    <h2 class="page-title {{ section.settings.heading_size }}">
       {%- if section.settings.page.title != blank -%}
         {{ section.settings.page.title | escape }}
       {%- else -%}
-        Page title
+        {{ 'sections.page.title' | t }}
```
ludoboludo, line 27: I wonder if we should offer the option to have none or overwrite it using a new setting for the section 🤷

ludoboludo (commented):
Some of the section you mentioned that already have the setting for the heading should be updated to have the 3 options: 
- Image banner has two options only
- Should it be added for the title of the feat. product section ? 🤔 
(might be out of scope with this PR tho).

Looks good otherwise and works as expected 👍  Though shouldn't the default be `h2` ?

sections/video.liquid
martinamarien: I cannot replicate this. If I remove text, the element is not outputted on the page for me. Can you try again and see if you are still seeing that Ludo?

ludoboludo: Im not sure why it happened in the first place. I can't replicate anymore 👍

locales/en.default.schema.json
```diff
@@ -1010,15 +1022,6 @@
...
               "label": "Heading"
-            },
-            "heading_size": {
-              "options__1": {
-                "label": "Medium"
-              },
-              "options__2": {
-                "label": "Large"
-              },
-              "label": "Heading size"
```
ludoboludo, line 1021: I think there are a few more you can remove no ? If I understand properly, we're using instead a general one instead of the same `"heading_size"` in all the section parts. 

I see 4 instances of `"heading_size"` but I believe we could have 1 only 🤔

sections/image-banner.liquid
```diff
@@ -84,7 +84,7 @@
       {%- for block in section.blocks -%}
         {%- case block.type -%}
           {%- when 'heading' -%}
-            <h2 class="banner__heading {% if block.settings.heading_size == 'medium' %}h1{% else %}h0{% endif %}" {{ block.shopify_attributes }}>
+            <h2 class="banner__heading {% if block.settings.heading_size == 'small' %}h2{% elsif 'medium' %}h1{% else %}h0{% endif %}" {{ block.shopify_attributes }}>
```
ludoboludo, line 87: ~Here you don't want to change the value to output directly the class and remove the conditional statements ?~
I see why in slack 😅  We don't want to introduce a breaking change. 👍

ludoboludo (commented):
Other than removing the rest of the translations, it looks good to me if we want to keep this approach 👍

assets/base.css
```diff
@@ -941,12 +941,6 @@ summary::-webkit-details-marker {
   flex-shrink: 0;
 }
 
-@media screen and (min-width: 990px) {
-  .title-wrapper-with-link.title-wrapper-with-link--no-heading {
-    display: none;
-  }
-}
-
```
tyleralsbury, line 949: Love to see this going away!

tyleralsbury (commented):
I love the consistency of the implementation - this is exactly the way we should handle all section parity work.

I'll test more thoroughly tomorrow, but it's looking good.

sections/image-banner.liquid
```diff
@@ -313,17 +313,21 @@
...
-              "label": "t:sections.image-banner.blocks.heading.settings.heading_size.options__1.label"
+              "label": "t:sections.all.heading_size.options__2.label"
             },
             {
               "value": "large",
-              "label": "t:sections.image-banner.blocks.heading.settings.heading_size.options__2.label"
+              "label": "t:sections.all.heading_size.options__3.label"
             }
           ],
           "default": "medium",
```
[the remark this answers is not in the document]
tyleralsbury: Just tested and got the same effect.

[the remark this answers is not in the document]
tyleralsbury: Question for @martinamarien - I see that the rest of the implementations are using `{{ block.settings.title_size }}` with the value being the heading class like `h1`, etc...

I'm assuming that the reason you didn't do that here is to prevent unwanted changes for merchants who are upgrading.  I'm wondering if the technical debt of having a slightly different implementation here is worth the better upgrade experience if we can include it in the release notes that they'd need to check this.

I guess it's not a huge difference implementation-wise, but I'd love for it to be consistent across the board.  The more exceptions we make during these parity exercises due to previous implementations, the more each setting will have two or more different implementations.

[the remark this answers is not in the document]
melissaperreault: When I am adding a new section, it's defaulting to medium for me!

melissaperreault (dismissed):
Thank you Martina! 🎉 🚢 All works fine! Lets catch up on the follow up PR for my unchecked list items below! 🙌

**Task recap**
- [x] Looking for defaults
- [x] Looking for Heading size application where I've noted they were missing from a parity audit, more details below

**Testing list**
- [x] Image banner
- [x] Collage
- [x] Video
- [x] Multicolumn
   - [ ] Column block heading to be part of the Footer heading update
- [x] Blog posts
   - [ ] To have different hierarchy for the blog post heading, additionally to the section. i.e. (`Section heading size `and `Blog post heading size`) heading
- [x] Featured product
   - [ ] Product page title block missing options but will be part of another round follow up.
- [x] Product recommendations
- [x] Email signup section
- [x] Email banner from password
- [x] Collapsible content
   - [ ] Collapsible row block heading to be part of the Footer heading update
- [x] Collection list
   - [ ] Collection list template missing options but will be part of another round follow up.
- [x] Contact form
- [ ] Collection template missing options but will be part of another round follow up.
- [ ] All pages missing options but will be part of another round follow up.

sections/image-banner.liquid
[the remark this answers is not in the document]
martinamarien: > I'm assuming that the reason you didn't do that here is to prevent unwanted changes for merchants who are upgrading. I'm wondering if the technical debt of having a slightly different implementation here is worth the better upgrade experience if we can include it in the release notes that they'd need to check this.

This is exactly the reason I didn't convert it to h0/h1/h2. Since we're already introducing a change to the default heading value (breaking), I'm happy to update Image Banner and Slideshow to use the classes instead. We just need to clearly communicate this to merchants in the release notes, like you said.

locales/en.default.schema.json
martinamarien: I only updated ones where I was already touching the file. 
I was planning to create a follow up PR where I could update the rest. Would that be okay with you?

sections/collage.liquid
```diff
@@ -161,6 +161,26 @@
...
+          "label": "t:sections.all.heading_size.options__1.label"
+        },
+        {
+          "value": "h1",
+          "label": "t:sections.all.heading_size.options__2.label"
+        },
+        {
+          "value": "h0",
+          "label": "t:sections.all.heading_size.options__3.label"
+        }
```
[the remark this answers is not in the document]
martinamarien: > but we should consider cleaning up our translation files to make them more manageable

100%. I thought the same as you when I was implementing this. "Tyler and I just talked about how we need to update this". Although it's a source of frustration for us, like you, I think that this update should happen in its own PR. This PR focuses on adding the size setting, and we can do a full `heading_size` pass on all files that use this translation - and update them all at once.

sections/featured-blog.liquid
```diff
@@ -32,17 +32,19 @@
 -%}
 <div class="blog color-{{ section.settings.color_scheme }} gradient{% if section.settings.heading == blank %} no-heading{% endif %}">
   <div class="page-width-desktop isolate{% if posts_displayed < 3 %} page-width-tablet{% endif %} section-{{ section.id }}-padding">
-    <div class="title-wrapper-with-link{% if section.settings.heading == blank %} title-wrapper-with-link--no-heading{% endif %} {% if posts_displayed > 2 %}title-wrapper--self-padded-tablet-down{% else %}title-wrapper--self-padded-mobile{% endif %} title-wrapper--no-top-margin">
-      <h2 class="blog__title">{{ section.settings.heading | escape }}</h2>
+    {%- unless section.settings.heading == blank -%}
+      <div class="title-wrapper-with-link {% if posts_displayed > 2 %}title-wrapper--self-padded-tablet-down{% else %}title-wrapper--self-padded-mobile{% endif %} title-wrapper--no-top-margin">
```
KaichenWang, line 36: Could adjust spacing for consistency with `collection-list.liquid` and `featured-collection.liquid`

```suggestion
      <div class="title-wrapper-with-link{% if posts_displayed > 2 %} title-wrapper--self-padded-tablet-down{% else %} title-wrapper--self-padded-mobile{% endif %} title-wrapper--no-top-margin">
```

KaichenWang (approved):
Looks good

KaichenWang (approved):
Looks good. 

Just curious, is there a reason the heading size setting is available for `featured-product` title, but not the product page's title?

martinamarien:
Hey Kai. Yeah. We'll be tackling the template headings in their own PR.

> Templates
All page templates (e.g., Page, product, etc) will be updated in their own PRs. Section heading updates and parity provide their own value to merchants, and we are okay shipping these two separately, if needed.

martinamarien:
> Note that this change also affects Email sign up banner so we should include it in the docs https://screenshot.click/25-36-ywq2m-czt2q.png

Yep! Can confirm that Kimli has done it 👍🏻

PR #1374, opened 2022-02-18
`{{ shop.url }}` → `{{ request.origin }}` when building absolute URLs on the current domain

cbothner:

`{{ shop.url }}` always returns the shop's primary domain, but it's used to generate `Share` links and structured data. When visiting a shop on a non-primary domain, they result in the wrong url.

image

The behavior of `{{ shop.url }}` cannot change without potentially breaking shops which are relying on it as a reference to the primary domain. So `{{ request.origin }}` was recently added as an alternative for use cases like this.

Replaced all instances of `{{ shop.url }}` with `{{ request.origin }}`.

**Other considerations**

Are there any cases where we should always be using the shop's primary domain? Gift cards for example, although I'm not sure why that one makes me suspicious.

**Demo links**

- Store: https://ca.catleaves.com/products/basic-cat-leaf?preview_theme_id=127614287894
- Editor: https://cat-leaves.myshopify.com/admin/themes/127614287894/editor

**Checklist**

### DIFF
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -227,7 +227,7 @@
     </ul>
 
     <script>
-      window.shopUrl = '{{ shop.url }}';
+      window.shopUrl = '{{ request.origin }}';
       window.routes = {
         cart_add_url: '{{ routes.cart_add_url }}',
         cart_change_url: '{{ routes.cart_change_url }}',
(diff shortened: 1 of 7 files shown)


ludoboludo (approved):
Seems good. I wonder why the predictive search is erroring on your store though 🤔  I don't think it's due to this PR as it was fine on my store though I don't have subdomains setup 🤷

LucasLacerdaUX (approved):
Tested on Header, Articles, Product Pages, Predictive Search and Gift Cards 👍

Issue #1376, opened 2022-02-18
Hamburger menu not opening, input fields have no borders (older model iPhones)


KaichenWang:
**Describe the current behavior**

Dawn not functioning on older model iPhones
- hamburger menu not opening
- input fields have no borders

**Describe the expected behavior**

- hamburger menu opens
- input fields have borders

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 3.0.0
- Safari Version 11
- iOS Version 11 (iPhone X)

**Possible solution**

- hamburger menu not opening

Update this in `global.js`
```
try {
  document.querySelector(":focus-visible");
} catch {
  focusVisiblePolyfill();
}
```
to
```
try {
  document.querySelector(":focus-visible");
} catch (e) {
  focusVisiblePolyfill();
}
```
- input fields have borders

Remove usage of `inset` CSS property

**Additional context/screenshots**
(Screenshot Removed)

PR #1377, opened 2022-02-18
Add binding to catch block to prevent syntax error in browsers that don't support es2019

KaichenWang:

Partially addresses https://github.com/Shopify/dawn/issues/1376

Add binding to catch block

**Other considerations**

- [Optional binding for catch blocks was introduced in es2019](https://github.com/tc39/proposal-optional-catch-binding)
- Omitting the binding causes syntax error in older browsers (including Safari in iOS 11)
- Although we don't officially support iOS 11, reasoning for fixing this:
  - Although we aren’t officially supporting it, it would allow older browsers to benefit without bloating the code.
  - If a change to support an older browser doesn’t add significant complexity and doesn’t cause any negative effects for the browsers we do support, then we should make the change.

**Testing steps/scenarios**

- [ ] Use preview on Browserstack, Simulator or iOS 11 device on mobile
- [ ] Open main mobile menu

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127614943254)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127614943254/editor)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -73,7 +73,7 @@ function trapFocus(container, elementToFocus = container) {
 // Here run the querySelector to figure out if the browser supports :focus-visible or not and run code based on it.
 try {
   document.querySelector(":focus-visible");
-} catch {
+} catch(e) {
   focusVisiblePolyfill();
 }
 


ludoboludo (approved):
tested and confirmed it works like a charm 👍 

Out of curiosity, could you add to cart on version 11 ? 😬 
Variant switching also seem to break ATC.

But technically not supported browser versions 🤔 

Im not sure if it's browserstack having issues loading or if it's actually erroring.

andrewetchen (approved):
Mobile menu is working on my end as well [video](https://screenshot.click/22-13-yqmy7-4du9zsf-f.mp4).

I also wasn't able to add to cart and experienced the same variant switching issue as Ludo. Is this is separate issue?

Issue #1364, opened 2022-02-15
Fix theme check warnings


andrewetchen:
**Describe the current behavior**

This is a follow up to @charlespwd 's PR: https://github.com/Shopify/dawn/pull/955

There are currently 291 offences.

- [x] **Auto-correctable** https://github.com/Shopify/dawn/pull/1371
  - **255** for `` Deprecated filter `img_url`, consider using `image_url`. ``
  - **7** for `SpaceInsideBraces`
  - **9** for `SchemaJsonFormat`
- [x] **Requires manual changes** https://github.com/Shopify/dawn/pull/1379
  - **10** for `` DeprecatedFilter: Deprecated filter `img_url`, consider using an alternative: `image_url`. ``
  - **8** for `ImgLazyLoading`
  - **2** for `LiquidTag: Use {% liquid ... %} to write multiple tags.`

**Describe the expected behavior**

As expected, `theme-check` is doing its job but the goal here is to fix these warnings.

**Version information (Dawn, browsers and operating systems)**

- Dawn Version: 3.0.0
- Chrome Version 98.0.4758.102 
- macOS Version 12.2.1

**Possible solution**

Use `shopify theme check` to auto-correct 271/291 offences and manually fix the rest.

Will create two separate PRs: one for **automatic** fixes and one for **manual** fixes.

**Additional context/screenshots**

**Screenshots**

<details>
  <summary>DeprecatedFilter</summary>

  ![image](https://user-images.githubusercontent.com/28404165/154155885-84fc2667-f907-46ad-90e7-f2f7be8ac072.png)

</details>

<details>
  <summary>SchemaJsonFormat</summary>

  ![image](https://user-images.githubusercontent.com/28404165/154375285-cefb95b9-dd4e-49ee-88f1-9be38629b2e9.png)

</details>

<details>
  <summary>SpaceInsideBraces</summary>

  ![image](https://user-images.githubusercontent.com/28404165/154375404-00ef803b-e6c4-4b34-8752-a2024c899b4a.png)

</details>

<details>
  <summary>LiquidTag</summary>

  ![image](https://user-images.githubusercontent.com/28404165/154375445-2832fd11-2dc6-4030-941b-b2bd642c8c75.png)

</details>

**Links**

- **Related PR**: https://github.com/Shopify/dawn/pull/955
- **Related issue**: https://github.com/Shopify/dawn/issues/1259

---

**Update: March 15 2022**

[...]

PR #955, opened 2021-11-30
Upgrade from img_url to image_url

charlespwd:
`img_url` has been deprecated. The new filter, `image_url` is more
straightforward to understand and doesn't require the use of the
'master' size specification.

The next version of theme-check will flag their use as warnings, this PR
fixes that before Dawn lights up with 250+ warnings.

And minor theme-check --autocorrect changes that can be fixed with the
JsonSchemaFormat autocorrector.

**Why are these changes introduced?**

Fixes #0.

**What approach did you take?**

**Other considerations**

**Demo links**

- [Store](https://os2-demo.myshopify.com?preview_theme_id=126964432918)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126964432918/editor) (linked to Github, avoid saving changes)
- [Editor - unlinked](https://os2-demo.myshopify.com/admin/themes/126964465686/editor)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #1379, opened 2022-02-18
Theme check manual fixes

andrewetchen:

This PR is part 2/2 for #1364 

Fixed all of the following theme check warnings:

- **10** for `` DeprecatedFilter: Deprecated filter `img_url`, consider using an alternative: `image_url` ``
- **8** for `ImgLazyLoading`
- **2** for `LiquidTag: Use {% liquid ... %} to write multiple tags`

And added the necessary `theme-check-disable` comments.

image

**Other considerations**

There are a few places where we don't want lazy loading added and `{% comment %}theme-check-disable ImgLazyLoading{% endcomment %}` should be added to these files if they contain lazy load attribute conditions to prevent `theme check` warnings:

- **Collection banner**
- **Product page**: Main image
- **Collection page**: First 2 images
- **Search page**: First 2 images

**CC**: @tyleralsbury 

- [Disable checks with comments](https://github.com/Shopify/theme-check#disable-checks-with-comments) README

**Testing steps/scenarios**
- [ ] Look at each of the files changes and ensure the following was fixed:
  - **10** warnings for deprecated `img_url`: these couldn't be auto-corrected due to `{% assign %}` variables.
  - **8** for `ImgLazyLoading`: the `theme-check-disable ImgLazyLoading` comment was added to the respective files.
  - **2** for `LiquidTag: Use {% liquid ... %} to write multiple tags.`: 2 instances in `main-product.liquid`.
- [ ] You can also make sure that running `shopify theme check` or `theme-check` returns:
  - `233 files inspected, 0 offenses detected, 0 offenses auto-correctable` 

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127615205398)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127615205398/editor)

**Checklist**

### DIFF
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -79,7 +79,7 @@
                       {%- assign image_size = block.settings.image_width | append: 'x' -%}
                       <img
                         srcset= "{{ block.settings.image | img_url: image_size }}, {{ block.settings.image | img_url: image_size, scale: 2 }} 2x"
-                        src="{{ block.settings.image | img_url: '400x' }}"
+                        src="{{ block.settings.image | image_url: width: 400 }}"
                         alt="{{ block.settings.image.alt | escape }}"
                         loading="lazy"
                         width="{{ block.settings.image.width }}"
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -77,7 +77,7 @@
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
                       {%- assign image_size = block.settings.image_width -%}
-                      {%- assign image_size_2x = block.settings.image_width | times: 2 -%}
+                      {%- assign image_size_2x = image_size | times: 2 | at_most: 5760 -%}
                       <img
                         srcset= "{{ block.settings.image | image_url: width: image_size }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
                         src="{{ block.settings.image | image_url: width: 400 }}"
@@ -85,7 +85,7 @@
                         loading="lazy"
                         width="{{ block.settings.image.width }}"
                         height="{{ block.settings.image.height }}"
-                        style="max-width: min(100%, {{ block.settings.image_width }}px);"
+                        style="max-width: min(100%, {{ image_size }}px);"
                       >
                     {%- else -%}
                       {{ 'image' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -76,9 +76,10 @@
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
-                      {%- assign image_size = block.settings.image_width | append: 'x' -%}
+                      {%- assign image_size = block.settings.image_width -%}
+                      {%- assign image_size_2x = block.settings.image_width | times: 2 -%}
                       <img
-                        srcset= "{{ block.settings.image | img_url: image_size }}, {{ block.settings.image | img_url: image_size, scale: 2 }} 2x"
+                        srcset= "{{ block.settings.image | image_url: width: image_size }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
                         src="{{ block.settings.image | image_url: width: 400 }}"
                         alt="{{ block.settings.image.alt | escape }}"
                         loading="lazy"
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -76,16 +76,15 @@
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
-                      {%- assign image_size = block.settings.image_width -%}
-                      {%- assign image_size_2x = image_size | times: 2 | at_most: 5760 -%}
+                      {%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
                       <img
-                        srcset= "{{ block.settings.image | image_url: width: image_size }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
+                        srcset= "{{ block.settings.image | image_url: width: block.settings.image_width }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
                         src="{{ block.settings.image | image_url: width: 400 }}"
                         alt="{{ block.settings.image.alt | escape }}"
                         loading="lazy"
                         width="{{ block.settings.image.width }}"
                         height="{{ block.settings.image.height }}"
-                        style="max-width: min(100%, {{ image_size }}px);"
+                        style="max-width: min(100%, {{ block.settings.image_width }}px);"
                       >
                     {%- else -%}
                       {{ 'image' | placeholder_svg_tag: 'placeholder-svg placeholder' }}
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -638,12 +638,12 @@
       {{ settings.social_youtube_link | json }},
       {{ settings.social_vimeo_link | json }}
     ],
-    "url": {{ shop.url | append: page.url | json }}
+    "url": {{ request.origin | append: page.url | json }}
   }
 </script>
 
 {%- if request.page_type == 'index' -%}
-  {% assign potential_action_target = shop.url | append: routes.search_url | append: "?q={search_term_string}" %}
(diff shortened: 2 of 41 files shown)


sections/footer.liquid
```diff
@@ -76,9 +76,10 @@
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
-                      {%- assign image_size = block.settings.image_width | append: 'x' -%}
+                      {%- assign image_size = block.settings.image_width -%}
+                      {%- assign image_size_2x = block.settings.image_width | times: 2 -%}
```
andrewetchen, line 80:
```diff
-{%- assign image_size_2x = block.settings.image_width | times: 2 -%}
+{%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
```

I think we also need to add `at_most: 5760` to avoid errors originally mentioned here: https://github.com/Shopify/dawn/pull/955#discussion_r759489837

sections/footer.liquid
```diff
@@ -77,15 +77,15 @@
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
                       {%- assign image_size = block.settings.image_width -%}
-                      {%- assign image_size_2x = block.settings.image_width | times: 2 -%}
+                      {%- assign image_size_2x = image_size | times: 2 | at_most: 5760 -%}
```
andrewetchen, line 80: `at_most: 5760` was added to the footer and header to avoid width errors:

- Originally mentioned here: https://github.com/Shopify/dawn/pull/955#discussion_r759489837

<details>
  <summary>Screenshot of width error</summary>

  image

</details>

sections/header.liquid
```diff
@@ -344,7 +344,7 @@
         <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
           {%- if section.settings.logo != blank -%}
             {%- assign image_size = section.settings.logo_width -%}
-            {%- assign image_size_2x = section.settings.logo_width | times: 2 -%}
+            {%- assign image_size_2x = image_size | times: 2 | at_most: 5760 -%}
```
andrewetchen, line 347: Added changes to the footer and header by reusing the assigned `image_size` variable instead of specifying `block.settings.image_width` and `section.settings.logo_width` (respectively) more than once.

sections/footer.liquid
```diff
@@ -76,15 +76,16 @@
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
-                      {%- assign image_size = block.settings.image_width | append: 'x' -%}
+                      {%- assign image_size = block.settings.image_width -%}
```
ludoboludo, line 79: I wonder if we actually need `image_size` 🤔 

Maybe we could use `block.settings.image_width` instead to remove one of the assign and only use `image_size_2x` as the assign. 

Ideally we use the least amount of `assign` since `block.settings.image_width` is easy to read anyway.

sections/header.liquid
```diff
@@ -343,9 +343,10 @@
     {%- endif -%}
         <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
           {%- if section.settings.logo != blank -%}
-            {%- assign image_size = section.settings.logo_width | append: 'x' -%}
-            <img srcset="{{ section.settings.logo | img_url: image_size }} 1x, {{ section.settings.logo | img_url: image_size, scale: 2 }} 2x"
-              src="{{ section.settings.logo | img_url: image_size }}"
+            {%- assign image_size = section.settings.logo_width -%}
```
ludoboludo, line 346: same here. Curious to see what others think too but I feel like we could use `section.settings.logo_width` without the assign but keep the `image_size_2x` one.

sections/main-article.liquid
```diff
@@ -279,7 +279,7 @@
         "logo": {
           "@type": "ImageObject",
           "height": {{ settings.share_image.height | json }},
-          "url": {{ settings.share_image | img_url: image_size | prepend: "https:" | json }},
+          "url": {{ settings.share_image | image_url: width: image_size | prepend: "https:" | json }},
```
ludoboludo, line 282: Same in here, no need for the assign I think.

sections/main-product.liquid
```diff
@@ -103,13 +103,18 @@
...
+                  capture media_index
+                    if featured_media.media_type == 'model'
+                      increment model_index
+                    elsif featured_media.media_type == 'video' or featured_media.media_type == 'external_video'
+                      increment video_index
+                    elsif featured_media.media_type == 'image'
+                      increment image_index
+                    endif
+                  endcapture
+                assign media_index = media_index | plus: 1
```
ludoboludo, line 116:
```suggestion
                  assign media_index = media_index | plus: 1
```

sections/main-product.liquid
```diff
@@ -612,9 +622,9 @@
     "name": {{ product.title | json }},
     "url": {{ request.origin | append: product.url | json }},
     {% if seo_media -%}
-      {%- assign media_size = seo_media.preview_image.width | append: 'x' -%}
+      {%- assign media_size = seo_media.preview_image.width -%}
```
ludoboludo, line 625: same here, maybe remove the assign and use `seo_media.preview_image.width` directly if that works 👍

sections/footer.liquid
andrewetchen: Thank you. I was hoping someone would point this out because I initially thought about doing this.

There are a couple instances of this in the footer, header, and main-article. I'll test and update.

sections/main-product.liquid
andrewetchen: I wasn't sure about these assign's within `<script type="application/ld+json">` on whether they needed to be like this. I'll test removing them, thanks.

sections/footer.liquid
```diff
@@ -76,9 +76,9 @@
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
-                      {%- assign image_size = block.settings.image_width | append: 'x' -%}
+                      {%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
                       <img
-                        srcset= "{{ block.settings.image | img_url: image_size }}, {{ block.settings.image | img_url: image_size, scale: 2 }} 2x"
+                        srcset= "{{ block.settings.image | image_url: width: block.settings.image_width }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
                         src="{{ block.settings.image | image_url: width: 400 }}"
```
andrewetchen, line 82: @ludoboludo I noticed that in the `header.liquid`, the width of the `src` attribute is set to the width of the image while this one is set to `400`. Any insight on why this was set to `400`? I'm happy to change it to `block.settings.image_width`.

- https://github.com/Shopify/dawn/pull/134/files#diff-0f3f44a79ab31e1d43ddbed9fdc379488cc710598937d73b57db6c9a8d90177bR80-R82

andrewetchen:
Adding a note to this PR. There's currently a [TemplateLength](https://shopify.dev/themes/tools/theme-check/checks/template-length) warning in the `header.liquid` section:

image

I'll hold off including a `TemplateLength` fix in this PR as I'm not certain where this fix should live.

### Context

@kmeleta, @tyleralsbury, and I agreed that template length can safely be ignored as there's very little reason to set a limit on the number of lines in a template.

Adding the following to `dawn/.theme-check.yml`, would ignore template length warnings when running `shopify theme check` or `theme-check` but only locally and not on PRs:

```yml
...
TemplateLength:
  enabled: false
...
```

To ignore `TemplateLength` on PRs and annotations: we have a couple of options

**Option 1**: Add the following `flags` to `./dawn/.github/workflows/ci.yml`:

```yml
...
  theme-check:
    ...
    steps:
        ...
        with:
          token: ${{ github.token }}
          flags: --exclude-category liquid
```

But this would ignore all [Liquid checks](https://shopify.dev/themes/tools/theme-check/checks#liquid-checks) as we're currently not able to specify `--exclude-category TemplateLength` as a flag.

**Option 2**: Add `{% comment %}theme-check-disable TemplateLength{% endcomment %}` to the top of the `header.liquid` section. But this would only pertain to that file and not others.

**Option 3**: Maybe the theme-check-action could be updated to allow specific category (html, json, liquid, performance, translation) > **checks** (TemplateLength, etc.) to be ignored as an optional flag. Adding something like `flags: --exclude-category liquid:TemplateLength` to ignore TemplateLength warnings on all files.

@charlespwd 👋🏼 

- Do you know if option 3 is possible or whether its something that could be added to [theme-check-action](https://github.com/Shopify/theme-check-action) in the future?
- Is there a better method of ignoring TemplateLength in the CI for PRs?

Issue #1307, opened 2022-02-03
[Product card] Product title and price spacing


Oliviammarcello:
Reduce the spacing between the product title and price on all screens. [Video](https://screenshot.click/03-23-i7j78-p2vh7.mp4) - when I change the margin-bottom to 0, the items look more closely related.

kimberlyoleiro:
@Oliviammarcello this looks to be resolved on Dawn 9.0

<img width="855" alt="Screenshot 2023-04-21 at 4 29 53 PM" src="https://user-images.githubusercontent.com/107572834/233729226-cc4c2c17-b3ca-4e53-b3f7-9e0bd583f192.png">
<img width="288" alt="Screenshot 2023-04-21 at 4 31 37 PM" src="https://user-images.githubusercontent.com/107572834/233729356-374cb63d-5f7b-4889-b9c1-4b95e54e8d56.png">

PR #1388, opened 2022-02-22
Quick Add

### LINKED ISSUE #882: Quick Add
**MVP Brief**: Add ability of "quick buy" on product cards for merchants with low intents to purchase. 
**Variations:** Product cards with no variant, Product cards with 1 or more variants. 
**Design File:** [Figma File](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/?node-id=14203%3A99898)

**1. Variations**
There will be 2 different named buttons for 2 variations. 

- [ ] **Variation 1:** Product Card with no variant. Buttons for this variant will be named "Add to cart"
- [ ] **Variation 2:** Product Cards with 1 or more variants. Buttons for this variant will be named "Choose Options". 
Screen Shot 2021-11-15 at 3 20 51 PM

**2. User Flows**
There will be 2 different flows for 2 variations. 

**Variation 1: Product Card with no variant.** 

Screen Shot 2021-11-16 at 4 26 06 PM

- [ ] When the buyer clicks "add to cart" from the product card, the product will be automatically added to the cart. The button will transition into "Added to cart" - stays in this state (also make the button disabled to avoid adds by mistake) for 5 milliseconds and turns back into default state. 

Screen Shot 2021-11-16 at 4 20 22 PM

- [ ] The cart pop up will be exposed once the product is added to cart and disappear. This interaction is currently existing in Dawn, so no change. In this case, the buyer can add only 1 product at once. If they would like to remove the product from the cart, they will need to navigate to the cart. 

novariant - added to cart

**Variation 1: Edge Case Scenario**
- [ ] The buyer adds a product more than once and hits the max limit of items in stock. In this case, when the limit is hit – change the button label to "Sold out" and to disabled state. If the buyer removes a product from their cart, take the button back to its default state "Add to cart".  _FYI this is reviewed and approved by the content designer and other theme designers._
Screen Shot 2021-11-16 at 4 35 53 PM

**Variation 2: Product Card with 1 or more variants.**

Screen Shot 2021-11-16 at 4 27 37 PM

- [ ] When the buyer clicks "choose options" from the product card, a quick view modal will be exposed for the buyer to select their options, and add to cart. In this case, the buyer can add more than 1 product at once. Both the product title and the view full details is clickable to the product page. (focus state: same as current Dawn specs - underlined)
Screen Shot 2021-11-16 at 4 22 46 PM

- [ ] When the buyer clicks "add to cart" from the quick view modal, the button will transition into "Added to cart" - s stays in this state (also make the button disabled to avoid adds by mistake) for 5 milliseconds and turns back into default state. 
Screen Shot 2021-11-16 at 4 24 14 PM

- [ ] After the buyer closes the quick view modal, the interaction is the same as variant 1. The cart pop up will be exposed once the product is added to cart and disappear. 
Screen Shot 2021-11-16 at 4 25 03 PM

**3. Settings**

- [ ] Add "Enable quick buy button" to the featured collection settings 

Screen Shot 2021-11-16 at 4 38 10 PM

**Note related to the impact on quick buy's cart view via new cart settings:** 
With the new cart settings, the merchant will be able to choose from different cart type options.
- Drawer
- Pop up notification
- Page
Screen Shot 2021-11-16 at 4 42 21 PM

**Impact on Quick Buy flow**
After the buyer adds and item to the cart, the cart type that will be exposed will depend on the merchant's cart type selection. Meaning, 

- If the merchant has selected a drawer cart type from new cart settings – show the cart in drawer.
- If the merchant has selected a pop up notification – show the cart in pop up. 
- If the merchant has selected a page – direct them to the checkout page.

- [ ] Add "Enable quick buy button" to the product grid settings.
product grid settings

**Button transition** – Here is the link to the [prototype](https://www.figma.com/proto/C7c5tmNp5qUNKyr11bAMHB/Quick-buy---button-transition?page-id=0%3A1&node-id=2%3A153&viewport=357%2C48%2C1&scaling=min-zoom).

**Full details** (desktop and mobile views, flows and all cart types) can be found in the [Figma file](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/?node-id=14203%3A99898).

LucasLacerdaUX:
Adds Quick Add to featured collections and collection pages.
Fixes #882 and #1307.

- Added a new `show_quick_buy` option to `card-product.liquid`. Because this card is used in multiple parts of the theme, it allows us to quickly add this functionality to any section.
- Products with no variants will show an `Add to Cart` button, while products with variants will display a `Choose options` that will load a modal once clicked.
- **Modal Content:** to load the modal content, we'll fetch the product page and query the `MainProduct-*` section. There are some pros and cons to this approach.
    - **Pro:** The content on the modal will follow the same order and customization from the product page. From the merchant perspective, this feels right.
    - **Con:** Because the modal content is supposed to be a leaner version of the product page, we must hide a lot of elements using CSS and disable some JS functionality (like URL updating).
- <s>A new state has been added to the `Add to Cart` button indicating a success: `Added to Cart`. In this state, an `aria-disabled` attribute will be applied to the button so the focus is not lost (like `disabled` would do). This success message will stay visible for two seconds. This applies to both quick add buttons & product page buttons.</s>
- If the `/cart/add` request results in an error on a Quick Buy `Add to Cart` (products with no variants), the button will be updated to display a `Sold Out` message and receive the `aria-disabled` attribute.

**Testing steps/scenarios**
- [ ] Product Cards
   - [ ] Product types
      - [ ] Product with no variants
      - [ ] Sold out product
      - [ ] Product with variants 
    - [ ] Viewports
      - [ ] Mobile
      - [ ] Tablet
      - [ ] Desktop 
    - [ ] Pages
      - [ ] Collection Page
      - [ ] Featured Collection section
- [ ] QuickBuy Modal
   - [ ] Test with different product page settings:
      - [ ] `Desktop layout` - should not change the look of the modal
      - [ ] `Desktop media size` - should not change the look of the modal
      - [ ] `Mobile layout` - should not change the look of the modal
   - [ ] Product types
      - [ ] Product with no media
      - [ ] Product with rich media as first image
      - [ ] Product with variant images + `Hide other variants’ media after selecting a variant` setting
    - [ ] Viewports
      - [ ] Mobile
      - [ ] Tablet
      - [ ] Desktop 

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127579160598)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127579160598/editor)

**Checklist**


assets/product-modal.js
```diff
@@ -0,0 +1,34 @@
+if (!customElements.get('product-modal')) {
```
LucasLacerdaUX, line 1: Nothing changed on this file. Just moved it from `featured-product` and `main-product`.

assets/quick-buy.css
```diff
@@ -0,0 +1,188 @@
+.quick-buy-modal {
```
LucasLacerdaUX, line 1: I'll do a follow up PR to combine modal styles in a single component. 

Decided to skip this here to maintain the scope smaller.

LucasLacerdaUX:
Reminder: turn on `Hide whitespace` when reviewing. It'll make your life a lot easier when reviewing `main-product.liquid` changes :)
image

katycobb:
Settings look good to me! 

One consideration: we've renamed the setting to reflect this being an add to cart action, not a straight-to-checkout action (which "quick buy" might suggest). So now there's a bit of a discrepancy with how it's reflected in the code vs. how it's displayed to the merchant. Maybe not a big deal? But if we were to add more of a true "quick buy/buy it now" feature in the future, we might need to revisit.

LucasLacerdaUX:
Thats a good point, @katycobb! Thanks for bringing this up, I fully agree that it's better to refer internally to it as `Quick Add` instead of `Quick Buy`. I'll change every reference to the old name here in GitHub as well.

I'll rename everything to `quick-add`, including the filenames, settings, keys and component naming. However, I'll wait until the `translation-plaform` pushes all the remaining translations to avoid any issues with locale keys.

bredowmax:
This is looking good functionally, but one issue design-wise: Depending on title length, this ATC can end up higher or lower, which ends up looking messy. I wish they would always be at the same height, even if that means a bit more white space for those product cards where the title was shorter

Screen Shot 2022-02-28 at 8 52 47 AM

LucasLacerdaUX:
Thanks for the @svinkle. Demo links have been updated - waiting for your confirmation if `role="presentation"` would be a good solution before pushing it, though :)

> When you click "Add to cart" then Esc out of the cart drawer, focus returns to the launcher as expected. After clicking "Choose options", adding to the cart, then Esc out of the cart drawer, focus is lost. I'd expect it to return to the "Choose options" control.

Great catch! Just updated the demo store to fix that.

> 2. The ul/li in the gallery viewer portion seems unnecessary as it only displays one image at a time.

This happens because I'm re-using the product page layout inside the modal. In order to avoid complex HTML manipulations with JS or having to add repeated elements to the page to match the expected structure, would it be a good solution to apply `role="presentation"` to the `<ul>` list and its `<li>` items? 

> I liked the other demo which included the auto close of the initial "Choose option" modal, sending focus to the cat drawer. Will this be implemented? I'd like to send this specifically for usability testing.

Just updated that! The demo links should now be matching that behaviour :) 
The `Added to Cart` success message is still visible on the product page, for the time being. Waiting on UX decision if we should also remove it from there.

svinkle:
@LucasLacerdaUX 

> would it be a good solution to apply `role="presentation"` to the `<ul>` list and its `<li>` items?

I believe this would do the trick, good thinking. Let me know when this is available to test.

Another issue: on add to card, I see focus moving from the "Choose options" modal, to the "Choose options" `button` then to the cart drawer. Ideally focus would move from modal to modal. When the cart drawer is dismissed, then shift focus to the "Choose options" `button`.

svinkle:
@LucasLacerdaUX The "Choose options" modal has an empty `aria-label` resulting in the screen reader announcing all content within. I'd suggest providing a label such as "Choose options for `{product.title}`".

LucasLacerdaUX:
@svinkle [Demo link ](https://os2-demo.myshopify.com/?preview_theme_id=127579160598) has now been updated to fix the following issues:
- [x] When auto-closing the modal, the focus will no longer shift to the `Choose options` button before moving to the cart popup.
- [x] The gallery list and its items now have a `role="presentation"` attribute.
- [x] Touch target for `Choose options` modal is no longer cut off by other elements.
- [x] The `Choose options` modal dialog now has a `aria-label` with `Choose options for {{ product.title }}`
- [x] `Add to Cart` and `Choose options` buttons are now receiving an `aria-label` with unique descriptions:
    - `Add to cart` now displays `Add {{ product.title }} to cart`
    - `Choose options` now displays `Choose options for {{ product.title }}`

For this last item, I opted for an `aria-label` instead of the `aria-labelledby` solution you suggested because: 
1. We're already adding a `Choose options for` translation for the modal dialog, so I thought it would make sense to re-use it there. 
2. In products with no media, the card title `h3` is hidden by `display: none` which resulted in the `aria-labelledby` not being announced (Chrome + VO)
3. I assumed that this would result in better translations.

However, while `Add {{ product.title }} to cart` is more grammatically correct, does it result in a worse experience for assistive technology users? Let me know if I should revert to `aria-labelledby` and try to fix problem number 2.

duygukalaycioglu:
Overall looking great! The interaction works perfectly. 

Couple quick questions: 

1. On mobile the add to cart is wrapping into 2 lines,  this is happening on 380px width. It is probably normal for screen sizes below <320px but is there a way to not wrap them until they reach iPhone SE or at least <360px = Android?

Screen Shot 2022-02-28 at 10 37 46 AM

3. When I test on my phone, I don't see the loading icon. Is this expected?

4. Spacing updates for modal on mobile: 

- Let's decrease the spacing above to 32px. I believe it is around 50px right now, it will help push the content in the modal above. The buyer has the "X" to close the modal. 
- Let's add some spacing below "view full details link" - on mobile device, it will help creating more space for the buyer to easily click to the link, it is a little bit squished right now. 40px should be good, let's also make sure the top of the image is set to 40 px as well. 
mobile spacings


4. When I was scrolling in the modal, the scroll of the page behind is still active. (double scrolling) Is this expected or a bug?

duygukalaycioglu:
> This is looking good functionally, but one issue design-wise: Depending on title length, this ATC can end up higher or lower, which ends up looking messy. I wish they would always be at the same height, even if that means a bit more white space for those product cards where the title was shorter
> 
> Screen Shot 2022-02-28 at 8 52 47 AM

Thanks for your feedback [bredowmax](https://github.com/bredowmax) we really appreciate it! Locking elements in specific positions on cards is not scalable for cards with large amount of metadata. i.e. When one card with less metadata is positioned next to a card with more metadata, the white space could become a very large gap which also wouldn't be an ideal experience for mobile. This is also problematic with longer product names. That said, when we visit product cards in the future, we'll definitely take your note into consideration. 
We encourage you to open a new issue if you would like to share your thoughts further. :pray:

svinkle:
@LucasLacerdaUX Looking good!

> Add to cart now displays `Add {{ product.title }} to cart`

This violates [2.5.3: Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) which includes this SC failure:

- > "Accessible name contains the visible label text, but the words of the visible label are not in the same order as they are in the visible label text"

This means adjusting the label `Add to cart {{ product.title }}`.

> In products with no media, the card title `h3` is hidden by `display: none`

Curious why this is? This means if there are no photos or anything the product title text is not displayed?

The reason why I suggested `aria-labelledby` is the fact that the text we need is already available in the DOM. It'd be a matter of referencing the `id`s as previously suggested. And when it comes to someone using something like Google Translate, there's a better change of successful translation due to the text being visible in the DOM.

LucasLacerdaUX:
@svinkle I've updated to use `aria-labelledby`. I'm happy that I made this mistake so I could learn about [2.5.3: Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html). Thank you! :)

The reason it was announcing empty wasn't because the content was hidden via `display: none`, but because we had two `<h3>`s with the same `id`.

On products with no image + card style set to `Standard`, the `<h3>` element is located in another container so it's displayed like a product media (see image below). The other, below the image, is then hidden using CSS.
image

I've added a check so no duplicated `id`s are ever in the DOM. Should be working as expected now! 

----

@duygukalaycioglu I have addressed most of your comments about mobile layout. Let me know if you have any additional feedback.

I tested on my phone and I couldn't reproduce the issues with loading spinners or the body being scrolled. I'll test on iPhone later today.

duygukalaycioglu:
The "enable quick add" checkbox on both featured collection section and collection page is missing the sub-copy. 
cc: @katycobb 

sub-copy

@LucasLacerdaUX  The updated button widths looks great! 
Pasting the additional spacing we talked today here as well. 

- Right now it is a little bit challenging to click the view full details on mobile device, adding more spacing (32px) below the modal to provide more space for the buyer to be able to easily click the view full details link. (see the new spacing marked in red) 

new spacing

duygukalaycioglu (changes requested):
@LucasLacerdaUX  The updated button widths looks great! 
Pasting the additional spacing we talked today here as well. 

- Right now it is a little bit challenging to click the view full details on mobile device, adding more spacing (32px) below the modal to provide more space for the buyer to be able to easily click the view full details link. (see the new spacing marked in red) 

new spacing

ludoboludo (commented):
Just sharing some initial thoughts and will continue reviewing by the end of the week: 
- mentioned by Scott already but having a little visual cue on the product card that it's already been added would be great. So we don't have to go to the cart to double check if we've added it already or not.
- would be good to be able to pick a quantity like on Express even for products without variants but that can be follow up work 👍 
- I'm not a big fan of doing the `sold out` button when all items have been added to the cart. I think there must be a better way to approach the situation, like on the product page maybe with an error message. Sold out can be confusing specifically when it's not behaving the same on the product page or if you refresh the page: [video](https://screenshot.click/02-04-uefpo-9ne2e.mp4)
- quick on product recommendation could be a nice follow up
- quick buy isn't coming up on the onboarding states, might not matter much and could be a lower priority follow up
- click outside the modal doesn't close it
- having different layout for the modal could be cool, so that merchants mostly using landscape type of images could show they assets in a better way: [screenshot](https://screenshot.click/02-18-sqq98-gl9ef.png). But if that's something we're doing I'm guessing it will something on its own and for the feat. product section as well. 
- the modal height changes based on the asset size, should we be worried about it ? [video](https://screenshot.click/02-01-n1ule-666jl.mp4)

I hope it's ok, I'm sharing some feedback that is out of scope for the PR but kinda ties back into it as well.

assets/product-form.js
```diff
@@ -35,23 +35,41 @@ if (!customElements.get('product-form')) {
...
+          const quickBuyModal = this.closest('quickbuy-modal');
+          if (quickBuyModal) quickBuyModal.hide(true);
+
+          if (!document.body.classList.contains('overflow-hidden')) {
+            this.cartNotification.renderContents(response);
+          } else {
+            document.body.addEventListener('modalClosed', () => {
+              setTimeout(() => { this.cartNotification.renderContents(response) });
+            }, { once: true });
+          }
```
ludoboludo, line 58: Could this be grouped in the `if (quickBuyModal)` or it wouldn't work 🤔 

So if `quickBuyModal` is `true`, then `hide` it and then have the `setTimeout()` ? Else `this.cartNotification.renderContents(response);`

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+
+.quick-buy-modal[open] {
+  opacity: 1;
+  visibility: visible;
+  z-index: 101;
+}
+
+.quick-buy-modal__content {
+  border-radius: var(--popup-corner-radius);
+  background-color: rgb(var(--color-background));
```
ludoboludo, line 24: Wondering if the modal should follow the section's color scheme:

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+  left: 50%;
+  transform: translateX(-50%);
+  margin-top: 3.2rem;
+  width: 92%;
+  position: absolute;
+  top: 0;
+  padding: 0 1.5rem 0 3rem;
+  border-color: rgba(var(--color-foreground), var(--popup-border-opacity));
+  border-style: solid;
+  border-width: var(--popup-border-width);
```
ludoboludo, line 36: Could this be maybe grouped? Or just easier like this for readability ? I don't have strong feeling about it. Im ok if you leave it as is.

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+  0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3),
+  var(--popup-shadow-horizontal-offset) var(--popup-shadow-vertical-offset) var(--popup-shadow-blur-radius) rgba(var(--color-shadow), var(--popup-shadow-opacity));
+}
+
+@media screen and (min-width: 750px) {
+  .quick-buy-modal__content {
+    padding-right: 1.5rem;
+    margin-top: 10rem;
+    width: 70%;
+    padding: 0 3rem 2rem 3rem;
```
ludoboludo, line 57:
```suggestion
    margin-top: 10rem;
    width: 70%;
    padding: 0 3rem 2rem 3rem;
```

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+quickbuy-modal .thumbnail-list,
+quickbuy-modal .product-popup-modal__opener,
+quickbuy-modal .product__modal-opener:not(.product__modal-opener--image),
+quickbuy-modal .product__media-item:not(:first-child),
+quickbuy-modal pickup-availability,
+quickbuy-modal .product__description,
+quickbuy-modal .product__accordion,
+quickbuy-modal .share-button,
+quickbuy-modal .slider-button,
+quickbuy-modal product-modal {
```
ludoboludo, line 135: We will have to keep this in mind when/if we ever add stuff to the product page. And add them in here if needed

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+}
+
+@media screen and (min-width: 750px) {
+  .quick-buy-modal__toggle {
+    margin: 0 0 0 auto;
+  }
+
+  quickbuy-modal .product--small:not(.product--no-media) .product__media-wrapper,
+  quickbuy-modal .product--medium:not(.product--no-media) .product__media-wrapper,
+  quickbuy-modal .product--large:not(.product--no-media) .product__media-wrapper {
```
ludoboludo, line 161: It might be interesting for the modal to have different sizing options or follow the product page settings. Though that would too "magical". But I could see it being useful when having landscape medias.

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+quickbuy-modal .product__media-item:not(:first-child),
+quickbuy-modal pickup-availability,
+quickbuy-modal .product__description,
+quickbuy-modal .product__accordion,
+quickbuy-modal .share-button,
+quickbuy-modal .slider-button,
+quickbuy-modal product-modal {
+  display: none !important;
+}
+
```
ludoboludo, line 138: There is something to figure out I think on mobile for the `view in your space` button. In this example there is a 3d model but it's not associated to the variants. So it creates extra space and shouldn't be there since from the modal you won't be able to get to the 3d model.

assets/quick-buy.js
```diff
@@ -0,0 +1,67 @@
...
+        .then((response) => response.text())
+        .then((responseText) => {
+          const responseHTML = new DOMParser().parseFromString(responseText, 'text/html');
+          const productInfo = responseHTML.querySelector('section[id^="MainProduct-"]');
+
+          this.setInnerHTML(this.modalContent, productInfo.innerHTML);
+          
+          if (window.Shopify && Shopify.PaymentButton) {
+            Shopify.PaymentButton.init();
+          }
```
ludoboludo, line 32: Something to maybe tweak as on the cart page the dynamic checkout button aren't coming up in modal.

sections/main-product.liquid
```diff
@@ -1,37 +1,37 @@
...
-  <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
-{%- endif -%}
+  {%- assign first_3d_model = product.media | where: "media_type", "model" | first -%}
+  {%- if first_3d_model -%}
+    {{ 'component-product-model.css' | asset_url | stylesheet_tag }}
+    <link id="ModelViewerStyle" rel="stylesheet" href="https://cdn.shopify.com/shopifycloud/model-viewer-ui/assets/v1.0/model-viewer-ui.css" media="print" onload="this.media='all'">
+    <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
+  {%- endif -%}
 
-<section class="page-width section-{{ section.id }}-padding">
```
ludoboludo, line 34: Curious to know why the section element was moved here 🤔

assets/quick-buy.css
```diff
@@ -0,0 +1,197 @@
...
+.quick-buy-modal__content {
+  border-radius: var(--popup-corner-radius);
+  background-color: rgb(var(--color-background));
+  overflow: auto;
+  margin: 0 auto;
+  left: 50%;
+  transform: translateX(-50%);
+  margin-top: 3.2rem;
```
ludoboludo, line 29:
```suggestion
  margin:3.2rem auto 0;
  left: 50%;
  transform: translateX(-50%);
```

Also on mobile it seems to cut the top border when there is one:

ludoboludo (commented):
A few more comments and questions :)

assets/quick-buy.css
duygukalaycioglu: Thanks for the feedback Ludo, that's an interesting thought! But I would park the modal explorations for now, our focus for mvp is the quick add per brief. cc: @LucasLacerdaUX

assets/quick-buy.css
duygukalaycioglu: are you looking at this from the editor or store? Because I remember catching something like this before, and I think @LucasLacerdaUX told me that this doesn't happen on store.

assets/quick-buy.css
duygukalaycioglu: good catch! I marked this spacing to be 40px on one of my previous comments above.

assets/quick-buy.css
ludoboludo: Yep! I double checked on my phone to confirm 👍

duygukalaycioglu:
> I'm not a big fan of doing the sold out button when all items have been added to the cart. I think there must be a better way to approach the situation, like on the product page maybe with an error message. 

I would park further explorations about "sold out" for now, and create a separate ticket to be revisited when we work product card explorations. I agree with the product title should go to the product page, but “sold out” shouldn't. It is using deactivated button style, it would be very confusing to a buyer why it is active and linking to a page. 

Re: adding a message that "this is not sold out", I don't think we have enough space to display it on product cards with images, I also think this is an edge case. @LucasLacerdaUX @ludoboludo 

> The modal height changes based on the asset size. 

Good catch, thanks @ludoboludo! @LucasLacerdaUX If we can force the modal height to be consistent no matter the asset size, that would be awesome!

sections/featured-collection.liquid
```diff
@@ -207,6 +213,12 @@
       "label": "t:sections.featured-collection.settings.show_rating.label",
       "info": "t:sections.featured-collection.settings.show_rating.info"
     },
+    {
+      "type": "checkbox",
+      "id": "enable_quick_buy",
+      "default": false,
+      "label": "t:sections.featured-collection.settings.enable_quick_buy.label"
+    },
```
andrewetchen, line 221: I ran `theme-check` on your `quick-buy` branch and it found 2 `style` suggestions, which can be auto-corrected. If you run `shopify theme check` or `theme-check` (using the standalone version), you'll see what I'm referring to but I'll try to outline them here:

**Offense 1 of 2** - `featured-collection.liquid`

```ruby
sections/featured-collection.liquid:103: style: SchemaJsonFormat: JSON formatting could be improved.
	{% schema %}
```

When you use the auto-correct flag `shopify theme check -a`, you'll be able to see what it fixed. For the `featured-collection.liquid`, it will [remove the empty whitespace from line 221](https://user-images.githubusercontent.com/28404165/156794380-814af91e-7ab3-44c3-9a7c-520f4d2e2989.png)

**Context**: There are currently **22 errors/offenses**. 2 of these can be auto-corrected: 1 mentioned here (above) and I'll add another comment for the other auto-correctable one). The other 20 errors/offenses are being fixed in the **Theme check manual fixes PR**.

sections/main-product.liquid
```diff
@@ -522,597 +531,560 @@
...
-          "label": "t:sections.main-product.settings.gallery_layout.options__3.label"
-        }
+  </script>
+</section>
+
+{% schema %}
+  {
+    "name": "t:sections.main-product.name",
+    "tag": "section",
+    "class": "section",
```
andrewetchen, line 620: **Offense 2 of 2** - `main-product.liquid`

I ran `theme-check` on your `quick-buy` branch and it found 2 `style` suggestions, which can be auto-corrected. If you run `shopify theme check` or `theme-check` (using the standalone version), you'll see what I'm referring to but I'll try to outline them here:

```ruby
sections/main-product.liquid:616: style: SchemaJsonFormat: JSON formatting could be improved.
	{% schema %}
```

When you use the auto-correct flag `shopify theme check -a`, you'll be able to see what it fixed. For the `main-product.liquid`, it will update the entire `{% schema %}` of this file by correcting the indentation, and it will also [remove the whitespace from line 933](https://user-images.githubusercontent.com/28404165/156797478-2c9bc060-6184-43ab-90f3-2dc979d73ff2.png).

sections/main-product.liquid
LucasLacerdaUX: We're now using this element to query and insert it into the modal. The idea is to consistently grab the `main-product` section. Originally, I had implemented a query selector that would use the `<section>` above (the one with `shopify-section`) by selecting the `id` that ends with `__main`. However, I feel like that could be change by the backend and could bring compatibility issues in the future.

Now, we're using an element that is defined by our theme.

assets/quick-buy.css
LucasLacerdaUX: I thought about grouping it too, but it's being used in many places in this exact same way for readability purposes:

image

assets/quick-buy.css
LucasLacerdaUX: I don't think we should and AFAIK, it's not in the plan for MVP at least.

I see modals as their own component and I feel like they should stay consistent across the theme. This same modal is being used for product media, product popup block and quick buy. I think it makes sense for them to stay consistent.

assets/global.js
```diff
@@ -847,14 +848,14 @@ class VariantSelects extends HTMLElement {
...
     const addButtonText = productForm.querySelector('[name="add"] > span');
-
     if (!addButton) return;
 
     if (disable) {
       addButton.setAttribute('disabled', 'disabled');
       if (text) addButtonText.textContent = text;
     } else {
       addButton.removeAttribute('disabled');
+      addButton.removeAttribute('aria-disabled');
```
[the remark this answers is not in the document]
LucasLacerdaUX: This was used for the success message. Good catch.

locales/en.default.json
```diff
@@ -103,6 +103,9 @@
   "products": {
     "product": {
       "add_to_cart": "Add to cart",
+      "added_to_cart": "Added to cart",
```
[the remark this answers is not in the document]
LucasLacerdaUX: This was the previous success message that was displayed on the button.

I won't remove it just yet because we were still discussing if we should bring it back, but good catch to remove before merging (and let the translation-platform take care of removing it later from other languages).

assets/section-main-product.css
```diff
@@ -122,7 +122,7 @@
   text-underline-offset: 0.3rem;
 }
 
-.shopify-payment-button__button--hidden {
+.shopify-payment-button__button + .shopify-payment-button__button--hidden {
```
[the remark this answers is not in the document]
LucasLacerdaUX: This style should've been specific to the second button (`More payment options`) but was being applied to both. That's why it requires two buttons now and targets only the second one.

assets/quick-buy.css
LucasLacerdaUX: Good catch. I'll check why it's not working 👀

assets/quick-buy.js
LucasLacerdaUX: I wonder if we can fix that on our side. I don't think we can, but I'll investigate and maybe report it to their team. 👁️👄👁️ 

It looks like the script behaves differently when inside a cart page, and as far as I know, `Shopify.PaymentButton` doesn't monitor new elements on the page by default.

LucasLacerdaUX:
@ludoboludo
> the modal height changes based on the asset size, should we be worried about it ? [video](https://screenshot.click/02-01-n1ule-666jl.mp4)

Do you have a suggestion on how to fix this? I honestly have no idea. I thought setting a fixed height for the modal, like `80%` height. However, that results in a lot of white space when the product content is smaller. I personally think this looks worse than the modal size changing to match the content.
image

I'll also note that, while it may not be a modal, the `featured-product` section behaves in the exact same way: it expands to fit the maximum content. There's no `max-height` or anything that forces the height of the section to stay the same.

TLDR: I couldn't find a solution and the only one I thought results in a worse-looking experience for most products with short height. Looking for suggestions. (cc @duygukalaycioglu)

LucasLacerdaUX:
I have updated the PR with:
- Fixed elements like `pickup-availability` and `product-modal` being inserted into the DOM.
- Renamed `quick-buy` to `quick-add` everywhere.
- View in your space button is now working.
- Applied `theme-check` automated suggestions
- Spacing fixes.

assets/quick-add.js
```diff
@@ -0,0 +1,79 @@
...
+
+    setInnerHTML(element, html) {
+      element.innerHTML = html;
+      element.querySelectorAll("script").forEach( oldScript => {
+        const newScript = document.createElement("script");
+        Array.from(oldScript.attributes).forEach(attribute => newScript.setAttribute(attribute.name, attribute.value));
+        newScript.appendChild(document.createTextNode(oldScript.innerHTML));
+        oldScript.parentNode.replaceChild(newScript, oldScript);
+      });
+    }
```
kmeleta, line 55: This is an obscure snippet to solve the obscure problem of scripts not executing when copying innerHTML. Could we leave a small comment to explain this? I don't think it would be clear why it's needed.

There's also a couple small formatting things like the unneeded space before oldScript and the use of double quotes instead of single.

assets/quick-add.css
```diff
@@ -0,0 +1,213 @@
+.quick-add-modal {
```
kmeleta, line 1: Anything we can do to ensure there's some margin below a scrollable modal? https://screenshot.click/15-40-ol71e-yb46p.mp4

kmeleta (commented):
Did a first pass. Really clever solution fetching the product page. Just had a few suggestions. Mostly nitpicky stuff.

assets/quick-add.css
duygukalaycioglu: @kmeleta  thanks for noticing this! @LucasLacerdaUX similar to the mobile spacing, if possible let's make sure there is some pre-set margin below the modal at all times.

danielvan:
> > This is looking good functionally, but one issue design-wise: Depending on title length, this ATC can end up higher or lower, which ends up looking messy. I wish they would always be at the same height, even if that means a bit more white space for those product cards where the title was shorter
> > Screen Shot 2022-02-28 at 8 52 47 AM
> 
> Thanks for your feedback [bredowmax](https://github.com/bredowmax) we really appreciate it! Locking elements in specific positions on cards is not scalable for cards with large amount of metadata. i.e. When one card with less metadata is positioned next to a card with more metadata, the white space could become a very large gap which also wouldn't be an ideal experience for mobile. This is also problematic with longer product names. That said, when we visit product cards in the future, we'll definitely take your note into consideration. We encourage you to open a new issue if you would like to share your thoughts further. 🙏

Hi D., when looking at our Figma explorations I always assumed the Add to Cart buttons would be always aligned based on the bottom of the tallest card in the row. I can see your point around large x smaller numbers of metadata.

Is there anywhere I can see a couple of examples of when this might happen, looking for: 
1) Smaller x large amounts of metadata in one row.
2) Cards with super long and short names, or with many pieces of metadata.

I think knowing the cases we are designing for would help us with this decision – from what I can see on the linked store above, the position of add to cart / metadata being shuffled isn't ideal. And it gets exacerbated by the 'adapt to image' style.

danielvan:
> I would park further explorations about "sold out" for now, and create a separate ticket to be revisited when we work product card explorations. I agree with the product title should go to the product page, but “sold out” shouldn't. It is using deactivated button style, it would be very confusing to a buyer why it is active and linking to a page.

Fully agree that Sold Out should not go to the product page. If there is any confusion around the button being present, I'd rather try a different treatment for sold out (e.g. not looking like a button), than removing it.

The messaging being present explains why that item can't be added to a cart, and being able to click on it may set the wrong expectations.

duygukalaycioglu:
> Hi @duygukalaycioglu, when looking at our Figma explorations I always assumed the Add to Cart buttons would be always aligned based on the bottom of the tallest card in the row. I can see your point around large x smaller numbers of metadata.
> 
> Is there anywhere I can see a couple of examples of when this might happen, looking for:
> 
> 1. Smaller x large amounts of metadata in one row.
> 2. Cards with super long and short names, or with many pieces of metadata.
> 
> I think knowing the cases we are designing for would help us with this decision – from what I can see on the linked store above, the position of add to cart / metadata being shuffled isn't ideal. And it gets exacerbated by the 'adapt to image' style.

Hi @danielvan, the Figma files doesn't have these mocks but here is the scenario we are trying to avoid. 

Screen Shot 2022-03-16 at 11 46 27 AM

LucasLacerdaUX:
@duygukalaycioglu Yeah, everything in Figma was designed with a single-media so it was my immediate assumption that it should work like the Featured Product (no sliders or thumbnails. buyers can change media by using variant-picker or going to the full product page via `View full details`)
image

Even the mobile slider is hidden on the mockups:
image

> Let's add the thumbnails under the image, otherwise the buyer is going to think that's the only product image. Modal should reflect all the images available in the actual product detail page.

Things brings me a few concerns:
- Does that mean [Stacked galleries](https://user-images.githubusercontent.com/37168033/158672157-5fcf4e61-55cf-4d60-9dca-39963f842213.png) should also show all images like the product pages? ([Video](https://screenshot.click/16-29-shzvg-ug6ik.mp4)). 
  - It's a minor concern, but this means loading more images and therefore makes the feature less "Quick" for buyers with slow internet connections
  - Users also won't be able to zoom the product image inside this modal. Is this okay?
- Should mobile sliders also be visible on quick-buy modals?
  - Unlike the product page, [where 3D models and videos open inside a modal on mobile](https://screenshot.click/16-33-nq2p7-h94yu.mp4) to prevent issues with swiping the slider, we can't do that inside the Quick Add modal. Media must be played inline. I don't think this results in a good UX.

Unfortunately, this feedback is also coming in very late, so it may also require some additional dev work, testing and UX validation that I'm still unsure of the effort. Can we descope this for the MVP? (cc @nicklepine)

LucasLacerdaUX:
> No, we just need the thumbnails for the modal.

Sounds good! :D

> I think we talked about zoom before, that is ok.

Ah, perfect. I asked because, when we discussed it before, we were considering only full-width images and not half-width like the ones below: 
image

But since we're not planning to show all images on `Stacked`, we don't have to worry about this at all.

> I don't believe the quick add feature is related to how fast the images load.

I was just flagging a minor concern: loading all the images will result in the modal taking longer to load - potentially a lot longer for slower connections. It shouldn't be a dealbreaker, but something to consider as part of the UX.

> I think media playing within the modal is okay, but do you foresee any limitations with the 3d models?

Just clarifying what I meant: the issue isn't the fact that the media will play within the modal (the current implementation already supports 3D models and videos if they're the first media), but that it will play inline on sliders. 

This is an issue because 3D models use the same swipe gesture that is used by the slider, and external videos also block swiping because they're an `<iframe>` element.

> I think this is a common and necessary feature modals have, I'm concerned about the buyer seeing one image and assuming that is the only product image. This may have an impact on the buyer's experience. 

I agree that it can be misleading. I have two follow-up questions to that, though:
- If we are *NOT* hiding other media in `Stacked` layout, isn't it still a confusing UX in that scenario? Would you have any suggestions to improve this?
- Do we want to revisit Featured Product for this same reason? Right now, the buyer only sees one media (the current variant or featured image). If so, I'd suggest creating an issue.

------

Anyway, I'll try to quickly implement the changes you requested, but if it's too much effort and we can't descope, I want to make sure @nicklepine is aware that we won't be able to ship it this week.

assets/quick-add.js
LucasLacerdaUX: Done! :D

assets/quick-add.css
```diff
@@ -0,0 +1,213 @@
...
+quick-add-modal .product__modal-opener:not(.product__modal-opener--image),
+quick-add-modal .product__media-item:not(:first-child),
+quick-add-modal pickup-availability,
+quick-add-modal .product__description,
+quick-add-modal .product__accordion,
+quick-add-modal .share-button,
+quick-add-modal .slider-button,
+quick-add-modal product-modal {
+  display: none !important;
+}
```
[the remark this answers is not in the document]
LucasLacerdaUX: That's a great suggestion! Thanks Ken! :DDD

Just implemented it and it looks so much easier to use now!

LucasLacerdaUX:
@duygukalaycioglu Okay, I did a refactor to the modal layout and I was able to get (most) issues fixed in order to implement the thumbnail switcher.

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127579160598)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127579160598/editor)

So, two remaining questions:
- You can see in the demo links above the issue I mentioned with mobile + 3D models. They use same swipe gesture as the slider, so you must pause the model before scrolling the slider. Sometimes, interacting with the model may scroll the slider.
- When using `Stacked`, additional images are hidden on desktop but shown on mobile via the slider. Is this how it should work? [Video](https://screenshot.click/17-30-z5wcg-ycsd7.mp4)

LucasLacerdaUX:
Update: we're reverting the quick thumbnail switching that I just implemented last night and working on it in a follow up PR.

duygukalaycioglu:
@LucasLacerdaUX  
**One quick update:** We decided to align the "add to cart buttons" always at the bottom of the tallest row, and image and copy stay together if possible. We have identified the use case below as an edge case and we are okay with it for MVP. 
edge case



(✅ approved by @danielvan) cc: @nicklepine

assets/base.css
```diff
@@ -1428,6 +1428,10 @@ button.shopify-payment-button__button--unbranded {
   display: flex;
 }
```
ludoboludo, line 1429: - [ ] I might have asked this before. Is the `add to cart` button text supposed to change once you've added the product to the cart ? [video](https://screenshot.click/17-50-fyjwy-xjqjv.mp4)

ludoboludo, line 1429: - [ ] Design question: i feel like the add to cart could be using a bit more space to not be so close to the price. [screenshot](https://screenshot.click/17-31-3aedw-6tso7.png)

duygukalaycioglu: it stays the same.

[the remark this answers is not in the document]
duygukalaycioglu: good catch!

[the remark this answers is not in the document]
duygukalaycioglu: @LucasLacerdaUX let's do this update if it is not high LOE. if it is, we can park.

duygukalaycioglu: Thanks Ludo! Let's park product card improvements for now, we can revisit once we start working on product cards.

duygukalaycioglu: no, it will stay the same.

ludoboludo: Great, thanks D!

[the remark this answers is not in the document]
LucasLacerdaUX: I agree. I'm not fully convinced that Stacked shouldn't be available on modal either.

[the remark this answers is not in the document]
LucasLacerdaUX: On the modal it's always 50/50 as far as I understand, but maybe this is something that wasn't in Figma so it's a good question to ask: @duygukalaycioglu can you confirm if the `Desktop media size` setting from the product page should be reflected inside the modal or not?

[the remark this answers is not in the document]
LucasLacerdaUX: Good catch. That's because the `modal-dialog` is checking for a `nodeName` = `modal-dialog`, but this is a custom element that has a different node name (`quick-add-modal`). I'll update it to check to compare the target to the element instead :P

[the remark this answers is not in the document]
LucasLacerdaUX: Yeah, I forgot to re-add it but that's a good point. We should bring the skip link when implementing thumbnails.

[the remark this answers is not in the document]
LucasLacerdaUX: That is expected and is related to how the product was set up in the admin. There's a `Size` variant, but with just one option.

If you go to this product's page, it also has a variant selector even if it there's only a single option to choose from. This is a multi-variant product.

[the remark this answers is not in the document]
LucasLacerdaUX: This is how it looks on the product page and I believe it was added to avoid shadows being cut. 

image

We should create an issue to revisit this instead of changing it on quick-add only.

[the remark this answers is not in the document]
duygukalaycioglu: no, let's leave it as is for now.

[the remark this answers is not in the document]
duygukalaycioglu: no, stacked shouldn't be available on modal. Modal should be always defaulted to one image @LucasLacerdaUX
When we revisit the thumbnails, we can take a look at this.

assets/base.css
```diff
@@ -1428,6 +1428,10 @@ button.shopify-payment-button__button--unbranded {
   display: flex;
```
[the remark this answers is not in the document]
LucasLacerdaUX: Yes, it will match the template that is assigned to the product. Description is never shown inside the quick modal, but it does reflect other settings.

Here's an example where each template has a different [variant picker type](https://screenshot.click/17-06-2pegd-vl2ox.mp4).

assets/base.css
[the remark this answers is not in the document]
LucasLacerdaUX: 👀 
I'll add this suggestion to the follow-up PR now that we decided to revert the thumbnail setting.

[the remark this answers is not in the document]
LucasLacerdaUX: I just fixed the focus trap issue when closing the modal! 

For the first issue, I couldn't reproduce it either.

LucasLacerdaUX:
@duygukalaycioglu @ludoboludo @kmeleta 

This should now be ready for another review.

- Thumbnails have been removed and will be added in a follow-up PR.
- `Choose options` was moved to the bottom of the product card. I'm not too familiar with this grid approach used for the card but hopefully I didn't break other cards.
- I believe I've fixed every bug reported so far, except for one I couldn't reproduce (modal failing to open sometimes 👁️ 👄 👁️).

duygukalaycioglu:
Hey Lucas, this is looking really great! 
**One quick feedback:** Since we have removed the thumbnails from the modal for MVP, there is no need to apply the fixed height on the modal. Because it won't go up and down anymore since there will be only 1 image, if it is possible - let's revert that back to not fixed.

Thank you for your hard work on this, super excited for this merge! :tada:

bredowmax:
@LucasLacerdaUX have you tested compatibility with subscriptions? 

I am migrating a store from Recharge (not-integrated checkout) to the new integrated subscription checkout, and it doesn't seem to work for me.

I tested further and found out:
- When using a stock Dawn theme (4.0.0), everything works
- But when using this branch, shopper can select the subscription variant, but when she adds it to the cart, the non-subscription variant is added to the cart

LucasLacerdaUX:
@bredowmax Good catch! I don't remember thoroughly testing compatibility with subscriptions, probably just did a quick pass. This feature injects the same code and scripts from the the product page, so I validated multiple product types, templates and configurations. I've probably missed any bugs related to this test scenario. 

I'll look into that on Monday so we can fix it before we ship. Thanks a lot for reporting this issue and for the overall contributions in this thread.

duygukalaycioglu (dismissed):
> Which one should we go with? :D

**Let's go with option 2** @LucasLacerdaUX = Setting height to max-height.
 The issue Ludo reported about modal jumping up and down won't happen in MVP , because we are not including thumbnails - there will be only 1 image in the modal. Let's park this to be revisited when we add the thumbnails.

sections/main-product.liquid
```diff
@@ -205,9 +205,14 @@
               {{- block.settings.text -}}
```
ludoboludo, line 205: We're not hiding the text block at the moment but I think maybe we should ?

snippets/card-product.liquid
```diff
@@ -138,6 +140,71 @@
...
+                        {%- endif -%}
+                      </span>
+                      <span class="sold-out-message hidden">
+                        {{ 'products.product.sold_out' | t }}
+                      </span>
+                      <div class="loading-overlay__spinner hidden">
+                        <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+                          <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+                        </svg>
+                      </div>
```
ludoboludo, line 175:
```suggestion
                    <span>
                      {%- if card_product.selected_or_first_available_variant.available -%}
                        {{ 'products.product.add_to_cart' | t }}
                      {%- else -%}
                        {{ 'products.product.sold_out' | t }}
                      {%- endif -%}
                    </span>
                    <span class="sold-out-message hidden">
                      {{ 'products.product.sold_out' | t }}
                    </span>
                    <div class="loading-overlay__spinner hidden">
                      <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
                        <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
                      </svg>
                    </div>
```

assets/quick-add.js
```diff
@@ -0,0 +1,83 @@
...
+          newScriptTag.setAttribute(attribute.name, attribute.value)
+        });
+        newScriptTag.appendChild(document.createTextNode(oldScriptTag.innerHTML));
+        oldScriptTag.parentNode.replaceChild(newScriptTag, oldScriptTag);
+      });
+    }
+
+    preventVariantURLSwitching() {
+      this.modalContent.querySelector('variant-radios,variant-selects').setAttribute('data-update-url', 'false');
+    }
```
ludoboludo, line 63: There is something happening when switching variant that are linked to images. It's related to the modal not cutting the content outside of it's limit/size. 
[video](https://screenshot.click/21-49-iqz5v-3377z.mp4)

assets/quick-add.css
```diff
@@ -0,0 +1,200 @@
...
+  transform: translateX(-50%);
+  margin-top: 3.2rem;
+  width: 100%;
+  position: absolute;
+  top: 0;
+  border-color: rgba(var(--color-foreground), var(--popup-border-opacity));
+  border-style: solid;
+  border-width: var(--popup-border-width);
+  box-shadow: var(--popup-shadow-horizontal-offset) var(--popup-shadow-vertical-offset) var(--popup-shadow-blur-radius) rgba(var(--color-shadow), var(--popup-shadow-opacity));
+  overflow: hidden;
```
ludoboludo, line 43: Here I think we have an issue when there are many variants to pick from or a pretty small screen:

assets/quick-add.css
```diff
@@ -0,0 +1,200 @@
...
+  text-underline-offset: 0.2rem;
+  text-decoration-thickness: 0.3rem;
+}
+
+quick-add-modal .product__view-details {
+  margin-bottom: 4rem;
+}
+
+.quick-add__submit {
+  padding: 0 2rem;
```
ludoboludo, line 199: I wonder if we should have some padding top here. When I grabbed the translation from another language it looked like this:

ludoboludo (commented):
Left a few comments. 

The main thing I think if for situation where the modal is cutting some of the content and you're not able to scroll within the modal.

sections/main-product.liquid
LucasLacerdaUX: Because the product vendor element (`OS 2.0`) or the product subtitle are simply a `Text` block with a linked metafield, I kept this type of block available on modals. I would personally vote to keep as it can also be used by merchants to add important information to the product's template. 

However, @duygukalaycioglu can confirm whether or not it should be part of the modal.

assets/quick-add.css
```diff
@@ -0,0 +1,200 @@
...
+  visibility: visible;
+  z-index: 101;
+}
+
+.quick-add-modal__content {
+  border-radius: var(--popup-corner-radius);
+  background-color: rgb(var(--color-background));
+  margin: 0 auto;
+  left: 50%;
+  transform: translateX(-50%);
```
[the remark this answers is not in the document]
LucasLacerdaUX: We can discuss it tomorrow when pairing. 

I'd be cool with that, but the modal would always be jumping positions depending on its size, resulting in the close button is never being on the same Y position.

assets/quick-add.css
LucasLacerdaUX: I think this is now fixed by setting `overflow-y: auto`. Let me know if it works.

LucasLacerdaUX:
Ready for new reviews.

sections/main-collection-product-grid.liquid
```diff
@@ -5,6 +5,11 @@
 
 <link rel="preload" href="{{ 'component-rte.css' | asset_url }}" as="style" onload="this.onload=null;this.rel='stylesheet'">
 
+{%- if section.settings.enable_quick_add -%}
+  <link rel="stylesheet" href="{{ 'quick-add.css' | asset_url }}" media="print" onload="this.media='all'">
+  <script src="{{ 'quick-add.js' | asset_url }}" defer="defer"></script>
+  <script src="{{ 'product-form.js' | asset_url }}" defer="defer"></script>
+{%- endif -%}
```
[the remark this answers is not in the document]
ludoboludo: oh 😅

ludoboludo (commented):
A couple things I noticed so far while testing further and thought I'd mention early: 

- when setting the cards to a thick border, the content information is somehow bigger than the container. Seems to be due with some of the grid styling on it 🤔  [video](https://screenshot.click/23-25-6dmxu-3ipkq.mp4)
- when adding a feat. collection on the product page I'm seeing a couple issue: 
  - on a product that doesn't have a variant. If I add a product to the cart with quick add, then try to add the product from the product page I'm on, it adds back the first item I added from the feat collection
  - on a product that has variants, switching the variant in the modal from quick buy, actually changes the variant from the main product in the background

[Here is a video about it](https://screenshot.click/23-36-ymf6e-68k28.mp4)

sections/main-product.liquid
ludoboludo: After meeting up we came to the conclusion that we'd keep the text block in the modal 👍

assets/quick-add.css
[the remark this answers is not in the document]
LucasLacerdaUX: Decision: We'll change from `max-height` to `height`, which no longer results in the modal being off-center. If we refactor to use `max-height` and always default the size to fit the max media height, we can revisit this :)

kmeleta (commented):
Maybe you're aware of this, but just in case for the sake of time I'll share this issue related to the featured collection on a product page. I think it's a different effect than before. Now some of the main product blocks are getting hidden just by enabling quick add. https://screenshot.click/23-59-a13d5-9pkj8.mp4

Oliviammarcello:
> This one looks the same as on the product page and I agree that it's too much. I believe this extra spacing was introduced a while ago when global settings were introduced and it is used to prevent media shadows from being cut. This is something we should refactor on a separate PR so it fixes both the product page and the quick modal.

> Agreed with your suggestion. I'd love to work on this on a separate PR in a fast follow, though. These adjustments can be a bit tricky to adjust because of the scrollbar, and I'd like to spend more time focused solely on that part. There's also a plan to combine similar modal styles (product page, product popup and quick-add) into a single component.
>
> I'll create an issue.

Makes sense sounds great for both, thanks!! 

I have a few small tweaks left:
- [ ] For some reason only one the shoe test product, the "Add to cart" button isn't outlined. [Screenshot](https://screenshot.click/24-06-gj2lk-fhf4j.png)
- [ ] I know there is a max height on the modal, but then you get this awkward small space below the image. How are we handling super tall images? Is there a max-height? If there is can we set it to allow even spacing above the image and below? [Video. ](https://screenshot.click/24-05-7rp9w-hyvxd.mp4)
- [ ] Not super pressing if we will change the 'X' pattern, so feel free to ignore if it isn't an easy fix. But the 'X' is still slightly misaligned. [Screenshot.](https://screenshot.click/24-08-q5jjl-5zx7i.png) 
- [ ] I realised my suggestion on 20px width of the modal is slightly too narrow. Can we change the width of the modal to align to the width of the content behind it? [Screenshot. ](https://screenshot.click/24-12-5dyxh-x4djh.png)

kmeleta (commented):
I'm noticing now that when quick add is first enabled, either on an existing featured collection or when adding a new one, the quick buy modal doesn't work until the page is re-initialized. I trust this isn't a problem in a production store, but it's probably not an ideal experience for merchants. I wonder if we can adjust how we include the quick add js 🤔 

https://screenshot.click/24-12-jsdmd-rkee4.mp4

assets/quick-add.css
```diff
@@ -0,0 +1,196 @@
...
+
+.quick-add-modal__content img {
+  max-width: 100%;
+}
+
+.quick-add-modal__content-info {
+  padding-right: 4.4rem;
+  display: flex;
+  overflow-y: auto;
+  padding: 2.5rem;
```
kmeleta, line 73: I think adding this padding here is preventing the product` ul.slider` element within from stretching the fullwidth. If the slider was being used as one, this would be obvious when you scrolled to the next image. The slider of course is still active as well https://screenshot.click/24-45-tju54-wfqcs.mp4. This is because we don't conditionally remove the slider classes or anything when the slider has only 1 _visible_ item. I'd advise we don't tackle that existing problem in this PR but the video helps show what the issue with the padding is.

kmeleta (commented):
Also found an issue that's noticeable with media shadows on mobile.

assets/quick-add.css
kmeleta: If the quick add modal is only showing 1 media no matter what right now though we can potentially just fully override the slider overflow within quick add for now.

assets/quick-add.css
```diff
@@ -0,0 +1,181 @@
...
+  text-decoration-thickness: 0.3rem;
+}
+
+quick-add-modal .product__view-details {
+  margin-bottom: 2.5rem;
+}
+
+.quick-add__submit {
+  padding: 0 2rem;
+}
```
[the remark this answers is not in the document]
LucasLacerdaUX: Updated to hide the button. 

As for the `View my cart` -> `Update my cart`, I agree that it would be a nice improvement but I believe we should leave this out of scope for this PR, especially if we want to avoid waiting for translations and design validation on a completely unrelated component.

LucasLacerdaUX:
@kmeleta 

> I'm noticing now that when quick add is first enabled, either on an existing featured collection or when adding a new one, the quick buy modal doesn't work until the page is re-initialized

I wonder if that's on us to fix, though. I'm using the `<script>` tag in the same way as every other `<script>` tag on the theme. The `<link>` just above it loads just fine, and both `<script>` tags are added to the DOM - they're just never requested (confirmed via the Network tab).

The same issue happens on the Search page. If you enable filtering and sorting for the first time, they won't work in the admin until you save and wait for the page to refresh:

https://user-images.githubusercontent.com/37168033/159995673-d4cd37d1-8aa1-4041-a527-9b9aa3ea171e.mp4

This sounds like a Theme Editor issue.

kmeleta:
> This sounds like a Theme Editor issue.

Sounds good 👍  Haven't noticed that issue before

sections/main-product.liquid
```diff
@@ -97,7 +97,7 @@
...
-          <slider-component id="GalleryThumbnails-{{ section.id }}" class="thumbnail-slider slider-mobile-gutter{% if section.settings.mobile_thumbnails == 'hide' %} small-hide{% endif %}{% if media_count <= 3 %} thumbnail-slider--no-slide{% endif %}">
+          <slider-component id="GalleryThumbnails-{{ section.id }}" class="thumbnail-slider slider-mobile-gutter quick-add-hidden{% if section.settings.mobile_thumbnails == 'hide' %} small-hide{% endif %}{% if media_count <= 3 %} thumbnail-slider--no-slide{% endif %}">
             <button type="button" class="slider-button slider-button--prev{% if media_count <= 3 %} small-hide{% endif %}{% if media_count <= 4 %} medium-hide large-up-hide{% endif %}" name="previous" aria-label="{{ 'general.slider.previous_slide' | t }}" aria-controls="GalleryThumbnails-{{ section.id }}" data-step="3">{% render 'icon-caret' %}</button>
             <ul id="Slider-Thumbnails-{{ section.id }}" class="thumbnail-list list-unstyled slider slider--mobile{% if section.settings.gallery_layout == 'thumbnail_slider' %} slider--tablet-up{% endif %}">
               {%- if featured_media != null -%}
```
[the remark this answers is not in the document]
ludoboludo: Im not sure it's something we should address here 🤔  We could be replacing the content of the size attribute with a new logic, only added through JS based on the max-width the modal is getting. 
But happy to have it as a follow up/thing to look into later so that we can settle on the what we currently have.

assets/quick-add.css
```diff
@@ -0,0 +1,201 @@
...
+
+quick-add-modal .product-form__buttons {
+  max-width: initial;
+}
+
+.quick-add__submit {
+  padding: 0 1rem;
+  min-width: 100%;
+}
```
ludoboludo, line 201: One thing which is already happening for feat. product section, is that it will show the `view in your space` button when it's displaying only one media that might not be the 3D model on a mobile device. 
[Screenshot](https://screenshot.click/24-38-n2li7-i5478.png). 
So it can be a bit odd to be able to load something you're not seeing on the page.

sections/main-product.liquid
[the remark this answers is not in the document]
kmeleta: fwiw I think this is fine to do as a follow up because these aren't displaying on page load or anything. This PR has enough going on and optimizations aren't crucial to the functionality here.

assets/quick-add.css
LucasLacerdaUX: Oh, that is intentional. I'll try to find our decision log from 2019 but we designed it so `View in your space` is always present for products with 3D models :D

All first-party themes that received Rich Media support followed this convention. If there is a 3D model, no matters if it's the currently selected media or not, we should display the `view in your space` button linking to the first model in the product.

assets/quick-add.css
```diff
@@ -0,0 +1,201 @@
...
+quick-add-modal .product--stacked .product__info-container--sticky {
+  top: 0;
+  position: relative;
+}
+
+@media screen and (min-width: 750px) {
+  quick-add-modal .product:not(.product--no-media) .product__media-wrapper {
+    max-width: 45%;
+    width: calc(45% - var(--grid-desktop-horizontal-spacing) / 2);
+  }
```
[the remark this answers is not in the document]
LucasLacerdaUX: Related to [this comment](https://github.com/Shopify/dawn/pull/1388#issuecomment-1076820852) by Olivia:
> [ ] Could we reduce the space between the bottom of the image and the first line of text on mobile by 1.5rem? It is a bit large. [Screenshot.](https://screenshot.click/23-25-hp5p5-bo9fs.png)

Would probably be better to fix them both at once in a separate PR.

ludoboludo (approved):
It's looking good to me 🤔 👍 🚀

LucasLacerdaUX:
Just reverted our decision to hide dynamic payment buttons on the cart page.

The button works if there is a featured product in that page, but it doesn't if there isn't:
https://user-images.githubusercontent.com/37168033/160188526-2197ad3a-7e1d-47bf-b329-dad69d7caf74.mp4

So, if we hide the payment buttons (previous):
- No Dynamic Checkout, regardless of Featured Product section being in the cart or not
- No whitespace

If we **DON'T** hide the payment buttons (current):
- Dynamic Checkout works when there's a Featured Product
- Whitespace if there isn't

I believe it's a better approach to keep the whitespace when not supported and to allow it to work when supported. When this issue gets fixed on the payment script side, all merchants will be able to use dynamic checkout buttons in that page without having to update their theme.

Oliviammarcello:
- [ ] The add to cart button for the shoe product still isn't outlined 

Otherwise I think its good to go! :)

LucasLacerdaUX:
@Oliviammarcello 

>  The add to cart button for the shoe product still isn't outlined

This is because that product is set up as a subscription in the platform but no subscription app is installed in the theme we're previewing. It was actually introduced by a PR from June 2021 that added variant pickers, so I don't have the full context on why we made this decision.

Ludo created an issue so we can revisit this separately (https://github.com/Shopify/dawn/issues/1540).

Oliviammarcello (approved):
LGTM! Just as an FYI I am creating an issue to add pickup availability to the modal. This is something that came up in the product review for us to explore =.

deseastudio:
Hello,

I replace my files by your and the "add to cart" button works very well (thank you so much!!)

BUT, when product have variants, when I click on "Choose an option" nothing happens... Can you tell me why ? 

Thanks :)

PR #1407, opened 2022-02-24
Add variant picker sold out UI [temporary solution]

### LINKED ISSUE #45: UI Polish Review: Product Template (Part 2)

cfxd:
This is getting more attention so here's a temporary solution that relies on some JavaScript (and Liquid).

Temporarily fixes #45 (Variant picker) and #105 (Add variant picker sold out UI).

Style radio buttons with `disabled` class; sold out select options can either be disabled, appended, or ignored via block settings.

**Demo links**
Password: `fleomo`
- [Product with one option](https://cfx-theme-playground.myshopify.com/products/billowing-glade)
- [Product with two options](https://cfx-theme-playground.myshopify.com/products/ancient-water)
- [Product with three options](https://cfx-theme-playground.myshopify.com/products/autumn-hill)

**Checklist**

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -832,8 +832,11 @@ class VariantSelects extends HTMLElement {
         const html = new DOMParser().parseFromString(responseText, 'text/html')
         const destination = document.getElementById(id);
         const source = html.getElementById(id);
+        const variantPickerDestination = document.querySelector('variant-radios') || document.querySelector('variant-selects');
+        const variantPickerSource = html.querySelector('variant-radios') || html.querySelector('variant-selects');
 
         if (source && destination) destination.innerHTML = source.innerHTML;
+        if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
 
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -199,7 +199,9 @@ fieldset.product-form__input .form__label {
   cursor: pointer;
   position: relative;
 }
-
+.product-form__input input[type='radio'].disabled + label {
+  opacity: 0.5;
+}
 .product-form__input input[type='radio'] + label:before {
   content: '';
   position: absolute;
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -199,9 +199,6 @@ fieldset.product-form__input .form__label {
   cursor: pointer;
   position: relative;
 }
-.product-form__input input[type='radio'].disabled + label {
-  opacity: 0.5;
-}
 .product-form__input input[type='radio'] + label:before {
   content: '';
   position: absolute;
@@ -233,10 +230,13 @@ fieldset.product-form__input .form__label {
   background-color: rgba(var(--color-background), 0.3);
 }
 
-.product-form__input input[type='radio']:disabled + label {
+.product-form__input input[type='radio']:disabled + label, .product-form__input input[type='radio'].disabled + label {
   border-color: rgba(var(--color-foreground), 0.1);
   color: rgba(var(--color-foreground), 0.4);
   text-decoration: line-through;
+}
+.product-form__input input[type='radio'].disabled:checked + label, .product-form__input input[type='radio']:disabled:checked + label {
+  color: rgba(var(--color-background),.4);
 }
 .product-form__input input[type='radio']:focus-visible + label {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -391,8 +391,8 @@
                                   {% endif %}
                               {% endcase %}
                             {% endfor %}
-                            <option value="{{ value | escape }}" {% if option.selected_value == value %}selected="selected"{% endif %}>
-                              {{ value }}{% if option_disabled %} [{{ 'products.product.unavailable' | t }}]{% endif %}
+                            <option value="{{ value | escape }}" {% if option.selected_value == value %}selected="selected"{% endif %}{% if option_disabled and 'disabled' == block.settings.select_soldout_type %} disabled{% endif %}>
+                              {{ value }}{% if option_disabled and 'append' == block.settings.select_soldout_type %} [{{ 'products.product.unavailable' | t }}]{% endif %}
                             </option>
                           {%- endfor -%}
                         </select>
@@ -758,6 +758,26 @@
           ],
           "default": "button",
           "label": "t:sections.main-product.blocks.variant_picker.settings.picker_type.label"
+        },
+        {
+          "type": "select",
+          "id": "select_soldout_type",
+          "options": [
+            {
+              "value": "none",
+              "label": "t:sections.main-product.blocks.variant_picker.settings.select_soldout_type.options__1.label"
+            },
+            {
+              "value": "disabled",
+              "label": "t:sections.main-product.blocks.variant_picker.settings.select_soldout_type.options__2.label"
+            },
+            {
+              "value": "append",
+              "label": "t:sections.main-product.blocks.variant_picker.settings.select_soldout_type.options__3.label"
+            }
+          ],
+          "default": "none",
+          "label": "t:sections.main-product.blocks.variant_picker.settings.select_soldout_type.label"
         }
       ]
     },
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -25,6 +25,11 @@
 
 <script src="{{ 'product-form.js' | asset_url }}" defer="defer"></script>
 
+{%- assign variants_available_arr = product.variants | map: 'available' -%}
+{%- assign variants_option1_arr = product.variants | map: 'option1' -%}
+{%- assign variants_option2_arr = product.variants | map: 'option2' -%}
+{%- assign variants_option3_arr = product.variants | map: 'option3' -%}
+
 {%- assign first_3d_model = product.media | where: "media_type", "model" | first -%}
 {%- if first_3d_model -%}
   {{ 'component-product-model.css' | asset_url | stylesheet_tag }}
@@ -318,20 +323,38 @@
               {%- if block.settings.picker_type == 'button' -%}
                 <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
                   {%- for option in product.options_with_values -%}
-                      <fieldset class="js product-form__input">
-                        <legend class="form__label">{{ option.name }}</legend>
-                        {%- for value in option.values -%}
-                          <input type="radio" id="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}"
-                                name="{{ option.name }}"
-                                value="{{ value | escape }}"
-                                form="{{ product_form_id }}"
-                                {% if option.selected_value == value %}checked{% endif %}
-                          >
-                          <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
-                            {{ value }}
-                          </label>
-                        {%- endfor -%}
-                      </fieldset>
+                    <fieldset class="js product-form__input">
+                      <legend class="form__label">{{ option.name }}</legend>
+                      {%- for value in option.values -%}
+                        {%- assign option_disabled = true -%}
+                        {% for option1_name in variants_option1_arr %}
+                          {% case option.position %}
+                            {% when 1 %}
+                              {% if variants_option1_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                            {% when 2 %}
+                              {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                            {% when 3 %}
+                              {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == product.selected_or_first_available_variant.option2 and variants_option3_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                {%- assign option_disabled = false -%}
+                              {% endif %}
+                          {% endcase %}
+                        {% endfor %}
+                        <input type="radio" id="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}"
+                              name="{{ option.name }}"
+                              value="{{ value | escape }}"
+                              form="{{ product_form_id }}"
+                              {% if option.selected_value == value %}checked{% endif %}
+                              {% if option_disabled %}class="disabled"{% endif %}
+                        >
+                        <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
+                          {{ value }}
+                        </label>
+                      {%- endfor -%}
+                    </fieldset>
                   {%- endfor -%}
                   <script type="application/json">
                     {{ product.variants | json }}
@@ -351,8 +374,25 @@
                           form="{{ product_form_id }}"
                         >
                           {%- for value in option.values -%}
+                            {%- assign option_disabled = true -%}
+                            {% for option1_name in variants_option1_arr %}
+                              {% case option.position %}
+                                {% when 1 %}
+                                  {% if variants_option1_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                    {%- assign option_disabled = false -%}
+                                  {% endif %}
+                                {% when 2 %}
+                                  {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                    {%- assign option_disabled = false -%}
+                                  {% endif %}
+                                {% when 3 %}
+                                  {% if option1_name == product.selected_or_first_available_variant.option1 and variants_option2_arr[forloop.index0] == product.selected_or_first_available_variant.option2 and variants_option3_arr[forloop.index0] == value and variants_available_arr[forloop.index0] == true %}
+                                    {%- assign option_disabled = false -%}
+                                  {% endif %}
+                              {% endcase %}
+                            {% endfor %}
                             <option value="{{ value | escape }}" {% if option.selected_value == value %}selected="selected"{% endif %}>
-                              {{ value }}
+                              {{ value }}{% if option_disabled %} [{{ 'products.product.unavailable' | t }}]{% endif %}
                             </option>
                           {%- endfor -%}
                         </select>
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1594,6 +1594,18 @@
               "options__2": {
                 "label": "Pills"
               }
+            },
+            "select_soldout_type": {
+              "label": "Soldout select type",
+              "options__1": {
+                "label": "None"
+              },
+              "options__2": {
+                "label": "Disabled"
+              },
+              "options__3": {
+                "label": "Append"
+              }
             }
           }
         },


gregjotau:
@tyleralsbury @ludoboludo I really think one should consider a temporary solution to the theme as a whole if a fix is not planned for this in the coming release. I would classify it as "javascript when needed" :) 

Thanks @cfxd

gregjotau:
@tyleralsbury @ludoboludo still no fix on this? 

It is still something annoying probably all merchants using Dawn.

gregjotau:
> @gregjotau I can confirm we often shy away from using variants altogether to not have to worry about such headaches, because we know Dawn isn't ready yet to be used with variants

Yeah, makes no sense that a "reference theme" does not support the most basic thing of a product page :(

cfxd:
@ludoboludo Took the opportunity to incorporate your feedback along with making the section generally more DRY (only loop through all options once).

sections/main-product.liquid
```diff
@@ -315,52 +322,100 @@
...
+
+                {% capture select_options_html %}
+                  {{ select_options_html }}
+                      </select>
+                      {% render 'icon-caret' %}
+                    </div>
+                  </div>
+                {% endcapture %}
+
+              {%- endfor -%}
```
ludoboludo, line 408: I think we could actually keep what you had prior to this but just do the variable assignment once. So basically taking out just the logic that sets `option_disabled`.

This I find make me look in a few different place to understand what's outputted where. There is a bit of the HTML structure here but also further down. 

Do you think that'd be doable ?

So maybe if you just kept the original code and added right after the logic you had going: 
```html
{%- when 'variant_picker' -%}
            {%- unless product.has_only_default_variant -%}
```

ludoboludo: Hmm yeah you're right. I think I prefer the first approach then. More readable even if it's the for loop is there twice. I'm not sure if I can get someone to review the PR right now to get their input but hopefully soon. 

What do you think/what's your take?

cfxd: I agree the original version was easier to read with the multiple loops but I have a brainstorm of a better way if I can get to it today then I'll try.

cfxd: @ludoboludo Moved the `input` and `select` elements into a separate snippet

ludoboludo: Hey @cfxd sorry I haven't had a chance to re review this yet. Got side tracked by other things. Next week I won't either but I keep it on my list of things to do and get to. 👍

gregjotau: Would be great to have this as part of a 6.0.0 reease 🙌

gregjotau: @ludoboludo since you need 2 reviewers, maybe someone else can look at it this week? This is more important IMO than most of the other PRs

ludoboludo: 👀 let me try and loop in another dev and designer as well to avoid having to rebase the PR over and over to keep up with the rest.

ludoboludo (commented):
A few more comments. 
I'll set a test store with those changes to share with our designers for them to test 👍 

Thanks again for your time and effort @cfxd

cfxd:
@ludoboludo Thank you for the review and comments. I haven't had a solid chunk of time to devote to this lately, but I should have some time this week and even though this feature isn't perfect I think it will be a good step in the right direction.

assets/global.js
```diff
@@ -840,7 +840,11 @@ class VariantSelects extends HTMLElement {
         const html = new DOMParser().parseFromString(responseText, 'text/html')
         const destination = document.getElementById(`price-${this.dataset.section}`);
         const source = html.getElementById(`price-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`);
```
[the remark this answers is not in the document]
cfxd: You tell me, I just went with the existing convention from up above :-/

snippets/product-variant-options.liquid
```diff
@@ -0,0 +1,81 @@
...
+      {{ value }}{% if option_disabled and 'append' == block.settings.select_soldout_type %} [{{ 'products.product.unavailable' | t }}]{% endif %}
+    </option>
+  {%- endcapture -%}
+{%- endfor -%}
+
+{%- if block.settings.picker_type == 'button' -%}
+  {{ button_options_html }}
+{%- elsif block.settings.picker_type == 'dropdown' -%}
+  {{ dropdown_options_html }}
+{%- endif -%}
```
LucasLacerdaUX, line 81: I think it would be better to have the `{%- for -%}` output the HTML directly instead of keeping it on a variable that will only be used once. Something like this.

For simplicity and general readability, we usually try to use less variables in Liquid code. If you look at other examples in Dawn, we try to restrict the use of variables to places where it's actually needed for reusability purposes - and even then, only if it's a big block of code that would cause readability issues if repeated too many times.

Thoughts, @ludoboludo?

```suggestion
  {%- if block.settings.picker_type == 'button' -%}
    <input type="radio" id="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}"
          name="{{ option.name }}"
          value="{{ value | escape }}"
          form="{{ product_form_id }}"
          {% if option.selected_value == value %}checked{% endif %}
          {% if option_disabled %}class="disabled" aria-label="Variant {{ 'products.product.sold_out' | t }} or {{ 'products.product.unavailable' | t }}"{% endif %}
    >
    <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
      {{ value }}
    </label>
  {%- elsif block.settings.picker_type == 'dropdown' -%}
    <option value="{{ value | escape }}"{% if option.selected_value == value %} selected="selected"{% endif %}{% if option_disabled and block.settings.select_soldout_type == 'disabled' %} disabled aria-label="Variant {{ 'products.product.sold_out' | t }} or {{ 'products.product.unavailable' | t }}{% endif %}>
      {{ value }}{% if option_disabled and block.settings.select_soldout_type == 'append' %} [{{ 'products.product.unavailable' | t }}]"{% endif %}
    </option>
  {%- endif -%}
{%- endfor -%}
```

assets/global.js
```diff
@@ -840,7 +840,11 @@ class VariantSelects extends HTMLElement {
...
+        const variantPickerDestination = document.querySelector('variant-radios') || document.querySelector('variant-selects');
+        const variantPickerSource = html.querySelector('variant-radios') || html.querySelector('variant-selects');
+
         if (source && destination) destination.innerHTML = source.innerHTML;
+        if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
```
LucasLacerdaUX, line 847: While this idea of using the fetched components to avoid using JavaScript to calculate available products is great, right now the main issue with this approach is that we're replacing the variant picker (currently selected element) with something else.

This results in focus being lost every time you interact with the component. Try using a keyboard to navigate between the variant pills - each time it reloads, the focus is lost.

I don't have any alternatives in mind, but as is, this breaks keyboard user experience and causes accessibility issues. Video example below.

https://user-images.githubusercontent.com/37168033/176344283-43d46db9-f67d-4b13-87fe-60866f863d0a.mp4

LucasLacerdaUX (commented):
Sorry for taking a long time to review this. I was finally able to complete this first round of feedback. Let me know if you have any questions or thoughts - every suggestion below is open for discussion and clarifications if needed! :D

Like mentioned in other PRs and issues, ideally this `option.disabled` calculation would be done on the platform side and not by some complex Liquid logic in the theme. But until we get there, I like your temporary solution and I think with a few adjustments and UX improvements it can be a great start to include this feature.

Thanks for your contribution and I hope we can continue to improve this temporary solution to get a solid implementation to be included in our themes.

cfxd:
@tyleralsbury what say you?

snippets/product-variant-options.liquid
ludoboludo: I like that 👍 Easier to understand I agree.

tyleralsbury:
> @tyleralsbury any timeline when this will be reviewed?

@LucasLacerdaUX and @ludoboludo are reviewing and are getting some UXer support for review as well.  I will not be reviewing since they've got it handled.

cfxd:
@ludoboludo Just rebased and pushed with all suggested changes. Only thing I'm unclear on is the focus solution. I see the problem, but I don't think there's a way to trigger `:focus-visible` specifically.

ludoboludo:
I feel like we could check on variant change right before we fetch the new variant info, what is the current `document.activeElement` and set the focus on it once we output the HTML we fetched. 
So checking for the active element in here: [updateVariantInput](https://screenshot.click/14-33-6v1vy-1m95g.png)

Or another potential option could be to pass in the `event` of the `change` event listener from `this.addEventListener('change', this.onVariantChange);` and grab the event target so we can reset focus on it later based on its value 🤔

assets/section-main-product.css
```diff
@@ -231,11 +231,17 @@ fieldset.product-form__input .form__label {
   background-color: rgba(var(--color-background), 0.3);
 }
 
-.product-form__input input[type='radio']:disabled + label {
+.product-form__input input[type='radio']:disabled + label,
+.product-form__input input[type='radio'].disabled + label {
   border-color: rgba(var(--color-foreground), 0.1);
   color: rgba(var(--color-foreground), 0.4);
```
ludoboludo, line 237:
```suggestion
  color: rgba(var(--color-foreground), 0.6);
```

I think we can bump up the opacity to be at `0.6` so that we're safer in terms of color contrast. Right now when it's disabled, the contrast is too low to be considered accessible:

edmund-teh:
I just wanted to add some feedback here from a UX perspective. Apologies that it's taken so long to get to.

- I agree with @LucasLacerdaUX here on removing the `Soldout select type` setting that, with the most recent changes, currently only shows `None` and `Disabled` options. Settings that only work conditionally for Type: `Dropdown` vs. Type: `Pills` aren't considered best practices for UX (even if we have these instances elsewhere, let's not continue this pattern).  

- Let's remove this setting and show all dropdown menu options by default with `[Unavailable]` appended. 

- We still want `[Unavailable]` options to be selectable (when Type is set to Dropdown) based on previous feedback in allowing 'back in stock' apps to function correctly. 

- +1 to @ludoboludo's recommendation in adjusting the disabled pill opacity to be at `0.6` to pass for accessibility contrast standards.

- @ludoboludo's updates in improving the experience with the keyboard focus being lost looks great. However, after discussing with @LucasLacerdaUX yesterday, we still need to validate things and lean on our accessibility experts to weigh in on things as there will still be focus loss with this solution... 

> "Even if we save the currently selected element, fetch, replace and re-focus, the focus will be lost for a second. That focus change could still be announced to assistive technologies like voiceover, and might cause confusion for users." - @LucasLacerdaUX  

Finally, I wanted to thank everyone for their contributions and for taking the time to discuss solutions, provide suggestions, etc.

ludoboludo:
I've booked some time on Friday to chat about accessibility with someone internally. I'll keep you all posted.

gregjotau:
> 

Just do a follow up IMO, we have experienced that issue on famme.no

cfxd:
@ludoboludo I believe this is a known shortcoming and one I've referred to that doesn't have a great solution. In the absence of a perfect solution or adding more client side scripting, I believe this is something that should be followed up on and improved/iterated upon once the PR is out in the wild.

ludoboludo:
We're going to aim to provide all the necessary feedback to move forward and hopefully have this merged by the end of next week 👍 
It will be nice to get this PR finally merged 🤞 😓

sections/main-product.liquid
```diff
@@ -336,29 +336,23 @@
...
+                        %}
                       </fieldset>
                   {%- endfor -%}
                   <script type="application/json">
                     {{ product.variants | json }}
                   </script>
                 </variant-radios>
               {%- else -%}
-                <variant-selects class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
+                <variant-selects class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }} id="variant-selects-{{ section.id }}">
```
ludoboludo, line 355:
```suggestion
                <variant-selects id="variant-selects-{{ section.id }}" class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
```

sections/main-product.liquid
```diff
@@ -336,29 +336,23 @@
           {%- when 'variant_picker' -%}
             {%- unless product.has_only_default_variant -%}
               {%- if block.settings.picker_type == 'button' -%}
-                <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
+                <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }} id="variant-radios-{{ section.id }}">
```
ludoboludo, line 339:
```suggestion
                <variant-radios id="variant-radios-{{ section.id }}" class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" {{ block.shopify_attributes }}>
```

sections/featured-product.liquid
```diff
@@ -181,29 +181,23 @@
...
+                          %}
                         </fieldset>
                     {%- endfor -%}
                     <script type="application/json">
                       {{ product.variants | json }}
                     </script>
                   </variant-radios>
                 {%- else -%}
-                  <variant-selects class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }}>
+                  <variant-selects class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }} id="variant-selects-{{ section.id }}">
```
ludoboludo, line 200:
```suggestion
                  <variant-selects id="variant-selects-{{ section.id }}" class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }}>
```

sections/featured-product.liquid
```diff
@@ -181,29 +181,23 @@
             {%- when 'variant_picker' -%}
               {%- unless product.has_only_default_variant -%}
                 {%- if block.settings.picker_type == 'button' -%}
-                  <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }}>
+                  <variant-radios class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }} id="variant-radios-{{ section.id }}">
```
ludoboludo, line 184:
```suggestion
                  <variant-radios id="variant-radios-{{ section.id }}" class="no-js-hidden" data-section="{{ section.id }}" data-url="{{ product.url }}" data-update-url="false" {{ block.shopify_attributes }}>
```

snippets/product-variant-options.liquid
```diff
@@ -0,0 +1,67 @@
...
+          form="{{ product_form_id }}"
+          {% if option.selected_value == value %}checked{% endif %}
+          {% if option_disabled %}class="disabled" aria-label="{{ 'products.product.variant_sold_out_or_unavailable' | t }}"{% endif %}
+    >
+    <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
+      {{ value }}
+    </label>
+  {%- elsif block.settings.picker_type == 'dropdown' -%}
+    <option value="{{ value | escape }}"{% if option.selected_value == value %} selected="selected"{% endif %}{% if option_disabled and block.settings.select_soldout_type == 'disabled' %} aria-label="{{ 'products.product.variant_sold_out_or_unavailable' | t }}"{% endif %}>
+      {{ value }}{% if option_disabled %} [{{ 'products.product.unavailable' | t }}]{% endif %}
```
ludoboludo, line 63: There should be a check here as well to see if the block setting for `select_soldout_type` is set to `none` or `disabled`:
```suggestion
      {{ value }}{% if option_disabled %} [{{ 'products.product.unavailable' | t }}]{% endif %}
```

sections/featured-product.liquid
```diff
@@ -215,11 +209,11 @@
...
-                              </option>
-                            {%- endfor -%}
+                            {% render 'product-variant-options',
+                              product: product,
+                              option: option,
+                              block: block
+                            %}
                           </select>
                           {% render 'icon-caret' %}
                         </div>
```
ludoboludo, line 219: We're missing the setting here, if we are to keep it, similar than what we have on `main-product.liquid` to choose `none` or `disabled` for `select_soldout_type`. 

@edmund-teh do you think we should offer that setting to merchant ? Basically it's a setting that only applies when the variant type is `dropdown` and here are the two visual output: 
- when set to none: [screenshot](https://screenshot.click/23-28-mq406-w621s.png)
- when set to disabled: [screenshot](https://screenshot.click/23-23-61xxa-10i5h.png)

Maybe it's good to keep both options for folks using apps to manage pre orders or out of stocks orders. 🤔

snippets/product-variant-options.liquid
```diff
@@ -0,0 +1,67 @@
...
+    <input type="radio" id="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}"
+          name="{{ option.name }}"
+          value="{{ value | escape }}"
+          form="{{ product_form_id }}"
+          {% if option.selected_value == value %}checked{% endif %}
+          {% if option_disabled %}class="disabled" aria-label="{{ 'products.product.variant_sold_out_or_unavailable' | t }}"{% endif %}
+    >
+    <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
+      {{ value }}
+    </label>
```
ludoboludo, line 60: Here we can do a small change for accessibility purposes. The aria label is taking over what's being announced. So instead of adding one, let's add that text in a `visually-hidden` `<span>` within the `label`.
```suggestion
          {% if option_disabled %}class="disabled"{% endif %}
    >
    <label for="{{ section.id }}-{{ option.position }}-{{ forloop.index0 }}">
      {{ value }}{% if option_disabled %}<span class="visually-hidden">{{ 'products.product.variant_sold_out_or_unavailable' | t }}</span>{% endif %}
    </label>
```

snippets/product-variant-options.liquid
```diff
@@ -0,0 +1,67 @@
...
+  {%- elsif block.settings.picker_type == 'dropdown' -%}
+    <option value="{{ value | escape }}"{% if option.selected_value == value %} selected="selected"{% endif %}{% if option_disabled and block.settings.select_soldout_type == 'disabled' %} aria-label="{{ 'products.product.variant_sold_out_or_unavailable' | t }}"{% endif %}>
```
ludoboludo, line 62: Here we can remove the aria label as well: 
```suggestion
    <option value="{{ value | escape }}"{% if option.selected_value == value %} selected="selected"{% endif %}>
```

ludoboludo (commented):
Alright, I think those should be my last round of comments. 

In terms of accessibility, could you add the fix I mentioned in my previous comment ? Here is [the commit](https://github.com/Shopify/dawn/commit/807a5562b5f1bd50b53771d18213c445a0a53066) with that fix on my test PR that reuses your approach. 

There is a rebase needed which is mostly due to formatting changes but there are a few other things to watch out for. You can [double check my PR](https://github.com/Shopify/dawn/pull/1779) as I did the rebase to see which changes were conflicting.

gregjotau:
Merge this week? 🤓

cfxd:
@ludoboludo @kmeleta Apologies for the delay, just was able to get to this now!

ludoboludo (commented):
Niice 👌 

There is that issue with unavailable variant that isn't sitting right with me 🤔 I'm sure it's not really likely for merchants to have unavailable variants on their product pages but it must still happen. 

The fact that if you currently have selected an `unavailable` variant and change the option above it (option2 or option1), then it won't re render and update accordingly the options and show you which ones are sold out/unavailable. 
[video](https://screenshot.click/16-36-ual3v-2ujoa.mp4)

gregjotau:
bump

sections/featured-product.liquid
```diff
@@ -894,13 +885,6 @@
         {
           "type": "title"
         },
-        {
-          "type": "text",
-          "settings": {
-            "text": "{{ section.settings.product.metafields.descriptors.subtitle.value }}",
-            "text_style": "subtitle"
-          }
-        },
```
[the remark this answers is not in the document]
ludoboludo: Nope actually you're good. It was just that your branch was not up to date with the latest but it's been removed on `main` too 👍

ludoboludo (approved):
It's looking good to me 🤔 

I might look into a follow up for that issue I mentioned with unavailable variants. Otherwise you got my approval. I'll loop in Ken to let him know.

edmund-teh (approved):

kmeleta (commented):
Looks like everything has been addressed and working as I'd expect. The one exception being the quirk that Ludo mentioned about the selected variant options not being re-rendered. Good eye, Ludo. I totally missed that. However I'm also ok with us opening an issue for this and handling it as a fast-follow on our end, if need be. I don't think it's serious enough to be blocking at this point.

We do need to wait for translations to come back, but I think we're good to go after that 🚀 I'll formally approve then.

kmeleta (approved):

fayskerritt1:
Hello, this code half works for me, it works in the sense that it disables the size (which is the second variant looped through in 'option in product.options_with_values) however it does not disable the colours (the first variant) when changing the size unless there are no sizes available at all. Is there a way to check which colours there are in the size that is selected?

To explain further using the below image as an example, when clicked on XS (which you can see is selected) how can we get the colours (the coloured variant images) to be disabled in the same way the sizes are when you click on colours.
image

cfxd:
This code works as expected. It assumes—which has been the Shopify way for quite some time—that shoppers will select an option from option 1  first (color in your case), and that any subsequent options become progressively more specific. So, in effect, shoppers are shopping "by option 1" or "by color." If size is more important, then size should be option 1.

The desired outcome you described is a bit easier and perhaps more intuitive to grasp at a glance from a shopper's perspective, where option states reflect which option _combinations_ are available by prioritizing the option which was most recently selected/changed. In a sense, your flow would detect which option shoppers are shopping by. Something like that is definitely possible and I have seen a few themes in the wild that do this, but it requires a bit more client side code and is probably best left as a customization for now.

ludoboludo:
Exactly what @cfxd mentioned. Thanks for jumping in 🙂 
The logic is top down and now in a different order. So for now that's the default logic applied. I'd argue that it'd be a bit confusing for a user to know what means what if the logic worked based on the last selected option 🤔 But I what might seem easier for me to grasp is different for everyone 😅

Issue #1404, opened 2022-02-24
Add a "Read more" button option to long descriptions


bacounts:
**Describe the enhancement you'd like**
A "Read more" button should be an option for long descriptions throughout the theme (i.e., Collection description, product description, custom meta fields on certain pages).

Some use longer text as an SEO strategy — while this practice is not ideal for most, it is still a common practice and should be accounted for as a setting within the editor.

**Additional context/screenshots**
<img width="1792" alt="Screen Shot 2022-02-23 at 4 33 14 PM" src="https://user-images.githubusercontent.com/6481954/155434927-9c7e04df-f874-4ad0-9cdb-76acf1493759.png">

PR #1443, opened 2022-03-02
Create vertical filter

sofiamatulis:
**PR Summary:** 

Fixes #0.

**Other considerations**

**Testing steps/scenarios**

**Demo links**

- [Editor](url)

**Checklist**

### DIFF
--- a/assets/component-facets.css
+++ b/assets/component-facets.css
@@ -906,3 +906,66 @@ input.mobile-facets__checkbox {
 .product-count__text.loading + .loading-overlay__spinner {
   display: block;
 }
+
+@media screen and (min-width: 990px) {
+.facets-vertical {
+  display: flex;
+}
+}
+
+@media screen and (min-width: 990px) {
+.facets-vertical-container {
+width: 33%;
+}
+}
+
+@media screen and (min-width: 990px) {
+#ProductGridContainer {
+  width: 66%;
+}
+}
+
+.facets-vertical .disclosure-has-popup[open] > summary::before {
+  content: none;
+}
+
+.facets-vertical .facets__display {
+  position: initial;
+  border: none;
+  max-height: none;
+}
+
+.facets-vertical .facet-filters {
+  padding-left: 0;
+}
+
+.facets-vertical .facets__form {
+  display: flex;
+  flex-direction: column;
+  width: 35rem;
+}
+
+.facets-vertical .facet-filters__field {
+  justify-content: flex-start;
+}
+
+.facets-vertical .facets__wrapper {
+  flex-direction: column;
+  align-items: initial;
+  align-self: initial;
+  flex-wrap: initial;
+}
+
+.facets-vertical .facets__heading {
+  margin: 0;
+}
+
+.facets-vertical .facets__header,
+.facets-vertical .facets__list {
+  padding-left: 0;
+}
+
+.facets-vertical .facets__disclosure {
+  margin-right: 0;
+}
+
--- a/sections/main-collection-product-grid.liquid
+++ b/sections/main-collection-product-grid.liquid
@@ -21,16 +21,16 @@
   }
 {%- endstyle -%}
 
-<div class="section-{{ section.id }}-padding">
+<div class="section-{{ section.id }}-padding{% if section.settings.filter_type == 'vertical' %} facets-vertical{% endif %}">
   {%- if section.settings.enable_filtering or section.settings.enable_sorting -%}
     {{ 'component-facets.css' | asset_url | stylesheet_tag }}
     <script src="{{ 'facets.js' | asset_url }}" defer="defer"></script>
   
-    <div class="page-width" id="main-collection-filters" data-id="{{ section.id }}">
-      {% render 'facets', results: collection, enable_filtering: section.settings.enable_filtering, enable_sorting: section.settings.enable_sorting, collapse_on_larger_devices: section.settings.collapse_on_larger_devices %}
+    <div class="page-width{% if section.settings.filter_type == 'vertical'%} facets-vertical-container{% endif %}" id="main-collection-filters" data-id="{{ section.id }}">
+      {% render 'facets', results: collection, enable_filtering: section.settings.enable_filtering, enable_sorting: section.settings.enable_sorting, collapse_on_larger_devices: section.settings.collapse_on_larger_devices, filter_type: section.settings.filter_type %}
     </div>
   {%- endif -%}
-  
+
   <div id="ProductGridContainer">
     {%- paginate collection.products by section.settings.products_per_page -%}
       {%- if collection.products.size == 0 -%}
@@ -154,6 +154,22 @@
       "default": true,
       "label": "t:sections.main-collection-product-grid.settings.enable_filtering.label",
       "info": "t:sections.main-collection-product-grid.settings.enable_filtering.info"
+    },
+    {
+      "type": "select",
+      "id": "filter_type",
+      "options": [
+        {
+          "value": "vertical",
+          "label": "t:sections.main-collection-product-grid.settings.filter_type.options__1.label"
+        },
+        {
+          "value": "horizontal",
+          "label": "t:sections.main-collection-product-grid.settings.filter_type.options__2.label"
+        }
+      ],
+      "default": "horizontal",
+      "label": "t:sections.main-collection-product-grid.settings.filter_type.label"
     },
     {
       "type": "checkbox",
--- a/snippets/facets.liquid
+++ b/snippets/facets.liquid
@@ -6,9 +6,10 @@
     - enable_filtering: {Boolean} Show filtering when true
     - enable_sorting: {Boolean} Show sorting when true
     - collapse_on_larger_devices: {Boolean} Collapse filtering/sorting into menu on larger devices when true
+    - filter_type: {String} Type of filter
 
     Usage:
-    {% render 'facets', results: collection, enable_filtering: true, enable_sorting: true, collapse_on_larger_devices: false %}
+    {% render 'facets', results: collection, enable_filtering: true, enable_sorting: true, collapse_on_larger_devices: false, filter_type: 'vertical' %}
 {% endcomment %}
 
 {%- liquid
@@ -26,7 +27,28 @@
   {%- unless collapse_on_larger_devices -%}
     <facet-filters-form class="facets small-hide">
       <form id="FacetFiltersForm" class="facets__form">
-
+        {%- if enable_sorting and filter_type == 'vertical' -%}
+          <div class="facet-filters sorting caption">
+            <div class="facet-filters__field">
+              <h2 class="facet-filters__label caption-large text-body">
+                <label for="SortBy">{{ 'products.facets.sort_by_label' | t }}</label>
+              </h2>
+              <div class="select">
+                {%- assign sort_by = results.sort_by | default: results.default_sort_by -%}
+                <select name="sort_by" class="facet-filters__sort select__select caption-large" id="SortBy" aria-describedby="a11y-refresh-page-message">
+                  {%- for option in results.sort_options -%}
+                    <option value="{{ option.value | escape }}"{% if option.value == sort_by %} selected="selected"{% endif %}>{{ option.name | escape }}</option>
+                  {%- endfor -%}
+                </select>
+                {% render 'icon-caret' %}
+              </div>
+            </div>
+      
+            <noscript>
+              <button type="submit" class="facets__button-no-js button button--tertiary">{{ 'products.facets.sort_button' | t }}</button>
+            </noscript>
+          </div>
+        {%- endif -%}
         {%- if results.terms -%}
           <input type="hidden" name="q" value="{{ results.terms | escape }}">
           <input name="options[prefix]" type="hidden" value="last">
@@ -205,7 +227,7 @@
           <input type="hidden" name="q" value="{{ results.current_vendor }}{{ results.current_type }}">
         {% endif %}
 
-        {%- if enable_sorting -%}
+        {%- if enable_sorting and filter_type == 'horizontal' -%}
           <div class="facet-filters sorting caption">
             <div class="facet-filters__field">
               <h2 class="facet-filters__label caption-large text-body">
--- a/assets/component-facets-vertical.css
+++ b/assets/component-facets-vertical.css
@@ -1,0 +1,189 @@
+.facets-vertical {
+    display: flex;
+}
+  
+.facets-vertical-container {
+width: 33%;
+}
+
+#ProductGridContainer {
+    width: 66%;
+}
+
+.facets__item label,
+.facets__item input[type='checkbox'] {
+  cursor: pointer;
+}
+
+.facet-checkbox {
+  padding: 1rem 2rem 1rem 0;
+  flex-grow: 1;
+  position: relative;
+  font-size: 1.4rem;
+  display: flex;
+  word-break: break-word;
+}
+
+.facet-checkbox input[type='checkbox'] {
+  position: absolute;
+  opacity: 1;
+  width: 1.6rem;
+  height: 1.6rem;
+  top: 0.7rem;
+  left: -0.4rem;
+  z-index: -1;
+  appearance: none;
+  -webkit-appearance: none;
+}
+
+.facet-checkbox > svg {
+  background-color: rgb(var(--color-background));
+  margin-right: 1.2rem;
+  flex-shrink: 0;
+}
+
+.facet-checkbox .icon-checkmark {
+  visibility: hidden;
+  position: absolute;
+  left: 0.3rem;
+  z-index: 5;
+  top: 1.4rem;
+}
+
+.facet-checkbox > input[type='checkbox']:checked ~ .icon-checkmark {
+  visibility: visible;
+}
+
+.facets__heading {
+display: block;
+color: rgba(var(--color-foreground), 0.85);
+font-size: 1.4rem;
+margin: -1.5rem 2rem 0 0;
+}
+
+.disclosure-has-popup[open] > summary::before {
+    content: none;
+}
+
+.active-facets {
+    display: flex;
+    flex-wrap: wrap;
+    width: 100%;
+    grid-column: 1 / -1;
+    grid-row: 2;
+    margin-top: -0.5rem;
+  }
+  
+  .active-facets__button {
+    display: block;
+    margin-right: 1.5rem;
+    margin-top: 1.5rem;
+    padding-left: 0.2rem;
+    padding-right: 0.2rem;
+    text-decoration: none;
+  }
+
+span.active-facets__button-inner {
+    color: rgb(var(--color-foreground));
+    box-shadow: 0 0 0 0.1rem rgb(var(--color-foreground));
+    border-radius: 2.6rem;
+    font-size: 1rem;
+    min-height: 0;
+    min-width: 0;
+    padding: 0.5rem 1rem;
+    display: flex;
+    align-items: stretch;
+  }
+  
+  span.active-facets__button-inner:before,
+  span.active-facets__button-inner:after {
+    display: none;
+  }
+  
+  .facets__summary {
+    color: rgba(var(--color-foreground), 0.75);
+    font-size: 1.4rem;
+    margin-bottom: 1.5rem;
+    padding: 0 1.75rem 0 0;
+  }
+
+  .facet-filters__field {
+    align-items: center;
+    display: flex;
+    flex-grow: 1;
+    justify-content: flex-end;
+  }
+  
+  .facet-filters__field .select {
+    width: auto;
+  }
+  
+  .facet-filters__field .select:after,
+  .facet-filters__field .select:before,
+  .mobile-facets__sort .select:after,
+  .mobile-facets__sort .select:before {
+    content: none;
+  }
+  
+  .facet-filters__field .select__select,
+  .mobile-facets__sort .select__select {
+    border-radius: 0;
+    min-width: auto;
+    min-height: auto;
+    transition: none;
+  }
+
+  .facet-filters {
+    align-items: flex-start;
+    display: flex;
+    grid-column: 2;
+    grid-row: 1;
+    padding-left: 2.5rem;
+  }
+  
+  @media screen and (min-width: 990px) {
+    .facet-filters {
+      padding-left: 3rem;
+    }
+  }
+  
+
+  .facet-filters button {
+    margin-left: 2.5rem;
+  }
+  
+  .facet-filters__sort {
+    background-color: transparent;
+    border: 0;
+    border-radius: 0;
+    font-size: 1.4rem;
+    height: auto;
+    line-height: calc(1 + 0.5 / var(--font-body-scale));
+    margin: 0;
+    padding-left: 0;
+    padding-right: 1.75rem;
+  }
+  
+  .facet-filters__sort + .icon-caret {
+    right: 0;
+  }
+  
+  @media screen and (forced-colors: active) {
+    .facet-filters__sort {
+      border: none;
+    }
+  }
+  
+  .facet-filters__sort,
+  .facet-filters__sort:hover {
+    box-shadow: none;
+    filter: none;
+    transition: none;
+  }
+
+  .facet-filters__label {
+    display: block;
+    color: var(--color-foreground-85);
+    font-size: 1.4rem;
+    margin: 0 2rem 0 0;
+  }
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1428,6 +1428,15 @@
           "label": "Enable filtering",
           "info": "Customize [filters](/admin/menus)"
         },
+        "filter_type": {
+          "label": "Filter type",
+          "options__1": {
+            "label": "Vertical"
+          },
+          "options__2": {
+            "label": "Horizontal"
+          }
+        },
         "enable_sorting": {
           "label": "Enable sorting"
         },
--- a/snippets/facets-vertical.liquid
+++ b/snippets/facets-vertical.liquid
@@ -1,0 +1,508 @@
+{% comment %}
+    Renders facets (filtering and sorting)
+
+    Accepts:
+    - results: {Object} Collection or Search object
+    - enable_filtering: {Boolean} Show filtering when true
+    - enable_sorting: {Boolean} Show sorting when true
+    - collapse_on_larger_devices: {Boolean} Collapse filtering/sorting into menu on larger devices when true
+
+    Usage:
+    {% render 'facets', results: collection, enable_filtering: true, enable_sorting: true, collapse_on_larger_devices: false %}
+{% endcomment %}
+
+{%- liquid
+    assign sort_by = results.sort_by | default: results.default_sort_by
+    assign total_active_values = 0
+    if results.url
+      assign results_url = results.url
+    else 
+      assign terms = results.terms | escape
+      assign results_url = '?q=' | append: terms | append: '&options%5Bprefix%5D=last&sort_by=' | append: sort_by
+    endif
+  -%}
+
+  <div class="facets-container">
+    {%- unless collapse_on_larger_devices -%}
+      <facet-filters-form class="facets small-hide">
+        <form id="FacetFiltersForm" class="facets__form">
+          {%- if enable_sorting -%}
+            <div class="facet-filters sorting caption">
+              <div class="facet-filters__field">
+                <h2 class="facet-filters__label caption-large text-body">
+                  <label for="SortBy">{{ 'products.facets.sort_by_label' | t }}</label>
+                </h2>
+                <div class="select">
+                  {%- assign sort_by = results.sort_by | default: results.default_sort_by -%}
+                  <select name="sort_by" class="facet-filters__sort select__select caption-large" id="SortBy" aria-describedby="a11y-refresh-page-message">
+                    {%- for option in results.sort_options -%}
+                      <option value="{{ option.value | escape }}"{% if option.value == sort_by %} selected="selected"{% endif %}>{{ option.name | escape }}</option>
+                    {%- endfor -%}
+                  </select>
+                  {% render 'icon-caret' %}
+                </div>
+              </div>
+        
+              <noscript>
+                <button type="submit" class="facets__button-no-js button button--tertiary">{{ 'products.facets.sort_button' | t }}</button>
+              </noscript>
+            </div>
+          {%- endif -%}
+
+          <div class="product-count light" role="status">
+            <h2 class="product-count__text text-body">
+              <span id="ProductCountDesktop">
+                {%- if results.results_count -%}
+                  {{ 'templates.search.results_with_count' | t: terms: results.terms, count: results.results_count }}
+                {%- elsif results.products_count == results.all_products_count -%}
+                  {{ 'products.facets.product_count_simple' | t: count: results.products_count }}
+                {%- else -%}
+                  {{ 'products.facets.product_count' | t: product_count: results.products_count, count: results.all_products_count }}
+                {%- endif -%}
+              </span>
+            </h2>
+            <div class="loading-overlay__spinner">
+              <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+                <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+              </svg>
+            </div>
+          </div>    
+  
+          {%- if results.terms -%}
+            <input type="hidden" name="q" value="{{ results.terms | escape }}">
+            <input name="options[prefix]" type="hidden" value="last">
+          {%- endif -%}
+  
+          {% if enable_filtering %}
+            <div id="FacetsWrapperDesktop" class="facets__wrapper">
+              {%- unless results.filters == empty -%}
+                <h2 class="facets__heading caption-large text-body">{{ 'products.facets.filter_by_label' | t }}</h2>
+              {%- endunless -%}
+  
+              {%- for filter in results.filters -%}
+                {%- assign total_active_values = total_active_values | plus: filter.active_values.size -%}
+                {% case filter.type %}
+                {% when 'list' %}
+                  <details id="Details-{{ forloop.index }}-{{ section.id }}" class="disclosure-has-popup facets__disclosure js-filter" data-index="{{ forloop.index }}">
+                    <summary class="facets__summary caption-large focus-offset">
+                      <div>
+                        <span>{{ filter.label | escape }}</span>
+                        {% render 'icon-caret' %}
+                      </div>
+                    </summary>
+                    <div id="Facet-{{ forloop.index }}-{{ section.id }}" class="facets__display">
+                      <div class="facets__header">
+                        <span class="facets__selected no-js-hidden">{{ 'products.facets.filters_selected' | t: count: filter.active_values.size }}</span>
+                        <facet-remove>
+                          <a href="{{ filter.url_to_remove }}" class="facets__reset link underlined-link">
+                            {{ 'products.facets.reset' | t }}
+                          </a>
+                        </facet-remove>
+                      </div>
+  
+                      <ul class="facets__list list-unstyled" role="list">
+                        {%- for value in filter.values -%}
+                          <li class="list-menu__item facets__item">
+                            <label for="Filter-{{ filter.label | escape }}-{{ forloop.index }}" class="facet-checkbox{% if value.count == 0 and value.active == false %} facet-checkbox--disabled{% endif %}">
+                              <input type="checkbox"
+                                name="{{ value.param_name }}"
+                                value="{{ value.value }}"
+                                id="Filter-{{ filter.label | escape }}-{{ forloop.index }}"
+                                {% if value.active %}checked{% endif %}
+                                {% if value.count == 0 and value.active == false %}disabled{% endif %}
+                              >
+  
+                              <svg width="1.6rem" height="1.6rem" viewBox="0 0 16 16" aria-hidden="true" focusable="false">
+                                <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
+                              </svg>
+  
+                              <svg class="icon icon-checkmark"
+                                width="1.1rem"
+                                height="0.7rem"
+                                viewBox="0 0 11 7"
+                                fill="none"
+                                xmlns="http://www.w3.org/2000/svg">
+  
+                                <path d="M1.5 3.5L2.83333 4.75L4.16667 6L9.5 1"
+                                  stroke="currentColor"
+                                  stroke-width="1.75"
+                                  stroke-linecap="round"
+                                  stroke-linejoin="round" />
+                              </svg>
+  
+                              {{ value.label | escape }} ({{ value.count }})
+                            </label>
+                          </li>
+                        {%- endfor -%}
+                      </ul>
+                    </div>
+                  </details>
+                {% when 'price_range' %}
+                  {% liquid
+                    assign currencies_using_comma_decimals = 'ANG,ARS,BRL,BYN,BYR,CLF,CLP,COP,CRC,CZK,DKK,EUR,HRK,HUF,IDR,ISK,MZN,NOK,PLN,RON,RUB,SEK,TRY,UYU,VES,VND' | split: ','
+                    assign uses_comma_decimals = false
+                    if currencies_using_comma_decimals contains cart.currency.iso_code
+                      assign uses_comma_decimals = true
+                    endif
+                  %}
+                  <details id="Details-{{ forloop.index }}-{{ section.id }}" class="disclosure-has-popup facets__disclosure js-filter" data-index="{{ forloop.index }}">
+                    <summary class="facets__summary caption-large focus-offset">
+                      <div>
+                        <span>{{ filter.label | escape }}</span>
+                        {% render 'icon-caret' %}
+                      </div>
+                    </summary>
+                    <div id="Facet-{{ forloop.index }}-{{ section.id }}" class="facets__display">
+                      <div class="facets__header">
+                        {%- assign max_price_amount = filter.range_max | money | strip_html | escape -%}
+                        <span class="facets__selected">{{ "products.facets.max_price" | t: price: max_price_amount }}</span>
+                        <facet-remove>
+                          <a href="{{ filter.url_to_remove }}" class="facets__reset link underlined-link" >
+                            {{ 'products.facets.reset' | t }}
+                          </a>
+                        </facet-remove>
+                      </div>
+                      <price-range class="facets__price">
+                        <span class="field-currency">{{ cart.currency.symbol }}</span>
+                        <div class="field">
+                          <input class="field__input"
+                            name="{{ filter.min_value.param_name }}"
+                            id="Filter-{{ filter.label | escape }}-GTE"
+                            {%- if filter.min_value.value -%}
+                              {%- if uses_comma_decimals -%}value="{{ filter.min_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"{%- else -%}value="{{ filter.min_value.value | money_without_currency | replace: ',', '' }}"{% endif %}
+                            {%- endif -%}
+                            type="number"
+                            placeholder="0"
+                            min="0"
+                            {%- if uses_comma_decimals -%}max="{{ filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' }}"{%- else -%}max="{{ filter.range_max | money_without_currency | replace: ',', '' }}"{% endif %}
+                          >
+                          <label class="field__label" for="Filter-{{ filter.label | escape }}-GTE">{{ 'products.facets.from' | t }}</label>
+                        </div>
+                        <span class="field-currency">{{ cart.currency.symbol }}</span>
+                        <div class="field">
+                          <input class="field__input"
+                            name="{{ filter.max_value.param_name }}"
+                            id="Filter-{{ filter.label | escape }}-LTE"
+                            {%- if filter.max_value.value -%}{%- if uses_comma_decimals -%}value="{{ filter.max_value.value | money_without_currency | replace: '.', '' | replace: ',', '.' }}"{%- else -%}value="{{ filter.max_value.value | money_without_currency | replace: ',', '' }}"{% endif %}
+                            {%- endif -%}
+                            type="number"
+                            min="0"
+                            {%- if uses_comma_decimals -%}
+                              placeholder="{{ filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
+                              max="{{ filter.range_max | money_without_currency | replace: '.', '' | replace: ',', '.' }}"
+                            {%- else -%}
+                              placeholder="{{ filter.range_max | money_without_currency | replace: ',', '' }}"
+                              max="{{ filter.range_max | money_without_currency | replace: ',', '' }}"
+                            {% endif %}
+                          >
+                          <label class="field__label" for="Filter-{{ filter.label | escape }}-LTE">{{ 'products.facets.to' | t }}</label>
+                        </div>
+                      </price-range>
+                    </div>
+                  </details>
+                {% endcase %}
+              {%- endfor -%}
+              <noscript>
+                <button type="submit" class="facets__button-no-js button button--tertiary">{{ 'products.facets.filter_button' | t }}</button>
+              </noscript>
+            </div>
+  
+            <div class="active-facets active-facets-desktop">
+              {%- for filter in results.filters -%}
+                {%- for value in filter.active_values -%}
+                  <facet-remove>
+                    <a href="{{ value.url_to_remove }}" class="active-facets__button active-facets__button--light">
+                      <span class="active-facets__button-inner button button--tertiary">
+                        {{ value.label | escape }}
+                        {% render 'icon-close-small' %}
+                        <span class="visually-hidden">{{ 'products.facets.clear_filter' | t }}</span>
+                      </span>
+                    </a>
+                  </facet-remove>
+                {%- endfor -%}
+                {% if filter.type == "price_range" %}
+                  {%- if filter.min_value.value != nil or filter.max_value.value != nil -%}
+                    <facet-remove>
+                      <a href="{{ filter.url_to_remove }}" class="active-facets__button active-facets__button--light">
+                        <span class="active-facets__button-inner button button--tertiary">
+                          {%- if filter.min_value.value -%}{{ filter.min_value.value | money }}{%- else -%}{{ 0 | money }}{%- endif -%}-{%- if filter.max_value.value -%}{{ filter.max_value.value | money }}{%- else -%}{{ filter.range_max | money }}{%- endif -%}
+                          {% render 'icon-close-small' %}
+                          <span class="visually-hidden">{{ 'products.facets.clear_filter' | t }}</span>
+                        </span>
+                      </a>
+                    </facet-remove>
+                  {%- endif -%}
+                {% endif %}
+              {%- endfor -%}
+              <facet-remove class="active-facets__button-wrapper">
+                <a href="{{ results_url }}" class="active-facets__button-remove underlined-link">
+                  <span>{{ 'products.facets.clear_all' | t }}</span>
+                </a>
+              </facet-remove>
+            </div>
+          {% endif %}
+  
+          {% if results.current_vendor or results.current_type %}
+            <input type="hidden" name="q" value="{{ results.current_vendor }}{{ results.current_type }}">
+          {% endif %}
+          
+        </form>
+      </facet-filters-form>
+    {%- endunless -%}
+  
+    <menu-drawer class="mobile-facets__wrapper {% unless collapse_on_larger_devices %} medium-hide large-up-hide{% endunless %}" data-breakpoint="mobile">
+      <details class="mobile-facets__disclosure disclosure-has-popup">
+        <summary class="mobile-facets__open-wrapper focus-offset">
+          <span class="mobile-facets__open">
+            {% render 'icon-filter' %}
+            <span class="mobile-facets__open-label button-label">
+              {%- if enable_filtering and enable_sorting -%}
+                {{ 'products.facets.filter_and_sort' | t }}
+              {%- elsif enable_filtering -%}
+                {{ 'products.facets.filter_button' | t }}
+              {%- elsif enable_sorting -%}
+                {{ 'products.facets.sort_button' | t }}
+              {%- endif -%}
+            </span>
+          </span>
+          <span tabindex="0" class="mobile-facets__close mobile-facets__close--no-js">{%- render 'icon-close' -%}</span>
+        </summary>
+        <facet-filters-form>
+          <form id="FacetFiltersFormMobile" class="mobile-facets">
+            <div class="mobile-facets__inner">
+              <div class="mobile-facets__header">
+                <div class="mobile-facets__header-inner">
+                  <h2 class="mobile-facets__heading">
+                    {%- if enable_filtering and enable_sorting -%}
+                      {{ 'products.facets.filter_and_sort' | t }}
+                    {%- elsif enable_filtering -%}
+                      {{ 'products.facets.filter_button' | t }}
+                    {%- elsif enable_sorting -%}
+                      {{ 'products.facets.sort_button' | t }}
+                    {%- endif -%}
+                  </h2>
+                  <p class="mobile-facets__count">
+                      {%- if results.results_count -%}
+                        {{ 'templates.search.results_with_count' | t: terms: results.terms, count: results.results_count }}
+                      {%- elsif results.products_count == results.all_products_count -%}
+                        {{ 'products.facets.product_count_simple' | t: count: results.products_count }}
+                      {%- else -%}
+                        {{ 'products.facets.product_count' | t: product_count: results.products_count, count: results.all_products_count }}
+                      {%- endif -%}
+                  </p>
+                </div>
+              </div>
+              <div class="mobile-facets__main">
+                {%- for filter in results.filters -%}
+                  {% case filter.type %}
+                  {% when 'list' %}
+                    <details id="Details-Mobile-{{ forloop.index }}-{{ section.id }}" class="mobile-facets__details js-filter" data-index="mobile-{{ forloop.index }}">
+                      <summary class="mobile-facets__summary focus-inset">
+                        <div>
+                          <span>{{ filter.label | escape }}</span>                        
+                          <span class="mobile-facets__arrow no-js-hidden">{% render 'icon-arrow' %}</span>
+                          <noscript>{% render 'icon-caret' %}</noscript>
+                        </div>
+                      </summary>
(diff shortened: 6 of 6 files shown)


sofiamatulis:
> Hi Sofia, thanks for sharing. A couple of content tweaks needed:


@katycobb Thanks Katy! The content has been updated. If it looks good, I will request translations 🙏

katycobb:
> > Hi Sofia, thanks for sharing. A couple of content tweaks needed:
> 
> @katycobb Thanks Katy! The content has been updated. If it looks good, I will request translations 🙏

Looks good, thanks!

sofiamatulis:
> * [x]  Price content is getting outside of the filters.
> 

@duygukalaycioglu This has been fixed :) But to allow the price to fit I had to increase the width of the vertical filter

duygukalaycioglu (changes requested):
I'll continue looking at spacings and text sizes next week, here are quick updates:

- [x] - change "clear all" to "remove all" as discussed with Scott
- [ ] - "remove all" link text size = increase to 14px
- [ ] - arrows should rotate once the filter tabs are opened
- [ ] - let's make the spacings within the tabs consistent (see below)
- [ ] – let's also make some adjustments for the price filter elements (see below) - if possible remove line below this is the highest price copy, left align, apply new spacing 

**global spacing** 
Screen Shot 2022-03-17 at 8 48 24 PM

**price filter adjustments**
Screen Shot 2022-03-17 at 8 50 40 PM

bacounts:
> Because of that, we are just decreasing the size of the inputs

Could the inputs stack at an earlier or 1-time breakpoint? Reducing the input size might be visually problematic for longer-priced products (i.e., $100,000.00).

The ideal solution would be to create a responsive range slider with smaller input fields below, but this would likely be tacked in a separate PR. See example from ticket: https://github.com/Shopify/dawn/issues/255#issuecomment-927820553

There is quite a bit of discussion around the price input fields in this ticket: https://github.com/Shopify/dawn/issues/255

duygukalaycioglu:
> @duygukalaycioglu For the spacing this is the current spacing for horizontal filter. Would you like me to update that one as well or just for vertical?

just for vertical, thank you! @sofiamatulis

KaichenWang (commented):
Added some initial comments. I'm continuing to test and focusing on the Liquid changes next.

At certain screen widths (i.e. 800px), some of the layout and styling seems off:
Screen Shot 2022-03-21 at 7 55 35 PM

sections/main-collection-product-grid.liquid
```diff
@@ -23,60 +23,108 @@
...
+          <div class="product-count light" role="status">
+            <h2 class="product-count__text text-body">
+              <span id="ProductCountDesktop">
+                {%- if collection.results_count -%}
+                  {{ 'templates.search.results_with_count' | t: terms: collection.terms, count: collection.results_count }}
+                {%- elsif results.products_count == results.all_products_count -%}
+                  {{ 'products.facets.product_count_simple' | t: count: collection.products_count }}
+                {%- else -%}
+                  {{ 'products.facets.product_count' | t: product_count: collection.products_count, count: collection.all_products_count }}
+                {%- endif -%}
```
andrewetchen, line 58: Can you help me better understand what the first condition is doing here? Theme check is throwing an error but maybe I'm missing some context on what `collection.results_count` is 🤔

andrewetchen:
Here are some things I've encountered:

Similar to what Kai mentioned here: https://github.com/Shopify/dawn/pull/1443#pullrequestreview-916166247

I noticed this between `750px` to `989px`:

- [Vertical filters layout](https://user-images.githubusercontent.com/28404165/159356723-e260c645-3749-4052-8d94-25ec44180c33.png)
- [Vertical filter with size small selected](https://user-images.githubusercontent.com/28404165/159356806-31dd7ff7-82e2-40a1-9f05-0685df6f191e.png)
- Here's a [video recording](https://screenshot.click/vertical-_-filters_video_.mp4) to help illustrate this.

**No JS**

- [Vertical filters layout](https://user-images.githubusercontent.com/28404165/159357032-50295a1d-85c6-44a9-bdb7-1b5cf253a071.png)
- [Vertical filter with size small selected](https://user-images.githubusercontent.com/28404165/159357076-76688719-e630-472a-82a7-01fc524c96e1.png)

I'll keep testing.

locales/en.default.json
```diff
@@ -163,6 +163,10 @@
         "one": "{{ count }} selected",
         "other": "{{ count }} selected"
       },
+      "filter_selected": " ({{ count }})",
+      "filter_selected_accessibility": "{{ type }} ({{ count }} filters selected)",
+      "show_more": "+Show more",
```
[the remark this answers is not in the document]
sofiamatulis: Oh yes, that is a good point 👍 Before moving it, @svinkle What are your thoughts on that? 🤔

assets/component-facets.css
```diff
@@ -233,6 +233,10 @@
   right: 0;
 }
 
+.facets-vertical .facets__disclosure[open] .facets__summary .icon-caret {
```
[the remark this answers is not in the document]
sofiamatulis: Only because we are only rotating for the vertical filter 🤔 Unless youd like to do it for the horizontal too cc @duygukalaycioglu

locales/en.default.json
```diff
@@ -163,6 +163,10 @@
         "one": "{{ count }} selected",
         "other": "{{ count }} selected"
       },
+      "filter_selected": " ({{ count }})",
```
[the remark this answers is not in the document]
KaichenWang: Ok i see. Unless we expect this string to look different when translated, the string doesn't need to be defined in `en.default.json`. Instead you could just hard code the parenthesis in Liquid. Something like:
```
({{ filter.active_values.size }})
```

sections/main-collection-product-grid.liquid
sofiamatulis: @andrewetchen I copied this over from `facets.liquid` (different order of elements so we can achieve a different layout) where you have this:

```
        <span id="ProductCountDesktop">
                {%- if results.results_count -%}
                  {{ 'templates.search.results_with_count' | t: terms: results.terms, count: results.results_count }}
                {%- elsif results.products_count == results.all_products_count -%}
                  {{ 'products.facets.product_count_simple' | t: count: results.products_count }}
                {%- else -%}
                  {{ 'products.facets.product_count' | t: product_count: results.products_count, count: results.all_products_count }}
                {%- endif -%}
              </span>

```

it is just a different variable 🤔 I will investigate why this error is showing

assets/component-facets.css
```diff
@@ -903,3 +926,148 @@ input.mobile-facets__checkbox {
...
+    justify-content: flex-start;
+    padding-bottom: 1rem;
+    padding-top: 2rem;
+  }
+
+  .facets-vertical .facets__price {
+    padding: 2rem 0.5rem 3.2rem 0;
+  }
+
+  .facets-vertical .facets__price .field:last-of-type {
```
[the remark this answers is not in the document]
sofiamatulis: I don't see it in the docs specifically saying it works or doesn't but it seems to work 🤔 : https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type the examples use elements, but it doesn't mention that you can't use it on a class and it seems to work

https://screenshot.click/21-40-zroeb-adzgm.mp4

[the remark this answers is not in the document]
KaichenWang: Yeah I think `type` is assumed to only work on types of elements and not class selectors. I think in this case, `.field:last-of-type` is equivalent to `div:last-of-type`. I'm ok leaving this as is, but just wanted to make sure this doesn't cause anything unexpected.

assets/component-facets.css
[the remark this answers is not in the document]
duygukalaycioglu: Good question! The decision on rotating the arrow is a decision made during mega menu project. 

I would park the horizontal filters as is. There are other arrows who don't rotate right now. It would make sense to create a separate ticket to rotate all arrows. cc: @nicklepine

duygukalaycioglu (changes requested):
**Quick feedback:** 
- [ ] Let's add space after "+" before Show on "+ Show more" and same for "- Show less"
- [x] **Padding below Size:** Remove `margin-top` 
padding below sizing

- [x] **Padding around "+ Show more" link:** Remove `margin-bottom` and decrease `margin-top` to 1.3 rem 
padding around show more

- [x] **Padding around price range:** `padding `1rem 0.5rem 2.9rem 0;
padding around price range

- [x] **Padding below price:** Remove `margin-top`
padding below price

assets/component-facets.css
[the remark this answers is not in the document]
sofiamatulis: I am aligned 👍

sections/main-collection-product-grid.liquid
sofiamatulis: I forgot to update the following line! Should be fixed now 👍

snippets/facets.liquid
```diff
@@ -22,78 +22,131 @@
...
+            {%- endif -%}
 
             {%- for filter in results.filters -%}
               {%- assign total_active_values = total_active_values | plus: filter.active_values.size -%}
               {% case filter.type %}
               {% when 'boolean' or 'list' %}
-                <details id="Details-{{ forloop.index }}-{{ section.id }}" class="disclosure-has-popup facets__disclosure js-filter" data-index="{{ forloop.index }}">
-                  <summary class="facets__summary caption-large focus-offset">
+                <details id="Details-{{ forloop.index }}-{{ section.id }}" class="disclosure-has-popup facets__disclosure js-filter" data-index="{{ forloop.index }}"{% if filter_type == 'vertical' and  forloop.index == 1 %} open{% endif %}>
+                  <summary class="facets__summary caption-large focus-offset" aria-label="{{ filter.label }} ({{ 'products.facets.filters_selected.one' | t: count: filter.active_values.size }})">
```
sofiamatulis, line 88: @svinkle I have added the aria label here but instead of adding `Size (2 filters selected)` I kept the existing translation `Size (2 selected)`. Let me know your thoughts. If not I can create a new string and request translations 👍 

https://screenshot.click/22-52-10efv-ejq9k.png

assets/component-facets.css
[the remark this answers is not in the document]
sofiamatulis: That is a fair point. I can also change it to a different approach if you feel that would be better. What are your thoughts?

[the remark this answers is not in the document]
sofiamatulis: We aligned to keep it this way since we will always be adding a `field` class to a `div`

assets/component-facets.css
```diff
@@ -903,3 +932,164 @@ input.mobile-facets__checkbox {
...
+  
+  .facets-vertical .facets__wrapper {
+    flex-direction: column;
+    align-items: initial;
+    align-self: initial;
+    flex-wrap: initial;
+  }
+  
+  .facets-vertical .facets__disclosure {
+    border-top: solid rgba(var(--color-foreground),var(--popup-border-opacity));
```
andrewetchen, line 966: Thoughts on using the same borders as the Collapsible content section's accordion?

- https://github.com/Shopify/dawn/blob/3e5c4f8fb5cdef5db1cc6f4bb1a5eb081c5c29f7/assets/component-accordion.css#L22-L27

I was testing on Firefox and Safari and noticed the borders looked "thicker":

<details>
  <summary>Firefox</summary>

  image

</details>

<details>
  <summary>Safari</summary>

  image

</details>

I noticed something re: the first filter being open (we chatted about this earlier) but I'll add that as a comment on your PR 👍🏼

sofiamatulis:
> * global.js:9



> * . The facet selectors are missing attributes to convey the disclosure pattern (I thought this was automatic with `id="Details-"` parent selector in [global.js:9](https://github.com/Shopify/dawn/blob/main/assets/global.js#L9)?)

@svinkle I just tested it and it seems to be working. Can you confirm? 

https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=127667044374

assets/facets.js
```diff
@@ -154,11 +160,31 @@ class FacetFiltersForm extends HTMLElement {
...
-    const searchParams = new URLSearchParams(formData).toString();
-    FacetFiltersForm.renderPage(searchParams, event);
+    const sortFilterForms = document.querySelectorAll('facet-filters-form form');
+    if (event.srcElement.className == 'mobile-facets__checkbox') {
+      const searchParams = this.createSearchParams(event.target.closest('form'))
+      this.onSubmitForm(searchParams, event)
+    } else {
+      const forms = [];
+      sortFilterForms.forEach((form) => {
+        if (form.id == 'FacetSortForm' || form.id == 'FacetFiltersForm') {
```
KaichenWang, line 181: This could even be a 1-liner

```
if (form.id === 'FacetSortForm' || form.id === 'FacetFiltersForm') forms.push(this.createSearchParams(form))
```

KaichenWang (commented):
Added a few more comments. Still reviewing the Liquid files.

assets/global.js
```diff
@@ -8,7 +8,7 @@ function getFocusableElements(container) {
 
 document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
   summary.setAttribute('role', 'button');
-  summary.setAttribute('aria-expanded', 'false');
+  summary.setAttribute('aria-expanded', summary.parentNode.hasAttribute('open'));
```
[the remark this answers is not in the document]
sofiamatulis: This is a good point about other parts of the theme 🤔 the reason I added this is because by default the first summary is open. So if we do this, we are always setting it to `false` and in this case the first one is open.

https://screenshot.click/23-39-ov5ns-a67ed.png
Otherwise by default this would be false: https://screenshot.click/23-40-jcfhx-u44zd.png

assets/component-facets.css
```diff
@@ -903,3 +932,164 @@ input.mobile-facets__checkbox {
...
+@media screen and (min-width: 750px) {
+  .facets-vertical {
+    display: flex;
+  }
+
+  .facets-vertical .disclosure-has-popup[open] > summary::before {
+    content: none;
+  }
+  
+  .facets-vertical .facets__display {
```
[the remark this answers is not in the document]
sofiamatulis: I see what you mean 🤔 I would still keep them in the same file though since there are still some classes that are shared.

From what I have seen most of the updates are spacing and the direction of classes that is why i kept most of it. And these are examples of some that are re-used: `.facets`, `..facets__summary`, `.facets__disclosure`, `facets__header`, `facets__heading`, `.active-facets__button` and `active-facets__button-inner` at a first glance.

I can separate it or duplicate the classes. I was just trying to re-use as much as I could.

What do you think? 🤔 Happy to go either way

assets/facets.js
```diff
@@ -8,6 +8,12 @@ class FacetFiltersForm extends HTMLElement {
     }, 500);
 
     this.querySelector('form').addEventListener('input', this.debouncedOnSubmit.bind(this));
+    this.querySelector('form').addEventListener('keydown', (event) => {
```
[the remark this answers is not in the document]
sofiamatulis: Do you mean that instead of doing the above id do something like saving it in a variable (pseudo code below)? 🤔 Trying to make sure I understand

`const allElements = this.querySelectorAll('li')`

```
    allElements.addEventListener('keydown', (event) => { 
      if (event.target.tagName === 'LI' && event.code === 'Space') {
        event.preventDefault();
        event.target.querySelector('input').click();
      }
    });
```

assets/component-facets.css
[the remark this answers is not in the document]
KaichenWang: Might be best to just flag this as a future refactor to-do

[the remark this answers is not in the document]
sofiamatulis: Sounds good! Issue created here https://github.com/Shopify/dawn/issues/1527

sofiamatulis:
@svinkle  Thanks for the feedback! :)
 
> * [ ]  1. Checkbox focus indicator design doesn't match other checkboxes (see horizontal filter for example.)

What do you mean by this exactly? they are both sharing the same html and css for focus so it should be the same 

> * [x]  2. I noticed the `li` is receiving focus via `tabindex="0"`. This shouldn't be applied since `li` are not actionable items. Rather, allow the `input` to receive focus in order to convey its presence, semantic meaning, and state (checked or unchecked.) Remove the `tabindex` attribute from both `li` and `input`.

I removed this, but now the inputs don't receive any focus when tabbing. what do you suggest?

sofiamatulis:
> * On "Show more" click, **shift focus to the first, newly revealed `input` element**. This will provide an indication that clicking the "Show more" control did _something_, that there are now more facets to explore, and that the user can navigate forward to explore said facet options.
> * On "Show less" click, focus is to remain on the `button` control.

Updated! Code here for reference :) https://github.com/Shopify/dawn/pull/1443/commits/4e8648b49fb8474dcf7dac727edc8d8356d0c36e

svinkle:
@sofiamatulis 

> they are both sharing the same html and css for focus so it should be the same

Not sure, but toying with the CSS, when I remove `zindex: -1` the focus indicator appears.

Chrome devtools, unticked z-index property reveals the checkbox focus indicator. A large red arrow points from the checkbox to this property in devtools.

assets/facets.js
```diff
@@ -7,7 +7,33 @@ class FacetFiltersForm extends HTMLElement {
...
+    const showMore = this.querySelector('.button-show-more');
+    if (showMore) {
+      showMore.addEventListener('click', (event) => {
+        showMore.closest('.facets__display').querySelector('.facets-wrap').classList.toggle('facets-wrap-show-more', !showMore.closest('.facets__display').querySelector('.facets-wrap').classList.contains('facets-wrap-show-more')); 
+        showMore.closest('.facets__display').querySelectorAll('.facets__item.facets__item-show-more').forEach(item => item.classList.toggle('hidden', !item.classList.contains('hidden')))
+      })
+      showMore.addEventListener('keydown', (event) => {
+        if (event.target.className === 'button-show-more link underlined-link' && event.code === 'Space') {
+          event.preventDefault();
+          showMore.closest('.facets__display').querySelector('.facets-wrap').classList.toggle('facets-wrap-show-more', !showMore.closest('.facets__display').querySelector('.facets-wrap').classList.contains('facets-wrap-show-more'));
```
andrewetchen, line 28:
```suggestion
    facetForm.addEventListener('keydown', (event) => {
      if (event.target.tagName === 'LI' && event.code === 'Space') {
        event.preventDefault();
        event.target.querySelector('input').click();
      }
    });

    const showMore = this.querySelector('.button-show-more');
    if (showMore) {
      showMore.addEventListener('click', (event) => {
        showMore.closest('.facets__display').querySelector('.facets-wrap').classList.toggle('facets-wrap-show-more', !showMore.closest('.facets__display').querySelector('.facets-wrap').classList.contains('facets-wrap-show-more'));
        showMore.closest('.facets__display').querySelectorAll('.facets__item.facets__item-show-more').forEach(item => item.classList.toggle('hidden', !item.classList.contains('hidden')))
      })
      showMore.addEventListener('keydown', (event) => {
        if (event.target.className === 'button-show-more link underlined-link' && event.code === 'Space') {
          event.preventDefault();
          showMore.closest('.facets__display').querySelector('.facets-wrap').classList.toggle('facets-wrap-show-more', !showMore.closest('.facets__display').querySelector('.facets-wrap').classList.contains('facets-wrap-show-more'));
```

Just some minor, extra whitespace that can be removed 👍🏼

assets/facets.js
```diff
@@ -214,12 +256,17 @@ customElements.define('price-range', PriceRange);
 class FacetRemove extends HTMLElement {
   constructor() {
     super();
-    this.querySelector('a').addEventListener('click', (event) => {
-      event.preventDefault();
-      const form = this.closest('facet-filters-form') || document.querySelector('facet-filters-form');
-      form.onActiveFilterClick(event);
+    this.querySelector('a').addEventListener('click', this.closeFilter);
+    this.querySelector('a').addEventListener('keydown', (event) => {
```
andrewetchen, line 260:
```suggestion
    this.querySelector('a').addEventListener('keydown', (event) => {
```

snippets/facets.liquid
```diff
@@ -22,78 +22,131 @@
...
         {% if enable_filtering %}
           <div id="FacetsWrapperDesktop" class="facets__wrapper">
-            {%- unless results.filters == empty -%}
-              <h2 class="facets__heading caption-large text-body">{{ 'products.facets.filter_by_label' | t }}</h2>
-            {%- endunless -%}
+            {%- if filter_type != 'vertical' -%}
+              {%- unless results.filters == empty -%}
+                <h2 class="facets__heading caption-large text-body" id="verticalTitle" tabindex="-1">{{ 'products.facets.filter_by_label' | t }}</h2>
+              {%- endunless -%}
+            {%- endif -%}
```
andrewetchen, line 40:
```suggestion
            {%- unless filter_type == 'vertical' and results.filters == empty -%}
              <h2 class="facets__heading caption-large text-body" id="verticalTitle" tabindex="-1">{{ 'products.facets.filter_by_label' | t }}</h2>
            {%- endunless -%}
```
I think we could make this into one line instead of two. It's up to you though on whether this impacts readability.

sofiamatulis: Good find! Grouped them but using `if`

duygukalaycioglu (commented):
Hey Sofia, the spacing are almost there!

Let's decrease the spacing below and above the price range. 
Screen Shot 2022-03-23 at 12 15 40 PM

Let's make sure the spacings are consistent in the filter groups. 
Screen Shot 2022-03-23 at 12 17 44 PM

sections/main-search.liquid
```diff
@@ -132,83 +132,131 @@
...
-            <h2 class="title title--primary">
-              {{ 'sections.collection_template.empty' | t }}<br>
-              {{ 'sections.collection_template.use_fewer_filters_html' | t: link: search_url, class: "underlined-link link" }}
-            </h2>
-          </div>
-        </div>
-      {%- else -%}
-        {% paginate search.results by 24 %}
-          <div class="template-search__results collection page-width" id="product-grid" data-id="{{ section.id }}">
+    <div class="{% if section.settings.filter_type == 'vertical' %} facets-vertical page-width{% endif %}">
```
andrewetchen, line 181:
```suggestion
    <div class="{% if section.settings.filter_type == 'vertical' %}facets-vertical page-width{% endif %}">
```

assets/component-facets.css
andrewetchen: Hey @duygukalaycioglu - I don't see this as a blocker myself, how about you? I can see this as something we can quickly fix in a follow up issue.

melissaperreault (commented):
Thanks Sofia! Great work so far! Here are my observations! Very open to iterate on some on a new PR where we see fit! 👍 

### General experience 
#### _(👆Update Mar 25: [Opened an issue](https://github.com/Shopify/dawn/issues/1537#issue-1181201675) to tackle another time)_
**Truncating enhancement** ([Video reference](https://screenshot.click/22-03-219v0-kgm3b.mp4))
- [x] If 10 filters or less, don't display `Show more`, we currently display the link when there are no more to reveal.
- [ ] ~Further more on that one above. The [recommendation is to truncate after 10 filters](https://shopify.dev/themes/navigation-search/filtering/storefront-filtering/storefront-filtering-ux#manage-filter-values). Although, clicking a `Show more` link just to display 1 or 2 more filters is not a great experience for buyers. I'd recommend to have a rule. At least doing it to not have to click to reveal 1 filter, I would be good with that. Happy to get your thoughts because it can be arbitrary. cc. @danielvan @nicklepine~
   - ~If filters = 12 or less don't display `Show more`~
   - ~If filters = 13 or greater display `Show more`~

**Collapsing groups**
- [ ] ~The recommendation is to collapse [after 5 from our internal docs](https://shopify.dev/themes/navigation-search/filtering/storefront-filtering/storefront-filtering-ux#manage-filter-groups).~

### Content

**Clear, Clear all, Remove all, Reset**
We've made the change from `Clear all` to `Remove all` but I think we need to look at the whole experience flow to make sure the language is efficient and consistent across. Also need to validate which word translates well to other languages to convey the same meaning.

I am not an expert of filtering experience yet but I hope we can get some help from @katycobb to support the ideal content. Also @a6haji might have additional insights from his expertise. I see that we display `Clear, Clear all` over in the [UX considerations docs](https://shopify.dev/themes/navigation-search/filtering/storefront-filtering/storefront-filtering-ux), but that might be just an example.

- [ ] Depending on which filter layout merchants choose, we use different words for the same action. ([Video reference](https://screenshot.click/22-03-w9eh5-91ewh.mp4))


### UI details
**Horizontal layout**
- [ ] [We can improve the breakpoint for small tablet or large mobile](https://screenshot.click/22-03-ezd6h-ubezg.png). Testing on landscape orientation helps discovering this opportunity.

**Vertical layout**
- [x] We can improve spacing above the first filter accordion when filter pills are visible. I've tested a 1rem value for that scenario and it balances the empty space all around. ([Video reference](https://screenshot.click/22-03-64q9v-676lz.mp4))
- [x] I have a preference to increase slightly the Vertical Filter title to help differentiate from the filter list below it. [Aiming to help the hierarchy](https://screenshot.click/22-03-razql-wb9a9.png).
- [x] While bumping the font size, the `Remove all` link [looks misaligned vertically with the title](https://screenshot.click/22-03-xe7ah-vaap0.png). Not sure how to recommend to improve it code-wise
- [x] Eagle eye adjustment to raise the bottom-padding to 1.6rem instead of 1.5 so when filters are open the visual space feels equally applied above and below the divider line. [Video reference](https://screenshot.click/22-03-30xvy-z685q.mp4)
- [x] I couldn't test Price, it is not displaying even though it is part of the filters in the admin. I'll investigate why 🤔 

**Drawer**
- [x] `mobile-facets__footer` is [overlapping filters](https://screenshot.click/22-03-puppq-8yyk2.png)
- [x] With the change to pull out Sorting on desktop/tablet, [Alignment with sorting and filters are lost](https://screenshot.click/22-03-h3h80-sm3y5.png). 
- [x] Smaller screen size needs more attention, it it [not resolved for those breakpoint](https://screenshot.click/22-03-8muxe-odrkm.png). ([Video reference](https://screenshot.click/22-03-hswo1-ecxf5.mp4))

**Boolean filters**
I'll investigate more to learn the details but I am not convinced the UI outcome of the experience is the best right now. The ideal scenario would be to have the label displayed only, not the value. If something is `Vegan Friendly`, just checking that box or a selecting a different UI would be a better buyer experience. It's like we are forcing am information architecture on an item that doesn't need that deeper layer of information.

i.e.
Vegan Friendly: Yes
Vegan Friendly: No

I don't think we should display the `No` value. I'll read through the documentation and report back!

duygukalaycioglu:
> Boolean filters
> I'll investigate more to learn the details but I am not convinced the UI outcome of the experience is the best right now. 

Hey @melissaperreault thank you so much for sharing your feedback! 
**Quick note on Boolean filters:** This won't be a part of vertical filters projects, updates to Boolean filters parked for now to be visited later. 

Approved by: @danielvan @nicklepine ✅

sofiamatulis:
> [ ] [We can improve the breakpoint for small tablet or large mobile](https://screenshot.click/22-03-ezd6h-ubezg.png). Testing on landscape orientation helps discovering this opportunity.

Since this is the existing experience for horizontal I think we can keep it for now and re-evaluate 🤔 the mobile experience only comes in below 750 at the moment

KaichenWang (commented):
Left a few code comments. 

Also would like to make some general suggestions that we can discuss and potentially turn into follow up issues:

1. The `facets.liquid` would benefit from code comments. The file has become quite large and it's difficult to keep track of what each section of the code is for; adding some comments would help. This was already an issue prior to this PR, but has been exacerbated. 

2. Opportunity to modularize: sort, filters and active filters could be turned into separate components. Removing some of the shared dependencies would make it easier to build different layouts and allow each component to work independently when the others aren't enabled or needed. Again, this was already an issue prior to this PR, but has been exacerbated. 

3. Re-evaluate proportion of classes/styles that's shared between each style of filter. For example, the horizontal/vertical filters should have a more independent set of classes in their markup to avoid nesting of selectors and overrides in CSS.

4. Isolate the show more/less functionality into its own component/pattern that is dissociated from facets. This way the same pattern can be re-used and is not mixed in with other facet functionality.

assets/facets.js
```diff
@@ -214,12 +265,17 @@ customElements.define('price-range', PriceRange);
...
+      if (event.code == 'Space') this.closeFilter(event);
```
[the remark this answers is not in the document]
sofiamatulis: Yes I updated it :)

assets/facets.js
```diff
@@ -7,12 +7,40 @@ class FacetFiltersForm extends HTMLElement {
...
+    const facetForm = this.querySelector('form');
+    facetForm.addEventListener('input', this.debouncedOnSubmit.bind(this));
+    facetForm.addEventListener('keydown', (event) => {
+      if (event.target.tagName === 'LI' && (event.code.topUpperCase() === 'SPACE' || event.code.toUpperCase()  === 'ENTER' )) {
+        event.preventDefault();
+        event.target.querySelector('input').click();
+      }
+    });
+
+    facetForm.addEventListener('click',  (event) => {
```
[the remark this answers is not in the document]
sofiamatulis: I added this because before when you selected a filter the `Show more` cease to function. This was due to the DOM being rewritten and the events being removed.

assets/facets.js
```diff
@@ -7,12 +7,40 @@ class FacetFiltersForm extends HTMLElement {
...
+      if (event.target.nodeName === 'SPAN' && event.target.closest('[id^="Show-More-"]')) {
+        this.expandShowMore(event.target.closest('[id^="Show-More-"]'));
+      }
+    });
+
+    facetForm.addEventListener('keydown',  (event) => {
+      if (event.target.className === 'button-show-more link underlined-link' && (event.code.toUpperCase() === 'SPACE' || event.code.toUpperCase() === 'ENTER' ) && event.target.closest('[id^="Show-More-"]')) {
```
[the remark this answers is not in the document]
sofiamatulis: Since this is the only button I updated it to be the node type. What do you think? https://github.com/Shopify/dawn/pull/1443/commits/2f3b9363db2f48dffa2093bdb857acd34c37c1e2

assets/facets.js
```diff
@@ -7,12 +7,40 @@ class FacetFiltersForm extends HTMLElement {
...
+        }
+      }
+    });
 
     const facetWrapper = this.querySelector('#FacetsWrapperDesktop');
     if (facetWrapper) facetWrapper.addEventListener('keyup', onKeyUpEscape);
   }
 
+  expandShowMore(showMore) {
+    showMore.closest('.facets__display').querySelector('.facets-wrap').classList.toggle('facets-wrap-show-more', !showMore.closest('.facets__display').querySelector('.facets-wrap').classList.contains('facets-wrap-show-more'));
```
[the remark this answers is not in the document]
bacounts: > * Going even further, I think the show more/less functionality should be detached from facets since it can be a useful standalone component/pattern

Agreed. A good proof of concept would be something like issue https://github.com/Shopify/dawn/issues/1404 — "Add a 'Read more' button option to long descriptions..."

assets/facets.js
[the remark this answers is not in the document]
sofiamatulis: I also grouped them since this was an accessibility concern on windows: https://github.com/Shopify/dawn/pull/1443/commits/a9207a288723cc5ed5d06b6afed8515d7738601c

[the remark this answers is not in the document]
KaichenWang: Ah yes you're right, since the DOM nodes inside the form are being replaced, the event listeners (if applied to the button) would need to be re-attached. My suggestion would be to leverage custom HTML elements. The `connectedCallback()` lifecycle method allows event listeners to automatically re-attach when an instance of the custom element is added to the DOM.

I've done some testing and an implementation like this should work for this scenario: https://codepen.io/KaichenWang/pen/eYyvmqv

assets/facets.js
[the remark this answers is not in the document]
sofiamatulis: That is fair 🤔 However it's a good point that class names can change. I think checking for the node type (button) makes more sense. Thoughts?

assets/facets.js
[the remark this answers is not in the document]
sofiamatulis: Just saw this but I moved it out in its own component. 👀 :

https://github.com/Shopify/dawn/pull/1443/commits/3406756dbb923e2335cc15aa7a422c8ce519441e 👀 

I can include `connectedCallback`. Thoughts?

assets/show-more.js
```diff
@@ -0,0 +1,26 @@
+class ShowMoreButton extends HTMLElement {
+  constructor() {
+    super();
+    this.addEventListener('click', (event) => {
+      if (event.target.nodeName === 'SPAN' && event.target.closest('[id^="Show-More-"]')) {
```
KaichenWang, line 5: Do we need separate event handling for `span` and `button`? I think it would be better to just attach 1 event listener to the button element.

assets/facets.js
```diff
@@ -154,11 +189,27 @@ class FacetFiltersForm extends HTMLElement {
...
     event.preventDefault();
-    const formData = new FormData(event.target.closest('form'));
-    const searchParams = new URLSearchParams(formData).toString();
-    FacetFiltersForm.renderPage(searchParams, event);
+    const sortFilterForms = document.querySelectorAll('facet-filters-form form');
+    if (event.srcElement.className == 'mobile-facets__checkbox') {
+      const searchParams = this.createSearchParams(event.target.closest('form'))
+      this.onSubmitForm(searchParams, event)
+    } else {
+      const forms = [];
```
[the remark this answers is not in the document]
KaichenWang: Yeah i agree, this wouldn't be possible as is. Would require re-thinking our overall approach

assets/facets.js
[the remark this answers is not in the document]
KaichenWang: I think your implementation with custom HTML element is good. In this case, attaching the event listeners in the `constructor()` also works.

assets/show-more.js
sofiamatulis: Yes that is because when you use space/enter the node type is `button` and when you click the type is `span

 `console.log(event.target.nodeName, 'nodename')`

https://screenshot.click/25-01-ae1u9-y4ui3.png

assets/facets.js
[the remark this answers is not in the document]
KaichenWang: Yeah in this case it's fine since the component is small and only has 1 button. But yeah whether we should ever reference classes in JS would be a good point to bring up in team discussion.

assets/facets.js
```diff
@@ -214,12 +244,18 @@ customElements.define('price-range', PriceRange);
 class FacetRemove extends HTMLElement {
   constructor() {
     super();
-    this.querySelector('a').addEventListener('click', (event) => {
-      event.preventDefault();
-      const form = this.closest('facet-filters-form') || document.querySelector('facet-filters-form');
-      form.onActiveFilterClick(event);
+    const facetLink = this.querySelector('a');
+    facetLink.addEventListener('click', this.closeFilter);
```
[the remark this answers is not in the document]
sofiamatulis: Reverting it, but I tested it and it seemed to work without it 🤔

 I think you'd only need when you need access to `this`, which I guess you do for the `event`

assets/facets.js
```diff
@@ -214,12 +244,18 @@ customElements.define('price-range', PriceRange);
 class FacetRemove extends HTMLElement {
   constructor() {
     super();
-    this.querySelector('a').addEventListener('click', (event) => {
-      event.preventDefault();
-      const form = this.closest('facet-filters-form') || document.querySelector('facet-filters-form');
-      form.onActiveFilterClick(event);
+    const facetLink = this.querySelector('a');
+    facetLink.addEventListener('click', this.closeFilter.bind(this));
+    facetLink.addEventListener('keydown', (event) => {
```
svinkle, line 249: WCAG [2.5.2 Pointer Cancellation](https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html) requires use of "up" events (`click()`, `keyup()`, `mouseup()`, `touchend()`) in order to allow users to "cancel" the event by swiping/mousing/<kbd>Esc</kbd> away from of the control.

```suggestion
    facetLink.addEventListener('keyup', (event) => {
```

assets/facets.js
```diff
@@ -7,7 +7,14 @@ class FacetFiltersForm extends HTMLElement {
       this.onSubmitHandler(event);
     }, 500);
 
-    this.querySelector('form').addEventListener('input', this.debouncedOnSubmit.bind(this));
+    const facetForm = this.querySelector('form');
+    facetForm.addEventListener('input', this.debouncedOnSubmit.bind(this));
+    facetForm.addEventListener('keydown', (event) => {
```
svinkle, line 12: Do we still need this?

andrewetchen:
For either filter layout (Horizontal, Vertical, or Drawer), the Price filter on mobile (`749px`) isn't working: [video](https://screenshot.click/25-35-b5w0t-t69a2-99s.mp4). It works on desktop.

assets/show-more.js
KaichenWang: Ok in that case, we may want to more explicitly separate the different types of events (keyboard vs. click). Relying on `target.nodeName` to distinguish event types doesn't seem like the correct way of achieving this.
```
const button = this.querySelector('button');
button.addEventListener('click', event => {
  this.expandShowMore(button);
});
this.addEventListener('keydown', event => {
  if (event.code.toUpperCase() == 'SPACE' || event.code.toUpperCase() == 'ENTER') {
    this.expandShowMore(button);
    const nextElementToFocus = event.target.parentNode.parentNode.querySelector(".facets__item-show-more")
    if (!nextElementToFocus.classList.contains('.hidden')) {
      nextElementToFocus.querySelector('input').focus();
    }
  }
});
```

assets/show-more.js
```diff
@@ -0,0 +1,26 @@
...
+        this.expandShowMore(event.target.closest('[id^="Show-More-"]'));
+      }
+      if (event.target.nodeName === 'BUTTON' && event.target.closest('[id^="Show-More-"]')) {
+        this.expandShowMore(event.target.closest('[id^="Show-More-"]'));
+        const nextElementToFocus = event.target.parentNode.parentNode.querySelector(".facets__item-show-more")
```
KaichenWang, line 10: If we've decided to split out show-more as a standalone component, I don't think its JS should refer to facets. This needs some work to be more generalized.

assets/show-more.js
```diff
@@ -0,0 +1,26 @@
...
+        if (!nextElementToFocus.classList.contains('.hidden')) {
+          nextElementToFocus.querySelector('input').focus()
+        }
+      }
+    });
+  }
+  expandShowMore(showMore) {
+    const facetsDiplay = showMore.closest('.facets__display-vertical');
```
KaichenWang, line 18: Again, if we've decided to split out show-more as a standalone component, I don't think its JS should refer to facets. This needs some work to be more generalized.

assets/show-more.js
```diff
@@ -0,0 +1,26 @@
...
+      if (event.target.nodeName === 'BUTTON' && event.target.closest('[id^="Show-More-"]')) {
+        this.expandShowMore(event.target.closest('[id^="Show-More-"]'));
+        const nextElementToFocus = event.target.parentNode.parentNode.querySelector(".facets__item-show-more")
+        if (!nextElementToFocus.classList.contains('.hidden')) {
+          nextElementToFocus.querySelector('input').focus()
+        }
+      }
+    });
+  }
+  expandShowMore(showMore) {
```
KaichenWang, line 17: Does showMore need to be passed as param? It appears in our usage, we always pass the same element to this method

assets/facets.js
sofiamatulis: With it: https://screenshot.click/25-55-fenh4-yw0jf.mp4

Without it: https://screenshot.click/25-54-a52ez-frkgr.mp4

That is because it doesn't listen to the `input ` event listener

ludoboludo (commented):
A few comments. Sorry if any is repeating the conversations you might have had before, feel free to disregard them. 

First though, functionality wise I didn't run into any bugs. 

- [ ] on mobile we're using a font of 10px. This seem quite small and we're applying quite a bit of whitespace around it. I take it to make sure it's easily clickable but wouldn't it look better if the pill was just a bit bigger with a bigger font rather than too much white space? [screenshot of pill padding](https://screenshot.click/25-05-zbg53-u7m6k.png) and [screenshot of white space with many pills](https://screenshot.click/25-09-jpyx8-ynzq8.png)
- [x] arrow icon isn't animated/changing when opening/closing the dropdowns. 
- [x] when opening long or many filters, the page's height keeps increasing and you don't get to see the visual updates of the product list being updated. Would be nice to keep the vertical filters the same height has the product grid and enable scrolling within it: [video comparison with Nike](https://screenshot.click/25-29-njgjc-bcgy3.mp4)
- [ ] the filter name seem to be announced a few times: [video](https://screenshot.click/25-54-e8tsz-x6f4i.mp4)

assets/show-more.js
sofiamatulis: Good catch! Before the event listener was on the form so I wanted to ensure we were on the correct element. I think now we can just use the click. We dont need to listen for keydown 🤔 Seems to work on tab with click

https://github.com/Shopify/dawn/pull/1443/commits/6e54b9142de6575847bc95e0d9239299387b9151

sofiamatulis:
> [ ] the filter name seem to be announced a few times: [video](https://screenshot.click/25-54-e8tsz-x6f4i.mp4)

I believe that is because of the `legend` @svinkle asked to be added 👀 
https://screenshot.click/25-39-16zok-7wz82.png

assets/facets.js
```diff
@@ -214,12 +252,19 @@ customElements.define('price-range', PriceRange);
...
     super();
-    this.querySelector('a').addEventListener('click', (event) => {
-      event.preventDefault();
-      const form = this.closest('facet-filters-form') || document.querySelector('facet-filters-form');
-      form.onActiveFilterClick(event);
+    const facetLink = this.querySelector('a');
+    facetLink.setAttribute('role', 'button');
+    facetLink.addEventListener('click', this.closeFilter.bind(this));
+    facetLink.addEventListener('keyup', (event) => {
+      if (event.code.toUpperCase() == 'SPACE') this.closeFilter(event);
```
svinkle, line 259: Need to add `event.preventDefault()` here since pressing <kbd>Space</kbd> causes the page to scroll.

assets/show-more.js
```diff
@@ -0,0 +1,22 @@
+class ShowMoreButton extends HTMLElement {
+  constructor() {
+    super();
+    const button = this.querySelector('button');
+    button.addEventListener('click', (event) => {
+      this.expandShowMore(event);
+      const nextElementToFocus = event.target.parentNode.parentNode.querySelector(".show-more-item")
+      if (!nextElementToFocus.classList.contains('.hidden')) {
```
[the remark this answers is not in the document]
andrewetchen: Also, I think you can remove the `.` from the `hidden` class.

sofiamatulis:
> [ ] We'll need to consider having a rule for the position of the `Price` input [with the global styles capabilities](https://screenshot.click/22-03-8gx6z-zq2wl.png). When inputs are side by side, it doesn't give enough room for large numbers efficiently. Worth revisiting stacking them in some scenarios.

Good catch Meli! created an issue here: https://github.com/Shopify/dawn/issues/1543

It can be a follow up or included here depending on what the desired outcome is

cc @nicklepine @duygukalaycioglu

sofiamatulis:
> [ ] The arrow is [not place in the same horizontal position compared](https://screenshot.click/22-03-eulhp-9o02m.png) to `Collapsible row` blocks. It makes them feel like they are falling off the right edge. On Filters the arrow is aligned at the [edge to the right](https://screenshot.click/22-03-6k05g-6m1ze.png), but on `Collapsible row` block, [there is some inner right padding](https://screenshot.click/22-03-lm91c-xhy18.png).

This was done based on the existing horizontal filter and not the collapsible component that is why it looks different. 👀 Thoughts on adding some padding? cc @duygukalaycioglu

andrewetchen:
> Also seeing a minor visual issue with the overlay that appears while filters are being applied. With vertical filters its somewhat likely a user will be interacting with these while scrolled the down page a bit. When I do this I can see the edge of the overlay and some items may not be covered by it https://screenshot.click/24-03-pq2xl-3b3i8.mp4

@sofiamatulis I noticed this as well and I think we can quickly fix this (without much effort) on the collection + search pages and accounting for all desktop filter layouts by using `position: fixed;` instead of `position: absolute;`.

My first thought was to change this here:

https://github.com/Shopify/dawn/blob/205232484be8a90a0fb690158f6b172b261db172/assets/component-loading-overlay.css#L2

... but doing this would affect the `.loading-overlay` in [main-cart-items](https://github.com/Shopify/dawn/blob/205232484be8a90a0fb690158f6b172b261db172/sections/main-cart-items.liquid#L149-L155) for the loading spinner:
- Loading spinner with `position: absolute;` [screenshot](https://user-images.githubusercontent.com/28404165/160446349-83c9c774-51da-47c9-ab7b-982525120231.png)
- Loading spinner with `position: fixed;` [screenshot](https://user-images.githubusercontent.com/28404165/160446389-0ec902e9-31b6-4c84-845d-6a62e0d7b0db.png)

This can be fixed by adding `position: fixed;` to `.collection.loading .loading-overlay` here:

https://github.com/Shopify/dawn/blob/205232484be8a90a0fb690158f6b172b261db172/assets/template-collection.css#L32-L34

```css
.collection.loading .loading-overlay {
  display: block;
  position: fixed;
}
```

Collection + search pages and accounting for all desktop filter layouts:

- [Before](https://user-images.githubusercontent.com/28404165/160446717-f93265ca-a4ed-4f83-b7b6-95c83c319fd1.png)
- [After](https://user-images.githubusercontent.com/28404165/160446832-0009ce1d-452b-4dd2-8d71-243dd63fe46f.png)

assets/facets.js
sofiamatulis: It originally was checking the input but not updating the products 🤔 retested and it works now probably. Maybe it was caching in the video above. Removed 🤔

sections/main-collection-product-grid.liquid
```diff
@@ -27,64 +27,112 @@
...
-            {%- endfor -%}
-          </ul>
-  
-          {%- if paginate.pages > 1 -%}
-            {% render 'pagination', paginate: paginate, anchor: '' %}
-          {%- endif -%}
-        </div>
-      {%- endif -%}
-    {%- endpaginate -%}
+  <div class="{% if section.settings.filter_type == 'vertical' %} facets-vertical page-width{% endif %}">
```
andrewetchen, line 76:
```suggestion
  <div{% if section.settings.filter_type == 'vertical' %} class="facets-vertical page-width"{% endif %}>
```

assets/show-more.js
```diff
@@ -0,0 +1,22 @@
+class ShowMoreButton extends HTMLElement {
+  constructor() {
+    super();
+    const button = this.querySelector('button');
+    button.addEventListener('click', (event) => {
+      this.expandShowMore(event);
+      const nextElementToFocus = event.target.parentNode.parentNode.parentNode.querySelector(".show-more-item")
```
KaichenWang, line 7: It would be best to not traverse the DOM via chained `parentNode`. This adds fragility and may lead to unexpected results if even a small change in the markup is made.

assets/component-show-more.css
```diff
@@ -0,0 +1,16 @@
...
+  justify-content: flex-start;
+  padding-bottom: 1.1rem;
+}
+
+.button-show-more, 
+.button-show-less {
+  margin-top: 1.5rem;
+}
+
+.show-more-wrap ~ show-more-button .label-show-less,
```
KaichenWang, line 13: This CSS seems unnecessarily complex to me, is there a way for the `show-more-button` element to have a class which indicates what state it is in instead?

assets/show-more.js
sofiamatulis: Good point https://github.com/Shopify/dawn/pull/1443/commits/e7fbe88461b3fb873381185f1e38c14f367b6dc6

sofiamatulis:
> If we don't want it to "overlay" on top of other sections, omitting the suggested `position: fixed;` and adding `position: relative;` to the parent containers would work better since we're already using `position: absolute;` for `.loading-overlay`:

I will let Meli give her thoughts. We can also do this as a follow up 👍

assets/component-show-more.css
sofiamatulis: How's this? https://github.com/Shopify/dawn/pull/1443/commits/f1d85ca24677df80bee6e7e0bc5fcffd68516b8d

sofiamatulis:
> Hi @sofiamatulis/ @nicklepine: I've done a quick look into the inconsistent Clear/Reset/Remove labels. Not sure if we can get this change in this time around, but tried to make a fast decision in case we can :)
> 
> From my understanding, all these buttons do the same thing: clear any applied filters. Is that correct?
> 
> If so, @svinkle and I aligned. We're happy to go ahead with aligning all instances of the button to `Clear all`, and `[category] [filter name] clear filter` for the individual filter buttons.
> 
> I did a quick competitive audit and "Clear all" is the resounding language used by ecomm retailers so will be familiar to our buyers.
> 
> Edit: happy to chat through this further if needed!

@katycobb @svinkle @Nicklepine since translations are already back I'd recommend to do this in a fast follow so it's not a blocker to merge it 🤔

melissaperreault:
> Also seeing a minor visual issue with the overlay that appears while filters are being applied. With vertical filters its somewhat likely a user will be interacting with these while scrolled the down page a bit. When I do this I can see the edge of the overlay and some items may not be covered by it https://screenshot.click/24-03-pq2xl-3b3i8.mp4

@sofiamatulis This is considered not cute, but can be tackled in another round!

katycobb:
> Re: updating language to close all. Aligned on tackling this separately, in a non-launch-critical follow up pr.

Sounds good. I'll add to the [enhancements ticket ](https://github.com/Shopify/dawn/issues/1537). Since it's not a rush to get in today, I'll also connect with translation folks on it :)

andrewetchen:
@sofiamatulis - Thank you for all the work getting vertical filters added! Amazing work! 👏🏼 
The addition of vertical filters offers greater flexibility on the collection and search pages and will greatly enhance the buyer experience while sifting through products - something to be really excited about and proud of 🎉 

I did one last UX/UI audit to test the many different combinations of settings:

- [x] Filters + sorting enabled and disabled.
- [x] Setting a filter on mobile and ensuring it's retained on desktop. And vice versa.
- [x] Test the show more/less on filters.
- [x] Test the search page.
- [x] Test with different global settings (dropdown and pop-ups).
- [x] No-JS functionality.
- [x] Desktop and mobile layouts.
- [x] Other browsers: Firefox, Safari.
- [x] Re-tested previously raised bugs: sorting not updating, price filtering on mobile, minor layout issues, etc.

There are 2 blockers that I think need to be addressed before this is merged. Apologies if any of these were already surfaced by someone else - it's hard to sift through the many comments to confirm 😅

### 1 of 2

**Why is this a blocker?**
It's quite visibly noticeable in comparison to the search page or when switching between the horizontal filter layout. I also believe you had fixed this previously in this PR. The vertical filter layout on the search page is good.

**Desktop filter layout** > **Vertical**:  The collection container `page-width` on the collection page. Looks like this one resurfaced: https://github.com/Shopify/dawn/pull/1443#discussion_r834519834

- [Collection container on collection page](https://user-images.githubusercontent.com/28404165/160497980-dba636bb-7af1-486a-a591-370e06a7eb9d.png)
- [Collection container on search page](https://user-images.githubusercontent.com/28404165/160497945-04b0f395-a290-4dfc-b63c-3308d45128ef.png)

### 2 of 2

**Why are these blockers?**
These were working previously and the search page issue leaves a user with a non-interactive page.

**a)** Sorting isn't working after filters are set. And when you sort first, setting a filter afterwards resets the sorting option.

https://user-images.githubusercontent.com/28404165/160500056-a66c2ade-4d95-4236-830b-d1d8d78dc78c.mp4

**b)** On the search page, the loading overlay remains and the page is non-interactive:

https://user-images.githubusercontent.com/28404165/160500758-25dd92d1-e996-4464-9cdc-3bba45df22aa.mp4

### Note

For all other bugs/issues that I brought up, they've been added or will be added here: https://github.com/Shopify/dawn/issues/1537 so that we can tackle these follow up style.

melissaperreault:
Thanks Sofia! Update from UX!

I tried to do a clean up pass and port over the remaining UX topics to the enhancement issue. They could all become a single issue if we see fit anyway! Here is the recap:

- Update labels to "Clear" issue was created here for quick assistance follow up https://github.com/Shopify/dawn/issues/1547
- Arrow padding question ([Comment](https://github.com/Shopify/dawn/pull/1443#issuecomment-1080676391)) No strong opinion for now but worth investigating in a follow up. I added the point to https://github.com/Shopify/dawn/issues/1537#issuecomment-1081367083
- Overlay question ([Comment](https://github.com/Shopify/dawn/pull/1443#issuecomment-1080914662)): _"This works but it means the whole page (including other sections) would have the overlay."_ 
   - I don't have a complete opinion yet. I have many questions such as the impact on performance depending of the solution we take. From a UX perspective, as buyers are filtering, I don't expect them to scroll further the product grid but there could be sections above and would like to explore a prototype with the recommended implementation before making a final UX call. 🙏 Adding this note to the [Enhancement issue](https://github.com/Shopify/dawn/issues/1537) as well!


### Other findings
**Other things I stress tested:**
- [x] No JS
- [x] Experience on Collection and Search, Test all of the filters on mobile and desktop and tablet
- [x] Adding sections above and below Product grid
- [x] Tab with keyboard
- [x] Safari

**Small desktop/tablet**
_Vertical layout_
- [ ] Filter pills are no longer aligned, neither the Remove all button. [Reference](https://screenshot.click/22-03-sjnmj-9i682.png)
   - [ ] The remove all button don't need to be scale down in this layout because its surrounding are keeping their size (Filter labels)
   - [ ] Gaps between filter pills can be improved
      - [ ] [Vertical filter, the pills get a lot of space ](https://screenshot.click/22-03-q84hf-0upg2.png)for touch target I wonder how smaller we could adjust. (Approx 44px gap vertically, 30px horizontally)
      - [ ] As opposed to Horizontal filter pills, they are spread out but not as much. (Approx 27px gap vertically, 35px horizontally)
- [ ] Not sure if this is related to Andrew's finding, but the tablet [layout has unwanted padding](https://screenshot.click/22-03-pkxea-ppybn.png)
- [ ] Vertical layout is really [not optimal on Tablet and smaller desktop](https://screenshot.click/22-03-nupxy-64ex5.png). The filter row alway remain the same size and the product cards become very small and not legible. ([reference from a real iPad](https://screenshot.click/22-03-cmx8u-19otb.png))
- [ ] Nothing to do about that but [this gets weird on smaller screens](https://screenshot.click/22-03-zshh0-wsliz.png) when everything is all around.

_Drawer layout_
- [ ] Product number [count is squeezed](https://screenshot.click/22-03-8kde8-i4dsc.png) because the sorting is taking more place with the recent change. Look at the Horizontal layout behaviour, the scale is smoother. ([Video reference](https://screenshot.click/22-03-8g13f-qd8k9.mp4))
- [ ] I can't scroll properly in the drawer without scrolling the page on iPad/tablet
- [ ] Filter pill labels [seems like wrapping sooner](https://screenshot.click/22-03-89uq8-a6596.png) on iPad portrait and [landscape](https://screenshot.click/22-03-zu4jw-fde80.png)
- [ ] Checkboxes [have odd double outlines/shadows](https://screenshot.click/22-03-dykdd-k98ux.png) on iPad

**Mobile**
- [ ] Number of products [is misaligned](https://screenshot.click/22-03-6az63-mzssi.png)

**No JS**
[The current live pattern](https://screenshot.click/22-03-9swre-ymbzt.png) for horizontal when JS is disabled is to have 1 button for `Filter` and 1 button for `Sort`. Something is missing on that implementation is the improvement of the tertiary button style where it should receive a small percentage of the foreground for its border.

When testing though, a few things are off to me:

_Drawer layout_
- [ ] Styling of button is [inconsistent and position of sort is off.](https://screenshot.click/22-03-k9x8t-t5cb4.png)

_Vertical layout_
- [ ] [This layout](https://screenshot.click/22-03-gqlv5-4a3v5.png) really needs love to look more professional. I'll provide a visual reference when I am ready Tuesday! 
- [ ] There is a "Filter" [button in between](https://screenshot.click/22-03-bswxs-0k7cb.png) that is missing it's styling.

sofiamatulis:
> **Why is this a blocker?** It's quite visibly noticeable in comparison to the search page or when switching between the horizontal filter layout. I also believe you had fixed this previously in this PR. The vertical filter layout on the search page is good.
> 
> **Desktop filter layout** > **Vertical**: The collection container `page-width` on the collection page. Looks like this one resurfaced:

@andrewetchen Good catch! 👍 It looks like this was overwritten when I rebased/fixed the conflicts. Should be fixed now

sofiamatulis:
> **a)** Sorting isn't working after filters are set. And when you sort first, setting a filter afterwards resets the sorting option.

I believe this is because of this 👀 @KaichenWang Would you be ok if I reverted the logic? https://github.com/Shopify/dawn/pull/1443/commits/9a63a62fdb38bf03ff11e332dc99be4b94957481

sofiamatulis:
> * [ ]  The remove all button don't need to be scale down in this layout because its surrounding are keeping their size (Filter labels)

@melissaperreault We change the size of the pills and the button for all filter types. Here are some videos:

https://screenshot.click/29-34-kx5jv-o7ure.mp4

https://screenshot.click/29-36-bkst9-odvyp.mp4

sofiamatulis:
> It's not the most crucial issue, just worth tracking. Maybe we can go back and clean this up after merge.

@kmeleta Could you create an issue for this 🙏 cc @nicklepine for visibility

sofiamatulis:
> * There is a "Filter" [button in between](https://screenshot.click/22-03-bswxs-0k7cb.png) that is missing it's styling.

The same happens on Horizontal and Drawer so it'd be good to get feedback for those as well 🙏

sofiamatulis:
Any bugs should be added here: https://github.com/Shopify/dawn/issues/1558 to avoid duplicates

KaichenWang (dismissed):

andrewetchen (approved):

sections/main-collection-product-grid.liquid
```diff
@@ -27,64 +27,112 @@
...
+    {{ 'component-facets.css' | asset_url | stylesheet_tag }}
+    <script src="{{ 'facets.js' | asset_url }}" defer="defer"></script>
+    {%- if section.settings.enable_filtering -%}
+      <aside aria-labelledby="verticalTitle" class="facets-wrapper{% if section.settings.filter_type != 'vertical' %} page-width{% endif %}" id="main-collection-filters" data-id="{{ section.id }}">
```
svinkle, line 80: Just noticed the `aria-labelledby` doesn't point to anything when set to Drawer, or mobile. I'd suggest simplifying to use `aria-label` with a string value.
```suggestion
      <aside aria-label="Filter" class="facets-wrapper{% if section.settings.filter_type != 'vertical' %} page-width{% endif %}" id="main-collection-filters" data-id="{{ section.id }}">
```

sofiamatulis: Does it make sense to create a hidden h2 for Drawer and mobile instead?

andrewetchen (approved):

KaichenWang (approved):

sections/main-collection-product-grid.liquid
sofiamatulis: Created an issue here: https://github.com/Shopify/dawn/issues/1561

PR #1456, opened 2022-03-07
Add support for boolean filters

carolineschnapp:
**PR Summary:** 

Add support for boolean filtering on collection and search results pages.

New feature.

**Checklist**


snippets/facets.liquid
```diff
@@ -41,7 +41,7 @@
             {%- for filter in results.filters -%}
               {%- assign total_active_values = total_active_values | plus: filter.active_values.size -%}
               {% case filter.type %}
-              {% when 'list' %}
+              {% when 'boolean' or 'list' %}
```
carolineschnapp, line 44: Looks like using “or” is the preferred syntax here: https://github.com/Shopify/liquid/blob/master/lib/liquid/locales/en.yml#L8

snippets/facets.liquid
```diff
@@ -459,7 +459,7 @@
         <facet-remove>
           <a href="{{ value.url_to_remove }}" class="active-facets__button active-facets__button--light">
             <span class="active-facets__button-inner button button--tertiary">
-              {{ value.label | escape }}
+              {{ filter.label }}: {{ value.label | escape }}
```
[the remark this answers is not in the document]
carolineschnapp: Thank you, @NathanPJF !

gregjotau:
https://famme.no/collections/treningstights a lot of our filters are either "on/off"

gregjotau:
> @gregjotau yes, those new filters will rely on the presence of metafields of type “boolean”

OK, hmmm, I guess it is more efficient than having text with 1 option.

snippets/facets.liquid
NathanPJF: Interesting. How'd you come across that? Memory recall? I didn't see the ["theme linter"](https://github.com/Shopify/dawn/runs/5451041955?check_suite_focus=true) catch it.

snippets/facets.liquid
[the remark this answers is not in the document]
tyleralsbury: Just a heads-up, @duygukalaycioglu will be looking into this.

snippets/facets.liquid
```diff
@@ -459,7 +459,7 @@
...
               {% render 'icon-close-small' %}
               <span class="visually-hidden">{{ 'products.facets.clear_filter' | t }}</span>
             </span>
```
martinamarien, line 465: If we add the option for a label, should it also be applied to the "price" filter?

martinamarien (commented):
Code-wise, this looks good to me. @duygukalaycioglu and @carolineschnapp, let me know if you need another review after the labels have been discussed.

snippets/facets.liquid
carolineschnapp: We figure seeing `Price:` is not needed:

image

The same could be said for the Availability filter.

The Price, Availability, Product Type, and Brand filters are default filters you can only have one of each. Their value is self-explanatory.

For custom filters, it might not be as obvious.

Imagine product options or metafield filters for dimensions, for products that have multiple dimensions e.g. Inseam, waist size. The value alone, without the label, is confusing:
 
image

A blanket theme setting — always show, never show — is not ideal. Maybe we can be smart about this.

- Show label for all production options and metafield-based settings.
- No label for:
    - Product Type
    - Brand
    - Availability
    - Price

duygukalaycioglu:
Hi! I’ve conducted some benchmarking, and found a few examples in regards to the best practices for labeling pills.

1) In this example, the filter group name is not repeated in pills but the filter name itself is establishing this differentiation. I think this is the ideal option I would propose, but since the filter names are being entered by the merchant, I assume we can’t come up with a logic to achieve this.
option 1a
option 1

2) These 2 examples below are replicating the same solution presented above, keeping the filter group names in the pill. One disadvantage with this is repetitiveness, but the advantage is that it is always clear, removing the risk of misinformation and possibly the best option for accessibility. 

Screen Shot 2022-03-09 at 2 10 06 PM
Screen Shot 2022-03-09 at 2 53 32 PM


I recommend we go with the 2nd option and if possible, introduce show more and less capability to horizontal filters, so that if the number of filters end up being too many, we are giving the buyer the option to show more/less.

show more-less

bacounts:
"Kida-of-sorta related" to this specific PR: While testing and researching, I came across the following notation about comma-separated lists for a "filter based on a specific variant metafield".... [https://shopify.dev/themes/navigation-search/filtering/storefront-filtering#variant-specific-filters](url).

However, this "comma-separated list" does not render as described on the theme. If I understand the documentation right, I should be able to:

1. Create a product metafield, for example, "Product material."
2. Inside the product page in the admin, I add Gold, Yellow gold, Enamel."
3. On the collection page, each comma-separated value should now be a separate clickable filter, but they are not.

What am I missing, or is this feature not fully supported yet?

**I have included some reference images below using my system.**

Screen Shot 2022-03-09 at 10 45 12 PM
Screen Shot 2022-03-09 at 10 50 31 PM
Screen Shot 2022-03-09 at 10 50 17 PM
Screen Shot 2022-03-09 at 10 50 58 PM

carolineschnapp:
@bacounts the comma-separation is in the URL when filtering a collection or search results page, not in the value of a metafield you set in your _admin._ It's one way to construct the filtering URL, so it is shorter. If you use commas in the value of a `single_line_text_field` metafield in your admin, it will be interpreted as a single value, that happens to contain commas.

carolineschnapp:
@bacounts this isn't the right place to discuss this. Please create an issue.

martinamarien (approved):
Looks good to me. Thanks Caro.

PR #1337, opened 2022-02-09
Focus outline enhancements

kmeleta:
**Why are these changes introduced?**

Fixes #0. 1218 1202 1193

- Enables rounded focus outlines for cards, quantity, search, and email inputs
- Enables rounded focus outlines for all elements in Safari
- Cleans up our focus styles to remove a lot of redundant lines of css throughout the theme

**What approach did you take?**

1. Updated general `*:focus` styles use `box-shadow` instead of `outline` so Safari can also focus corner radius.
2. Added a reusable `.focus-pseudo` class for safer reuse of focus styles or for elements with special circumstances.
3. Defined `--focus-box-shadow` var to ensure consistency and reduce repetition of the complex box-shadow value we're using for focus
4. Removed or reduced redundant focus styles where possible (button, card, product media, signup/newsletter, search, quantity button, 
5. Removed `component-search.css` and relocated the couple remaining non-focus related styles to `base.css`

**Other considerations**

**Testing steps/scenarios**
- [ ] _List all the testing tasks that applies to your fix and help peers to review your work._

**Demo links**
_Please include a link to a demo store that includes preconfigured sections and settings to allow reviewers to easily test the features you are working on._

- [Store](url)
- [Editor](url)

**Checklist**
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

Issue #922, opened 2021-11-19
[Global settings] Adjust adjacent borders and corners


melissaperreault:
### Some considerations

- [ ]  Like other sections where there is 0 grip gap, logic needs to be added to ensure collapsing of adjacent borders + elimination of border-radius for adjacent corners. This it not yet implemented.
   - [ ] Product cards
   - [ ] Blog cards
   - [ ] Image with text
- [ ] This is also sometimes due to `Vertical spacing between sections` setting from `Layout` that could be set to 0
   - [ ] Image with text
   - [ ] ...TBD

KaichenWang:
I will explore behaviour of section child elements when `Vertical spacing between sections` is set to zero together with https://github.com/Shopify/dawn/issues/925

tyleralsbury:
> I will explore behaviour of section child elements when Vertical spacing between sections is set to zero together with #925

I see that this is from 9 days ago but I don't imagine that it will be easy to handle these cases.  Are we still considering it?  I would strongly advise against trying to create any logic like this.  If we did have section settings that corresponded to the entire section container, it would be different, but since it's elements within the section it's very sketchy.

kmeleta:
Opened a separate ticket to account for the section-specific collapsing work https://github.com/Shopify/dawn/issues/1524. This work is slightly different than the content collapsing within sections and I feel they can be handled and merged asynchronously.

Stephanie-Chou:
We are closing this issue. This issue was reported on an old version of Dawn (we are now on 10.0. If this is still a high priority issue today, please re open with updated reproduction steps

PR #1478, opened 2022-03-10
Implement corner collapsing

kmeleta:

This is a follow up to #1464 and should either be merged after the `grid-collapse` branch or merged into it upon approval.
Refs #922 

These changes build on the 0px grid spacing and border collapsing added in the other PR to also [collapse adjacent rounded corners](https://screenshot.click/10-26-zsqor-osyew.png) in grids. Without this change we end up with [this effect](https://screenshot.click/10-25-pjnpt-br3mc.png).

**Enabling independent radius values**
All applicable elements (cards, media, content containers) received additional selectors scoped to the `grid-collapsed` class that define each corner radius independently using css vars to represent each corner (e.g. `--top-left-radius`).

**:nth-child mayhem**
The selectors in `grid-collapsed.css` do not set any properties, only variables which serve as the single source of truth for independent corner radii within the `.grid .grid__item` context. A group of selectors exists for each possible column number configuration (2 cols, 3 cols, etc). These groups of selectors are unfortunately verbose and cryptic. However, all groups are completely identical to each other. The only difference being the integer used which correlates to the number of columns. This makes it easy to copy and paste to support additional column numbers in the future. If CSS allowed vars to be used in selectors, we could reduce this logic to perhaps a single ~50 line group of selectors per breakpoint.

**Exceptions to the single source of truth**
The only place were radius vars are overridden outside of `grid-collapsed.css` is `component-slider.css`. Slider grids disregard all `grid--#-col` rules to form a horizontal list so we also must override all radius values associated with a particular number of columns.

As with the border collapsing PR, the other notable exceptions are product media gallery, main blog, and collage. Each of these use their own irregular "grid" and are excluded from the global grid collapse rules for integrity purposes. Corner collapsing logic has been included in their respective stylesheets, but I could be convinced to move it all into `grid-collapsed.css` if that sounds better.

**Why all the variables?**
We _could_ explicitly set the properties (e.g. `border-top-left-radius: 0px`) directly from the selectors in `grid-collapsed.css`. However, we'd need to expand the selectors to target the specific elements within grid items that have the border radius, like cards, media, content containers, and any future types. Additionally, each of these elements already re-purpose radius values in children and pseudo elements for shadows, focus outlines, calculations etc so we'd need to further expand every single one of the ~100 selectors that are already needed to account for every single one of those cases (read: not feasible). **I feel that maintaining an agnostic context for grids that children can choose to access and utilize how they need is a more sane approach.**

It's also worth pointing out that if we hope to consolidate focus outline styles that also respect collapsed corners, we will most likely need to use this type of approach.

**Possible alternative approach**
<details>
<summary>Alternative approach details..</summary>
Rather than declaring all the necessary styles for each possible grid column configuration, it may be possible to dynamically generate only those needed and render them via inline style tags in the liquid. <a href="https://gist.github.com/kmeleta/c230ffef6a5c54176ba8306bf891010b">It might look like this</a>
<br/><br/>
This would be nice because if we increase the max number of columns above 6, no additional work would be required and no additional css would need to be added to account for the new configuration. Additionally, on a page with only a single grid, this would have the benefit of surely requiring less css than the current approach.
<br/><br/>
However on a more complicated page, you could end up with duplicated styles resulting in even more css that what the grid-overrides.css would. The biggest issue here though may be that these inline rendered style tags wouldn't be cached by the browser. This _might_ still benefit first load for a user (which is important) but overall I don't foresee it being a net gain. I think for now, with our max columns limit what it is and our grid classes what they are are currently (tablet-down and desktop-up) we're fine without this alternative approach. However with future changes to either of those 2 points, this approach could definitely be worth exploring and performance testing.
</details>

**Other considerations**

**Known issue: Product media orphans**
Because there'd be no way of knowing if the last item would be full width or not, we can't safely apply corner radius to the second to last item. I think it's a fairly minor issue and only applicable when there is a non-full-width orphan in the gallery.

**Known issue: Focus outlines**
This PR doesn't introduce any new logic to enable rounded focus outlines. A possible approach to this that can be easily expanded to account for collapsed corners can be seen in #1337 

**Known issue: Mixed card types in main blog, search, collection list section**
In situations where card style cards (text--only) can be rendered along with standard cards, we can't realistically predict which grid items will be touching other items enough to collapse corners. These cases will collapse borders, but we can't take action on the corners without risking the proper display of that grid in the "best case scenario" (all standard cards have images). We can open an investigation to look into this more, see how close to ideal we can get it, and consider the trade off in complexity in isolation.

**Testing steps/scenarios**

Affected areas..
- [ ] Main collection
- [ ] Search results
- [ ] Main collections list
- [ ] Collection list section
- [ ] Featured collection
- [ ] Multicolumn
- [ ] Featured blogs
- [ ] Main blog
- [ ] Collage
- [ ] Product media gallery

Permutations..
- [ ] zero horizontal spacing
- [ ] zero vertical spacing
- [ ] zero horizontal and vertical spacing
- [ ] with and without sliders enabled
- [ ] card/standard style cards
- [ ] various corner radius values
- [ ] various number of columns
- [ ] with drop shadows 

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127704596502/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2854,11 +2854,11 @@ details-disclosure > details {
 }
 
 .grid--collapsed-x.contains-content-container .content-container:after,
-.grid--collapsed-y.contains-media .global-media-settings:after {
+.grid--collapsed-x.contains-media .global-media-settings:after {
   margin-left: var(--collapsed-shadow-offset-x);
 }
 
-.grid--collapsed-x.contains-content-container .content-container:after,
+.grid--collapsed-y.contains-content-container .content-container:after,
 .grid--collapsed-y.contains-media .global-media-settings:after {
   margin-top: var(--collapsed-shadow-offset-y);
 }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1002,6 +1002,13 @@ summary::-webkit-details-marker {
   }
 }
 
+.grid--collapsed .grid__item {
+  --top-left-radius: var(--border-radius);
+  --top-right-radius: var(--border-radius);
+  --bottom-right-radius: var(--border-radius);
+  --bottom-left-radius: var(--border-radius);
+}
+
 .grid--gapless.grid {
   column-gap: 0;
   row-gap: 0;
@@ -2847,13 +2854,30 @@ details-disclosure > details {
 }
 
 .grid--collapsed-x.contains-content-container .content-container:after,
-.grid--collapsed-x.contains-media .global-media-settings:after {
+.grid--collapsed-y.contains-media .global-media-settings:after {
   margin-left: var(--collapsed-shadow-offset-x);
 }
 
-.grid--collapsed-y.contains-content-container .content-container:after,
+.grid--collapsed-x.contains-content-container .content-container:after,
 .grid--collapsed-y.contains-media .global-media-settings:after {
   margin-top: var(--collapsed-shadow-offset-y);
+}
+
+.grid--collapsed.contains-content-container .content-container,
+.grid--collapsed.contains-media .global-media-settings,
+.grid--collapsed.contains-content-container .content-container:after,
+.grid--collapsed.contains-media .global-media-settings:after {
+  border-radius: var(--top-left-radius) var(--top-right-radius) var(--bottom-right-radius) var(--bottom-left-radius);
+}
+
+.grid--collapsed.contains-media .global-media-settings img,
+.grid--collapsed.contains-media .global-media-settings iframe,
+.grid--collapsed.contains-media .global-media-settings model-viewer,
+.grid--collapsed.contains-media .global-media-settings video {
+  border-top-left-radius: calc(var(--top-left-radius) - var(--media-border-width));
+  border-top-right-radius: calc(var(--top-right-radius) - var(--media-border-width));
+  border-bottom-right-radius: calc(var(--bottom-right-radius) - var(--media-border-width));
+  border-bottom-left-radius: calc(var(--bottom-left-radius) - var(--media-border-width));
 }
 
 /* check for flexbox gap in older Safari versions */
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -270,50 +270,76 @@
 
 @media screen and (min-width: 750px) {
   .collage--collapsed .collage__item:only-child,
-  .collage--collapsed-x .collage__item--left:first-child {
+  .collage--collapsed .collage__item--left:first-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--right:last-child {
     --top-left-radius: var(--card-corner-radius);
     --bottom-left-radius: var(--card-corner-radius);
   }
 
   .collage--collapsed .collage__item:only-child,
-  .collage--collapsed-x .collage__item--right:last-child {
+  .collage--collapsed .collage__item--right:last-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--left:first-child {
     --top-right-radius: var(--card-corner-radius);
     --bottom-right-radius: var(--card-corner-radius);
   }
 
-  .collage--collapsed-x .collage__item--right:first-child {
+  .collage--collapsed-x .collage__item--right:first-child,
+  .collage--collapsed-x:not(.collage--collapsed-y) .collage__item--right:nth-last-child(2),
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--left:nth-child(2),
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--right:first-child,
+  .collage--collapsed-y.contains-card--standard .collage__item--right:first-child:where(.collage__item--product, .collage__item--collection) + .collage__item:nth-last-child(2),
+   .collage--collapsed-y.contains-card--standard:not(.collage--collapsed-x) .collage__item--left:nth-child(2):where(.collage__item--product, .collage__item--collection) + .collage__item {
     --top-left-radius: var(--card-corner-radius);
   }
 
-  .collage--collapsed-x .collage__item--left:nth-child(2) {
+  .collage--collapsed-x .collage__item--left:nth-child(2),
+  .collage--collapsed-x:not(.collage--collapsed-y) .collage__item--left:last-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--left:nth-child(2),
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--right:first-child,
+  .collage--collapsed-y.contains-card--standard .collage__item--left:nth-child(2):where(.collage__item--product, .collage__item--collection) + .collage__item,
+  .collage--collapsed-y.contains-card--standard:not(.collage--collapsed-x) .collage__item--right:first-child:where(.collage__item--product, .collage__item--collection) + .collage__item {
     --top-right-radius: var(--card-corner-radius);
   }
 
-  .collage--collapsed-x .collage__item--left:last-child {
+  .collage--collapsed-x .collage__item--left:last-child,
+  .collage--collapsed-x:not(.collage--collapsed-y) .collage__item--left:nth-child(2),
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--left:last-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--right:nth-last-child(2),
+  .collage--collapsed-y.contains-card--standard .collage__item--left:nth-child(2):where(.collage__item--product, .collage__item--collection),
+  .collage--collapsed-y.contains-card--standard:not(.collage--collapsed-x) .collage__item--right:first-child:where(.collage__item--product, .collage__item--collection) {
     --bottom-right-radius: var(--card-corner-radius);
   }
 
   .collage--collapsed-x .collage__item--right:nth-last-child(2):first-child,
-  .collage--collapsed-x .collage__item--right:nth-child(3n - 1):not(:last-child),
-  .collage--collapsed-x .collage__item--right:nth-child(3n - 2):not(:first-child)  {
+  .collage--collapsed-x .collage__item--right:nth-child(2):not(:last-child),
+  .collage--collapsed-x:not(.collage--collapsed-y) .collage__item--right:first-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--left:last-child,
+  .collage--collapsed-y:not(.collage--collapsed-x) .collage__item--right:nth-last-child(2),
+  .collage--collapsed-y.contains-card--standard .collage__item--right:first-child:where(.collage__item--product, .collage__item--collection),
+  .collage--collapsed-y.contains-card--standard:not(.collage--collapsed-x) .collage__item--left:nth-child(2):where(.collage__item--product, .collage__item--collection) {
     --bottom-left-radius: var(--card-corner-radius);
   }
+
 }
 
 @media screen and (max-width: 749px) {
-  .collage--collapsed-y .collage__item--left:nth-last-child(3n),
+  .collage--collapsed-y .collage__item--left:nth-last-child(3),
   .collage--collapsed-y:not(.collage--mobile) .collage__item:first-child,
+  .collage--collapsed-x:not(.collage--collapsed-y):not(.collage--mobile) .collage__item,
   .collage--collapsed-x:not(.collage--collapsed-y).collage--mobile .collage__item--left:nth-last-child(3),
   .collage--collapsed-x:not(.collage--collapsed-y).collage--mobile .collage__item--right:nth-child(3),
-  .collage--collapsed-y:not(.collage--collapsed-x).collage--mobile .collage__item--right:nth-child(-n + 2) {
+  .collage--collapsed-y:not(.collage--collapsed-x).collage--mobile .collage__item--right:nth-child(-n + 2),
+  .collage--collapsed-y.contains-card--standard:not(.collage--mobile) .collage__item:where(.collage__item--product, .collage__item--collection) + .collage__item {
     --top-left-radius: var(--card-corner-radius);
     --top-right-radius: var(--card-corner-radius);
   }
 
   .collage--collapsed-y:not(.collage--mobile) .collage__item:last-child,
+  .collage--collapsed-x:not(.collage--collapsed-y):not(.collage--mobile) .collage__item,
   .collage--collapsed-x:not(.collage--collapsed-y).collage--mobile .collage__item--left:nth-last-child(3),
   .collage--collapsed-x:not(.collage--collapsed-y).collage--mobile .collage__item--right:nth-child(3),
-  .collage--collapsed-y:not(.collage--collapsed-x).collage--mobile .collage__item--left:nth-last-child(-n + 2) {
+  .collage--collapsed-y:not(.collage--collapsed-x).collage--mobile .collage__item--left:nth-last-child(-n + 2),
+  .collage--collapsed-y.contains-card--standard:not(.collage--mobile) .collage__item:where(.collage__item--product, .collage__item--collection) {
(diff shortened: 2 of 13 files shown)


assets/component-card.css
```diff
@@ -37,14 +37,32 @@
...
+.blog--collapsed .card--standard .card__inner,
+.blog--collapsed .card--card:after,
+.blog--collapsed .card--standard .card__inner:after,
+.collage--collapsed .card--card,
+.collage--collapsed .card--standard .card__inner,
+.collage--collapsed .card--card:after,
+.collage--collapsed .card--standard .card__inner:after {
+  border-radius: var(--top-left-radius) var(--top-right-radius) var(--bottom-right-radius) var(--bottom-left-radius);
+}
+
```
kmeleta, line 58: Anyone prefer this?
```suggestion
.grid[class*='--collapsed'].contains-card .card--card,
.grid[class*='--collapsed'].contains-card .card--standard .card__inner,
.grid[class*='--collapsed'].contains-card .card--card:after,
.grid[class*='--collapsed'].contains-card .card--standard .card__inner:after {
  border-radius: var(--top-left-radius) var(--top-right-radius) var(--bottom-right-radius) var(--bottom-left-radius);
}
```

assets/section-main-blog.css
```diff
@@ -96,3 +96,85 @@
...
+@media screen and (min-width: 750px) {
+  /* collage layout */
+  .blog--collapsed.blog-articles--collage .article:first-child,
+  .blog--collapsed.blog-articles--collage.contains-card--standard .article:nth-child(3n + 1) {
+    --top-right-radius: var(--border-radius);
+    --top-left-radius: var(--border-radius);
+  }
+  .blog--collapsed.blog-articles--collage .article:nth-child(3n + 1):last-child,
+  .blog--collapsed.blog-articles--collage .article:nth-child(3n + 2):nth-last-child(-n + 2),
+  .blog--collapsed.blog-articles--collage.contains-card--standard .article:nth-child(3n + 1) {
```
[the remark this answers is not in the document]
ludoboludo: I won't comment the rest, I think it's address or mentioned in the other PR 😅

assets/section-main-blog.css
```diff
@@ -96,3 +96,85 @@
...
+  .blog--collapsed.blog-articles--collage.contains-card--standard .article {
+    --bottom-right-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+}
+
+@media screen and (min-width: 750px) {
+  /* collage layout */
+  .blog--collapsed.blog-articles--collage .article:first-child,
+  .blog--collapsed.blog-articles--collage.contains-card--standard .article:nth-child(3n + 1) {
```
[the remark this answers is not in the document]
kmeleta: :first-child isn't the same as 3n + 1. We need 3n + 1 here.

assets/section-main-blog.css
```diff
@@ -96,3 +96,85 @@
...
+    --top-right-radius: var(--border-radius);
+    --top-left-radius: var(--border-radius);
+  }
+  .blog--collapsed.blog-articles--collage .article:nth-child(3n + 1):last-child,
```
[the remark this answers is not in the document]
kmeleta: This means the last child is also one of full width articles that are in the 3n + 1 pattern.

assets/section-main-blog.css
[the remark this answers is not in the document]
kmeleta: This means every this child starting at the first, meaning all the full width ones.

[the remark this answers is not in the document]
kmeleta: You should look at what these are doing in the dev tools to learn how they're working. It's also the best way to find potentially unneeded ones or ones that are applying too broadly.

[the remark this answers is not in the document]
ludoboludo: nah I know how they work, it's just that my mind was stuck on the collage section 😓

assets/collage.css
```diff
@@ -249,3 +262,94 @@
...
+    --bottom-right-radius: var(--card-corner-radius);
+  }
+
+  .collage--collapsed-x .collage__item--right:nth-last-child(2):first-child,
+  .collage--collapsed-x .collage__item--right:nth-child(3n - 1):not(:last-child),
+  .collage--collapsed-x .collage__item--right:nth-child(3n - 2):not(:first-child)  {
+    --bottom-left-radius: var(--card-corner-radius);
+  }
+}
+
```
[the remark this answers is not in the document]
kmeleta: That should be fixed. I also tried to account for mixed use of standard cards and media cards. Like if a standard card sits above a media card with 0px y spacing, the media card should still have radius along the top. I'm a little torn about adding the additional lines of code for it, but since I mostly can solve for it in this section (because of the limited layout options) I figured I should.

assets/component-slider.css
```diff
@@ -114,16 +132,34 @@ slider-component.slider-component-full-width {
...
+    --top-left-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+
+  .slider.slider--tablet.grid.grid.grid--collapsed-x .slider__slide:last-child {
+    --top-right-radius: var(--border-radius);
+    --bottom-right-radius: var(--border-radius);
+  }
+
 }
```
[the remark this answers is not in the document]
kmeleta: Same type of overrides needed to be copied from the mobile/tablet slider code for desktop. Should be fixed.

assets/base.css
```diff
@@ -1002,6 +1002,13 @@ summary::-webkit-details-marker {
   }
 }
 
+.grid--collapsed .grid__item {
+  --top-left-radius: var(--border-radius);
+  --top-right-radius: var(--border-radius);
+  --bottom-right-radius: var(--border-radius);
+  --bottom-left-radius: var(--border-radius);
+}
```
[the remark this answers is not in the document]
kmeleta: I could also consider including the equivalent blog and collage selectors in the collapsed grid file as well if that is appealing. The original thought being they should mostly live in those specific files instead of a generic `.grid` overrides file. But the collage code at least has certainly bloated a bit, so that could be a conversation worth having.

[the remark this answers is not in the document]
ludoboludo: Yeah 🤔  I think maybe we can do some testing once the functionality and code is approved. To compare the different approaches in terms of where to keep that CSS.

assets/section-main-product.css
```diff
@@ -1178,3 +1195,65 @@ a.product__text {
...
+    --top-left-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+
+  .product--stacked .grid--collapsed-x:not(.grid--collapsed-y) .grid__item:nth-child(odd),
+  .product--stacked .grid--collapsed-x:not(.grid--collapsed-y) .grid__item:last-child:nth-child(even) {
+    --top-right-radius: var(--border-radius);
+    --bottom-right-radius: var(--border-radius);
+  }
+}
```
ludoboludo, line 1259: Not added in here but I guess coming from main maybe ? There is some padding right applied to the first product image which messes with zero grid space:

ludoboludo, line 1259: Here when there is no vertical space but some horizontal one, we're losing the border radius:

assets/component-slider.css
```diff
@@ -226,6 +262,23 @@ slider-component.slider-component-full-width {
...
+  .slider.slider--desktop.grid.grid--collapsed-x .slider__slide.grid__item:first-child {
+    --top-left-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+
+  .slider.slider--desktop.grid.grid.grid--collapsed-x .slider__slide.grid__item:last-child {
+    --top-right-radius: var(--border-radius);
+    --bottom-right-radius: var(--border-radius);
+  }
 }
```
ludoboludo, line 282: Noticed some issues with slider items when there is no vertical spacing but there is some horizontal one: 
[collection list as cards or standard](https://screenshot.click/22-37-q4map-go8gi.png)

[blog post slider video](https://screenshot.click/22-35-ovk3w-ffu61.mp4)

[multicolumn video](https://screenshot.click/22-26-d0vbe-1dyhg.mp4)

assets/component-card.css
```diff
@@ -81,6 +99,11 @@
...
 }
 
+.card.card--card .card__inner .card__media {
+  border-bottom-right-radius: 0;
+  border-bottom-left-radius: 0;
+}
+
 .card--standard.card--text {
   background-color: transparent;
 }
```
ludoboludo, line 109: NOt sure if we care about this scenario since it's for standard cards with horizontal spacing and no vertical one: 


Same here: [screenshot](https://screenshot.click/22-21-om5go-z4vw2.png). But im pretty sure we're good 👍

assets/component-slider.css
```diff
@@ -226,6 +262,23 @@ slider-component.slider-component-full-width {
...
+
+  .slider.slider--desktop.grid.grid--collapsed-x .slider__slide.grid__item:first-child {
+    --top-left-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+
+  .slider.slider--desktop.grid.grid.grid--collapsed-x .slider__slide.grid__item:last-child {
+    --top-right-radius: var(--border-radius);
+    --bottom-right-radius: var(--border-radius);
+  }
```
ludoboludo, line 281: Ill create an issue I think where we can look if including the border width into the calculation of the grid item width:

assets/base.css
```diff
@@ -2830,16 +2837,28 @@ details-disclosure > details {
 
 .grid--collapsed.contains-content-container .content-container,
 .grid--collapsed.contains-media .global-media-settings {
+  border-radius: var(--top-left-radius) var(--top-right-radius) var(--bottom-right-radius) var(--bottom-left-radius);
   margin-bottom: var(--collapsed-offset-y);
   margin-right: var(--collapsed-offset-x);
```
ludoboludo, line 2842: here I noticed that for multicolumn, when it's not set to swipe on mobile, the blocks gets too wide due to a negative margin:

ludoboludo (commented):
Noticed a few things. Some I think are known and expected. 

It'd be nice to probably have the same treatment for collapsible content when set to `row container` 🤔  [screenshot](https://screenshot.click/22-14-r4zdj-s1khq.png)

assets/component-slider.css
kmeleta: Oh good catch. Added another slider override to each breakpoint. This case got missed because sliders are only concerned with x spacing so I forgot to make sure the original radius was applied when collapsed base styles were still applied by a y-only configuration.

assets/base.css
kmeleta: This was only on multicolumn right? The issue looked to be a width: 100% on multicolumn list. I was hesitant to remove it in this PR, but it really didn't seem like it was needed for any current multicolumn configurations.

assets/section-main-product.css
kmeleta: Should be fixed along with the other sliders

kmeleta: That actually turned in to quite a thing haha. We apparently have a case where there may be only one _visible_ media item but it isn't the only `<li>` child so we can't target it with CSS alone. Separately, but also related, is even when we have this case, we still render the slider classes for some reason, which means it inherits all the slider related margins, widths, and of course collapsed radius values.. This section really needs a cleanup. 

I think this is safe to include here, but let me know what you think. The i_deal_ solution to all this is a much larger task.
https://github.com/Shopify/dawn/pull/1478/commits/678cb77372e81d75e83dc1a1ecc10bfc70be9705

kmeleta: The issue you screenshot is because that variant:first-child rule with the padding actually needed be referring to the "only visible media" item.

kmeleta:
> It'd be nice to probably have the same treatment for collapsible content when set to `row container` 🤔 [screenshot](https://screenshot.click/22-14-r4zdj-s1khq.png)

@ludoboludo Collapsible content isn't using global spacing currently so I'm not going to account for it here. Opened an issue to track it though https://github.com/Shopify/dawn/issues/1522

kmeleta:
> I noticed that we can't display the rounded radius from the post-last thumbail if a PPD media galery (view screenshot below) but we manage to do it when it's the same in a collection list. There is no way to do it in the PPD like we do in the collection list?

@YoannJailin I certainly can, the problem is more that I wouldn't be able to prevent that same corner from being rounded in this situation 


I suppose I created this problem for myself when I merged the PR prevented the last orphan item from being full width no matter what regardless of if it was a video/3d model or not (which I still think is the right thing to do).

kmeleta:
> I noticed that the mobile right padding is different when we have a card list vs the end of a slider. Left spacing seems narrower on every mobile slider's end.

@YoannJailin this should be resolved now. I technically fixed it on the other commit because the issue related more to that code, but the change should be reflected here now too.

melissaperreault:
>  [The result is really not cute](https://user-images.githubusercontent.com/95925370/159774239-9f44396e-57ce-47c3-8e4c-b738cb33a3d4.png) everytime an image has another ratio. It looks even worse when we intend to have zero padding between cards. Do we intend to propose a setting for image gallery to we can control the cropping of the images? (like we do for products cards with the "Image ratio" droplist (portrait, square, adapt to image).

@YoannJailin Sorry for replying late to this. Can you open an issue to discuss this one and build a case of why it requires  additional thoughts? 🙏 Thank you

cc. @wiktoriaswiecicka Since you'll explore product template. 🙏

assets/section-main-product.css
```diff
@@ -1211,3 +1236,65 @@ a.product__text {
...
```
ludoboludo, line 1300: I noticed some issues in here when the horizontal spacing is set to 0 and the vertical one is set to something greater than 0: 


And the other way around as well with some horizontal spacing and no vertical one:

assets/section-main-blog.css
```diff
@@ -96,3 +96,85 @@
...
+  .blog--collapsed-x.contains-card--standard:not(.blog-articles--collage) .article:nth-child(2n + 2) {
+    --top-right-radius: var(--border-radius);
+    --bottom-right-radius: var(--border-radius);
+  }
+  .blog--collapsed-x:not(.blog-articles--collage):not(.blog--collapsed-y) .article:nth-child(odd),
+  .blog--collapsed-x.contains-card--standard:not(.blog-articles--collage) .article:nth-child(2n + 1) {
+    --top-left-radius: var(--border-radius);
+    --bottom-left-radius: var(--border-radius);
+  }
+}
```
ludoboludo, line 180: Some horizontal space but no vertical one: 


No horizontal space but some vertical one:

ludoboludo (commented):
Noticed a few scenarios that needs to be tweaked for the main product images and main blog. 

I think a new rebase would be good to get [the fix for the quick add button on mobile](https://screenshot.click/07-43-g5ix9-0s3k1.png) where the atc button is wider than the card.

sections/main-list-collections.liquid
```diff
@@ -15,7 +15,7 @@
       assign collections = collections | reverse
     endif
   -%}
-  <ul class="collection-list contains-card grid grid--{{ section.settings.columns_desktop }}-col-desktop grid--{{ section.settings.columns_mobile }}-col-tablet-down{% render 'collapsed-classes' %}" role="list">
+  <ul class="collection-list contains-card{% if settings.card_style == 'standard' %} contains-card--standard{% endif %} grid grid--{{ section.settings.columns_desktop }}-col-desktop grid--{{ section.settings.columns_mobile }}-col-tablet-down{% render 'collapsed-classes' %}" role="list">
     {%- for collection in collections -%}
       <li class="collection-list__item grid__item">
         {% render 'card-collection', card_collection: collection, media_aspect_ratio: section.settings.image_ratio, columns: 3 %}
```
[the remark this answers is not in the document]
kmeleta: Yeah good find. Probably something I'd opt to do in a follow up. We don't even know how to handle the lack of product heading space in standard cards yet and I see that as a bigger issue than this. FYI @YoannJailin

PR #1504, opened 2022-03-17
Fixed condensed mega menu

tyleralsbury:
Quick fix for the specificity that handles the condensed menu when there are only 2 levels.


sofiamatulis (dismissed):
Good catch thank you! 🎉

andrewetchen (dismissed):

PR #1520, opened 2022-03-22
Using font size and spacing for mega menu hierarchy

tyleralsbury:

Further explorations on how to improve the hierarchy of the different navigation levels for the mega menu.  In this PR we are trying removing the `bold` and instead using **font size** and **spacing** to help to emphasize the hierarchy.

Added a modifier.  I tried to use some fancy selectors, but this was cleaner.

**Other considerations**

I used `:not(:only-child)` to ensure that L2 links on their own won't have additional margins.

We may want to fine tune the numbers for the margin below the L2 link as well as potentially the vertical gap between the rows.  The total space below the L2 links including their padding is `1.8rem` and the total space between rows is `2.4rem` which is quite similar.

**Testing steps/scenarios**

- [ ] Add a navigation that has 3 levels
- [ ] In header -> apply "Mega Menu" setting to menu type setting

**Note:** This PR shouldn't affect any other menu types or make any changes to menus that only have 1 or 2 levels.

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/127752568854/editor)

**Checklist**

### DIFF
--- a/assets/component-mega-menu.css
+++ b/assets/component-mega-menu.css
@@ -49,11 +49,17 @@
 .mega-menu__link {
   color: rgba(var(--color-foreground), 0.75);
   display: block;
+  font-size: 1.2rem;
   line-height: calc(1 + 0.3 / var(--font-body-scale));
   padding-bottom: 0.8rem;
   padding-top: 0.8rem;
   text-decoration: none;
   transition: text-decoration var(--duration-short) ease;
+}
+
+.mega-menu__link--level-2:not(:only-child) {
+  font-size: 1.4rem;
+  margin-bottom: 1rem;
 }
 
 .header--top-center .mega-menu__list {
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -431,7 +431,7 @@
                         <ul class="mega-menu__list page-width{% if link.levels == 1 %} mega-menu__list--condensed{% endif %}" role="list">
                           {%- for childlink in link.links -%}
                             <li>
-                              <a href="{{ childlink.url }}" class="mega-menu__link link font-body-bold{% if childlink.current %} mega-menu__link--active{% endif %}"{% if childlink.current %} aria-current="page"{% endif %}>
+                              <a href="{{ childlink.url }}" class="mega-menu__link mega-menu__link--level-2 link{% if childlink.current %} mega-menu__link--active{% endif %}"{% if childlink.current %} aria-current="page"{% endif %}>
                                 {{ childlink.title | escape }}
                               </a>
                               {%- if childlink.links != blank -%}


allylane (dismissed):
Visually this works as expected. The only other thing we need to do is remove the animation from the mega menu opening which we could also include in this PR cc @danielvan

tyleralsbury:
> Visually this works as expected. The only other thing we need to do is remove the animation from the mega menu opening which we could also include in this PR cc @danielvan

Are we removing that animation entirely?

danielvan:
> Are we removing that animation entirely?

I was discussing this with @allylane, and both of us had different experiences when the animation triggered and when it didn't, as well as the transition animation. @tyleralsbury, are we able to remove the animation for now and revisit on a different PR once we have time set aside to work on it?

tyleralsbury:
> > Are we removing that animation entirely?
> 
> I was discussing this with @allylane, and both of us had different experiences when the animation triggered and when it didn't, as well as the transition animation. @tyleralsbury, are we able to remove the animation for now and revisit on a different PR once we have time set aside to work on it?

I think it should be easy enough to remove. Will try now.

tyleralsbury:
@danielvan @allylane Animation has been removed.

assets/component-mega-menu.css
```diff
@@ -49,13 +42,19 @@
 .mega-menu__link {
   color: rgba(var(--color-foreground), 0.75);
   display: block;
+  font-size: 1.2rem;
```
ludoboludo, line 45: I find this a little small 😬 
From what i'm seeing on other websites, having a bold or different color level 2 seems like the most common use case ([H&M](https://www2.hm.com/en_us/index.html), [Quicksilver](https://www.quiksilver-shop.ca/), [Fao Schwarz](https://faoschwarz.com/), etc). 
T

ludoboludo (dismissed):
The code seem to work as expected but not a huge fan of the new approach/tweaks.

I find that it was easier to read with the bold level2. It did look not as good when you had too many level2 without any level3 menu item in them but I don't think this approach is any better in those cases, things kinda look maybe connected maybe not: 



So I'd say I'm more in favour of our current approach with bold text but if we're to keep this approach then the code looks good 👍

assets/component-mega-menu.css
melissaperreault: Something is off here, from the original logic, they would be styled as L2s, not L3s when in presence of L2s and L3s

assets/component-mega-menu.css
```diff
@@ -49,13 +42,22 @@
...
   text-decoration: none;
   transition: text-decoration var(--duration-short) ease;
 }
 
+.mega-menu__link--level-2 {
+  font-size: 1.4rem;
+}
+
+.mega-menu__link--level-2:not(:only-child) {
+  margin-bottom: 1rem;
```
ludoboludo, line 58: Looks better now 👌 

But I think maybe it could be a little less margin.

melissaperreault:
> The code seem to work as expected but not a huge fan of the new approach/tweaks.

@ludoboludo I hear you and happy to share the context a bit! We've been bouncing back and forth for that one based on exploration test that bold wasn't scaling well for all fonts. We explored whether to keep bold or not and tested various outcome: Underline, Opacity, Uppercase, Divider columns, Bold and Size change, etc.

From the feedback we received, changing the size was the less opinionated styling we could go with until we provide better alternative to create hierarchy controls for merchants. But in the end, yes, indeed it's quite a subjective change!

assets/component-mega-menu.css
melissaperreault: I also think they can be the same value!

melissaperreault (commented):
Thank you Tyler!

I've stress tested:
- [x] Long menu names
- [x] Tall menu list
- [x] Uneven menu items
- [x] Ipad
- [x] Android
- [x] Various fonts

### Observations
Capturing here but will open a new issue for them if they don't exist yet!

- [x] Long menu label [overlap with the next column](https://screenshot.click/22-03-5xl8y-x86to.png)
- [ ] When menu list becomes long, combined with scroll, the experience is annoying, the mega menu closes on users pretty drastically. ([Video reference](https://screenshot.click/22-03-gn1lu-ykxqg.mp4))
- [ ] Scrolling inside the mega nav doesn't prevent from scrolling the page behind. This could convey there are still menu items to discover but in theory there are not. ([Video reference](https://screenshot.click/22-03-g66xa-anhtc.mp4))
- [x] Mobile drawer nested back link label [is centered instead of left aligned](https://screenshot.click/22-03-yhx10-b1plj.png). Noticeable when menu text link is very long. 


Other than that, very close to approve! Haven't found anything else! 🙏

assets/component-mega-menu.css
LucasLacerdaUX: I also find `1.2rem` way too small. I've looked at other websites ([H&M](https://www2.hm.com/en_us/index.html), [Quicksilver](https://www.quiksilver-shop.ca/), [Fao Schwarz](https://faoschwarz.com/), [Gap](https://www.gapcanada.ca/), [Simons](https://www.simons.ca/en), [Nordstrom](https://www.nordstrom.ca/)) and they're all using minimum of `13px`. This led me to test with `1.3rem` and it looks so much more readable. `1.3rem` would match our link sizes for product links, which I personally find at the limit of being easily readable. 

## 1.2rem
image

## 1.3rem
image

It does lose some of the hierarchy, so I'm not sure if there's something we can do about it. 

I don't have the strongest opinion here, but just sharing my personal point of view as a user. I find it kinda hard to read menus in Dawn and Studios on desktop when using `1.2rem` fonts, and I'd prefer having less distinction between the hierarchy levels than barely being able to read without having to zoom / get closer to the screen

------

<details>
<summary>Other themes @ 12px</summary>
image
image
image
image
image
</details>

LucasLacerdaUX (dismissed):
I'm not a big fan of the new `1.2rem` font size ([see comment](https://github.com/Shopify/dawn/pull/1520#discussion_r837861475)) but code looks good and things are working as expected. If we're okay with the design, consider this an approval :D

🎩
✅ 👍
👖

tyleralsbury:
> Long menu label [overlap with the next column](https://screenshot.click/22-03-5xl8y-x86to.png)

Fixed.

> When menu list becomes long, combined with scroll, the experience is annoying, the mega menu closes on users pretty drastically. ([Video reference](https://screenshot.click/22-03-gn1lu-ykxqg.mp4))

> Scrolling inside the mega nav doesn't prevent from scrolling the page behind. This could convey there are still menu items to discover but in theory there are not. ([Video reference](https://screenshot.click/22-03-g66xa-anhtc.mp4))

Blocked scrolling when the menu is open.  This fixes both of these points.

> Mobile drawer nested back link label [is centered instead of left aligned](https://screenshot.click/22-03-yhx10-b1plj.png). Noticeable when menu text link is very long.

Fixed.

I also adjusted the `margin-bottom` to `0.8rem`.

tyleralsbury:
@melissaperreault Not sure if we want to revisit the use of `1.2rem` and switch it to `1.3rem`.  We could bump the L2s to `1.5rem` if we did that to keep it `0.2` larger.  I don't have a strong opinion.

[Lucas's comment](https://github.com/Shopify/dawn/pull/1520#discussion_r837861475) is a great exploration.

assets/details-disclosure.js
```diff
@@ -39,6 +39,8 @@ class HeaderMenu extends DetailsDisclosure {
   }
 
   onToggle() {
+    document.body.classList.toggle('overflow-hidden', this.querySelectorAll('.mega-menu[open]').length);
```
ludoboludo, line 42: this class, when applied and not only in this scenario, is creating a bit of a jump that makes the experience feel subpar 🤔 

It does it as well on the drawer for pick up availability and probably anytime we apply the `overflow-hidden` class in the theme. 

[video](https://screenshot.click/29-57-wcssz-97g0x.mp4)

ludoboludo: When applying the `overflow: hidden;` it hides the scroll bar from the page so it creates that shift.

tyleralsbury: One option that pops into my head but which I'm not sure would work... `overflow-y: scroll` might force a scrollbar (again, might depend on OS), and then make the height of the container `100vh` so that it has nowhere to go... you'd also need to add a `transform` to it to position it in the spot it was in before adding the overflow.  Feels complicated, but something like that might work. 🤷 

We can look into if there are any more elegant solutions.

ludoboludo: Hmm yeah I was reading suggestions yesterday but nothing seemed very good or straight forward 🤔 
This sounded interesting but didn't try it yet: https://aykevl.nl/2014/09/fix-jumping-scrollbar

melissaperreault (dismissed):
Thanks Tyler! Looks  good to me! 🚢 

Related to font size, I hear the feedback, we'll monitor and adjust later as we explore font hierarchy.

danielvan:
I agree with the comments above on the L3s feeling small. I would be ok in shipping with 1.3rem for L3 to see if the comment repeats once this is live and in dynamic contexts. It seems also a gain for a11y and click target areas. 

@melissaperreault ultimately this is a decision on the design system, so keen to hear your thoughts.

melissaperreault:
@danielvan 

No strong opinion apart from following accessible recommendations to aim for bigger all the time on my end! We can ship with a 1px difference and monitor, instead of the 2px difference proposed during feedback time.

Less risk for merchants and buyers to go from 1.3 to 1.2 instead of the opposite. I'd also rather offer a bigger font size if we are aligned! cc. @nicklepine

LucasLacerdaUX (commented):
Just to add to my comment above:

melissaperreault:
> So this pattern that mega menu open = scroll blocked isn't really a thing buyers are used to.

Aligned, not the desired outcome I think, thanks for sharing this angle Lucas! We can park the use case/edge case and monitor because the point you are making is reallly not what we would want to refrain buyers to do! I'll open an issue!

I don't know how this can be improved but Mega menu are meant for large inventory that encourages long menu list. That following reference was difficult to experience considering these factors. Maybe we are pushing the edge cases a little too hard here but worth the conversation.

> _Scrolling inside the mega nav doesn't prevent from scrolling the page behind. This could convey there are still menu items to discover but in theory there are not. ([Video reference](https://screenshot.click/22-03-g66xa-anhtc.mp4))_

assets/details-disclosure.js
```diff
@@ -35,14 +35,15 @@ customElements.define('details-disclosure', DetailsDisclosure);
...
   constructor() {
     super();
-    this.header = document.querySelector('header');
+    this.header = document.querySelector('.header-wrapper');
   }
 
   onToggle() {
-    document.body.classList.toggle('overflow-hidden', this.querySelectorAll('.mega-menu[open]').length);
+    if (!this.header) return;
+    this.header.preventHide = !!this.querySelectorAll('.mega-menu[open]').length;
```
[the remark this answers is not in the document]
melissaperreault: Love it! The only trade off that this creates is that if buyer triggers the menu and don't close it then go explore down the page, the header will not appear like expected on scroll up. Did I get this right?

If so, I am less concerned because the menu was open on user behalf and they can close it by clicking/tapping a link or click outside the menu (dropdown or megamenu)

sections/header.liquid
```diff
@@ -618,6 +618,7 @@
       if (this.predictiveSearch && this.predictiveSearch.isOpen) return;
 
       if (scrollTop > this.currentScrollTop && scrollTop > this.headerBounds.bottom) {
+        if (this.preventHide) return;
```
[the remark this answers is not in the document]
LucasLacerdaUX: Thanks for sharing this. This is indeed one of the expected trade offs with this approach. [Meli shared it above](https://github.com/Shopify/dawn/pull/1520#discussion_r840017048) and said it shouldn't be a concern.

assets/component-mega-menu.css
```diff
@@ -18,6 +17,10 @@
   z-index: -1;
 }
 
+.shopify-section-header-sticky .mega-menu__content {
```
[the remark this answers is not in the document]
melissaperreault: Yes! This is only when the mega menu will take the full height of the view space right - I am happy that we can click outside of the menu most of the time for mid range height menu

sections/header.liquid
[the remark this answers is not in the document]
tyleralsbury: Switched it to `1.3` (and put a typo in the commit that made it look like I made it smaller).

melissaperreault (approved):
Thank you team! Looks good! 🎉 🚢

sections/header.liquid
ludoboludo, line 621: One of the other effect, I say "issue" in the video but it isn't one. Is that the sticky header will stay stick until you close the submenu: [video](https://screenshot.click/01-49-3tk40-3hzmt.mp4)

But since it's easy to close I don't see it being too much of an issue 🤔  Just sharing so we know all the behavioural changes with this.

ludoboludo (approved):
Seems good 👍

LucasLacerdaUX (approved):
👍🏻

PR #1530, opened 2022-03-23
Implement collapsible borders for sections

### LINKED ISSUE #1524: [Global settings] Collapse adjacent section borders
This is an extension of https://github.com/Shopify/dawn/issues/922

Similarly to how grid items within sections should "collapse" adjacent borders and rounded corners, adjacent sections should be able to collapse as well. 

It should be possible to collapse borders of adjacent sections, but I don't believe it's technically possible to collapse rounded corners, as a section would need to know if the following section is eligible to collapse or not. I'm not aware of anyway in liquid to parse sections and their settings before rendering them.

kmeleta:
**PR Summary:** 

Adds the ability to collapse adjacent section borders to cleanly stack sections with 0px of vertical spacing.

Fixes #1524 

Each section has been given conditional liquid logic to determine its collapsible eligibility at the top and bottom of the section. For the most part, sections with 0px top padding and no heading are "eligible" to collapse upwards, for example.

The actual "collapsing" effect is accomplished similarly to how we implemented it for grids, by using negative margins to offset one element onto another using the amount of border width as the offset distance. The problem is there's no real way for one section to know that a sibling above or below is eligible to collapse because we can't modify classes on the outer `.section` element based on a section's settings. (css :has() could probably help with this problem in the distant future)

The solution I found was to always have top-eligible sections offset upwards and use bottom-eligibility to enable the offset of the following section by setting an inline style context `#my-section-id + .section { --prev-section-offset: {{ my_section_border_width }}`. This approach does 2 key things..
1) Allows the offset to be conditionally set, so the next sibling doesn't offset unnecessarily or potentially mess up non-zero spacing values.
2) Allows the next section to offset only as much as the previous section's border requires. If a section offsets by its own width, a section above with less border width would have some of its content covered as a result. Having differing border widths involved isn't an ideal situation anyway, but it'd be nice to handle it gracefully.

**Other considerations**

This feature is highly reliant on consistent, predictable margins around grids. Ideally grids don't have any generic margins associated them and the conditionally non-existent headings, buttons, etc around them should manage their own spacing in relation to the grid. **This PR doesn't attempt to correct for any inconsistencies in grid margins**. Doing so would likely involve either systematic changes or section specific adjustments. Both of which would ideally be implemented and tested in isolation (notice no section-specific stylesheets are modified in this PR). Any issues in this regard can be recorded and addressed separately alongside this PR or as follow ups.

**Testing steps/scenarios**
- [ ] No regressions using non-zero vertical spacings (a 4px value is easily visually validated)
- [ ] No regressions when titles and/or CTA buttons are present
- [ ] Different border widths
- [ ] Sliders collapse properly above and preserve original margins below

**Demo links**

- [Store](https://jwdkm8cvrh3yg87j-45840990230.shopifypreview.com)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127758073878/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -981,6 +981,11 @@ summary::-webkit-details-marker {
   row-gap: var(--grid-mobile-vertical-spacing);
 }
 
+/* remove once heading/grid spacing is better managed */
+.section-collapsed .grid {
+  margin-top: 0;
+}
+
 @media screen and (min-width: 750px) {
   .grid {
     column-gap: var(--grid-desktop-horizontal-spacing);
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -1,4 +1,5 @@
 slider-component {
+  --focus-outline-padding: 0.5rem;
   --desktop-margin-left-first-item: max(5rem, calc((100vw - var(--page-width) + 10rem - var(--grid-desktop-horizontal-spacing)) / 2));
   position: relative;
   display: block;
@@ -27,7 +28,6 @@ slider-component.slider-component-full-width {
 }
 
 .slider__slide {
-  --focus-outline-padding: 0.5rem;
   --shadow-padding-top: calc(var(--shadow-vertical-offset) * -1 + var(--shadow-blur-radius));
   --shadow-padding-bottom: calc(var(--shadow-vertical-offset) + var(--shadow-blur-radius));
   scroll-snap-align: start;
@@ -47,6 +47,10 @@ slider-component.slider-component-full-width {
     margin-bottom: 1rem;
   }
 
+  .section-collapsed .slider--mobile {
+    top: calc(var(--focus-outline-padding) * -1);
+  }
+
   .slider.slider--mobile .slider__slide {
     margin-bottom: 0;
     padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
@@ -57,7 +61,7 @@ slider-component.slider-component-full-width {
     padding-bottom: var(--focus-outline-padding);
   }
 
-  .slider.slider--mobile.contains-content-container .slider__slide {
+  .slider.slider--mobile.contains-content-container {
     --focus-outline-padding: 0rem;
   }
 }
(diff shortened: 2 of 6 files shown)


kmeleta:
@melissaperreault let's walk through this one together.

YoannJailin:
@kmeleta  Okay no problem. Just one last questions tough 😅 is it even not possible between the same section? To me more specific, I am thinking about subsequent -image with text- sections with no titles. Here is an example of what i was trying to achieve. No problem if it's still not possible, just trying to push my luck!

Capture d’écran, le 2022-04-11 à 18 14 53

kmeleta:
> is it even not possible between the same section? To me more specific, I am thinking about subsequent -image with text- sections with no titles. @YoannJailin 

A case like image with text is _more_ possible I suppose, in that there are fewer variables, but it still can't be done safely because each image with text section can have its own top and bottom section padding. Theoretically we probably could do it for sections like Rich text, where this padding is applied _inside_ the section container. However I don't feel there are enough of those sections for the feature to be worthwhile.

melissaperreault (commented):
### UX Update

Ken and I aligned that we would need the 2 other collapsing PRs to merge before reviewing this one so we can have a big picture of the flexibility and limitations all around.

**Our shared opinions**
- Optimizing for collapsing corner radius is not a priority compared to the impact of just collapsing successfully borders. _(We can create various styles with just the border collapsing successfully between sections)_
- No gap grid is mostly appealing in a sharp/square corners
- When a full list of sections is stacked, the open and closing section that could receive the corner radius is not strong enough to invest in a solution to collapse the corners contained in between and to fix all the edge cases and exceptions. Meaning we don't believe the visual outcome/visual impact it provides justifies such an investment for now. If we want no gap, sharp corners should be prioritized for now.

**Quick review for things to look for**
- [ ] Collapsible content section when no heading and various styling
- [ ] Product media gallery tall layout above a no gap section. When not sticky, even if all settings are set to zero, [we still have a gap](https://screenshot.click/22-04-h7u07-s66mo.png)

PR #1535, opened 2022-03-25
[WIP] Prettier plugin demo

charlespwd:
This PR exists to share what running the prettier-plugin-liquid would do to the Dawn repo.

This way we can see what would happen and discuss rules / spot bugs.


layout/password.liquid
```diff
@@ -4,16 +4,28 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1">
-    <meta name="theme-color" content="">
+    <meta name="theme-color" content=""">
```
charlespwd, line 7: This needs a unit test.

layout/theme.liquid
```diff
@@ -193,20 +205,38 @@
...
     {%- endif -%}
 
-    <script>document.documentElement.className = document.documentElement.className.replace('no-js', 'js');
-    if (Shopify.designMode) {
-      document.documentElement.classList.add('shopify-design-mode');
-    }
+    <script>
+      document.documentElement.className = document.documentElement.className.replace('no-js', 'js');
+          if (Shopify.designMode) {
+            document.documentElement.classList.add('shopify-design-mode');
```
charlespwd, line 238: Indentation is off by 1

sections/cart-icon-bubble.liquid
```diff
@@ -8,9 +8,9 @@
 <span class="visually-hidden">{{ 'templates.cart.cart' | t }}</span>
 {%- if cart != empty -%}
   <div class="cart-count-bubble">
-      {%- if cart.item_count < 100 -%}
-        <span aria-hidden="true">{{ cart.item_count }}</span>
-      {%- endif -%}
-      <span class="visually-hidden">{{ 'sections.header.cart_count' | t: count: cart.item_count }}</span>
+    {%- if cart.item_count < 100 -%}<span aria-hidden="true">{{ cart.item_count }}</span>{%- endif -%}
```
charlespwd, line 11: This should probably break. But why? Because there was whitespace in the input or because it's too nested?

sections/cart-live-region-text.liquid
```diff
@@ -1 +1,3 @@
-{{ 'sections.cart.new_subtotal' | t }}: {{ cart.total_price | money_with_currency }}
+{{ 'sections.cart.new_subtotal' | t }}
+:
+{{ cart.total_price | money_with_currency }}
```
charlespwd, line 3: Oh shit. Top level as paragraph? Didn't see this one coming.

sections/pickup-availability.liquid
```diff
@@ -1,4 +1,6 @@
-{% comment %}theme-check-disable UndefinedObject{% endcomment %}
+{% comment %}
+  theme-check-disable UndefinedObject
+{% endcomment %}
```
charlespwd, line 3: Probably don't need to break unconditionally. Unless it's breaking because that's more than 100 chars? 🤔

sections/pickup-availability.liquid
```diff
@@ -46,28 +76,32 @@
...
           </p>
 
           {%- assign address = availability.location.address -%}
           <address class="pickup-availability-address">
             {{ address | format_address }}
 
-            {%- if address.phone.size > 0 -%}
-              <p>{{ address.phone }}</p>
-            {%- endif -%}
+            {%- if address.phone.size > 0 -%}<p>{{ address.phone }}</p>{%- endif -%}
```
charlespwd, line 104: That should probably break.

sections/video.liquid
```diff
@@ -17,29 +17,39 @@
 
 <div class="color-{{ section.settings.color_scheme }} gradient">
   <div class="video-section isolate{% unless section.settings.full_width %} page-width{% endunless %} section-{{ section.id }}-padding">
-    <div{% if section.settings.full_width %} class="page-width"{% endif %}>
+    <div {% if section.settings.full_width %}
+        class="page-width"
+      {%- endif %}>
```
charlespwd, line 22: 1 attribute that force breaks here.

sections/footer.liquid
```diff
@@ -231,41 +257,79 @@
...
+                            {% if country.iso_code == localization.country.iso_code %}
+                              aria-current="true"
+                            {% endif %}
+                            data-value="{{ country.iso_code }}"
+                          >
+                            {{ country.name }}
+                            <span class="localization-form__currency"
+                              >({{ country.currency.iso_code }}
+                              {{ country.currency.symbol }})</span
+                            >
```
charlespwd, line 321: Kind of funky looking but _correct_. prettier-html would do the same.

charlespwd: You could _manually_ add whitespace stripping characters and get this instead:
```
<span class="localization-form__currency"
  >(
  {{- country.currency.iso_code }}
  {{ country.currency.symbol -}}
  )</span
>
```

ludoboludo: Hmm yeah a little funky 🤔  but I get the logic

snippets/icon-accordion.liquid
```diff
@@ -1,5 +1,14 @@
...
+  <svg
+    class="icon icon-accordion color-foreground-{{ settings.accent_icons }}"
+    aria-hidden="true"
+    focusable="false"
+    role="presentation"
+    xmlns="http://www.w3.org/2000/svg"
+    width="20"
+    height="20"
+    viewBox="0 0 20 20"
+  >
```
ludoboludo, line 11: I wonder if this should also apply to SVG elements in liquid files like this for icons. 🤔  Definitely easier to read but I feel like it isn't necessary.

templates/customers/register.liquid
```diff
@@ -110,7 +117,8 @@
         <svg aria-hidden="true" focusable="false" role="presentation">
           <use href="#icon-error" />
         </svg>
-        {{ form.errors.translated_fields['password'] | capitalize }} {{ form.errors.messages['password'] }}.
+        {{ form.errors.translated_fields['password'] | capitalize }}
+        {{ form.errors.messages['password'] }}.
       </span>
```
ludoboludo, line 122: I think we need some funky looking placement here cause we're getting some whitespace issue. Which we were getting prior to the reformatting.

templates/customers/reset_password.liquid
```diff
@@ -60,12 +60,13 @@
           <svg aria-hidden="true" focusable="false" role="presentation">
             <use href="#icon-error" />
           </svg>
-          {{ form.errors.translated_fields['password'] | capitalize }} {{ form.errors.messages['password'] }}
+          {{ form.errors.translated_fields['password'] | capitalize }}
+          {{ form.errors.messages['password'] }}
         </small>
```
ludoboludo, line 65: Same here but I don't if it's something the plugin could catch and tweak 🤔  It's probably more on us to know and make sure that inline elements will add white space in those cases.

templates/gift_card.liquid
```diff
@@ -118,13 +120,22 @@
         {%- assign gift_card_balance = gift_card.balance | money -%}
       {% endif %}
       {%- if gift_card.balance != gift_card.initial_value -%}
-        <p class="gift-card__label caption-large">{{ 'gift_cards.issued.remaining_html' | t: balance: gift_card_balance }}</p>
+        <p class="gift-card__label caption-large">
+          {{ 'gift_cards.issued.remaining_html' | t: balance: gift_card_balance }}
+        </p>
```
ludoboludo, line 125: I fee like here there is a need for `-}}` here too no ?

ludoboludo (commented):
Did a quick review of some files but not all 76 😅

sections/featured-collection.liquid
```diff
@@ -47,20 +47,35 @@
...
+      {%- if section.settings.title != blank -%}
+        <h2 class="title {{ section.settings.heading_size }}">{{ section.settings.title | escape }}</h2>
+      {%- endif -%}
+      {%- if section.settings.description != blank or section.settings.show_description and section.settings.collection.description != empty -%}
+        <div class="collection__description {{ section.settings.description_style }}">
+          {%- if section.settings.show_description -%}
+            {{ section.settings.collection.description }}
+          {%- else -%}
+            {{ section.settings.description -}}
+          {%- endif %}
```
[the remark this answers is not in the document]
charlespwd: That sounds like a bug 👀

templates/customers/register.liquid
charlespwd: Same thing here. The _source_ had whitespace in it so we understood that it was meaningful and thus didn't change the output. I think it's working as intended. If the source _lacked_ whitespace here, it would have added a whitespace stripping character to the error message drop.

templates/gift_card.liquid
charlespwd: No because `p` tags are `display: block`. Which means they are not inner leading whitespace sensitive or inner trailing whitespace sensitive.

sections/featured-collection.liquid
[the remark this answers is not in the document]
charlespwd: Ahh I see it now. 

It's because the _source_ had the whitespace stripping characters.

```liquid
{%- if section.settings.show_description -%}{{ section.settings.collection.description }}{%- else -%}{{ section.settings.description }}{% endif %}</div>
```

So when we _broke_ the code into a new line, the plugin determined that the `{{ section.settings.collection.description }}` was not leading whitespace sensitive (because we're stripping it).

And our logic is to only _add_ whitespace stripping characters is to add one iff there wasn't whitespace before and it is sensitive.

In this case, it was not sensitive because of the _source_, thus no new whitespace stripping characters were added.

sections/collage.liquid
```diff
@@ -52,14 +64,16 @@
...
               </div>
             {%- when 'product' -%}
-              {% render 'card-product',
+              {%
+                render 'card-product',
                 card_product: block.settings.product,
                 media_aspect_ratio: 'adapt',
                 show_secondary_image: block.settings.second_image,
                 extend_height: true
               %}
```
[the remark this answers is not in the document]
charlespwd: This seems like a good candidate for a GitHub discussion :D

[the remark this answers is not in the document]
charlespwd: https://github.com/Shopify/prettier-plugin-liquid/discussions/13

snippets/icon-accordion.liquid
charlespwd: https://github.com/Shopify/prettier-plugin-liquid/discussions/15

Issue #732, opened 2021-10-04
Option to go straight to cart upon add to cart.


nicklepine:
A significant amount of merchants are requesting the possibility to direct their customers to the cart page after clicking the add to cart button rather than surfacing the cart notification module. 

Use cases where this is needed: 
- Shop has only 1 product
- Type of product where you only buy one unit (where the buyer is not expected to keep browsing after adding to cart) 
- Shops that have made customizations to their cart flow 

-----

Debut currently includes a setting to turn/off the cart notification. 
![image](https://user-images.githubusercontent.com/8127145/135906330-ada6e29a-2cbf-4894-87da-d6991f3530ff.png)

---

Perhaps we use a dropdown format for the "Cart behaviour" setting given additional options may be added down the road (i.e. cart drawer)

nicklepine:
Hey! @ilseatloop It remains in progress. We are aiming for this to be part of an eventual release in 1-2 months, but might be merged into main earlier than that.

tyleralsbury:
cc @ludoboludo assigning this to you, it ties into the cart drawer work you've been doing.  I would definitely do it in a separate PR but it makes sense to add both things at once.

https://github.com/Shopify/dawn/pull/744 there is an open PR - you can use it or just use it as inspiration, there's a good amount of context in that PR (some provided by yourself!) so it's useful either way. :)

ludoboludo:
Merged this feature in https://github.com/Shopify/dawn/pull/1638 👍

Issue #599, opened 2021-09-09
[Global] Inserted links which open in a new window do not include a warning


svinkle:
## Issue summary

Links which are set to open a new window from the Admin interface do not include a warning. This could confuse visually impaired users who, when discover and interact with the link, believe they are following a link in the same browser window. This also causes extra work for sighted, keyboard-only users to switch back to the previous tab or window.

<details>
  <summary>Screenshots</summary>

  <img width="288" alt="Dawn theme settings. Heading with text section, edit link form shows link text 'your brand with your customers' linking to All products with the 'Open this link in a new window' checkbox selected." src="https://user-images.githubusercontent.com/1392632/132725616-f2c0fe75-9416-43a3-8d2d-b34c77871d1f.png">
  <img width="786" alt="Dawn homepage, heading with text section. An underlined link with the text 'your brand with your customers' is highlighted." src="https://user-images.githubusercontent.com/1392632/132725745-d11f15d8-362b-4e97-923d-8b1f6addf473.png">

</details>

#### HTML

```html
<p>
  Share information about
  <a href="/collections/all" target="_blank" title="All products">
    your brand with your customers
  </a>
  . Describe a product, make announcements, or welcome customers to your store.
</p>
```

## Steps to reproduce

0. In Dawn theme editor, insert a new link into a text area, making sure to select the "Open this link in a new window" checkbox
1. Go to a Dawn store with any supported browser
2. With an active screen reader, <kbd>Tab</kbd> to a link which opens a new window

### Expected

- Links which open a new window will be accompanied by a visual and aural warning.

### Actual

- Links which open a new window do not include a warning.

## Recommendation

1. Add a new `hidden` element to the DOM which includes the warning text.
2. Add the `aria-describedby` attribute to links, setting its value to the `id` of the warning text container, providing an audible warning.
3. Include a "new window" SVG providing a visual warning.

With this change:

- Sighted mouse and touch users will see the visual icon, providing the visual warning.
- Screen readers will hear the aural warning, providing context on the result of interacting with the link.

#### HTML

```html
<div hidden id="warning-text">Opens in a new window.</div>

<p>
  Share information about
  <a href="/collections/all" target="_blank" aria-describedby="warning-text">
    your brand with your customers
    <svg class="icon icon-new-window" aria-hidden="true" focusable="false">
      <use href="#icon-new-window"></use>
    </svg>
  </a>
  . Describe a product, make announcements, or welcome customers to your store.
</p>
```

<!-- #### Diff

[...]

PR #1544, opened 2022-03-28
Cart drawer

tyleralsbury:
**PR Summary:** 

Fixes #0.

**Other considerations**

**Testing steps/scenarios**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**


bredowmax:
@nicklepine can you please also consider the opposite of #732: a setting to prevent shoppers to bypass the cart page. I wrote more about this in #1609

ludoboludo:
@Oliviammarcello here is some feedback on some of your items: 

> - [ ] I am not super clear on how the border stylings in theme settings work so feel free to challenge this, but shouldn't the drawer border lines apply to the separator lines? [Screenshot.](https://screenshot.click/11-23-niviz-ifwfu.png) 

It's not being applied on the customer account page, it's not for the pick up availability so I think it's safe to keep it this way. Otherwise we could end up with big lines that take more space and limit what can be viewed in the drawer.

> - [ ] The Cart type Page setting isn't working. When I add something to the cart, nothing happens

This will be done in a separate PR: 

> - [ ] Can we add the same animations used for the menu item drop downs to cart notes.

For this, I looked at the collapsible blocks on the product page and the collapsible content section and in those sections/parts, there isn't any animation when opening/closing. I'm happy to look into adding it for cart note but wanted to bring this up so we compare to other parts of the theme where the same behaviour exists. 

> - [ ] Do we have control over the roundness of the dynamic checkout buttons, they are inconsistent? [Screenshot](https://screenshot.click/11-02-na3xp-wqz8j.png)

Paypal isn't one we have been able to properly tweak for the border unfortunately while the others are good. Though I noticed we apply the `box-shadow` for paypal but not on the other ones. 

> - [ ] On mobile, dynamic checkout buttons aren't aligned horizontally. [Screenshot.](https://screenshot.click/11-10-y5kgt-lzvt6.png) Also on tablet, a button is missing or the alignment is broken. [Screenshot](https://screenshot.click/11-12-iudfe-z8o3c.png)

I haven't been able to replicate this 🤔 You were testing on your laptop ? I tested as well on my android phone and didn't run into the issue.

Oliviammarcello:
> It's not being applied on the customer account page, it's not for the pick up availability so I think it's safe to keep it this way. Otherwise we could end up with big lines that take more space and limit what can be viewed in the drawer.
Sounds good

> > * [ ]  Can we add the same animations used for the menu item drop downs to cart notes.
> 
> For this, I looked at the collapsible blocks on the product page and the collapsible content section and in those sections/parts, there isn't any animation when opening/closing. I'm happy to look into adding it for cart note but wanted to bring this up so we compare to other parts of the theme where the same behaviour exists.
Ah I see, let's match collapsible blocks and flip the arrow in the same pattern. 

> > * [ ]  On mobile, dynamic checkout buttons aren't aligned horizontally. [Screenshot.](https://screenshot.click/11-10-y5kgt-lzvt6.png) Also on tablet, a button is missing or the alignment is broken. [Screenshot](https://screenshot.click/11-12-iudfe-z8o3c.png)
> 
> I haven't been able to replicate this 🤔 You were testing on your laptop ? I tested as well on my android phone and didn't run into the issue.

Yes I was on my laptop.

Oliviammarcello:
- [ ] For the vertical space issue, instead of the cart note remaining sticky to the footer of the drawer can we add it below the last product to show more of the products?
- [ ] Can we move the variants onto one line and separate them with a comma 
- [ ] There is still room to reduce the space between the variants and quantity selector. [Screenshot](https://screenshot.click/12-23-g4fiv-itzoc.png)
- [ ] We can reduce the space between the separator line and the first product by another 1rem. [Screenshot](https://screenshot.click/12-27-irjnr-4rwor.png)
- [ ] Reduce space around tax/shipping text. [Screenshot](https://screenshot.click/12-29-6tqqx-8bkju.png)

svinkle:
@ludoboludo Looking great! Nice work here.

Here's a few more things I noticed:

- [ ] <kbd>Space</kbd> on cart header control scrolls the page (even though `preventDefault()` is there?)
- [ ] Since this is a dynamic environment, remove item controls should be conveyed as a button (`role="button"`) and support the <kbd>Space</kbd> key.
- [ ] When removing an item via remove button, focus is lost. On remove, if there are other items, shift focus to the _next_ product link (`a.cart-item__name`). If there are no more items (empty state) shift focus to the "Continue shopping" link.
- [ ] It's still difficult to use in landscape mode: cannot see any products.
- [ ] Empty state features an image link and text link to what I assume is a collection, but this is not conveyed. "13 products." What are these links for? Should there be a collection name?
- [ ] `span` is invalid child of `dl` – swap for `div` instead.
- [ ] Table header content doesn't align with body cell content. Adjust header columns:
   1. Product image (visually hidden)
   2. Product
   3. Total
   4. Quantity (visually hidden)
- [ ] When item quantity is adjusted, the new Subtotal value is not announced. Add `role="status"` to the `div.totals` container to create a live region which will announce the change.

bacounts:
Can you raise the empty cart's collection card? My screenshot shows that it will get covered up by chat or app icons when a longer collection name is present.

**Other Thoughts:** There might be an argument for making the collection have two options:
1. Collection card - that displays the collection's first product or image when present (basically, how it is now)
2. Collection carousel - displaying product cards with card count options, etc.

PS: Adding an option to display the collection directly below the product list when the cart is full or scrolled will allow a better cross-sell option for sellers. Adding a header text option would be ideal (i.e., Frequently bought together).

Screen Shot 2022-04-21 at 1 40 22 PM

bacounts:
It could be tacked in a separate PR for all pop-ups, windows, and drawers, but the close buttons should be synced visually. The circle with the "x" inside should be the default (as seen in the Quick add pop-up).

Screen Shot 2022-04-21 at 2 19 08 PMScreen Shot 2022-04-21 at 2 19 31 PM

Oliviammarcello:
Looking good! 

Few other notes:
- [ ] Can we reduce the opacity of the scroll bar to 55% (or the lowest percent that still ensures that there is enough colour contrast) and round the top and bottom of the line? 

> For this, the [label isn't there on the cart page for mobile](https://screenshot.click/19-31-aln8m-ah5cn.png) so we were reusing the same approach. It also reduces the space taken per items which seem like a win 🤔 You think I should add it anyway ?

Good catch, lets leave it out thanks!

> I have removed some spacing that was there. Now it still doesn't look quite align due to the difference in font size and line height. [Is this ok ?](https://screenshot.click/19-47-gd89d-tl2pn.png) or should we be tweaking the line height or something else ?

- [ ] Could we tweak it to visually look aligned? It should be a bit higher up

assets/cart-drawer.js
```diff
@@ -0,0 +1,106 @@
...
+
+  setHeaderCartIconAccessibility() {
+    const cartLink = document.querySelector('#cart-icon-bubble');
+    cartLink.setAttribute('role', 'button');
+    cartLink.setAttribute('aria-haspopup', 'dialog');
+    cartLink.addEventListener('click', (event)=> {
+      event.preventDefault();
+      this.open(cartLink)
+    });
+    cartLink.addEventListener('keydown', (event) => {
```
svinkle, line 18:
```suggestion
    cartLink.addEventListener('keyup', (event) => {
```

To satisfy [2.5.2: Pointer Cancellation](https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html).

Oliviammarcello (dismissed):

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+      </div>
+      <cart-drawer-items {% if cart == empty %}class=" is-empty"{% endif %}>
+        <form action="{{ routes.cart_url }}" class="cart__contents cart-drawer__form" method="post" id="cart-drawer-form">
+          <div class="drawer__contents js-contents" id="drawer-main-cart-items">
+            {%- if cart != empty -%}
+              <div class="drawer__cart-items-wrapper">
+                <table class="cart-items" role="table">
+                  <thead role="rowgroup">
+                    <tr role="row">
+                      <th id="ColumnProductImageDrawer" class="visually-hidden" role="columnheader">{{ 'sections.cart.headings.image' | t }}</th>
```
svinkle, line 47: Some environments don't convey `visualyhidden` cell semantics. Best to place this content in a `span. 
```suggestion
                      <th id="ColumnProductImageDrawer" role="columnheader"><span class="visually-hidden">{{ 'sections.cart.headings.image' | t }}</span></th>
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+                      <th id="ColumnProductDrawer" class="caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.product' | t }}</th>
+                      <th id="ColumnTotalDrawer" class="right caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.total' | t }}</th>
+                      <th id="ColumnQuantityDrawer" class="visually-hidden" role="columnheader">{{ 'sections.cart.headings.quantity' | t }}</th>
```
svinkle, line 50: Same issue here.

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+                              {%- endif -%}
+
+                              {%- for property in item.properties -%}
+                                {%- assign property_first_char = property.first | slice: 0 -%}
+                                {%- if property.last != blank and property_first_char != '_' -%}
+                                  <div class="product-option">
+                                    <dt>{{ property.first }}: </dt>
+                                    <dd>
+                                      {%- if property.last contains '/uploads/' -%}
+                                        <a href="{{ property.last }}" class="link" target="_blank">
```
svinkle, line 117: Does this include a "new window" warning? (Ex., https://github.com/Shopify/dawn/issues/599.)

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+                          <div class="loading-overlay hidden">
+                            <div class="loading-overlay__spinner">
+                              <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+                                <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+                              </svg>
+                            </div>
+                          </div>
+
+                          <div class="cart-item__price-wrapper">
+                            {%- if item.original_line_price != item.final_line_price -%}
```
svinkle, line 152: Could we update this to reflect the same markup as in lines 78-97?

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+                            </svg>
+                          </div>
+                        </td>
+                      </tr>
+                    {%- endfor -%}
+                  </tbody>
+                </table>
+              </div>
+            {%- endif -%}
+            <p class="visually-hidden" id="cart-drawer-live-region-text" aria-live="polite" role="status"></p>
```
svinkle, line 235: `aria-live` is redundant here and the next line. Elements with the role `status` have an implicit `aria-live` value of `polite`.

assets/cart-drawer.js
ludoboludo: Hmm ok, ill re test then. The reason I changed it to `keydown` was because with key up it was scrolling down the page until the space bar is released.

sections/cart-drawer.liquid
ludoboludo: Added one now. I haven't added an SVG icon though. I think this can be done as a follow up so we can get an icon designed by designer and have them look into it.

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,345 @@
...
+                                {{ discount.title }}
+                              </li>
+                            {%- endfor -%}
+                          </ul>
+                        </td>
+
+                        <td class="cart-item__totals right" role="cell" headers="ColumnTotalDrawer">
+                          <div class="loading-overlay hidden">
+                            <div class="loading-overlay__spinner">
+                              <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
ludoboludo: I just double check and we don't have a label next to the spinning icon but we have a `p` within the cart to convey the loading state that we show/hide when loading. 
So when I change the quantity, I do here a `loading...` and then it mention the new totals. We should be good 👍

snippets/cart-drawer.liquid
```diff
@@ -0,0 +1,11 @@
+{% comment %}
+    Renders cart drawer
+
+    Accepts:
+    - color_scheme: {String} sets the color scheme of the notification (optional)
```
LucasLacerdaUX, line 5: Looks like this was copied from `cart-notification.liquid` and we don't accept this parameter here.

locales/en.default.json
```diff
@@ -74,7 +74,8 @@
     "loading": "Loading...",
     "total_reviews": "total reviews",
     "star_reviews_info": "{{ rating_value }} out of {{ rating_max }} stars",
-    "collapsible_content_title": "Collapsible content"
+    "collapsible_content_title": "Collapsible content",
+    "open_in_new_window": "Opens in a new window."
```
LucasLacerdaUX, line 78: We already have this string under `link_messages.new_window` just above (line 71).

snippets/cart-drawer.liquid
ludoboludo: Oh this is because it was part of the design but will need to be added in a different setting for more than just the cart drawer. Thanks for flagging tho, ill remove it for now.

Here are the designs for context: Colour schemes. [Figma.](https://www.figma.com/file/JcYdswrk2JhqBTGBVWPM7J/Dawn-2.0?node-id=13884%3A111446) |  Other drawers colour schemes. [Figma]

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,344 @@
+<script src="{{ 'cart.js' | asset_url }}" defer="defer"></script>
```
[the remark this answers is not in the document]
ludoboludo: Need for the section rendering API calls 👍

assets/cart.js
```diff
@@ -72,43 +73,55 @@ class CartItems extends HTMLElement {
...
+        } else if (document.querySelector('.cart-item') && cartDrawerWrapper) {
+          trapFocus(cartDrawerWrapper, document.querySelector('.cart-item__name'))
+        }
         this.disableLoading();
-      }).catch(() => {
+      }).catch((e) => {
         this.querySelectorAll('.loading-overlay').forEach((overlay) => overlay.classList.add('hidden'));
-        document.getElementById('cart-errors').textContent = window.cartStrings.error;
+        const errors = document.getElementById('cart-errors') || document.getElementById('drawer-cart-errors');
+        console.log('window cartStrings errors', e,  window.cartStrings.error)
```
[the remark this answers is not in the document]
ludoboludo: Yeah I just found it super useful while I was working on it to know what was failing where. I can remove it before merging 👍

sections/main-cart-footer.liquid
```diff
@@ -79,18 +79,18 @@
...
-  }
-
-  customElements.define('cart-note', CartNote);
+        this.addEventListener('change', debounce((event) => {
+          const body = JSON.stringify({ note: event.target.value });
+          fetch(`${routes.cart_update_url}`, {...fetchConfig(), ...{ body }});
+        }, 300))
+      }
+    });
+  };
```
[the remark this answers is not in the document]
ludoboludo: I moved it in the `cart.js` file. Ill push the changes and test again tomorrow to make sure it's all good.

sections/header.liquid
```diff
@@ -85,6 +85,9 @@
 <script src="{{ 'details-disclosure.js' | asset_url }}" defer="defer"></script>
 <script src="{{ 'details-modal.js' | asset_url }}" defer="defer"></script>
 <script src="{{ 'cart-notification.js' | asset_url }}" defer="defer"></script>
+{%- if settings.cart_type == "drawer" -%}
+  <script src="{{ 'cart-drawer.js' | asset_url }}" defer="defer"></script>
+{%- endif -%}
```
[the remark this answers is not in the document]
ludoboludo: So here what I'm seeing is that when I load the CSS in the header, I get a delay and the drawer first shows up as open. So I moved some of the styling to be in the `cart-drawer.liquid` section within a `<style>`. 
Let me know if you think I should go another way about it :)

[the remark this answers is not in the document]
LucasLacerdaUX: @ludoboludo Do we need to have all the styling in the `.liquid` file, in that case? Or can we move back non-critical properties to the `.css` file?

Would something like
```css
.drawer {
    visibility: hidden;
}
```
be enough to prevent this delay you noticed? Didn't test, just thinking out loud.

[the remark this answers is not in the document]
ludoboludo: Nah that's a good point. Probably works that way. Ill make an update locally and test.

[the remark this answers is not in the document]
ludoboludo: done, ready for another look 👍

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+
+<style>
+  .drawer {
+    visibility: hidden;
+  }
+</style>
+
+<cart-drawer class="drawer{% if cart == empty %} is-empty{% endif %}">
+  <div class="cart-drawer" id="cart-drawer">
```
LucasLacerdaUX, line 10: This `id` is also not using the same pattern we've been using on other parts of the theme.

Should we rename it to `CartDrawer`?

```suggestion
  <div id="CartDrawer" class="cart-drawer">
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+    <div class="cart-drawer__overlay" id="cart-drawer-overlay"></div>
```
LucasLacerdaUX, line 11:
```suggestion
    <div id="CartDrawer-Overlay" class="cart-drawer__overlay"></div>
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+          <div class="drawer__contents js-contents" id="drawer-main-cart-items">
+            {%- if cart != empty -%}
+              <div class="drawer__cart-items-wrapper">
+                <table class="cart-items" role="table">
+                  <thead role="rowgroup">
+                    <tr role="row">
+                      <th id="ColumnProductImageDrawer" role="columnheader"><span class="visually-hidden">{{ 'sections.cart.headings.image' | t }}</span></th>
+                      <th id="ColumnProductDrawer" class="caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.product' | t }}</th>
+                      <th id="ColumnTotalDrawer" class="right caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.total' | t }}</th>
+                      <th id="ColumnQuantityDrawer" role="columnheader"><span class="visually-hidden">{{ 'sections.cart.headings.quantity' | t }}</span></th>
```
LucasLacerdaUX, line 56:
```suggestion
                      <th id="CartDrawer-ColumnProductImage" role="columnheader"><span class="visually-hidden">{{ 'sections.cart.headings.image' | t }}</span></th>
                      <th id="CartDrawer-ColumnProduct" class="caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.product' | t }}</th>
                      <th id="CartDrawer-ColumnTotal" class="right caption-with-letter-spacing" scope="col" role="columnheader">{{ 'sections.cart.headings.total' | t }}</th>
                      <th id="CartDrawer-ColumnQuantity" role="columnheader"><span class="visually-hidden">{{ 'sections.cart.headings.quantity' | t }}</span></th>
```

Can we also standardize these `id`s?

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+                    </tr>
+                  </thead>
+
+                  <tbody role="rowgroup">
+                    {%- for item in cart.items -%}
+                      <tr class="cart-item" id="Drawer-cartItem-{{ item.index | plus: 1 }}" role="row">
```
LucasLacerdaUX, line 62:
```suggestion
                      <tr class="cart-item" id="CartDrawer-Item-{{ item.index | plus: 1 }}" role="row">
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+                              {%- endif -%}
+
+                              {%- for property in item.properties -%}
+                                {%- assign property_first_char = property.first | slice: 0 -%}
+                                {%- if property.last != blank and property_first_char != '_' -%}
+                                  <div class="product-option">
+                                    <dt>{{ property.first }}: </dt>
+                                    <dd>
+                                      {%- if property.last contains '/uploads/' -%}
+                                        <div hidden id="newWindowCartDrawer">{{ 'accessibility.link_messages.new_window' | t }}</div>
```
LucasLacerdaUX, line 123: Can we move this to the `<ul hidden>` (line 230) in `theme.liquid` with `id="a11y-new-window-message"` instead of printing it for each element?

In #1612, I'm replacing that `<ul hidden>` with a `<div hidden>`, by the way.

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+          <details id="Details-Cart-Drawer">
+            <summary>
+              <span class="summary__title">
+                {{ 'sections.cart.note' | t }}
+                {% render 'icon-caret' %}
+              </span>
+            </summary>
+            <cart-note id="cart-drawer-note" class="cart__note field">
+              <label class="visually-hidden" for="drawer-Cart-note">{{ 'sections.cart.note' | t }}</label>
+              <textarea class="text-area text-area--resize-vertical field__input" name="note" id="drawer-Cart-note" placeholder="{{ 'sections.cart.note' | t }}">{{ cart.note }}</textarea>
```
LucasLacerdaUX, line 257:
```suggestion
              <textarea id="CartDrawer-Note" class="text-area text-area--resize-vertical field__input" name="note"placeholder="{{ 'sections.cart.note' | t }}">{{ cart.note }}</textarea>
```
For consistency, `id`s first and a more standardized naming for the value.

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+        <!-- CTAs -->
+
+        <div class="cart__ctas" {{ block.shopify_attributes }}>
+          <noscript>
+            <button type="submit" class="cart__update-button button button--secondary" form="cart-drawer-form">
+              {{ 'sections.cart.update' | t }}
+            </button>
+          </noscript>
+
+          <button type="submit" id="drawer-checkout" class="cart__checkout-button button" name="checkout"{% if cart == empty %} disabled{% endif %} form="cart-drawer-form">
```
LucasLacerdaUX, line 307:
```suggestion
          <button type="submit" id="CartDrawer-Checkout" class="cart__checkout-button button" name="checkout"{% if cart == empty %} disabled{% endif %} form="cart-drawer-form">
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+        super();
+
+        this.addEventListener('change', debounce((event) => {
+          const body = JSON.stringify({ note: event.target.value });
+          fetch(`${routes.cart_update_url}`, {...fetchConfig(), ...{ body }});
+        }, 300))
+      }
+    });
+  };
+{% endjavascript %}
```
LucasLacerdaUX, line 329: Already defined in `cart.js` which is imported at the top of this file.

layout/theme.liquid
```diff
@@ -215,6 +215,10 @@
       {{ "accessibility.skip_to_text" | t }}
     </a>
 
+    {%- if settings.cart_type == 'drawer' -%}
+      {%- render 'cart-drawer' -%}
```
LucasLacerdaUX, line 219:
```suggestion
      {%- section 'cart-drawer' -%}
```
Would that work? :D

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+                            </svg>
+                          </div>
+                        </td>
+                      </tr>
+                    {%- endfor -%}
+                  </tbody>
+                </table>
+              </div>
+            {%- endif -%}
+            <p class="visually-hidden" id="cart-drawer-live-region-text" role="status"></p>
```
LucasLacerdaUX, line 240:
```suggestion
            <p id="CartDrawer-LiveRegionText" class="visually-hidden" role="status"></p>
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+            <p class="visually-hidden" id="shopping-cart-drawer-line-item-status" aria-hidden="true" role="status">{{ 'accessibility.loading' | t }}</p>
```
LucasLacerdaUX, line 241:
```suggestion
            <p id="CartDrawer-LineItemStatus" class="visually-hidden" aria-hidden="true" role="status">{{ 'accessibility.loading' | t }}</p>
```

LucasLacerdaUX (commented):
Mostly nitpicks related to naming consistency in `id`s. We have ids starting with `Cart-drawer`, `Drawer-`, `cart-drawer`, `shopping-cart-drawer`, `drawer-`, etc.

sections/cart-drawer.liquid
ludoboludo: I'd be nice to align with the rest of the team on this at some point cause there is a few different ways IDs are written across the theme 😬

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,350 @@
...
+          </div>
+          <div id="drawer-cart-errors" role="alert"></div>
+        </form>
+      </cart-drawer-items>
+      <div class="drawer__footer">
+        {%- if settings.show_cart_note -%}
+          <details id="Details-Cart-Drawer">
```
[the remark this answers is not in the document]
ludoboludo: This I can't change as it's part of the JS pattern: 
```javascript
document.querySelectorAll('[id^="Details-"] summary').forEach((summary) => {
  summary.setAttribute('role', 'button');
  summary.setAttribute('aria-expanded', 'false');

  if(summary.nextElementSibling.getAttribute('id')) {
    summary.setAttribute('aria-controls', summary.nextElementSibling.id);
  }

  summary.addEventListener('click', (event) => {
    event.currentTarget.setAttribute('aria-expanded', !event.currentTarget.closest('details').hasAttribute('open'));
  });

  if (summary.closest('header-drawer')) return;
  summary.parentElement.addEventListener('keyup', onKeyUpEscape);
});
```

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,334 @@
+<script src="{{ 'cart.js' | asset_url }}" defer="defer"></script>
+
+<style>
+  .drawer {
+    visibility: hidden;
+  }
+</style>
+
+<cart-drawer class="drawer{% if cart == empty %} is-empty{% endif %}">
```
LucasLacerdaUX, line 9: Did we consider at any point using the `details` / `summary` pattern on this component, similarly to what we're doing in menu drawers and search? The a11y properties could then be applied via JS to make sure it conveys the correct semantics.

One benefit of that would be for slow connections. Currently, if you tap the cart icon before the scripts finish loading it'll redirect you to the `cart` page. Redirecting to another page isn't really something good for users on slow connections.

Without knowing the full details behind this decision to not use this pattern, I think refactoring this as `details`/`summary` could result in a better experience while keeping our web native + progressive enhancement principles.

This is more a suggestion for improvement in the form of a question than a must have to ship, btw.

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,334 @@
...
+              </div>
+            {%- endif -%}
+            <p id="CartDrawer-LiveRegionText" class="visually-hidden" role="status"></p>
+            <p id="CartDrawer-LineItemStatus" class="visually-hidden" aria-hidden="true" role="status">{{ 'accessibility.loading' | t }}</p>
+          </div>
+          <div id="CartDrawer-CartErrors" role="alert"></div>
+        </form>
+      </cart-drawer-items>
+      <div class="drawer__footer">
+        {%- if settings.show_cart_note -%}
```
LucasLacerdaUX, line 246: Do cart notes show in checkout at any point? I just wanted to confirm if they're working as expected, because I don't see them anywhere.

image

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,334 @@
...
+        <!-- CTAs -->
+
+        <div class="cart__ctas" {{ block.shopify_attributes }}>
+          <noscript>
+            <button type="submit" class="cart__update-button button button--secondary" form="CartDrawer-Form">
+              {{ 'sections.cart.update' | t }}
+            </button>
+          </noscript>
+
+          <button type="submit" id="CartDrawer-Checkout" class="cart__checkout-button button" name="checkout"{% if cart == empty %} disabled{% endif %} form="CartDrawer-Form">
```
LucasLacerdaUX, line 306:
```suggestion
          <button type="submit" id="CartDrawer-Checkout" class="cart__checkout-button button" name="checkout" form="CartDrawer-Form"{% if cart == empty %} disabled{% endif %}>
```

LucasLacerdaUX (commented):
Looking great so far! I think this should be my "final" code review.

Now, back to testing to make sure we didn't break anything when changing ids!

sections/cart-drawer.liquid
ludoboludo: Not they don't show there. But you can complete an order and see the order in the admin with the cart note. Our payment setup is set to test mode so you can checkout following those instructions: [testing shopify payments](https://help.shopify.com/en/manual/payments/shopify-payments/testing-shopify-payments)

sections/cart-drawer.liquid
LucasLacerdaUX: Oh, agree 100% with your comment, Scott. Thanks for this perfect explanation. 

Looking back at my comment, I didn't really realize that displaying a `details`/`summary` would basically result in displaying a cart drawer that is not really functional at all without JS. That alone makes redirecting to the cart page a better choice. It's pointless to show the user a certain component that does not provide any functionality and is also using some pretty confusing semantics. This is also a key difference between this component and the search menu or the navigation links, which "works" without JS - they're just enhanced by JS.

The better A11Y UX is where I thought the JS could be used as a progressive enhancement, but I now see that we already have a dedicated page that already solves this problem haha. 

I'm resolving this conversation, please my first comment in this thread above.

assets/component-cart-drawer.css
```diff
@@ -0,0 +1,379 @@
...
+}
+
+cart-drawer-items::-webkit-scrollbar-thumb {
+  background-color: rgba(var(--color-foreground), 0.7);
+  border-radius: 100px;
+}
+
+cart-drawer-items::-webkit-scrollbar-track-piece {
+  margin-top: 31px;
+}
```
andrewetchen, line 379: I'm assuming this might have been added to combat this - [screenshot](https://user-images.githubusercontent.com/28404165/166509596-7c180787-c4bc-43e1-a38b-213c20ee1ca9.png)

My first reaction was "why wasn't the scrollbar fading away / disappearing?" but now I understand 👍🏼 

**Scrollbar**

This existed before your PR but I wanted to share since `-webkit-scrollbar` isn't supported in Firefox and we'll see similar patterns wherever we have vertical scrolling (or similar drawer components).

Different experience using Chrome and Safari compared to Firefox:

- Chrome and Safari - [video](https://screenshot.click/03-57-jhiex-g6poq-safari__ss.mp4)
- Firefox - [video](https://screenshot.click/03-56-7tc13-8boqe-firefox__2.mp4)

assets/component-cart-drawer.css
```diff
@@ -0,0 +1,379 @@
...
+.cart-drawer .cart-items, .cart-drawer tbody {
+  display: block;
+  width: 100%;
+}
+
+.cart-drawer thead {
+  display: inline-table;
+  width: 100%;
+  position: sticky;
+  top: -1px;
```
andrewetchen, line 188: I noticed with the drawer opened, when you scroll, "Product" and Total" seems to "shake" - [video](https://screenshot.click/03-10-z80jq-28d4t-s-scrolls.mp4)

Do you have any context on whether `top: 0;` isn't used? This seems to fix the "shake".

andrewetchen (commented):
Everything is looking great!

I will keep testing a things but I wanted to ask some questions on the following CSS / drawer:

assets/product-form.js
```diff
@@ -6,16 +6,17 @@ if (!customElements.get('product-form')) {
...
+      if (document.querySelector('cart-drawer')) this.submitButton.setAttribute('aria-haspopup', 'dialog');
     }
 
     onSubmitHandler(evt) {
       evt.preventDefault();
       if (this.submitButton.getAttribute('aria-disabled') === 'true') return;
 
       this.handleErrorMessage();
-      this.cartNotification.setActiveElement(document.activeElement);
+      this.cart.setActiveElement(document.activeElement);
```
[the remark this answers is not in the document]
ludoboludo: ah, do you want to check this PR then 😬  https://github.com/Shopify/dawn/pull/1638

assets/component-cart-drawer.css
ludoboludo: I had it like this originally because when set to `0` I was seeing products behind it when scrolling. It's like it had a mini gap between the header of the table and header of the cart. 
But doesn't seem to be the case anymore, so Ill set it back to `0` 👍

assets/product-form.js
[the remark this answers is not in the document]
LucasLacerdaUX: @ludoboludo 
My bad. I just didn't realize that was still the case once I saw the setting.
It's literally in the PR description that it will be tackled in a separate PR. 🤦‍♂️

I've added myself as a reviewer there and I'm ready to approve this one now 👀

snippets/cart-drawer.liquid
```diff
@@ -0,0 +1,8 @@
+{% comment %}
+    Renders cart drawer
+
+    Usage:
+    {% render 'cart-drawer' %}
+{% endcomment %}
+
+{% section 'cart-drawer' %}
```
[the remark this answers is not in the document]
andrewetchen: Curious to know if it's possible to not show this 🤔 but its most likely a platform limitation.

andrewetchen:
I was doing some mobile testing and found a small issue with scrolling. I wouldn't mark this as a blocker but it could be a quick fix 🤔 

This is happening on my iPhone and iPad:
- Add a few items to your cart.
- Add a cart note.
- Collapse the cart note and scroll.

https://user-images.githubusercontent.com/28404165/166564673-08a7489a-d5fe-46f3-af3c-4675e62dedba.mov

sections/cart-drawer.liquid
```diff
@@ -0,0 +1,334 @@
...
+    <div class="drawer__inner" role="dialog" aria-modal="true" aria-label="{{ 'sections.cart.title' | t }}" tabindex="-1">
+       {%- if cart == empty -%}
+        <div class="drawer__inner-empty">
+          <div class="cart-drawer__warnings center{% if settings.cart_drawer_collection != blank %} cart-drawer__warnings--has-collection{% endif %}">
+            <div class="cart-drawer__empty-content">
+              <h2 class="cart__empty-text">{{ 'sections.cart.empty' | t }}</h2>
+              <button class="drawer__close" type="button" onclick="this.closest('cart-drawer').close()" aria-label="{{ 'accessibility.close' | t }}">{% render 'icon-close' %}</button>
+              <a href="{{ routes.all_products_collection_url }}" class="button">
+                {{ 'general.continue_shopping' | t }}
+              </a>
```
LucasLacerdaUX, line 21: So, currently, this button redirects the user to `collections/all`. But let's say I'm browsing a product, I open the cart drawer and it's empty. Should this button:
1. Redirect me to the `collections/all` page, like in the Cart Page **OR**
2. Simply close the drawer, like in the Cart Popup?

I think **Option 1** makes sense in the context of the cart page, where you don't usually have anything to shop on that page. But in a cart drawer that opens as an overlay, I'm not fully convinced of this. I think **Option 2** kinda feels more natural to me - just acting as an additional close button.

This is not really a suggestion as I'm not fully convinced of the best UX. But I just wanted to share this to get more context behind this decision and hear thoughts from others.

### Cart Drawer

https://user-images.githubusercontent.com/37168033/166566890-8352c51f-ec0f-49e6-b1f3-b8550a14836d.mp4

### Popup notification

https://user-images.githubusercontent.com/37168033/166567031-235ce782-32cf-429d-80f1-6622edc1065c.mp4

LucasLacerdaUX (approved):
Looks good to me. Tested on mobile, desktop and tablet and in multiple browsers.

I left one question related to the `Continue shopping` button for discussion.

sections/cart-drawer.liquid
LucasLacerdaUX: Just to add: I remember that during one of our accessibility testing, the user told us that they couldn't easily tell the difference between `Close` and `Continue shopping` buttons in the cart popup and that it was not immediately obvious that they performed the same action.

So I'm not sure replicating that behaviour here is the solution. But just wanted to bring this to discussion, as we currently have different behaviours for the same label in these two different cart types.

andrewetchen: > 1. Redirect me to the `collections/all` page, like in the Cart Page **OR**
> 2. Simply close the drawer, like in the Cart Popup?
> 
> I think **Option 1** makes sense in the context of the cart page, where you don't usually have anything to shop on that page. But in a cart drawer that opens as an overlay, I'm not fully convinced of this. I think **Option 2** kinda feels more natural to me - just acting as an additional close button without interrupting the user's shopping flow.

Thanks for bringing this up, Lucas. Here's my take on the empty state "Continue shopping" button:

I agree that the "Continue shopping" button should simply close the drawer as it does feel more natural to me as well. There could be instances where users are browsing products (on the 5th or 6th collection page for example) and in the current state, if they click Continue shopping, they'd be redirected to the another collection of the same collection but on the first page. They could just hit the back button but that's probably not ideal and also leads to my next point.

Having worked on the other side of the fence (theme support), one customization request we often got was adding a setting to allow merchants to use their own/custom `link` for the continue shopping button. The same setting could then be added to the page and popup cart types since they also include "Continue shopping" - _might_ be something we could add in the future perhaps.

andrewetchen (dismissed):
Everything is looking and working great!

We could probably improve the mobile table header row spacing:

- [Desktop](https://user-images.githubusercontent.com/28404165/166969098-3cb445ca-b06a-4e79-abdb-009d8acd13d4.png) 👍🏼 
- [Mobile](https://user-images.githubusercontent.com/28404165/166969160-1f238335-e2f4-4fa0-8dcf-dac26e91b68a.png) 🤔 

Might have to do with:

https://github.com/Shopify/dawn/blob/a8ded5267f8ecce6bb6757fc8b907bb93431b1aa/assets/component-cart-items.css#L216-L221

andrewetchen (approved):
Thanks for fixing the mobile table rows. The cart drawer is [looking good](https://user-images.githubusercontent.com/28404165/167028476-78ca3f11-16d2-4a41-9527-969f88d31091.png) 🎉

LucasLacerdaUX (approved):
Both Cart Drawer and Cart Page are working as expected! :DD

Issue #1556, opened 2022-03-30
[Bug] 'Continue shopping' button on Gift card page is broken


bredowmax:
**Describe the current behavior**
After receiving an Email with a gift card, shopper arrives on a page that looks something like https://checkout.shopify.com/gift_cards/[id1]/[id2]
![Screen Shot 2022-03-30 at 9 06 01 AM](https://user-images.githubusercontent.com/30573532/160841318-457925b7-d472-4a36-b5de-376c8bbc097b.png)

The green button on that website leads to checkout.shopify.com, which means that the user will see this:
![Screen Shot 2022-03-30 at 9 06 19 AM](https://user-images.githubusercontent.com/30573532/160841341-b93e9ffa-a6d1-4925-9c3e-641fb016e312.png)

**Describe the expected behavior**
The button should lead to the store.

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 4.0.0

**Possible solution**
- Revert a recent commit and change `{{ request.origin }}` to `{{ shop.url }}` in gift_card.liquid

**Additional context/screenshots**
It looks like this commit broke the page: https://github.com/Shopify/dawn/commit/1098fd8137aa7659bd73c015a7237ce5fe258b1b

PR #1557, opened 2022-03-30
Fix shop link on giftcard page

bredowmax:
**PR Summary:** 
Quick fix for the 'Continue Shopping' link on the gift card page by reverting a previous change

Fixes https://github.com/Shopify/dawn/issues/1556

Reverts https://github.com/Shopify/dawn/commit/1098fd8137aa7659bd73c015a7237ce5fe258b1b on the Gift Card page

### DIFF
--- a/templates/gift_card.liquid
+++ b/templates/gift_card.liquid
@@ -149,7 +149,7 @@
         {%- endif -%}
         <div class="gift-card__buttons no-print">
           <a
-            href="{{ request.origin }}"
+            href="{{ shop.url }}"
             class="button"
             target="_blank"
             rel="noopener"


martinamarien (approved):
The team confirmed that this is the best approach. 
Thanks for reporting this bug and creating a fix. 

I have tested and it is working as expected. 👍🏻

LucasLacerdaUX (approved):
Looks good to me! The `request.origin` was redirecting to `checkout.shopify.com`, which is the expected URL for that context, so we should move forward with reverting it to `shop.url`.

I was a reviewer on the original PR and this was definitely a mistake on my end when testing this page. On the admin's Theme Editor, it redirects to the store but not on the real page. Good catch! :)

Thanks for making this change.

bredowmax:
@martinamarien I signed the CLA. Is there any other action required from my side?

martinamarien:
Hey @bredowmax. Thanks for signing the CLA. 

For today, I can merge this for you, but you will need to complete one more step to be able to merge a PR yourself in the future. The CLA is asking "Connect your email address with a GitHub account" 

Currently, it looks like your commits are being associated with your computer's "non-email" and not your email connected to your github account. I imagine that you've now signed the CLA using your github email, but your commits are being associated with a different one. 

If you'd like to try updating this for a future PR, you can update your [git config](https://git-scm.com/docs/git-config):  
Specifically, I think you will need to update your config email. `--global` would be optional in this case.
```
git config --global user.email "your_email@example.com"
```

Once you complete this step, your future commits should be associated with your github email. 

If you give this a try, it would be great if you could let us know if it solves your issue with the CLA. That way we can provide these steps in our README for other contributors. Thanks!

PR #1586, opened 2022-04-05
Allow eager loading to improve LCP

felixb:
**PR Summary:** 

Allow eager loading to improve LCP for hero images. The default behavior is lazy loading as before.

Lighthouse complains about lazy loading the LCP.

Create an option to force eager loading for the first image.

**Testing steps/scenarios**
- [ ] Load the index page and check that banner image has still an `loading="lazy"` attribute.
- [ ] Change settings of that banner image to eager loading.
- [ ] Load the index page and check that banner image has now an `loading="eager"` attribute.

**Checklist**


ludoboludo:
Hey @felixb, thanks for creating this option to tackle some image loading issues. At this time I don't think this is something we would add to our section options as it is a pretty technical aspect that most merchants might be confused by. 

We could revisit this approach in the future though.

Thanks again for you're involvement in making our theme better!

tyleralsbury:
@felixb just as a follow-up to what Ludo has mentioned, I actually think this is a really good idea in theory and is something I've been leaning towards in recent weeks.  The ideal solution would be for the platform itself to have some way to help with this, but there are a lot of challenges to handling something like this automatically.  In the meanwhile a setting would really help and would massively improve loading speeds for image focused sections when they are at the top of the page.

We would want to ensure two things, though.

1. It's not too complex.  Eager v.s. Lazy loading is not something an average merchant is likely to really understand if they see it as a setting.  Especially because it doesn't make a noticeable change on the page.  We'd want to give it phrasing that's a little easier to understand like "this is the first section on the page" or "this section is above the fold" or something
2. It has parity throughout the theme.  We would audit the theme and update it for all sections with images.

PR #1595, opened 2022-04-06
fix(predictive-search): remove font from image

Maybach91:
**PR Summary:** 
There was an unnecessary `font-family` property on the search image with the content of `'object-fit: container'`, which is invalid at first and second we don’t really need font-family on the search image explicit.

clean code

**Testing steps/scenarios**
- [x] No difference at production, just syntax sugar.

**Checklist**
- [x] Notified the [help.shopify.com](https://help.shopify.com) team about updates to theme settings (current contact: Kimli)
- [] Linted with [Theme Check](https://github.com/Shopify/theme-check)

### DIFF
--- a/assets/component-predictive-search.css
+++ b/assets/component-predictive-search.css
@@ -172,5 +172,4 @@ predictive-search[loading] .predictive-search__results-list:first-child {
 .predictive-search__image {
   grid-area: product-image;
   object-fit: contain;
-  font-family: 'object-fit: contain';
 }


ludoboludo (approved):
Hello @Maybach91, thanks for looking Dawn and ways to make it better 🙂 
In this case I think we had this declaration as a work around to support the css property `object-fit` in older browsers. Though it's only used in this instance 🤔 
So I think we can actually get rid of it. Thanks for taking care of it.

Looks like most browser do support the property and don't need a work around: [screenshot](https://screenshot.click/06-45-nytym-ujr7c.png)

Maybach91:
Any plan to merge this @ludoboludo @KaichenWang @martinamarien ?

PR #1610, opened 2022-04-10
1-click Quick Add for products with 1 non-default variant

### LINKED ISSUE #1604: [Quick Add] Enable 1-click Quick Add for any product in product grid where only 1 option exists
I'm looking into transitioning from the Warehouse theme to Dawn but came across one major limitation regarding Quick Add #1388: Similar to the Empire theme, Dawn appears to be incompatible with Product Options. In exercise, there is no "Add to Cart" button for products that have exactly 1 variant defined.

In the store I'm talking about, shoppers on average buy 10-20 products, and select their products mainly from the collection grid, since products are low-consideration, such as "Banana". 

Assuming 2 clicks difference between the "Add to Cart" vs. "Open Modal" versions of the Quick Add button, and 15 products per cart, a shopper saves 30 clicks with the "Add to Cart" version vs. the "Open Modal" version.

The Warehouse theme does _not_ have this limitation, and with a product inventory of 500 different SKUs, it would be quite an effort for the merchant to make the inventory compatible with the new Quick Add feature.

For any product with only 1 option, I'd expect shoppers to be able to add this variant to the cart with only one click. That means: _without_ opening the Quick Add modal.

I'd expect the same behavior for products where several variants exist, but only 1 option is valid with applied facet filters. Example: A shirt which exists in Blue, Red & Yellow, while color facet filter is set to "blue". An additional message would say: _"More colors available"_

- Dawn Version: 4.0.0

**Possible solution**
Option 1: 
- My proposed solution treats any product with only 1 variant the same - no matter if this one variant was specifically defined or not. 

Option 2:
- I wouldn't be able to produce an example, but I could imagine that some merchants may want to have this particular setup (that is: opening a Quick Add modal even if only 1 variant is available). Therefore, option 2 proposes a theme setting to let merchants choose the preferred behavior

Screen Shot 2022-04-09 at 4 09 45 PM
Screen Shot 2022-04-09 at 4 09 24 PM
Screen Shot 2022-04-09 at 4 09 01 PM

bredowmax:
**PR Summary:** 
Decreases clicks necessary to add products to the cart in cases where a product would have exactly one non-default variant or product option

Fixes #1604.

**Other considerations**
An alternative could be to provide an option for the merchant to toggle this on or off.

**Checklist**

### DIFF
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -143,7 +143,7 @@
         {%- if show_quick_add -%}
           <div class="quick-add">
             {%- assign product_form_id = 'quick-add-' | append: section_id | append: card_product.id -%}
-            {%- if card_product.has_only_default_variant -%}
+            {%- if card_product.has_only_default_variant or card_product.card_product.options.size == 1 -%}
               <product-form>
                 {%- form 'product', card_product, id: product_form_id, class: 'form', novalidate: 'novalidate', data-type: 'add-to-cart-form' -%}
                   <input type="hidden" name="id" value="{{ card_product.selected_or_first_available_variant.id }}" disabled>
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -143,7 +143,7 @@
         {%- if show_quick_add -%}
           <div class="quick-add">
             {%- assign product_form_id = 'quick-add-' | append: section_id | append: card_product.id -%}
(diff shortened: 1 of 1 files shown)


ludoboludo (approved):

ludoboludo:
Thanks a bunch @bredowmax 👍

Issue #1460, opened 2022-03-07
Rich text 2.0


wiktoriaswiecicka:
Currently Rich text section has limited options to style heading and paragraph text or change text alignment. We are adding settings that allow merchant to achieve more layout options.

As a merchant:

- I can add more than 3 blocks to my rich text section
- I can align the content to be right, center of left aligned
- I have more options for position or layout of the content 
- I can italicize and bold words in my headings.

### 1. Section settings
- [ ] Text alignment (Left, Center, Right)
- [ ] Text position on desktop (Left, Center, Right)

### 2. Improve Heading block settings

  - [ ] Adding extra large text size 
  - [ ] Changing short text field to Rich text RTE to enable merchant to use bold / italic in large headings.

### 3. New block

- [ ] Adding caption block
- [ ] Lift block limits

andrewetchen:
>### 2. Improve Heading block settings
>
>  - [ ] Adding extra large text size 
>  - [ ] Changing short text field to Rich text RTE to enable merchant to use bold / italic in large headings.
>
>### 3. New block
>
>- [ ] Adding caption block
>- [ ] Lift block limits

Setting a reminder for myself to look into the following:

**Extra large text size**
- We currently have `h0` for Large; `h1` for Medium; `h2` for Small.
- How will Extra large be implemented?

**Caption block**
- Should we use the same caption block from the Image with text section by including Text style and Text size?

**Block limit**
- Set a higher limit or remove it completely?

Issue #1615, opened 2022-04-12
[Rich text] Empty label still shows button


andrewetchen:
**Describe the current behavior**

When you add a button block and remove the label, an empty button appears.

**Describe the expected behavior**

No button when button label is empty. The Image with text section has a button block and a check to not show the button when the label is empty.

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 4.0.0
- Chrome Version 100.0.4896.75
- macOS Version 12.3.1

**Possible solution**

I'll need to add the following condition for the rich text button block:

```liquid
{%- if block.settings.button_label != blank -%}
  <a{% if block.settings.button_link == blank %} role="link" aria-disabled="true"{% else %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}" {{ block.shopify_attributes }}>
    {{ block.settings.button_label | escape }}
  </a>
{%- endif -%}
```

**Additional context/screenshots**

**Rich text - empty button label**

![image](https://user-images.githubusercontent.com/28404165/163028866-20e06634-b718-485b-8079-fcd72fcf64f6.png)

---

**Image with text - empty button label**

![image](https://user-images.githubusercontent.com/28404165/163029045-6383a640-7dfb-4878-b2bc-d4fcdd73b5dd.png)

PR #1625, opened 2022-04-14
Rich text updates

andrewetchen:
### Draft PR

Rich text updates - UX feedback

cc: @wiktoriaswiecicka 

Related: https://github.com/Shopify/dawn/issues/1460

**Section level settings**
- Desktop context position
- Text/content alignment

**Block level settings and additions**
- Add a new "Extra large" heading option
- Change the heading type from `text` to `richtext` to allow merchants to use bold, italic, links, dynamic source, and underline
- Add a caption block
- Remove limit for each block

### Other considerations

Range slider widths will need to be explored further / section parity etc.

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/127844286486/editor)

---

### Questions

**Section settings**
- Text position on desktop:
  - This setting currently doesn't exist in our theme. The only similar setting that does is "Desktop content position" which I think conveys the same meaning. Let me know your thoughts on this.
- Text or content alignment:
  - Most other sections have "Desktop content alignment" and "Mobile content alignment".
  - We also have a global "Text alignment" setting for Cards.
  - Since this setting will set the alignment of all block elements, I think it makes sense to go with something that includes "content" instead of "text". If this ends up being the case:
    - Should we keep "Content alignment"?
    - For section parity, the **Image with text** section has "Desktop content alignment" and "Mobile content alignment" - if there are plans to include a mobile adjustment for text/content alignment, would it make more sense to offer "Desktop content alignment" and "Mobile content alignment" for rich text instead of just one text/content alignment adjustment for desktop and mobile. _Thoughts?_

**Extra large heading size**
- Let me know if this needs to be made smaller or larger.

**Caption block**
- I used the same caption block from the **Image with text** section which includes _text style_ and _text size_. Let me know if this is okay or if I should remove either _text style_ or _text size_.
- The default language for the caption block is "Add a tagline".
  - Should we change this to something else like, "Add a caption" or "Caption"?

**Block limit removal**
- I've removed the block limit for each block: Heading, Text, Caption, and Button. Looking for feedback on whether we should keep the limit for certain blocks and if so, which ones? I'm perfectly okay with the way it is now but I just think having multiple button blocks _might_ cause confusion or increased cognitive load for buyers. My mental model is one button being associated per rich text section.

**Presets**
- The current presets when adding a new Rich text section includes one of each: Heading, Text, Button
  - Should we include the new caption block?

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -387,6 +387,16 @@ h5,
   word-break: break-word;
 }
 
+.hxl {
+  font-size: calc(var(--font-heading-scale) * 5rem);
+}
+
+@media only screen and (min-width: 750px) {
+  .hxl {
+    font-size: calc(var(--font-heading-scale) * 6.2rem);
+  }
+}
+
 .h0 {
   font-size: calc(var(--font-heading-scale) * 4rem);
 }
@@ -551,6 +561,10 @@ table:not([class]) th {
   .large-up-hide {
     display: none !important;
   }
+}
+
+.left {
+  text-align: left;
 }
 
 .center {
--- a/sections/rich-text.liquid
+++ b/sections/rich-text.liquid
@@ -20,19 +20,27 @@
 
 <div class="isolate{% unless section.settings.full_width %} page-width{% endunless %}">
   <div class="rich-text content-container color-{{ section.settings.color_scheme }} gradient{% if section.settings.full_width %} rich-text--full-width content-container--full-width{% endif %} section-{{ section.id }}-padding">
-    <div class="rich-text__blocks">
-      {%- for block in section.blocks -%}
-        {%- case block.type -%}
-          {%- when 'heading' -%}
-            <h2 class="{{ block.settings.heading_size }}" {{ block.shopify_attributes }}>{{ block.settings.heading | escape }}</h2>
-          {%- when 'text' -%}
-            <div class="rich-text__text rte" {{ block.shopify_attributes }}>{{ block.settings.text }}</div>
-          {%- when 'button' -%}
-            <a{% if block.settings.button_link == blank %} role="link" aria-disabled="true"{% else %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}" {{ block.shopify_attributes }}>
-              {{ block.settings.button_label | escape }}
-            </a>
-        {%- endcase -%}
-      {%- endfor -%}
+    <div class="rich-text__wrapper rich-text__wrapper--{{ section.settings.desktop_content_position }}{% if section.settings.full_width %} page-width{% endif %}">
+      <div class="rich-text__blocks {{ section.settings.content_alignment }}">
+        {%- for block in section.blocks -%}
+          {%- case block.type -%}
+            {%- when 'heading' -%}
+              <h2 class="rich-text__heading rte {{ block.settings.heading_size }}" {{ block.shopify_attributes }}>{{ block.settings.heading }}</h2>
+            {%- when 'caption' -%}
+              <p class="rich-text__caption {{ block.settings.text_style }} {{ block.settings.text_style }}--{{ block.settings.text_size }} {{ block.settings.text_style }}" {{ block.shopify_attributes }}>{{ block.settings.caption }}</p>
+            {%- when 'text' -%}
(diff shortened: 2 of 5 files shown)


assets/base.css
```diff
@@ -553,6 +563,10 @@ table:not([class]) th {
   }
 }
 
+.left {
+  text-align: left;
+}
+
```
andrewetchen, line 569: Although this isn't required (`text-align: left;`) being the default value, it might be used for other elements and for instances where we'd need to override an existing `text-align` value.

sections/rich-text.liquid
```diff
@@ -20,19 +20,27 @@
...
-              {{ block.settings.button_label | escape }}
-            </a>
-        {%- endcase -%}
-      {%- endfor -%}
+    <div class="rich-text__wrapper rich-text__wrapper--{{ section.settings.desktop_content_position }}{% if section.settings.full_width %} page-width{% endif %}">
+      <div class="rich-text__blocks {{ section.settings.content_alignment }}">
+        {%- for block in section.blocks -%}
+          {%- case block.type -%}
+            {%- when 'heading' -%}
+              <h2 class="rich-text__heading rte {{ block.settings.heading_size }}" {{ block.shopify_attributes }}>{{ block.settings.heading }}</h2>
```
andrewetchen, line 28: I'll need to update the `h2` so that it doesn't include `<p>` tags for the **richtext**, to prevent this:

<details>
  <summary>Screenshot</summary>

  image

</details>

Either of the following works but I'll need to confirm if its the solution we want to go with for rich text headings:

```liquid
<h2 class="rich-text__heading rte {{ block.settings.heading_size }}" {{ block.shopify_attributes }}>{{ block.settings.heading | replace: '<p>', '<span>' | replace: '</p>', '</span>' }}</h2>
```

Or

```liquid
<h2 class="rich-text__heading rte {{ block.settings.heading_size }}" {{ block.shopify_attributes }}>{{ block.settings.heading | replace: 'p>', 'span>' }}</h2>
```

<details>
  <summary>Screenshot</summary>

  image

</details>

Confirmed with: https://validator.w3.org/nu/#textarea

sections/rich-text.liquid
```diff
@@ -20,19 +20,27 @@
...
+            {%- when 'text' -%}
+              <div class="rich-text__text rte" {{ block.shopify_attributes }}>{{ block.settings.text }}</div>
+            {%- when 'button' -%}
+              {%- if block.settings.button_label != blank -%}
+                <div class="rich-text__button">
+                  <a{% if block.settings.button_link == blank %} role="link" aria-disabled="true"{% else %} href="{{ block.settings.button_link }}"{% endif %} class="button{% if block.settings.button_style_secondary %} button--secondary{% else %} button--primary{% endif %}" {{ block.shopify_attributes }}>
+                    {{ block.settings.button_label | escape }}
+                  </a>
+                </div>
+              {%- endif -%}
```
andrewetchen, line 40: The `{%- if block.settings.button_label != blank -%}` check was added to prevent this bug: https://github.com/Shopify/dawn/issues/1615.

Additionally, I added a `<div>` to make sure button blocks are placed on new lines instead of being inline.

templates/index.json
```diff
@@ -53,7 +53,7 @@
         "heading": {
           "type": "heading",
           "settings": {
-            "heading": "Talk about your brand",
+            "heading": "<p>Talk about your brand</p>",
```
andrewetchen, line 56: **Related**: https://github.com/Shopify/dawn/pull/1625/files#r852183418

[`richtext` input setting](https://shopify.dev/themes/architecture/settings/input-settings#richtext)

`p` tags must be used to wrap the `default` richtext. It will result in an error if it's not:

**`shopify-cli`**

```ruby
11:12:21 ERROR  » update templates/index.json:
  Setting 'heading' is invalid. All top level nodes must be '<p>' tags
```

**Code editor**

image

andrewetchen:
> The ability to have a setting to set up a position (top left, top top, top right, left middle…) is very very common for many sections, and having this as a select cause lot of duplication of wording.
> 
> i think this may be a new built in setting that Shopify could add, with nice visual icons to convey the different position. We (theme developers) should be able to exclude some choices.
> 
> this setting type could also be used for text positioning (left, venter, right).
> 
> what do you think ? :)

Hey @bakura10,

Regarding the extra positional settings, this would definitely add more flexibility and similarity between some of our sections but unfortunately, I don't think it will be added to the rich text section, at least not for this iteration. I will record your feedback/suggestion - thanks so much!

Issue #1642, opened 2022-04-27
Text box on slide from slideshow has background even when set transparent  due to gradient


ludoboludo:
**Describe the current behavior**

When setting the slide's text box to be transparent, there is actually a background due to the gradient. 

**Describe the expected behavior**

Background should be transparent

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 5.0.0

**Possible solution**

We would most likely need to set a `background: transparent;` on top of the already existing `background-color: transparent;`. 
Since gradients are applied through the `background` property, we would need to set it to transparent for it to work.

**Additional context/screenshots**
![](https://screenshot.click/27-26-7rhpe-li6uz.png)

PR #1643, opened 2022-04-27
Remove gradient when slide container is set to be hidden on desktop

ludoboludo:
**PR Summary:** 

This is a bug fix where if you were using a gradient and setting your slide to have a transparent background on desktop, it would instead show the gradient as the background wasn't set to `transparent`.

Fixes https://github.com/Shopify/dawn/issues/1642

Added `background: transparent;` to the container in CSS as we were resetting the `background-color` but not the `background` which is the property used for the gradients.

**Testing steps/scenarios**
- [ ] Check out the slideshow's slides and make sure the background is transparent when set to while a gradient is also added in the theme settings.

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=127919685654)
- [Editor](https://os2-demo.myshopify.com/admin/themes/127919685654/editor)

**Checklist**

### DIFF
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -190,7 +190,7 @@
   }
 
   .banner:not(.banner--mobile-bottom):not(.email-signup-banner) .banner__box {
-    background-color: transparent;
+    background: transparent;
     --color-foreground: 255, 255, 255;
     --color-button: 255, 255, 255;
     --color-button-text: 0, 0, 0;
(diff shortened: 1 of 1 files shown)


assets/section-image-banner.css
```diff
@@ -317,6 +317,7 @@
 
 @media screen and (min-width: 750px) {
   .banner--desktop-transparent .banner__box {
+    background: transparent;
```
andrewetchen, line 320: I also checked the Image banner section and its working great!

Would there be any downsides to replacing `background-color: transparent;` with `background: transparent;` vs declaring both? I tested this and although it works, I'm wondering if there's a scenario where both are needed 🤔

ludoboludo: Yeah you're right. I can probably just keep the one 🤔

Issue #1641, opened 2022-04-26
Multicolumn text is not aligned on mobile swipe view


katycobb:
**Describe the current behavior**
When multiple column blocks are used in multicolumn section, and the "enable swipe on mobile" is checked, the last column's content alignment is not as expected. The second-to-last column's text spills into the last column.  

**Describe the expected behavior**
The final column's content would be centered, and the second-to-last column's text would not be visible when viewing the final column. 

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 5.0.0
- Chrome Version 100.0.4896.127 
- macOS Version 12.3.1

**Possible solution**

**Additional context/screenshots**
![image](https://user-images.githubusercontent.com/93404168/165347872-eb85802f-3577-4364-8190-d4c0ad63b281.png)

ludoboludo:
This should be fixed with this https://github.com/Shopify/dawn/issues/1641 PR merged 👍 Closing the issue.

PR #1648, opened 2022-04-28
Fix Safari issue with right margin on multicolumn slider

ludoboludo:
**PR Summary:** 

There is an issue in Safari that's been there since the dawn of our theme launching which is the right margin we apply to the last item of the sliders isn't applied: 

In this PR I fix this specific issue but for the multicolumn only. We could look into applying this fix to all sliders but I didn't want to broaden the scope of this fix for now.

Fixes https://github.com/Shopify/dawn/issues/1641

I'm not exactly sure why the spacing at the end of the slider won't work in Safari but found via some forum chats a solution that create a pseudo `:after` element to the `ul` on which we set a width (0.1px) and calculate a margin left based on the grid gap value and the fact that we want the spacing to be `1.5rem`.

I got do some math with `calc()` because the pseudo element is seen as a new flex item so the gap value is applied between the last list item and that element.  

**Other considerations**

I think this could potentially be something we apply across the board for all sliders but since it's not as problematic visually on other sliders I focused on multicolumn. 

**Testing steps/scenarios**
- [ ] Check out the multicolumn slider on mobile specifically on iOS to confirm there is some spacing at the end of the slider
- [ ] Check as well that nothing funky is happening in other browsers on desktop and android
- [ ] Check different screen sizes

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/assets/section-multicolumn.css
+++ b/assets/section-multicolumn.css
@@ -108,8 +108,18 @@
     padding-left: 1.5rem;
     padding-right: 1.5rem;
   }
-}
 
+  .multicolumn-list.slider:after {
+    /* Fix for margin right not showing on slider for Safari */
+    content: "";
+    flex: 0 0 0.1px;
+    margin-left: calc(1.5rem - var(--grid-mobile-horizontal-spacing));
+  }
+
+  .multicolumn-list.grid--peek.slider .grid__item:last-of-type {
+    margin-right: 0;
+  }
+}
 
 @media screen and (min-width: 750px) {
   .multicolumn-list.slider,


YoannJailin:
It's perfect thank you!

YoannJailin (dismissed):

ludoboludo:
@bakura10 thanks a lot for all this info and link to the issues related to it. When I test the fix on chrome (laptop and android phone) it seem to work as expected and same for firefox on my laptop. Ill look into the issue you created and add a comment to it 👍

ludoboludo:
[In this PR](https://github.com/Shopify/dawn/pull/1669) I'm actually adding this fix for all sliders. So Ill probably close this PR

Issue #1445, opened 2022-03-03
[Slider-Component] No white space after last element of slider component on some browsers (introduced with Dawn 4.0.0)


bredowmax:
EDIT: Hotfix `.slider::after { content: ''; min-width: 1px; }` in `component-slider.css` solved it for me, for now.

When viewing a clean Dawn 4.0.0 theme on a physical iPhone, the last element of any section with mobile slider (in this case: a Multicolumn section) appears too much to the right - as if it doesn't have any padding to the right. I tested on an iPhoner 13 Pro with both latest Chrome & Safari. Meanwhile, I couldn't replicate the issue in the Theme Editor or the Chrome Inspector tool. Also, I do not experience this issue on a clean Dawn 3.0.0 theme.

**Describe the current behavior**
<img src="https://user-images.githubusercontent.com/30573532/156622267-2c97efa7-8235-4fc8-9143-6168ef36dd3e.PNG" width="390" height="844">
Screenshot from Dawn 4.0.0

This seems to be an issue introduced lately: Downloading the latest 4.0.0 version of Dawn on Github, you can replicate this issue on an iPhone on both Chrome & Safari. But in the currently published 3.0.0 version of Dawn, the mobile slider behaves as expected

**Describe the expected behavior**
<img src="https://user-images.githubusercontent.com/30573532/156622275-d4c0a99e-05b2-4ec4-a23d-b059b0be4e42.PNG" width="390" height="844">
Screenshot from Dawn 3.0.0

bredowmax:
Further testing: this screenshot makes it more obvious
![IMG_2184](https://user-images.githubusercontent.com/30573532/156623057-a3bf30cf-6776-4ebc-a0c7-304709c27a0b.PNG)

bredowmax:
Meanwhile, the same issue does NOT appear when in the Theme Editor on MacOS / Chrome / Mobile View

<img width="797" alt="Screen Shot 2022-03-03 at 12 53 09 PM" src="https://user-images.githubusercontent.com/30573532/156623087-01d59b0f-d35e-4cda-8994-0070e0a6b774.png">

bredowmax:
Further testing: _cannot_ replicate the issue on Chrome / MacOS when using the "Inspect" tool & simulating a mobile device

[...]

Issue #1668, opened 2022-05-06
Slider count issue on product page


ludoboludo:
**Describe the current behavior**
On the product page, the slider can show a wrong item count. 

**Describe the expected behavior**

It should be displaying the right amount of slides

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 5.0.0

**Possible solution**
Change the selector in CSS that applies a padding right to the first image if it's an image linked to a variant. 

**Additional context/screenshots**
[video](https://screenshot.click/06-37-4w334-797s0.mp4)

PR #1669, opened 2022-05-06
Fix slider padding and whitespace issue on Safari

ludoboludo:
**PR Summary:** 

This is a PR that fixes a couple visual issues: 
- in Safari there is a bug where the space applied at the end of our sliders isn't coming up while it's fine in Chrome and Firefox
- Extra padding is applied on the first image tied to a variant which create extra space that isn't necessary: [video recording](https://screenshot.click/08-20-t1btv-29ta0.mp4)

Fixes https://github.com/Shopify/dawn/issues/1445 and https://github.com/Shopify/dawn/issues/1641

I'm applying a visual fix so that sliders look nicer in Safari and match what's coming up in Chrome and Firefox. 
This also help with the removal of some extra space added to the first image linked to a variant. 

**Other considerations**

New issue found. When you have a hidden variant as the last child. On mobile the slider's last visible item won't have spacing and be touching the edge of the screen.  But can be tackled in another PR with the similar approach we have for the spacing fix on Safari: https://github.com/Shopify/dawn/pull/1648

**Testing steps/scenarios**

- [ ] Test all sliders in different browsers. Make sure the spacing is good and the counting as well. There is another PR fixing the counter issues -> https://github.com/Shopify/dawn/pull/1696
- [ ] Test different breakpoints 
- [ ] Test product with only variant images
- [ ] test products with only one image

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128028409878)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128028409878/editor)

**Checklist**

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1141,6 +1141,7 @@ summary::-webkit-details-marker {
     margin-left: 1.5rem;
   }
 
+  /* Fix to show some space at the end of our sliders in all browsers */
   .grid--peek.slider:after {
     margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));
   }
--- a/assets/base.css
+++ b/assets/base.css
@@ -1141,8 +1141,8 @@ summary::-webkit-details-marker {
     margin-left: 1.5rem;
   }
 
-  .grid--peek.slider .grid__item:last-of-type {
-    margin-right: calc(-1 * var(--grid-mobile-horizontal-spacing));
+  .grid--peek.slider:after {
+    margin-left: calc(-1 * var(--grid-mobile-horizontal-spacing));
   }
 
   .grid--2-col-tablet-down .grid__item {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1142,7 +1142,7 @@ summary::-webkit-details-marker {
   }
 
   .grid--peek.slider .grid__item:last-of-type {
-    margin-right: 1.5rem;
+    margin-right: calc(-1 * var(--grid-mobile-horizontal-spacing));
   }
 
   .grid--2-col-tablet-down .grid__item {
@@ -1176,10 +1176,6 @@ summary::-webkit-details-marker {
 
   .slider--tablet.grid--peek .grid__item:first-of-type {
     margin-left: 1.5rem;
-  }
-
-  .slider--tablet.grid--peek .grid__item:last-of-type {
-    margin-right: 1.5rem;
   }
 
   .grid--2-col-tablet-down .grid__item {
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -100,10 +100,7 @@ slider-component.slider-component-full-width {
     content: "";
     width: 0;
     padding-left: 1.5rem;
-  }
-
-  .slider.slider--tablet .slider__slide:last-child {
-    margin-right: calc(-2 * var(--grid-mobile-horizontal-spacing));
+    margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing));
   }
 
   .slider.slider--tablet .slider__slide {
@@ -118,12 +115,6 @@ slider-component.slider-component-full-width {
 
   .slider.slider--tablet.contains-content-container .slider__slide {
     --focus-outline-padding: 0rem;
-  }
-}
-
-@media screen and (max-width: 749px) {
-  .slider.slider--tablet .slider__slide:last-child {
-    margin-right: calc(-1 * var(--grid-mobile-horizontal-spacing));
   }
 }
 
@@ -162,6 +153,7 @@ slider-component.slider-component-full-width {
     content: "";
     width: 0;
     padding-left: 5rem;
+    margin-left: calc(-1 * var(--grid-desktop-horizontal-spacing));
   }
 
   .slider.slider--desktop .slider__slide {
@@ -173,10 +165,6 @@ slider-component.slider-component-full-width {
   .slider--desktop .slider__slide:first-child {
     margin-left: var(--desktop-margin-left-first-item);
     scroll-margin-left: var(--desktop-margin-left-first-item);
-  }
-
-  .slider.slider--desktop .slider__slide:last-child {
-    margin-right: calc(-1 * var(--grid-desktop-horizontal-spacing));
   }
 
   .slider-component-full-width .slider--desktop {
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -47,6 +47,12 @@ slider-component.slider-component-full-width {
     margin-bottom: 1rem;
   }
 
+  .slider--mobile:after {
+    content: "";
+    width: 0px;
+    padding-left: 1.5rem;
+  }
+
   .slider.slider--mobile .slider__slide {
     margin-bottom: 0;
     padding-top: max(var(--focus-outline-padding), var(--shadow-padding-top));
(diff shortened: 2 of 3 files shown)


kmeleta (commented):
I think an open PR of mine will fix this, but if we end up with a nicer fix in this PR or want to cherrypick the code from mine to get this resolved independently, by all means. Here's the relevant commit from that PR https://github.com/Shopify/dawn/pull/1478/commits/62497547642baa025003f1d53146034574718a42

assets/section-main-product.css
```diff
@@ -533,9 +533,13 @@ a.product__text {
 }
 
 @media screen and (max-width: 749px) {
-  .product__media-item--variant:first-child {
+  .product__media-item--variant:nth-last-child(2):is(.product__media-item--variant) {
```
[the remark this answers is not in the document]
ludoboludo: And I was just thinking. If we apply the fix for the spacing on sliders for Safari, then that would fix this issue as well. 
And I think it might make sense to apply it to all sliders 🤔 
The fix is adding a pseudo element to the `ul` to add the spacing at the end of the slider. https://github.com/Shopify/dawn/pull/1648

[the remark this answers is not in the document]
ludoboludo: I made some changes to fix all sliders actually so they get the whitespace at the end in Safari as well as it's fixing this specific issue as well: https://github.com/Shopify/dawn/issues/1668

I haven't test a whole a lot but it seems to work from what I could tell. 

The number of slides in the counter seem to be off when we're using big values for the grid horizontal spacing but I can replicate them on main as well 🤔 (different scenarios tho)  So I don't think it's introduced in here.

ludoboludo:
I had created a specific fix for multicolumn here: https://github.com/Shopify/dawn/pull/1648 But if we're going with this PR instead it fixes all instances of sliders.

YoannJailin:
All sliders seems good to me on Safari. The ending on each of them matches the grid.

assets/base.css
```diff
@@ -1142,7 +1142,7 @@ summary::-webkit-details-marker {
   }
 
   .grid--peek.slider .grid__item:last-of-type {
-    margin-right: 1.5rem;
+    margin-right: calc(-1 * var(--grid-mobile-horizontal-spacing));
```
LucasLacerdaUX, line 1145: I haven't investigated yet to see which line is causing this issue, but the thumbnail slider now looks broken. I wonder if it's because we changed from a fixed margin to a variable, while the thumbnails are not using this variable. Will share an update if I find it more details.

Videos below. As you can see, the scrolling is really off when it reaches the last slide.

**Before:**

https://user-images.githubusercontent.com/37168033/169359718-b42b6415-536d-44a9-a97d-f74b19216bd9.mp4

**After**

https://user-images.githubusercontent.com/37168033/169359728-49659935-b691-4381-9e2d-4e46e2ee2aba.mp4

LucasLacerdaUX (commented):
This fix seems to break thumbnail slider on both Chrome and Safari. I don't have a lot of details yet on the cause, but shared a video below.

assets/base.css
ludoboludo: Got this fixed. It was because I was applying the fix to it when it's actually not needed in that context. So I'm overwriting it now in `section-main-product.css` with: 
```css
.slider--mobile.thumbnail-list:after {
  content: none;
}
```

assets/component-slider.css
```diff
@@ -47,6 +47,12 @@ slider-component.slider-component-full-width {
     margin-bottom: 1rem;
   }
 
+  .slider--mobile:after {
+    content: "";
+    width: 0;
+    padding-left: 1.5rem;
```
kmeleta, line 53: This doesn't entirely fix the single-visible variant case. It seems to be receiving the gap value (I imagine because the are technically more items in the grid) and the 1.5rem from the pseudo padding.

kmeleta (commented):
Seems to fix the safari specific weirdness, but the single-visible variant case still has issues.

assets/component-slider.css
ludoboludo: ill double check on it. It's supposed to calculate the spacing based on gap value too 🤔

assets/component-slider.css
```diff
@@ -160,8 +176,8 @@ slider-component.slider-component-full-width {
     scroll-margin-left: 1.5rem;
   }
 
-  .slider-component-full-width .slider--desktop .slider__slide:last-child {
-    margin-right: 1.5rem;
+  .slider-component-full-width .slider--desktop:after {
+    padding-left: 1.5rem;
```
kmeleta, line 180: Is there any reason we couldn't use either margin OR padding for this? Using both makes me think they serve different purposes. Do they? I'd slightly prefer if we could just account for the 1.5rem in the same calculation as the negative spacing offset rather than relying on one to counteract the other.

assets/component-slider.css
```diff
@@ -90,6 +96,13 @@ slider-component.slider-component-full-width {
     margin-bottom: 1rem;
   }
 
+  .slider--tablet:after {
```
kmeleta, line 99: I even know what we're doing here (kind of) and these rules confuse me. Can we add a comment above this selector that explains what this is for?

assets/component-slider.css
```diff
@@ -136,6 +149,13 @@ slider-component.slider-component-full-width {
     scroll-padding-left: var(--desktop-margin-left-first-item);
   }
 
+  .slider--desktop:after {
```
kmeleta, line 152: A little redundant but I think it's worth duplicating the comment for each of these main `slider--mobile/tablet/desktop:after` instances. Or else group them together below a common comment. Lot of this stuff is shared between them anyway.

This sort of unidiomatic workaround is a shining example of the kind of thing where comments are 100% justified as far as I'm concerned.

assets/section-main-product.css
```diff
@@ -974,6 +968,10 @@ a.product__text {
   grid-gap: 1rem;
 }
 
+.slider--mobile.thumbnail-list:after {
```
kmeleta, line 971: Being in another file I think this should have a comment too.

kmeleta (commented):
Haven't managed to break it with the latest changes so far. Couple minor notes though.

assets/component-slider.css
ludoboludo: So the thing by adding a pseudo `:after` element is that it will add some spacing using the gap like if it was another `li` element. I need the `padding` value to be positive in order for that pseudo element to be showing.

Now due to the `gap` value coming up, I want to cancel it by adding a negative value into the mix. I'm using the padding to set the value of whatever we use for `.page-width` and the margin to cancel whatever the gap value is. 
I can't set a negative value on padding so I can't use that property alone. And margin doesn't get the pseudo element to come up visually so I can't use that alone either. 
Seems to be the best way to go about it from what I tested 🤔

assets/component-slider.css
ludoboludo: Added some comments, let me know if they're clear enough 🦖

assets/section-main-product.css
```diff
@@ -974,6 +968,11 @@ a.product__text {
   grid-gap: 1rem;
 }
 
+/* Prevent fix to show some space at the end of our sliders in all browsers to be applied on thumbnails */
```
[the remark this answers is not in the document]
ludoboludo: 😅 Yeah I can remove it 👍

metamoni (approved):

assets/section-main-product.css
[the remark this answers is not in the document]
kmeleta: So we don't have to go back and correct this 😅  but I actually think the "prevent" language was intended, as this instance overrides/prevents the fix applied globally to the .slider classes, doesn't it? But either way, the comment adequately denotes that this rule has a purpose so either way 👍

kmeleta (approved):
Looks good from my testing. Thanks for the comments :shipit:

PR #1703, opened 2022-05-17
Fix syntax error caused by missing calc in product CSS

boltgolt:
**PR Summary:** 

Fix CSS syntax error.

There was a syntax error in `assets/section-main-product.css` where a border-radius calculation was done outside of a calc() function. This was accidentally introduced in #1124 and is done correctly in other places in the file.

When running a CSS linter this was the only error it found in our fork of Dawn. As it was a simple fix i thought it was worth fixing in upstream too.

**Other considerations**

Visually this changes very little, only a small non-inherited change to the border-radius on focus. I've looked at the other PRs and as far as i can tell none of them make a change to this file, so i think this PR is not a duplicate.

**Checklist**
- [X] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -366,7 +366,7 @@ a.product__text {
   .product__info-container {
     max-width: 60rem;
   }
-
+  
   .product__info-container .price--on-sale .price-item--regular {
     font-size: 1.6rem;
   }
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -651,8 +651,12 @@ a.product__text {
   box-shadow: none;
 }
 
-.product__media-toggle.focused:after,
 .product__media-toggle:focus-visible:after {
+  box-shadow: 0 0 0 0.3rem rgb(var(--color-background)), 0 0 0rem 0.5rem rgba(var(--color-foreground), 0.5);
+  border-radius: var(--media-radius);
+}
+
+.product__media-toggle.focused:after {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)), 0 0 0rem 0.5rem rgba(var(--color-foreground), 0.5);
   border-radius: var(--media-radius);
 }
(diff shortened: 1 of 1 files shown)


assets/section-main-product.css
```diff
@@ -653,7 +653,7 @@ a.product__text {
 
 .product__media-toggle:focus-visible:after {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0rem 0.5rem rgba(var(--color-foreground), 0.5);
-  border-radius: var(--media-radius) - var(--media-border-width);
+  border-radius: calc(var(--media-radius) - var(--media-border-width));
```
ludoboludo, line 656: it's actually overwritten below on line 665 and from what I can tell it looks better with the border radius that matches the `media-radius` only 🤔 
[video](https://screenshot.click/17-21-ppgtj-8mcd5.mp4)

I'd suggest updating to:
```suggestion
  border-radius:var(--media-radius);
```

And then removing the declaration that's on line `664-666`

boltgolt: Thanks for the feedback and the clear video!

I agree that having just the `--media-radius` as radius looks a lot better. After removing the duplicate `product__media-toggle:focus-visible:after` declaration we're left with two blocks with exactly the same properties:

image

I've now pushed a commit implementing your suggestions and merging these blocks into one. For consistency maybe it's better to also merge the two identical blocks above it?

image

ludoboludo: Ah actually we can't have the two grouped together. It's not obvious at first but basically we're using the `.focused` class specifically for browsers that don't support `:focus-visible` yet. 
Safari only supports it in the latest `15.4` version and we're supporting the last 2 versions (`14` and `15`). 

When grouped what happens is, the browser sees the pseudo class and doesn't `understand` it so it skips the whole declaration event if the other part of it should be `understood`. 

All that to say that we didn't group them on purpose for browser support 🙂

boltgolt: Aah that makes sense, reverted that change now

LucasLacerdaUX (approved):
Thanks for making this contribution to Dawn, @boltgolt. 

Looks good!

PR #1720, opened 2022-05-20
[Product page] Add desktop media position

andrewetchen:
_Work in progress_ 🚧

**PR Summary**

Fixes https://github.com/Shopify/dawn/issues/1690

**Other considerations**

**Testing steps/scenarios**

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -1059,6 +1059,23 @@
     },
     {
       "type": "select",
+      "id": "media_position",
+      "options": [
+        {
+          "value": "left",
+          "label": "t:sections.main-product.settings.media_position.options__1.label"
+        },
+        {
+          "value": "right",
+          "label": "t:sections.main-product.settings.media_position.options__2.label"
+        }
+      ],
+      "default": "left",
+      "label": "t:sections.main-product.settings.media_position.label",
+      "info": "t:sections.main-product.settings.media_position.info"
+    },
+    {
+      "type": "select",
       "id": "mobile_thumbnails",
       "options": [
         {
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1817,6 +1817,16 @@
             "label": "Large"
           }
         },
+        "media_position": {
+          "label": "Desktop media position",
+          "info": "Position is automatically optimized for mobile.",
+          "options__1": {
+            "label": "Left"
+          },
+          "options__2": {
+            "label": "Right"
+          }
+        },
         "mobile_thumbnails": {
           "label": "Mobile layout",
           "options__1": {


andrewetchen:
> Looks great and works as expected! The focus order is off when the image is on the right side, but I am not sure if that is considered a problem? [Video](https://screenshot.click/06-07-gkeh6-sr3zh.mp4).

Thanks for your review, @Oliviammarcello and thanks for pointing that out as well.

That was on my radar when I was first looking into this. I think the "Skip to product information" link that's currently on the first media image is a good solution though. Another solution would be to move the media wrapper code into a snippet so the DOM order (and tab/focus order) wouldn't be affected between the `left` and `right` desktop media position. The issue with this approach is that it would add more complexity to the code since the featured product section's media wrapper is slightly different than the one on the product page. We'd also have to add some logic to differentiate between desktop and mobile which could also further complicate things. That being said, I'm totally open to exploring this route if needed.

Oliviammarcello:
> > Looks great and works as expected! The focus order is off when the image is on the right side, but I am not sure if that is considered a problem? [Video](https://screenshot.click/06-07-gkeh6-sr3zh.mp4).
> 
> Thanks for your review, @Oliviammarcello and thanks for pointing that out as well.
> 
> That was on my radar when I was first looking into this. I think the "Skip to product information" link that's currently on the first media image is a good solution though. Another solution would be to move the media wrapper code into a snippet so the DOM order (and tab/focus order) wouldn't be affected between the `left` and `right` desktop media position. The issue with this approach is that it would add more complexity to the code since the featured product section's media wrapper is slightly different than the one on the product page. We'd also have to add some logic to differentiate between desktop and mobile which could also further complicate things. That being said, I'm totally open to exploring this route if needed.

Good point about "Skip to product information" I'd assume that is sufficient.

Oliviammarcello (approved):

assets/section-main-product.css
```diff
@@ -11,13 +11,18 @@
   margin: 0 auto;
 }
 
+@media screen and (min-width: 750px) {
+  .product--reverse {
+    flex-direction: row-reverse;
+  }
+}
+
```
ludoboludo, line 19: I guess it'd be good to loop in @svinkle to confirm it's ok. 
Context: 
>We're adding a setting to choose whether you want your product media showing first on desktop (so to the left like it's been so far) or second (to the right).
When using `flex-direction: row-reverse;` it means that when tabbing through you're going to land on the images/medias first, whatever the order chosen is. 
[Here is Olivia's video that shows the tabbing order](https://screenshot.click/06-07-gkeh6-sr3zh.mp4)

A word around would be to look into using grid and assigning the two different blocks to a specific column. 🤔

ludoboludo (commented):
Looking good 👍 
Left some nitpick comments and I'm curious to hear from Scott to see if the tab order should be something we worry about or not 🙂

assets/section-main-product.css
andrewetchen: Thanks for mentioning Scott.

As we wait for his feedback, do you happen to have more context on why the "Skip to product information" was originally added? See the video in my PR description > [Accessibility](https://github.com/Shopify/dawn/pull/1720#:~:text=keep%20things%20simple.-,Accessibility,-For%20accessibility%20) but I couldn't find any prior issues etc.

assets/section-main-product.css
```diff
@@ -11,13 +11,18 @@
   margin: 0 auto;
 }
 
+@media screen and (min-width: 750px) {
+  .product--reverse {
+    flex-direction: row-reverse;
+  }
+}
```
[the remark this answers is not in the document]
andrewetchen: >I got not strong feelings for this...

Same tbh. The main reasons why I chose to place this here were because 1) there are currently 10 instances of `@media screen and (min-width: 750px) {` within this stylesheet and thought that I really wouldn't be improving things much by doing this (unless I did it for all `@media screen and (min-width: 750px) {` declarations. 2) The current [cascading order of this](https://github.com/Shopify/dawn/blob/c91b37fcb819542350cf4f9be7bd5d6a70afc34c/assets/section-main-product.css#L1-L27) perfectly matches the order of elements in `main-product.liquid` and I didn't want to "break" this. 

I'm totally cool to reposition this under line 28 but I feel like we can improve things a lot more in this stylesheet -- something that could make more sense to tackle later as a refactor perhaps (calibration)?

assets/section-main-product.css
```diff
@@ -11,13 +11,18 @@
...
+
 .product__media-wrapper {
   padding-left: 0;
 }
 
 .product__info-wrapper {
-  padding-left: 0;
-  padding-bottom: 0;
+  padding: 0;
 }
```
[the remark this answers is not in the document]
ludoboludo: No context 🤔 I think it's been added a while back. Probably needed at the time 🤷 
But it didn't seem used/necessary while testing. Probably safe to remove

assets/section-main-product.css
ludoboludo: K so it seems that ideally we wouldn't rely on flex's `row-reverse` and keep the natural tab order. 

So I'm guessing we could capture the content of one of the "column" into a `capture` and output it either before or after the other column based on the setting. 

```html
{% capture media_column %}
  <div class="grid__item product__media-wrapper">
    ...
```

andrewetchen: Thanks Ludo. I'm using the `capture` method you suggested (running it locally) and it's working great. Since this is a "Desktop only" enhancement, do you have offhand suggestions on how to maintain the "image on top" / have the media wrapper "on top" on mobile? I could be missing something that's probably really simple 🤔 

https://user-images.githubusercontent.com/28404165/173429965-a77b42db-53a5-4967-bea4-d82e4ea95188.mp4

assets/section-main-product.css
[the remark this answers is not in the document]
andrewetchen: Removed here https://github.com/Shopify/dawn/pull/1720/commits/1a3a80c646f5429b89ca2ccaa773c2cb695ad2b1

assets/section-main-product.css
ludoboludo: Hmm then I guess it wouldn't be with a capture 🤔 But duplicate code. Always keep the original one and apply a hiding class on desktop if the medias aren't set to come up first/on the left. 

And add `small-hide` to the block containing the medias when set to be on the right. That's going to make it a big file 😬 
I wonder if this could create some issues in terms of IDs and behaviours (showing the image variant when it's being selected, etc).

sections/main-product.liquid
```diff
@@ -31,7 +31,7 @@
     <link id="ModelViewerOverride" rel="stylesheet" href="{{ 'component-model-viewer-ui.css' | asset_url }}" media="print" onload="this.media='all'">
   {%- endif -%}
 
-  <div class="product product--{{ section.settings.media_size }} product--{{ section.settings.gallery_layout }} grid grid--1-col {% if product.media.size > 0 %}grid--2-col-tablet{% else %}product--no-media{% endif %}">
+  <div class="product product--{{ section.settings.media_size }} product--{{ section.settings.gallery_layout }} grid grid--1-col {% if product.media.size > 0 %}grid--2-col-tablet{% else %}product--no-media{% endif %}{% if section.settings.media_position == 'right' %} product--reverse{% endif %}">
     <div class="grid__item product__media-wrapper">
       <media-gallery id="MediaGallery-{{ section.id }}" role="region" {% if section.settings.enable_sticky_info %}class="product__media-gallery"{% endif %} aria-label="{{ 'products.product.media.gallery_viewer' | t }}" data-desktop-layout="{{ section.settings.gallery_layout }}">
         <div id="GalleryStatus-{{ section.id }}" class="visually-hidden" role="status"></div>
```
ludoboludo, line 37: Scott had mentioned as well that we could remove the skip link when the media column is to the right. 
So this would only come up when medias are set to be on the left: 
```
<a class="skip-to-content-link button visually-hidden quick-add-hidden" href="#ProductInfo-{{ section.id }}">
  {{ "accessibility.skip_to_product_info" | t }}
</a>
```

andrewetchen: Thanks - I did see your convo with Scott and I agree that removing the skip to link seems like the _right_ thing to do 😉

andrewetchen:
> @andrewetchen, any updated on this PR's merge?

Hey @bacounts, I'm hoping to get this PR merge by the end of next week 🤞🏼

assets/section-main-product.css
```diff
@@ -11,13 +11,14 @@
   margin: 0 auto;
 }
 
-.product__media-wrapper {
-  padding-left: 0;
+@media screen and (min-width: 750px) {
+  .product--reverse {
+    flex-direction: row-reverse;
```
LucasLacerdaUX, line 16: Have we validated with accessibility folks if there's any issue with this approach as it changes the tab order? I tested and it seems to works fine, but just wanted to know if there's any conversation I'm missing.

I like that we have `Skip to product information` always there so it's never an issue to tab through items :)

LucasLacerdaUX (approved):
Tested on:
- [x] Product Page
- [x] Featured Product
- [x] Quick Add

They all seem to work as expected! Spacing looks good too.

I just wanted to confirm and get more context on the accessibility discussion, to make sure we've validated that this tab order changes won't cause issues for certain users :D

assets/section-main-product.css
andrewetchen: That's a good point. I spoke to Scott about this and confirmed that we shouldn't be using `flex-direction: row-reverse;` for this type of layout option change in order to maintain our accessibility standards + have the correct DOM / tab order.

**TL;DR**
The "Skip to product information" link will remain when using the default `left` media position and will not be shown when using the `right` layout.

assets/section-main-product.css
andrewetchen: Thanks Ludo. I've added some more changes based on your suggestions here. Let me know your thoughts / feedback.

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
         </a>
       </div>
     </div>
+    {%- if section.settings.media_position == 'right' -%}
```
LucasLacerdaUX, line 507: I think we'll need to follow this same approach on `featured-product` now that we removed `product--reverse` class. Otherwise, the setting won't do anything.

Oliviammarcello (dismissed):
Focus states all look great and work as expected 👍

andrewetchen:
I've signed the CLA!

andrewetchen:
I've signed the CLA!

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
...
+              data-shopify-model3d-id="{{ first_3d_model.id }}"
+              data-shopify-title="{{ product.title | escape }}"
+              data-shopify-xr-hidden
+            >
+              {% render 'icon-3d-model' %}
+              {{ 'products.product.xr_button' | t }}
+            </button>
+          {%- endif -%}
+          {%- if media_count > 1 and section.settings.gallery_layout != 'stacked' or section.settings.mobile_thumbnails == 'show' -%}
+            <slider-component id="GalleryThumbnails-{{ section.id }}" class="thumbnail-slider slider-mobile-gutter quick-add-hidden{% if section.settings.mobile_thumbnails == 'hide' %} small-hide{% endif %}{% if media_count <= 3 %} thumbnail-slider--no-slide{% endif %}">
```
LucasLacerdaUX, line 576:
```suggestion
            <slider-component id="GalleryThumbnails-{{ section.id }}-right" class="thumbnail-slider slider-mobile-gutter quick-add-hidden{% if section.settings.mobile_thumbnails == 'hide' %} small-hide{% endif %}{% if media_count <= 3 %} thumbnail-slider--no-slide{% endif %}">
```

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
...
+                  -%}
+                  <li id="Slide-Thumbnails-{{ section.id }}-0" class="thumbnail-list__item slider__slide{% if section.settings.hide_variants and variant_images contains featured_media.src %} thumbnail-list_item--variant{% endif %}" data-target="{{ section.id }}-{{ featured_media.id }}" data-media-position="{{ media_index }}">
+                    <button class="thumbnail global-media-settings global-media-settings--no-shadow {% if featured_media.preview_image.aspect_ratio > 1 %}thumbnail--wide{% else %}thumbnail--narrow{% endif %}"
+                      aria-label="{%- if featured_media.media_type == 'image' -%}{{ 'products.product.media.load_image' | t: index: media_index }}{%- elsif featured_media.media_type == 'model' -%}{{ 'products.product.media.load_model' | t: index: media_index }}{%- elsif featured_media.media_type == 'video' or featured_media.media_type == 'external_video' -%}{{ 'products.product.media.load_video' | t: index: media_index }}{%- endif -%}"
+                      aria-current="true"
+                      aria-controls="GalleryViewer-{{ section.id }}"
+                      aria-describedby="Thumbnail-{{ section.id }}-0"
```
LucasLacerdaUX, line 597:
```suggestion
                      aria-controls="GalleryViewer-{{ section.id }}-right"
                      aria-describedby="Thumbnail-{{ section.id }}-0-right"
```

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
...
+                    >
+                      <img id="Thumbnail-{{ section.id }}-0" srcset="{% if featured_media.preview_image.width >= 54 %}{{ featured_media.preview_image | image_url: width: 54 }} 54w,{% endif %}
```
LucasLacerdaUX, line 599:
```suggestion
                      <img id="Thumbnail-{{ section.id }}-0-right" srcset="{% if featured_media.preview_image.width >= 54 %}{{ featured_media.preview_image | image_url: width: 54 }} 54w,{% endif %}
```

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
...
+                          increment model_index
+                        elsif media.media_type == 'video' or media.media_type == 'external_video'
+                          increment video_index
+                        elsif media.media_type == 'image'
+                          increment image_index
+                        endif
+                      endcapture
+                      assign media_index = media_index | plus: 1
+                    -%}
+                    <li id="Slide-Thumbnails-{{ section.id }}-{{ forloop.index }}" class="thumbnail-list__item slider__slide{% if section.settings.hide_variants and variant_images contains media.src %} thumbnail-list_item--variant{% endif %}" data-target="{{ section.id }}-{{ media.id }}"  data-media-position="{{ media_index }}">
```
LucasLacerdaUX, line 631:
```suggestion
                    <li id="Slide-Thumbnails-{{ section.id }}-{{ forloop.index }}-right" class="thumbnail-list__item slider__slide{% if section.settings.hide_variants and variant_images contains media.src %} thumbnail-list_item--variant{% endif %}" data-target="{{ section.id }}-{{ media.id }}"  data-media-position="{{ media_index }}">
```

sections/main-product.liquid
```diff
@@ -504,6 +504,171 @@
...
+                        <span class="thumbnail__badge" aria-hidden="true">
+                          {%- render 'icon-play' -%}
+                        </span>
+                      {%- endif -%}
+                      <button class="thumbnail global-media-settings global-media-settings--no-shadow {% if media.preview_image.aspect_ratio > 1 %}thumbnail--wide{% else %}thumbnail--narrow{% endif %}"
+                        aria-label="{%- if media.media_type == 'image' -%}{{ 'products.product.media.load_image' | t: index: media_index }}{%- elsif media.media_type == 'model' -%}{{ 'products.product.media.load_model' | t: index: media_index }}{%- elsif media.media_type == 'video' or media.media_type == 'external_video' -%}{{ 'products.product.media.load_video' | t: index: media_index }}{%- endif -%}"
+                        {% if media == product.selected_or_first_available_variant.featured_media or product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} aria-current="true"{% endif %}
+                        aria-controls="GalleryViewer-{{ section.id }}"
+                        aria-describedby="Thumbnail-{{ section.id }}-{{ forloop.index }}"
```
LucasLacerdaUX, line 645:
```suggestion
                        aria-controls="GalleryViewer-{{ section.id }}-right"
                        aria-describedby="Thumbnail-{{ section.id }}-{{ forloop.index }}-right"
```

LucasLacerdaUX (commented):
We shouldn't have any duplicate IDs. 

While most of them are not actually used to target an specific element (most of our components use a "starts with" approach (`[id^="Gallery-"]) for querying), it is still a bad practice that can lead to errors. 

Some of them are being associated with `aria-describedby`, `aria-labelledby` so these IDs should either be changed to have unique or we should remove the "copy" and rely solely on the original element.

I've highlighted most, if not all, elements with duplicated IDs below. Let me know if you have any questions and we can pair to go through this list :D

sections/main-product.liquid
andrewetchen: Thanks Lucas, as I mentioned in our chat, I also added `-right` the `img` tag below ^this on line 647 as I noticed [this using the WAVE tool](https://user-images.githubusercontent.com/28404165/180304540-29e95f37-7841-47b7-a9b0-b887a8fc0432.png)

sections/featured-product.liquid
```diff
@@ -343,6 +340,47 @@
...
+                <div class="product__media-item" data-media-id="{{ section.id }}-{{ media.id }}">
+                  {% render 'product-thumbnail', media: media, position: 'featured', loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: false, media_width: 0.5 %}
+                </div>
+              {%- endif -%}
+              {%- liquid
+                assign media_to_render = product.featured_media.id
+                for variant in product.variants
+                  assign media_to_render = media_to_render | append: variant.featured_media.id | append: ' '
+                endfor
+              -%}
```
ludoboludo, line 358: I think this might not be needed twice as well. Since it's already done above on line 45-50

ludoboludo (commented):
Noted a couple things but the rest looks good to me.

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+                  {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
+                    {%- liquid
+                      capture media_index
+                        if media.media_type == 'model'
+                          increment model_index
+                        elsif media.media_type == 'video' or media.media_type == 'external_video'
+                          increment video_index
+                        elsif media.media_type == 'image'
+                          increment image_index
+                        endif
```
LucasLacerdaUX, line 868: These need to be separate variables, or it will continue from the number assigned before.

If we look at `data-media-position` for the right column, it starts at `8`:
image

Reassigning the variable to `0` won't work, because variables created using increment are independent from the ones created using `assign`.

```suggestion
                        if media.media_type == 'model'
                          increment model_index_right
                        elsif media.media_type == 'video' or media.media_type == 'external_video'
                          increment video_index_right
                        elsif media.media_type == 'image'
                          increment image_index_right
                        endif
```

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+                  {%- liquid
+                    capture media_index
+                      if featured_media.media_type == 'model'
+                        increment model_index
+                      elsif featured_media.media_type == 'video' or featured_media.media_type == 'external_video'
+                        increment video_index
+                      elsif featured_media.media_type == 'image'
+                        increment image_index
+                      endif
+                    endcapture
```
LucasLacerdaUX, line 820:
```suggestion
                    capture media_index
                      if featured_media.media_type == 'model'
                        increment model_index_right
                      elsif featured_media.media_type == 'video' or featured_media.media_type == 'external_video'
                        increment video_index_right
                      elsif featured_media.media_type == 'image'
                        increment image_index_right
                      endif
                    endcapture
```

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+              <button
+                type="button"
+                class="slider-button slider-button--next"
+                name="next"
+                aria-label="{{ 'general.slider.next_slide' | t }}"
+              >
+                {% render 'icon-caret' %}
+              </button>
+            </div>
+          </slider-component>
```
LucasLacerdaUX, line 776: Do we need this? This is never going to be a slider on mobile.

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+              {%- endif -%}
+              {%- for media in product.media -%}
+                {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
+                  <li
+                    id="Slide-{{ section.id }}-{{ media.id }}-right"
+                    class="product__media-item grid__item slider__slide{% if product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} is-active{% endif %}{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}"
+                    data-media-id="{{ section.id }}-{{ media.id }}"
+                  >
+                    {%- liquid
+                      assign media_position = media_position | default: 0 | plus: 1
```
LucasLacerdaUX, line 741: We need to reset this above, or if the product's first available variant has no featured media, the count won't be reset (line 729) and will start at the wrong value.

image
(it starts at `media 13` instead of `media 1`)

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+          data-desktop-layout="{{ section.settings.gallery_layout }}"
+        >
+          <div id="GalleryStatus-{{ section.id }}-right" class="visually-hidden" role="status"></div>
+          <slider-component id="GalleryViewer-{{ section.id }}-right" class="slider-mobile-gutter">
+            <ul
+              id="Slider-Gallery-{{ section.id }}-right"
+              class="product__media-list contains-media grid grid--peek list-unstyled slider slider--mobile"
+              role="list"
+            >
+              {%- if product.selected_or_first_available_variant.featured_media != null -%}
```
LucasLacerdaUX, line 722: Follow up to the above:
```suggestion
              {%- assign media_position = 0 -%}
              {%- if product.selected_or_first_available_variant.featured_media != null -%}
```

sections/main-product.liquid
```diff
@@ -701,6 +701,235 @@
...
+                    class="product__media-item grid__item slider__slide{% if product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} is-active{% endif %}{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}"
+                    data-media-id="{{ section.id }}-{{ media.id }}"
+                  >
+                    {%- liquid
+                      assign media_position = media_position | default: 0 | plus: 1
+                      assign lazy_load = false
+                      if media_position > 1
+                        assign lazy_load = true
+                      endif
+                      render 'product-thumbnail', media: media, position: media_position, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width, lazy_load: lazy_load
```
LucasLacerdaUX, line 746: We can revisit this in a separate follow up PR, but I would suggest adding a variable to `product-thumbnail` to remove `<modal-opener>` structure from `video` and `model` medias.

The reason for that is because, for these two media types, the `<modal-opener>` is only used on mobile. Our approach, however, never uses the right column on mobile.

sections/main-product.liquid
andrewetchen: Thanks for catching this and for the details.

sections/main-product.liquid
andrewetchen: Another good catch! This is definitely not needed.

sections/main-product.liquid
andrewetchen: Thanks Lucas!

sections/main-product.liquid
andrewetchen: Thanks for the detailed explanation. I agree that this can be taken care of in a separate PR so I'll create an issue for this 👍🏼

LucasLacerdaUX (approved):
👍 ✅

Oliviammarcello (approved):

PR #1728, opened 2022-05-23
Add video button aria label 

### LINKED ISSUE #1161: [Video] Button missing accessible name without image
## Issue summary

When the video section features a video without an image, the load `button` is missing an accessible name. The exclusion of an accessible name will not convey the purpose of the `button` to assistive technology users, leading to a confusiung user experience.

<details>

  Video section, button to load video is highlighted.

</details>

#### HTML

```html
<button
  id="Deferred-Poster-Modal-_9VUPq3SxOc"
  class="video-section__poster media deferred-media__poster media--landscape"
  type="button"
>
  <!-- SVG… -->
</button>
```

## Steps to reproduce

1. Go to a Dawn theme store with any supported browser
2. With an active screen reader, <kbd>Tab</kbd> through a Video section, `button` withouth an image

### Expected

- `button` to include an accessible name to convey its purpose.

### Actual

- `button` missing an accessible name.

## Recommendation

Apply the `aria-label` attribute to the `button` element, setting its value to `Load video: {video.alt}`. With this, screen readers will announce the video description along with the callout, "Load video."

If the video does not include a description, the fallback of "Load video" will suffice as an accessible name.

#### HTML

```html
<button
  id="Deferred-Poster-Modal-_9VUPq3SxOc"
  class="video-section__poster media deferred-media__poster media--landscape"
  type="button"
  aria-label="Load video: {{ video.alt }}"
>
  <!-- SVG… -->
</button>
```

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **Browser:** Any
- **Dawn version:** 2.5.0
- **Component:** Video
- **WCAG Principal:** Perceivable
- **WCAG SC:** [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
- **Severity:** High
- **Effort:** Low

### PR DESCRIPTION
**Pull request by pizzaz93**
Fixes #1161

### DIFF
--- a/sections/video.liquid
+++ b/sections/video.liquid
@@ -60,7 +60,7 @@
         id="Deferred-Poster-Modal-{{ section.settings.video_url.id }}"
         class="video-section__poster media deferred-media__poster media--landscape"
         type="button"
-        aria-label="Load video: {{ video.alt }}"
+        aria-label="{{ 'sections.video.load_video' | t: description: section.settings.description | escape }}"
       >
         {%- if section.settings.cover_image != blank -%}
           <img
--- a/sections/video.liquid
+++ b/sections/video.liquid
@@ -60,6 +60,7 @@
         id="Deferred-Poster-Modal-{{ section.settings.video_url.id }}"
         class="video-section__poster media deferred-media__poster media--landscape"
         type="button"
+        aria-label="Load video: {{ video.alt }}"
       >
         {%- if section.settings.cover_image != blank -%}
           <img
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -95,7 +95,7 @@
                 </noscript>
                 <modal-opener class="no-js-hidden" data-modal="#PopupModal-{{ block.id }}">
                   <div class="deferred-media">
-                    <button class="deferred-media__poster full-unstyled-link{% if block.settings.image_padding %} collage-card-spacing{% endif %}" type="button" aria-haspopup="dialog" data-media-id="{{ block.settings.video_url.id }}">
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: sections/video.liquid [hunk 1]
```diff
@@ -60,6 +60,7 @@
         id="Deferred-Poster-Modal-{{ section.settings.video_url.id }}"
         class="video-section__poster media deferred-media__poster media--landscape"
         type="button"
+        aria-label="Load video: {{ video.alt }}"
```
ludoboludo, line 63: So in the theme we don't use many english strings directly in the HTML as it's information that's going to need to be translated in whatever language the merchant is using for their storefront. 

We have a specific way to deal with translation internally so in this case here, it can be hard to understand the way to go about it. It would need to be: 
```suggestion
        aria-label="{{ 'sections.video.load_video' | t: description: section.settings.description | escape }}"
```

And for the video in the collage section, you could probably use: 
```
aria-label="{{ 'sections.video.load_video' | t: description: block.settings.description | escape }}"
```

On the `<button>` element.

ludoboludo (commented):
So here in the issue it mentions the video and collage section. So the same would need to be applied on the video block of the collage section.

sections/video.liquid
pizzaz93: Added to the video button in the collage section.

ludoboludo:
Let me loop in @svinkle about something here. 

We're adding a label to the play buttons for the video section and the video block in the collage section. The issue is we do not have access from what I understand to the `alt` of the video itself as the setting is using the `url`. So we can't use the `video` object in this case and `video.alt`. 

So what we have in this PR is: `Load video: {{ section.settings.description }}` which is referring to: 

Which could end up being a bit of a verbose experience. I could see it being `Load video: video about poneys jumping fences during a competition in the summer of 2020, etc`

svinkle:
@ludoboludo Rereading my ticket, `video.alt` was meant as pseudo-code; I did mean the value set from `section.settings.description`. 🙂

Yes, it could end up being a little verbose, but I don't see another option to include a unique, meaningful label for the `button`.

ludoboludo (approved):
🚀

LucasLacerdaUX (approved):

PR #1736, opened 2022-05-25
Removed Screen Reader Span for UX

ChinazoOnwukaike:
**PR Summary:** 

Removed the hidden screen reader span for the h2 in order to reduce verbose user experience. This will ensure that the reader doesn't repeat the header unnecessarily.

In order to reduce verbose user experience and eliminate the repetition of the h2.

Deleted the span.
**Other considerations**

**Testing steps/scenarios**
- [ ] Needs continuous testing using screen reader

**Checklist**

### DIFF
--- a/sections/main-collection-banner.liquid
+++ b/sections/main-collection-banner.liquid
@@ -13,7 +13,6 @@
   <div class="collection-hero__inner page-width">
     <div class="collection-hero__text-wrapper">
       <h1 class="collection-hero__title">
-        <span class="visually-hidden">{{ 'sections.collection_template.title' | t }}: </span>
         {{- collection.title | escape -}}
       </h1>
 


sections/main-collection-banner.liquid
```diff
@@ -13,7 +13,6 @@
   <div class="collection-hero__inner page-width">
     <div class="collection-hero__text-wrapper">
       <h1 class="collection-hero__title">
-        <span class="visually-hidden">{{ 'sections.collection_template.title' | t }}: </span>
```
ludoboludo, line 16: Hello @ChinazoOnwukaike, 

Thanks for looking into Dawn and participating in making a better theme 🙂 

I was looking at the PR and this `<span>` is actually needed. The way it's written seem like it's repeating the collection title twice but it isn't. 

When being called like this: `'sections.collection_template.title' | t`, the theme look for a string in the language files:
```JSON
'sections': {
  'collection_template': {
    'title': 'Collection'
```

So in this case the output when using a screen reader is:

PR #1753, opened 2022-05-27
Replace keydown with keyup for custom interactions

### LINKED ISSUE #776: [Global] Custom interactions rely on the keydown event
## Issue summary

Some custom control events rely on the JavaScript `keydown` event. This event does not allow the user to "cancel" by swiping/mousing away from of the control as it activates immediately. This could potentially lead to a frustrating user experience.

#### JS

```js
// global.js:33 (and other instances)
trapFocusHandlers.keydown = function(event) {
  // …
};

// predictive-search.js:79 (and other instances)
onKeydown(event) {
  // …
}
```

## Steps to reproduce

1. Go to a Dawn theme store with any supported browser
2. <kbd>Tab</kbd> to the Search control in the header
3. Activate the Search control

### Expected

- Control to activate on the `keyup` event.

### Actual

- Control activates immediately on `keydown`.

## Recommendation

Consider adjusting the custom JavaScript events to use the `keyup` event. This would allow the user to "cancel" the event by swiping/mousing/<kbd>Esc</kbd>aping away from of the control.

#### JS

```js
// global.js:33 (and other instances)
trapFocusHandlers.keyup = function(event) {
  // …
};

// predictive-search.js:79 (and other instances)
onKeyup(event) {
  // …
}
```

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **Browser:** Any
- **Dawn version:** 2.2.0
- **Component:** Global
- **WCAG SC:** [2.5.2 Pointer Cancellation](https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html)
- **Severity:** Medium
- **Effort:** Medium

pizzaz93:
Fixes #776

### DIFF
--- a/assets/cart-drawer.js
+++ b/assets/cart-drawer.js
@@ -15,7 +15,7 @@ class CartDrawer extends HTMLElement {
       event.preventDefault();
       this.open(cartLink)
     });
-    cartLink.addEventListener('keydown', (event) => {
+    cartLink.addEventListener('keyup', (event) => {
       if (event.code.toUpperCase() === 'SPACE') {
         event.preventDefault();
         this.open(cartLink);
--- a/assets/global.js
+++ b/assets/global.js
@@ -39,14 +39,14 @@ function trapFocus(container, elementToFocus = container) {
     )
       return;
 
-    document.addEventListener('keydown', trapFocusHandlers.keydown);
+    document.addEventListener('keyup', trapFocusHandlers.keyup);
   };
 
   trapFocusHandlers.focusout = function() {
-    document.removeEventListener('keydown', trapFocusHandlers.keydown);
+    document.removeEventListener('keyup', trapFocusHandlers.keyup);
   };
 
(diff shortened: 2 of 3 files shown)


ludoboludo:
This is going to need quite a bit of testing I think in order to make sure we don't introduce odd behaviours on keydown. I don't have time to test right now though.

ludoboludo:
Yeah maybe that's a good idea @pizzaz93. I'm not sure we need to change the `keydown` to `keyup` in all those places too 🤔 
I'd like to double check with Scott where it's exactly needed. He mentions the some control button to open the predictive search.
But for focus handler I feel like it's probably not necessary as it's meant to be for keyboard users and I don't think there is a way to prevent a key from being pressed once it's pressed down. So if you're tabbing through the site it should be triggered on `keydown` still I'd think.

pizzaz93:
Closing and leaving as attached to the issue as a proof of concept/first draft.

PR #1763, opened 2022-06-02
Remove menu title requirement

allylane:
[Link to test](https://os2-demo.myshopify.com/admin/themes/128122224662/editor)

**PR Summary:** 
Remove the requirement for footer menus to have Headings in order to show.
Screen Shot 2022-06-02 at 10 22 17 AM
Screen Shot 2022-06-02 at 10 22 12 AM

There are circumstances where we want to display footer menu's, but a header isn't desirable.

I removed the conditional statements, and all of the applicable translations from the store.
I removed the info text under heading that mentioned it was a requirement.

**Other considerations**

**Testing steps/scenarios**
- [ ] Open the theme in the editor
- [ ] Remove a heading from one of the footer menus

### DIFF
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -51,7 +51,7 @@
         -%}
         <div class="footer__blocks-wrapper grid grid--1-col grid--2-col grid--4-col-tablet {{ footer_grid_class }}">
           {%- for block in section.blocks -%}
-            <div class="footer-block grid__item{% if block.type{% == 'link_list' %} footer-block--menu{% endif %}" {{ block.shopify_attributes }}>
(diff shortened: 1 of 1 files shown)


sections/footer.liquid
```diff
@@ -402,8 +402,7 @@
           "type": "text",
           "id": "heading",
           "default": "Quick links",
```
melissaperreault, line 404: I was wondering if we still needed to keep a default heading `Quick links` when adding the `Menu` block? We could leave it empty if we don't find added value. Although it does provide a hint of what text style this might take by default. Probably worth keeping after thinking out loud.

andrewetchen (approved):
This looks good to me 🎉

Issue #1273, opened 2022-01-26
Incompatibility with long strings


KaichenWang:
**Edge-case**
- [ ] [Collage's Collection block](https://screenshot.click/22-01-j2esn-wybu0.png) going off screen. The parent container `card__information` [is stretching](https://screenshot.click/22-01-8yuxf-0n1jm.png), it seems.

**Other text findings**
- [ ] Rich text button [off screen](https://screenshot.click/22-01-ir55g-13tmq.png) 
- [ ] Featured Collection button [off screen](https://screenshot.click/22-01-1nf1z-k6fl3.png) 
- [ ] Blog posts [View all link](https://screenshot.click/22-01-exi7x-bqce0.png) 
- [ ] Blog comment [off screen](https://screenshot.click/22-01-cxi21-jp17x.png)
- [ ] [View full details on Featured product](https://screenshot.click/22-01-9qy6b-z0m6g.png) layout not optimal
- [ ] Image with text button [off screen](https://screenshot.click/22-01-r6dpk-eyhcy.png)
- [ ] Filter label on Collection template [would push the product count](https://screenshot.click/22-01-1w4bg-pth1n.png)
- [ ] Cart `Continue shopping` button [off screen](https://screenshot.click/22-01-gcgey-wamft.png)
- [ ] Cart notification: [All CTAs go off screen ](https://screenshot.click/22-01-rnu7b-jp61k.png)
- [ ] Cart template Continue shopping link [off screen](https://screenshot.click/22-01-fy7e4-l0jdw.png)
- [ ] Cart template Checkout button [off screen](https://screenshot.click/22-01-v1m35-0g6rz.png)
- [ ] Login: Forgot password long label [should remain left aligned ](https://screenshot.click/22-01-mh65b-47jkw.png)
- [ ] Login Sign in button [off screen](https://screenshot.click/22-01-ofjt5-dbb47.png)
- [ ] Account details: Long name or long address go [off screen](https://screenshot.click/22-01-ft91v-l76lj.png)
- [ ] Addresses: Return to Account and Add new address go [off screen](https://screenshot.click/22-01-s3x4q-wik2r.png) + to check the Address and details and the other buttons (Edit and Delete) going [off screen](https://screenshot.click/22-01-ozchf-it90l.png)

_Originally posted by @melissaperreault in https://github.com/Shopify/dawn/pull/1271#pullrequestreview-864156770_

KaichenWang:
Additional related issues:
[...]

PR #1770, opened 2022-06-07
Announcment alignment

allylane:
**PR Summary:** 

Introducing text alignment options to our headers.

Copied functionality from the rich text section, but edited it to work at a block level.
Removed the centre alignment by default from the announcement bar class.

**Testing steps/scenarios**
- [ ] Open [this link](https://os2-demo.myshopify.com/admin/themes/128150470678/editor)
- [ ] Click on Announcement bar, then an announcement.
- [ ] Change alignment of text


sections/announcement-bar.liquid
```diff
@@ -63,6 +63,26 @@
...
+              "label": "t:sections.announcement-bar.blocks.announcement.settings.text_alignment.options__2.label"
+            },
+            {
+              "value": "right",
+              "label": "t:sections.announcement-bar.blocks.announcement.settings.text_alignment.options__3.label"
+            }
+          ],
+          "default": "center",
+          "label": "t:sections.announcement-bar.blocks.announcement.settings.text_alignment.label"
+        },
```
melissaperreault, line 85: Let's put it right under the `Text` setting to follow our latest content hierarchy ([Internal Figma file](https://www.figma.com/file/igRnZQceCieIPmM3NYZJyC/Flagship-Themes-Settings-UI?node-id=1093%3A16090))

Color scheme would also become the last to follow those logic too. Here is a reference


cc. @katycobb for visibility

allylane:
> It looks good and works as expected, but I am wondering if we should make the text align within the margins on the left and right side. [Screenshot](https://screenshot.click/07-09-r4lfk-fogvc.png)

Yea that makes more sense. Updated to include `page-width`, so it aligns with the rest of the text.
@melissaperreault I've also updated the order of the controls.

locales/th.schema.json
```diff
@@ -382,6 +382,17 @@
...
+              "label": "การจัดวางเนื้อหาบนเดสก์ท็อป",
+              "option_1": {
+                "label": "ซ้าย"
+              },
+              "option_2": {
+                "label": "กึ่งกลาง"
+              },
+              "option_3": {
+                "label": "ขวา"
+            },
```
andrewetchen, line 395:
```suggestion
              "option_3": {
                "label": "ขวา"
              }
            },
```

To fix the JSON error we need to close `option_3`.

assets/base.css
```diff
@@ -2183,8 +2183,7 @@ input[type='checkbox'] {
 }
 
 .announcement-bar__message {
-  text-align: center;
-  padding: 1rem 2rem;
+  padding: 1rem 0;
```
andrewetchen, line 2186: We should probably make sure the `.announcement-bar__link` gets the same treatment [when using links](https://user-images.githubusercontent.com/28404165/173389567-74ca66ec-01f4-4bd3-a11c-c5eb63ec07cb.png) for the announcement text:

**Left aligned announcement text with link**



**Right aligned announcement text with link**

andrewetchen (commented):
Things are looking good.

Just a couple of things:

allylane:
Good catch @andrewetchen.
Updated.

assets/base.css
```diff
@@ -2161,7 +2161,7 @@ input[type='checkbox'] {
 .announcement-bar__link {
   display: block;
   width: 100%;
-  padding: 1rem 2rem;
+  padding: 1rem 0;
```
andrewetchen, line 2164: My initial thought was combine both selectors and declare the padding just once like this:

```css
.announcement-bar__link,
.announcement-bar__message {
  padding: 1rem 0;
}
```

... but the way you have it now works perfectly and it'll make it easier if we ever needed to separately adjust `announcement-bar__link` or `announcement-bar__message` in the future.

andrewetchen (approved):
Thanks @allylane - everything looks good to me! 🎩

melissaperreault:
Thanks @allylane ! I'll do a final pass as a follow up when this other PR gets merge to validate where color scheme should live and sync with Katy.

I noticed that the color scheme [doesn't apply to the entire width](https://screenshot.click/13-16-owzx3-ptmtx.png) right now.

sections/announcement-bar.liquid
```diff
@@ -2,11 +2,12 @@
   {%- case block.type -%}
     {%- when 'announcement' -%}
       <div class="announcement-bar color-{{ block.settings.color_scheme }} gradient" role="region" aria-label="{{ 'sections.header.announcement' | t }}" {{ block.shopify_attributes }}>
+        <div class="page-width">
```
andrewetchen, line 5: We should adjust some of the spacing/indentation from `{%- if block.settings.text != blank -%}`:
- [Lines 6-19](https://github.com/Shopify/dawn/blob/e14c33fe6f1b4ecbc37233dcea0a3dc529417b0a/sections/announcement-bar.liquid#L6-L19) ideally should be changed **from**:

```liquid
        <div class="page-width">
        {%- if block.settings.text != blank -%}
          {%- if block.settings.link != blank -%}
            <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
          {%- endif -%}
              <p class="announcement-bar__message {{ block.settings.text_alignment }} h5">
                {{ block.settings.text | escape }}
                {%- if block.settings.link != blank -%}
                  {% render 'icon-arrow' %}
                {%- endif -%}
              </p>
          {%- if block.settings.link != blank -%}
            </a>
          {%- endif -%}
        {%- endif -%}
        </div>
```

To this:

```liquid
        <div class="page-width">
          {%- if block.settings.text != blank -%}
            {%- if block.settings.link != blank -%}
              <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
            {%- endif -%}
            <p class="announcement-bar__message {{ block.settings.text_alignment }} h5">
              {{ block.settings.text | escape }}
              {%- if block.settings.link != blank -%}
                {% render 'icon-arrow' %}
              {%- endif -%}
            </p>
            {%- if block.settings.link != blank -%}
              </a>
            {%- endif -%}
          {%- endif -%}
        </div>
```

andrewetchen (commented):
This looks good to me.

1. Just one small spacing/indentation adjustment needs to be made before this can merge. See `page-width` code comment below.
2. I also added some feedback for things we could tackle in a follow up.

Regarding Melissa's comment:
> I would have one more wish but can be done as a follow up if we would like to ship this prior; limiting the `line-length` to avoid [extra long lines](https://screenshot.click/14-51-gvitg-eoz4g.png) of text in our fluid layouts to follow best [usability](https://baymard.com/blog/line-length-readability) and [accessibility practices](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html). We've set it to `89rem` when the Image banner box is hidden.

- We could add this to the list in the following issue:
  - https://github.com/Shopify/dawn/issues/1273

Lastly, I wasn't sure if this is something we'll also want to address in a follow up or not -- thought I'd share just in case. Have a look at the following before/after. I'm totally fine with the way it is -- what are your thoughts? The focus and hover background is slightly different when comparing to the latest version of Dawn:

- [Before](https://user-images.githubusercontent.com/28404165/174121906-fcdc6b52-a9e5-4d4b-8733-56780991fe71.mp4)
- [After](https://user-images.githubusercontent.com/28404165/174121982-766da172-4063-48c8-93d5-8017a706a5a4.mp4)

Oliviammarcello:
One note sorry
- [ ] The hover state doesn't fill the full width of the bar. [Video](https://screenshot.click/16-58-w7o9o-9z73x.mp4)

melissaperreault:
> The focus and hover background is slightly different when comparing to the latest version of Dawn

Good catch Olivia and Andrew!

For the Focus I would make it the same unless there are technical limitations. If so, adding more padding so the focus is [not close cutting](https://screenshot.click/16-35-ezswq-sfp0s.png) so close to the content would be best. 👍

andrewetchen:
> > The focus and hover background is slightly different when comparing to the latest version of Dawn
> 
> Good catch Olivia and Andrew!
> 
> For the Focus I would make it the same unless there are technical limitations. If so, adding more padding so the focus is [not close cutting](https://screenshot.click/16-35-ezswq-sfp0s.png) so close to the content would be best. 👍

Thinking about this more, I also agree with it being the same. One way we can achieve this is by adding the `page-width` container div (`<div class="page-width">`) before the paragraph tag in `announcement-bar.liquid` section:

```liquid
{%- for block in section.blocks -%}
  {%- case block.type -%}
    {%- when 'announcement' -%}
      <div
        class="announcement-bar color-{{ block.settings.color_scheme }} gradient"
        role="region"
        aria-label="{{ 'sections.header.announcement' | t }}"
        {{ block.shopify_attributes }}
      >
        {%- if block.settings.text != blank -%}
          {%- if block.settings.link != blank -%}
            <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
          {%- endif -%}
          {% comment %} Place it here {% endcomment %}
          <div class="page-width">
            <p class="announcement-bar__message {{ block.settings.text_alignment }} h5">
              {{ block.settings.text | escape }}
              {%- if block.settings.link != blank -%}
                {% render 'icon-arrow' %}
              {%- endif -%}
            </p>
          </div>
          {%- if block.settings.link != blank -%}
            </a>
          {%- endif -%}
        {%- endif -%}
      </div>
  {%- endcase -%}
{%- endfor -%}
```

https://user-images.githubusercontent.com/28404165/174349673-749bab2e-001e-4ee1-82ec-c28618e37281.mp4

allylane:
I've applied that fix @andrewetchen.
I've also added justify as an option as it's something I was hoping to include in a new theme we have coming.

We'll just need translations for that.

assets/base.css
```diff
@@ -572,6 +572,11 @@ table:not([class]) th {
   text-align: center;
 }
 
+.justify {
+  text-align: justify;
+  text-align-last: justify;
```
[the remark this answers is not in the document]
ludoboludo: it feels so weird that such a basic text formatting thing doesn't work on a browser... 🤷 


There is a work around mentioned there but it sounds wacky 🤔

ludoboludo (commented):
Looks good except for the `justify` option

assets/base.css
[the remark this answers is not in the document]
melissaperreault: I am curious to why we are adding a `Justify` option?

I don't believe justifying text works well on the web at all where we can't control the amount of words, the language and the width of devices the content will be read. Merchants won't be able to control the additional spacing between the words that this will create. But I would love to learn more of use cases where this might be successful though!

We also need to consider that if we are adding this in one place as an option, we would want to offer it elsewhere and would require more in depth explorations and testing.

ludoboludo (approved):
Looks good to me 👍

andrewetchen (approved):
>I've applied that fix @andrewetchen.
~I've also added justify as an option as it's something I was hoping to include in a new theme we have coming.~
>
>We'll just need translations for that.

Thanks @allylane, this is looking great!

ludoboludo:
I requested translations 👍 So once they're back and we get a green checkmark [here](https://screenshot.click/04-16-va3zz-1l1dc.png), we're good 👍

bacounts:
@allylane - I noticed a follow-up for when the Desktop logo position is set to Top center — the announcement and the header icons do not lineup. It seems fine when the text is right justified, but not left (I did not test every instance). See screenshot below. CC: @Oliviammarcello for alignment on issue.

**Screenshot:**
Screen Shot 2022-07-05 at 10 38 07 PM

PR #1776, opened 2022-06-09
Add a "Show separator lines" option to the footer

kjellr:
**PR Summary:** 

Added a "Show separator lines" option to the footer. 

separator-lines

We currently offer this option in the header, but we do not offer it in the footer. This feels inconsistent. 🙂

- I added a setting to the Footer area, mimicking the approach taken for the header (adding a class, etc.). 
- I chose to use a single setting to control both separators in the footer. I thought this made sense since the two separators are already linked today (if you have no content in the top footer area, the bottom separator disappears). 

**Other considerations**

- This setting is really only relevant if "Background 1" is chosen as the background. It would be cool to make this new setting show up conditionally only if that is the selected color (The header should probably work this way too, since it's also only relevant if "Background 1" is selected).

I'd love feedback on these three items specifically: 

- [x] I'm not sure about combining the two separators into one option, but adding two options felt like overkill. :shrug: What do you all think?
- [x] I wasn't entirely sure where to place the new option, so I just placed it below the background color option. Do folks think this makes sense?
- [x] I updated the default English locale file to reflect that this controls two separators, not just one. I'm honestly not sure how to generate all the other locale files (this is my first PR, and I just manually updated those! 🙈). The rest all still refer to just a single "separator" and need to be updated. 

**Testing steps/scenarios**

- [Visit this link](https://os2-demo.myshopify.com/admin/themes/128168624150/editor)
- Select the Footer
- Toggle the new "Show separator lines" option. 

**Checklist**

### DIFF
--- a/assets/section-footer.css
+++ b/assets/section-footer.css
@@ -1,4 +1,4 @@
-.footer--border-top:not(.color-background-1),
+.footer--border-top,
 .footer--border-top .footer__content-bottom:not(:only-child) {
   border-top: 0.1rem solid rgba(var(--color-foreground), 0.08);
 }
--- a/assets/section-footer.css
+++ b/assets/section-footer.css
@@ -1,5 +1,5 @@
-.footer-wrapper--border-top:not(.color-background-1),
-.footer-wrapper--border-top .footer__content-bottom:not(:only-child) {
+.footer--border-top:not(.color-background-1),
+.footer--border-top .footer__content-bottom:not(:only-child) {
   border-top: 0.1rem solid rgba(var(--color-foreground), 0.08);
 }
 
--- a/assets/section-footer.css
+++ b/assets/section-footer.css
@@ -1,5 +1,5 @@
 .footer-wrapper--border-top,
-.footer__content-bottom:not(:only-child) {
+.footer-wrapper--border-top .footer__content-bottom:not(:only-child) {
   border-top: 0.1rem solid rgba(var(--color-foreground), 0.08);
 }
 
--- a/assets/section-footer.css
+++ b/assets/section-footer.css
@@ -1,4 +1,5 @@
-.footer-wrapper--border-top {
+.footer-wrapper--border-top,
+.footer__content-bottom:not(:only-child) {
   border-top: 0.1rem solid rgba(var(--color-foreground), 0.08);
 }
 
@@ -37,12 +38,7 @@
 }
 
 .footer__content-bottom {
-  border-top: solid 0.1rem rgba(var(--color-foreground), 0.08);
   padding-top: 3rem;
-}
-
-.footer__content-bottom:only-child {
-  border-top: 0;
 }
 
 .footer__content-bottom-wrapper {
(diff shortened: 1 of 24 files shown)


kjellr:
> I updated the default English locale file to reflect that this controls two separators, not just one. I'm honestly not sure how to generate all the other locale files (this is my first PR, and I just manually updated those! 🙈). The rest all still refer to just a single "separator" and need to be updated.

Judging by the failed `translation-platform` check here, I'm guessing I should actually revert all those non-english translations. We'll cross that translation bridge when we vet the rest of the PR. 👍

assets/section-footer.css
```diff
@@ -1,11 +1,8 @@
-.footer {
+.footer-wrapper--border-top,
+.footer-wrapper--border-top .footer__content-bottom:not(:only-child) {
   border-top: 0.1rem solid rgba(var(--color-foreground), 0.08);
 }
 
-.footer:not(.color-background-1) {
-  border-top: none;
-}
```
ludoboludo, line 7: This means that it would have a visual impact on merchants' stores potentially where if they were using a `background-1` `color scheme` it would hide the separator lines. Or the other way around if we set them to `true` by default.
So any merchant using a `color scheme` other than `background-1` will now get some separator lines showing. 

Not a big issue, just wanting to flag this as something that we will need to be aware when doing new releases.

assets/section-footer.css
```diff
@@ -1,11 +1,8 @@
-.footer {
+.footer-wrapper--border-top,
```
ludoboludo, line 1: Small nitpick:
Maybe the class could be something like `footer--separator-lines` ?

I think the current name sound like it's applying border only to the wrapper and at the top when it's doing it in a couple places. Might be more self explanatory this way. 

Also if we try and follow [BEM conventions](https://github.com/Shopify/themes/wiki/BEM-convention) for class naming, we shouldn't be using `footer-wrapper--border-top` as we don't have a `footer-wrapper` class on this element 👍

ludoboludo (commented):
Looking good 👌 Added a few comments

assets/section-footer.css
kjellr: Yeah, I wasn't sure about this. I borrowed the nomenclature from the header separator implementation, which uses `header-wrapper--border-bottom`. If we want to rename to `separator-lines`, I think we should update that to match. 

Either way, good call on the `footer-wrapper` name though... I've updated this classname to `footer--border-top` in https://github.com/Shopify/dawn/pull/1776/commits/67f0744b6f2fd71101e96e8aef6339b0dc8b517c.

ludoboludo (dismissed):
Looks good to me 👌 🙂

kjellr:
I just reverted the incorrect labels that I'd manually added in https://github.com/Shopify/dawn/pull/1776/commits/51940de16d2de91b99f7da4bdd57e49152fe2db0. Should be ready to send over for the real translations now. 😄

ludoboludo (approved):
🚀

PR #1777, opened 2022-06-10
Try adding a CSS-only announcement bar marquee option

kjellr:
**PR Summary:** 

Adds a "Scrolling marquee" option to the announcement bar. 

https://user-images.githubusercontent.com/1202812/173064906-6b8a16f9-d5bd-4181-aabb-2b50afd1b24d.mp4

Just a proof of concept. 

The PR uses a simple CSS animation to create an endlessly scrolling marquee. It borrows [this idea](https://codepen.io/fcalderan/pen/GRJeYOL) for using text shadows to repeat the text. This generally works for text, but breaks if you use emoji. 😞 

Aside from the emoji issue, I don't think this is actually a viable option for a couple reasons though:

1. It relies on a hardcoded width for the text. This breaks on smaller screens. 
2. The speed of the animation generally to be adjusted depending on the amount of text present as well. 

Both of these could theoretically be handled by JS, but I think we'd want to explore that outside of this PR anyway. Keeping this as a draft for now, but it should probably be closed once we've taken a look at it. 

**Other considerations**

The PR automatically turns the scrolling off if the user's device has `Prefers reduced motion` set. 

**Testing steps/scenarios**

- [Visit the link here.](https://os2-demo.myshopify.com/admin/themes/128171835414/editor?section=announcement-bar&block=announcement-bar%2Fannouncement-bar-0)
- Try toggling on and off the marquee setting.
- Try changing the text. (Add lots of text, notice that it breaks)
- Check it out at other screen sizes (Notice that it breaks)

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2192,6 +2192,41 @@ input[type='checkbox'] {
   letter-spacing: 0.1rem;
 }
 
+.announcement-bar.marquee {
+  overflow: hidden;
+}
+
+.announcement-bar.marquee .announcement-bar__message {
+  display: inline-block;
+  white-space: nowrap;
+  color: #00112C;
+  width: 25vw;
+  text-align: initial;
+  animation: announcement-bar-marquee 2s linear infinite;
+
+  /* Use shadows to repeat the text */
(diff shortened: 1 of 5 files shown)


Oliviammarcello:
+1 to Melissa's comment. I think we will need to add a play/pause button. When working on header improvements, I designed one [here](https://www.figma.com/file/VIshelzL4USycwoA3QU4Ix/Header?node-id=449%3A5265) if that is helpful.

Issue #1711, opened 2022-05-19
Missing sold out UI


gregjotau:
I am making a new issue on this one, as it really is *not OK* that the main theme recommended to merchants and a great theme overall does not have a functioning sold out UI built in. 

_Edited part_

**Current problem**
Right now on the product page there isn't a way to tell if a variant combination is available or sold out. You'd have to click around and find out if the add to cart button is going to be disabled or not. 

**Describe the enhancement you'd like**
Ideally as you land on a product page you can see right away which variant aren't available. 

**Describe alternatives you've considered**
There is an existing alternative that @cfxd came up with in this PR: https://github.com/Shopify/dawn/pull/1407
Otherwise no other alternatives. 

**Additional context/screenshots**
![](https://screenshot.click/19-22-rxi6z-fndim.png)

The proposed change in the PR linked above looks like this: 
![](https://screenshot.click/19-14-lsmh8-3kaib.png)

As another example we can see how Nike goes about it on their product page: 
![](https://screenshot.click/19-25-l8ri3-kvzrg.png)

gregjotau:
Related to:

https://github.com/Shopify/dawn/pull/1407
https://github.com/Shopify/dawn/issues/45

cfxd:
> There is an existing alternative that @cfxd came up with in this PR: https://github.com/Shopify/dawn/pull/1407

Thanks, I can't take all the credit... but I will 😉

gregjotau:
I see @ludoboludo is already a saving angel and has started reviewing it, awesome 🙌

Issue #1693, opened 2022-05-12
[Variant picker block] Add checkbox to hide sold out variants


Oliviammarcello:
1. We can make an enhancement to how we tackle surfacing sold out variants. [Currently](https://screenshot.click/27-45-s1ceh-yq1j1.png), if a variant is sold out, the buyer only has the disabled add to cart button as a reference. We should have the variant pill  / dropdown indicate that the variant is sold out.  [See Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=202%3A15976) [See current TS Snippet for code reference](https://ts-code-snippets.com/2.0_Themes/Product_Page/Disable_Items_Sold_Out_Dropdown/)

2. Add checkbox to the `Variant picker` block on the PDP to enable and disable the ability to hide sold out variants. [Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=208%3A23536)

For Parity, ensure this is implemented on: 
- [ ] Product Page Template
- [ ] Featured product Section
- [ ] Quick Add Modal

bacounts:
This topic has been a huge discussion in three already existing PR's. There seems to be several holdup's, including a platform based error. See #1407 & #105 & #876

Issue #861, opened 2021-11-09
hide variants that have 0 in stock 


Miktau:
We have all color variant under the same product. If one variant is sold out it still showing and is annoying for the customers to see it. Is it possible to make a feature so when one variant have 0 in stock it goes automatically and hide this variant?

![Screen Shot 2021-11-09 at 15 22 50](https://user-images.githubusercontent.com/61705825/140941530-23db30b0-1263-4107-8438-14347a65697e.png)

PR #1779, opened 2022-06-10
Test variant picker soldout

ludoboludo:
**PR Summary:** 

This PR is just a test one to replicate the code changes and approach used in https://github.com/Shopify/dawn/pull/1407
The intent of my PR is just to share internal links with our team to be able to test more easily. 

Thos changes made so that you can see at first glance when landing on a product page, which variant are or aren't available. 

Related issue: https://github.com/Shopify/dawn/issues/1693, https://github.com/Shopify/dawn/issues/1711, https://github.com/Shopify/dawn/issues/861, https://github.com/Shopify/dawn/issues/648

Check out the original PR for details.


bakura10:
Hi,

we implemented that kind of features long time ago on our themes, and I recommend you to keep it clickable. By preventing the click you prevent any back in stock app to ever work (which stays a very good way to recover sales).

bakura10:
There is still no plan to add a clean solution in Liquid that would remove those hard to understand and inefficient Liquid loops ?

bakura10:
I do not consider this solution really clean. A clean way would be for Shopify to add something like this:

{% for value in option.values %}
  value.disabled
{% endfor %}

Where the disabled would resolve to true if, for a given value by taking into acccount the options order, there is at least one variant available.

edmund-teh:
I just wanted to add some feedback here from a UX perspective. Apologies that it's taken so long to get to.

- I agree with @LucasLacerdaUX here on removing the `Soldout select type` setting that, with the most recent changes, currently only shows `None` and `Disabled` options. Settings that only work conditionally for Type: `Dropdown` vs. Type: `Pills` aren't considered best practices for UX (even if we have these instances elsewhere, let's not continue this pattern).  

- Let's remove this setting and show all dropdown menu options by default with `[Unavailable]` appended. 

- We still want `[Unavailable]` options to be selectable (when Type is set to Dropdown) based on previous feedback in allowing 'back in stock' apps to function correctly. 

- +1 to @ludoboludo's recommendation in adjusting the disabled pill opacity to be at `0.6` to pass for accessibility contrast standards.

- @ludoboludo's updates in improving the experience with the keyboard focus being lost looks great. However, after discussing with @LucasLacerdaUX yesterday, we still need to validate things and lean on our accessibility experts to weigh in on things as there will still be focus loss with this solution... 

> "Even if we save the currently selected element, fetch, replace and re-focus, the focus will be lost for a second. That focus change could still be announced to assistive technologies like voiceover, and might cause confusion for users." - @LucasLacerdaUX  

Finally, I wanted to thank everyone for their contributions and for taking the time to discuss solutions, provide suggestions, etc.

ludoboludo:
Im gonna close this PR since the community one shipped. I have a follow up to fix a remaining issue which is a draft right now as I have it working but now I need to polish everything rather than duplicate my logic: https://github.com/Shopify/dawn/pull/2031

PR #1786, opened 2022-06-13
Introduce link to whole of card

allylane:
**PR Summary:** 
This introduces links to multicolumn when they are added.

Because the tappable area is currently quite small.

Followed convention elsewhere and added an :after to the link to cover the whole column

**Testing steps/scenarios**
- [ ] Open the [Editor](https://os2-demo.myshopify.com/admin/themes/128188612630/editor)
- [ ] Add a Multicolumn
- [ ] Add a link
- [ ] Add an image (optional)
- [ ] The whole column should be a link

### DIFF
--- a/assets/section-multicolumn.css
+++ b/assets/section-multicolumn.css
@@ -88,6 +88,16 @@
   margin-top: 1rem;
 }
 
+.multicolumn-card__info a:after {
+  bottom: 0;
+  content: "";
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  z-index: 1;
+}
+
 .multicolumn-list__item.center .media--adapt,
 .multicolumn-list__item .media--adapt .multicolumn-card__image {
   width: auto;
--- a/assets/section-multicolumn.css
+++ b/assets/section-multicolumn.css
@@ -88,16 +88,6 @@
   margin-top: 1rem;
 }
 
-.multicolumn-card__info a:after {
-  bottom: 0;
-  content: "";
-  left: 0;
-  position: absolute;
-  right: 0;
-  top: 0;
-  z-index: 1;
-}
-
 .multicolumn-list__item.center .media--adapt,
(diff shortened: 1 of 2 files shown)


allylane:
Uploading LinkArea.mp4…

allylane:
Uploading LinkArea.mp4…

templates/index.json
```diff
@@ -192,9 +202,9 @@
         "background_style": "none",
         "button_label": "",
         "button_link": "",
-        "swipe_on_mobile": false,
         "color_scheme": "background-1",
         "columns_mobile": "1",
+        "swipe_on_mobile": false,
         "padding_top": 36,
         "padding_bottom": 36
       }
```
ludoboludo, line 210: Hey @allylane, when you create a theme that's linked to your PR, you want to enable one way sync so that any edits made in the theme editor while testing your work isn't updated back to the PR. 
So here for example, there shouldn't be any change to this file 🤔

assets/section-multicolumn.css
```diff
@@ -88,6 +88,16 @@
...
 
+.multicolumn-card__info a:after {
+  bottom: 0;
+  content: "";
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  z-index: 1;
+}
```
ludoboludo, line 99: I don't think it's ideal to make the whole thing a link 🤔 
Doing so makes it that you can't select the text content and copy it.

On top of that you can use the column rich text setting to add some links into the text which become un-clickable because we're now covering the whole card with our link label  

I think the clicking area is decent enough the way we have it so I'm not sure it needs changing. 

[video](https://screenshot.click/14-45-5q45g-p30wc.mp4)

ludoboludo (commented):
Left a couple comments 👍

templates/index.json
allylane: Yea, whoops 🤦‍♂️

PR #1787, opened 2022-06-13
Switch to clamp for (most) responsive font sizes

kjellr:
## PR Summary:

This PR converts some (most) of Dawn's breakpoint-based font sizes to use `clamp()` instead. This provides a more natural set of font sizes for devices using a middle viewport.

The PR takes the font sizes currently used for large and small breakpoints, and uses those values as the `min` and `max` inputs for `clamp()` instead. **The "preferred" middle values are generalized just to show the effect.** We'd need to fine-tune them if we actually wanted to implement this. 

**Before**

Note that there's an abrupt change in font sizes around `750px`.

https://user-images.githubusercontent.com/1202812/173438276-eee2043b-bdd3-4543-b6fa-7165a215c5ad.mp4

**After**

Note that the abrupt change is gone, and font sizes transition fluidly from the large font size down to the smaller one.

https://user-images.githubusercontent.com/1202812/173438405-51b33f85-7822-4e67-bb57-c9b3d07f4f21.mp4

## Testing steps/scenarios

1. [Visit the OS-2 demo site](https://os2-demo.myshopify.com/admin/themes).
2. Find the `try/responsive-type-sizes` branch in the theme library.
3. Preview the theme, try a variety of viewports.

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -391,78 +391,36 @@ h5,
 }
 
 .hxl {
-  font-size: calc(var(--font-heading-scale) * 5rem);
-}
-
-@media only screen and (min-width: 750px) {
-  .hxl {
-    font-size: calc(var(--font-heading-scale) * 6.2rem);
-  }
+  font-size: clamp(calc(var(--font-heading-scale) * 5rem), 6.2vw, calc(var(--font-heading-scale) * 6.2rem));
 }
 
 .h0 {
-  font-size: calc(var(--font-heading-scale) * 4rem);
-}
-
-@media only screen and (min-width: 750px) {
-  .h0 {
-    font-size: calc(var(--font-heading-scale) * 5.2rem);
-  }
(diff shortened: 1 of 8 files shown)


ludoboludo:
I love this 😍
Something that helped me just now visualize more what `4vw` can represent is this [article from Smashing](https://www.smashingmagazine.com/2022/01/modern-fluid-typography-css-clamp/). I like thinking of it as `4vw = 4% of the viewport width` and having a table with our current breakpoints to show what is the applied value in terms of pixels will help anyone reviewing/trying to understand how it works. 

[Here is what I found super useful in that article mentioned above](https://screenshot.click/29-47-ygiye-7082y.png)

I think you mentioned the tool brought up in the article before but I finally understand a bit better 😅 
It also mentions some drawback in terms of accessibility and zooms and recommends using this technic for more titles and less for the body text. 

Maybe we could start by only applying that approach to our titles and see where else it could be useful/needed.

PR #1796, opened 2022-06-14
Allow transparent backgrounds for banners

allylane:
**PR Summary:** 

Removing the 0.1 opacity default background, and "magic" colours when overlays are applied so they stay true to the colour palletes.

More creativity with banners

**Testing steps/scenarios**
- [ ] Open the [editor](https://os2-demo.myshopify.com/admin/themes/128199163926/editor?section=template--15275289149462__image_banner)
- [ ] Click on the image banner
- [ ] Upload a transparent png for an image

### DIFF
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -318,19 +318,10 @@
 @media screen and (min-width: 750px) {
   .banner--desktop-transparent .banner__box {
     background: transparent;
-    --color-foreground: 255, 255, 255;
-    --color-button: 255, 255, 255;
-    --color-button-text: 0, 0, 0;
     max-width: 89rem;
     border: none;
     border-radius: 0;
     box-shadow: none;
-  }
-
-  .banner--desktop-transparent .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
-    --alpha-button-background: 0;
   }
 
   .banner--desktop-transparent .content-container:after {
--- a/assets/base.css
+++ b/assets/base.css
@@ -1196,7 +1196,6 @@ summary::-webkit-details-marker {
 /* component-media */
 .media {
   display: block;
-  background-color: rgba(var(--color-foreground), 0.1);
   position: relative;
   overflow: hidden;
 }


Oliviammarcello (commented):
Looks good and works as expected! 

- [ ] We should now remove the help text under the colour scheme picker since the selector will always work regardless of the container. [Screenshot](https://screenshot.click/15-21-7rls0-dhlwn.png)

assets/section-image-banner.css
```diff
@@ -318,21 +318,12 @@
...
     max-width: 89rem;
     border: none;
     border-radius: 0;
     box-shadow: none;
   }
 
-  .banner--desktop-transparent .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
-    --alpha-button-background: 0;
```
ludoboludo, line 333: I don't think we should be removing the alpha value here. If the button is using the outline style it should keep a transparent background 🤔

ludoboludo: On mobile it looks good but not on desktop: [recording](https://screenshot.click/04-23-cdj2f-7n92x.gif).

assets/section-image-banner.css
```diff
@@ -318,21 +318,12 @@
 @media screen and (min-width: 750px) {
   .banner--desktop-transparent .banner__box {
     background: transparent;
-    --color-foreground: 255, 255, 255;
-    --color-button: 255, 255, 255;
-    --color-button-text: 0, 0, 0;
```
ludoboludo, line 323: Those changes are on declarations within a media query `@media screen and (min-width: 750px) {` so they're only impacting the wider screens. 
If you're on mobile the colour scheme is never applied and the default is white. 

So it can make for a confusing experience. As it's working on desktop and applying the colour scheme accordingly when a container is and isn't showing but not on mobile. 

Also the info next to the colour scheme setting needs to be updated:

ludoboludo (commented):
I think we might need to change the default values of that section with a colour scheme using a white text, [screenshot](https://screenshot.click/04-12-xb69a-viwvb.png). But that would need to be done in the private dawn repo I believe as it would be an update on the `.json` file.

andrewetchen:
Hey team,

The [probot-CLA](https://user-images.githubusercontent.com/28404165/181069776-4926e5bd-7c45-4f8b-8cd0-9998fcc2c641.png) was recently deprecated and replaced with the [shopify-cla](https://user-images.githubusercontent.com/28404165/181069526-e35e4fd4-a151-45d3-9cfe-babfe4d31829.png) action.

To successfully use the new [shopify-cla](https://user-images.githubusercontent.com/28404165/181069526-e35e4fd4-a151-45d3-9cfe-babfe4d31829.png) status check, this PR will be closed and reopened.

Thanks and sorry for the minor inconvenience. Message me if you have any questions 😄 

Please refer to the following to for more information:

- https://github.com/Shopify/dawn/pull/1859
- https://github.com/Shopify/dawn/pull/1874

PR #1817, opened 2022-06-22
Fix Quick Add when cart type is set to Page (remove data-cart-type)

LucasLacerdaUX:
Fix Quick Add when cart type is set to page.

Instead of adding `data-cart-type` to all product forms, I've decided to remove this property:

- This attribute doesn't represent a `product-form` setting, but a theme-level one. That means it'll always have the SAME repeated value across all `product-form` elements.
- The only purpose of this attribute was to check if the cart type was set to `Page` or not.
- We already had a property inside our JS logic that gives us this answer: `this.cart`, the result of querying both `cart-notification` and `cart-drawer` elements.
  - `this.cart` will be empty if type is set to `page`
  - `this.cart` will be non-empty if type is NOT set to `page`.

**Testing steps/scenarios**
- [ ] Test product forms on
  - [ ] Featured Product
  - [ ] Product Page
  - [ ] Quick Add
- [ ] For each of the above, test the three different options for cart type:
  - [ ] Drawer
  - [ ] Notification
  - [ ] Page

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128252084246)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128252084246/editor)

### DIFF
--- a/assets/product-form.js
+++ b/assets/product-form.js
@@ -46,9 +46,7 @@ if (!customElements.get('product-form')) {
             soldOutMessage.classList.remove('hidden');
             this.error = true;
             return;
-          } 
-          
-          if (!this.cart) {
+          } else if (!this.cart) {
             window.location = window.routes.cart_url;
             return;
           }


ludoboludo (approved):
Tested different settings. 
On slow connection, my phone and my ipad as well as Safari and Chrome 👍 

Worked as expected. Thanks for the fix!

PR #1774, opened 2022-06-09
[Card Settings] Break card global settings into Blog, Collection and Product Cards

sofiamatulis:
**PR Summary:** 

Before, all card settings were condensed and affected all cards in the theme. With this update, they still live in the global settings, but they are separated by blog, product and collection

It should be working exactly the same, but the settings are now separated

Figma: https://www.figma.com/file/WzjG1Eld7XJBwm8xBjj8Il/Product-Cards?node-id=235%3A16313
Project Log: https://docs.google.com/document/d/1tASZYuY8RUZQcj_vc1S1q5mCeYtmp1gAZieWlyjXuog/edit#

**Why are these changes introduced?**

Fixes #1775 .

**What approach did you take?**

I added new variables for blog, product and collection cards, which are set in `base.css` and removed the cards settings.

**Other considerations**

**Testing steps/scenarios**

Please test the password page as well

_For products_
- [ ] Featured collection
- [ ] Prod recommendations
- [ ] Collection page
- [ ] Collage
- [ ] Search results

_For blog_
- [ ] Search results
- [ ]  Featured blog
- [ ] Main blog

_For collection_
- [ ] Collections list
- [ ] Cart drawer
- [ ] Main list collection
- [ ] Collage

**Demo links**
_Please include a link to a demo store that includes preconfigured sections and settings to allow reviewers to easily test the features you are working on._

- [Editor](https://os2-demo.myshopify.com/admin/themes/128174817302/editor)

**Checklist**
- [ ] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [ ] Notified the [help.shopify.com](https://help.shopify.com) team about updates to theme settings (current contact: Kimli)
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #1834, opened 2022-06-29
New card settings: collage match all cards

### LINKED ISSUE #1826: [Card settings] Add “match all cards” dropdown setting to the collage section
[With the new update](https://github.com/Shopify/dawn/pull/1774) to the global card settings, now collage section's card settings is tied globally to:

- Product cards will be tied to product card settings
- Collection cards to will be tied to collection card settings
- Media cards will be tied to media global settings 

Additionally, 
- [ ] We will be adding a new section level dropdown (select) input setting to the collage section called "Card Style" with options: 
- [ ] Use individual card styles
- [ ] Style all as product cards
- [ ] Adding a help-text below "Product, collection, and blog card styles can be updated in theme settings."

Screen Shot 2022-07-12 at 3 53 12 PM

**Dropdown options:** 
- **Option 1 (Default) | Use individual card styles:** Media will be tied back to the media global settings.
- **Option 2 | Style all as product cards:** Media will match the product card global settings (Including text size and style, color scheme, and card corner styles) 

**Why are we adding this section level setting?** For merchants to match their card styles at once without having to go to the individual card settings above. This setting will match all cards on collage into the product card style if the merchant chooses to. This is also the fix for possible media padding issue – if it occurs, the merchant can choose "Option 2: Style all as product cards"  and edit the media padding settings from product card global padding settings. 

**Why don't we have other card types such as collection or blog cards to match to?**
Because this will be similar to a section level override. In the back end the setting will only see the existing card types to match to. When the merchant selects product cards, the code sees the applied product cards setting at a section level and matches what it sees. Doesn’t go to the global settings to pull from. Collages section don’t have blog cards so the code can’t match what doesn’t exist at section level. And collection cards has only 1 line of metadata, so it won't be good fit for 1:1 match.

sofiamatulis:
**PR Summary:** 

Introduced a new dropdown setting labeled "card style" to the collage section settings to match all card styles. 
There will be 2 options for the merchants to choose from under this new dropdown setting: 

- **Use individual card styles** (should be selected by default) 
Media cards should be using the media global settings and all other card types (product and collection) should be using their own respective card type settings. 
- **Style all as product cards**
Media and collection cards should be using product card global settings and their text size, style, color scheme, and corner style should match the product card within the collage section. 

We are also adding a help text "Product, collection, and blog card styles can be updated in theme settings."

collage new section

**How does the settings work for collage right now?** 
- Product cards is always following the global product cards settings
- Collection cards is following the collection card settings (or product cards settings depending on the selection made))
- The media cards were always following product card settings ([before this update](https://github.com/Shopify/dawn/pull/1774#issuecomment-1170592382) - now it is following media global setting or product cards depending on the selection made.) 

This is introduced because we are breaking card settings (blog, product and collection https://github.com/Shopify/dawn/pull/1774). Because of this, the merchant wouldn't have a way anymore to make all cards have the same look if they are pulling from different settings now. This gives the merchant the opportunity to match all cards in the collage if desired. 

Fixes #1826  and https://github.com/Shopify/dawn/pull/1774#issuecomment-1170592382 

Passed an empty or product card class depending on the setting selected

**Other considerations**

- Note that this means that image and video in the collage will only have the image padding/color option setting when following the product card settings (not available in the global media settings currently) 
- Stand alone media (blog pages, image with text) will continue to be tied to the global media settings under theme settings. No change to its current behavior. 

**Testing steps/scenarios**
- [ ] Card style setting under collage section should be set to "use individual card style" by default 
- [ ] When the card style setting set to "use individual card style" media cards should be using the media global settings and all other card types (product and collection) should be using their own respective card type settings. Product cards = product cards global setting, Collection cards: collection cards global setting. 
- [ ] When the card style setting set to "style all as product cards" media and collection cards should be using product card global settings and their style, color scheme, and corner style should match the product card within the collage section. 

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/128295403542/editor)

**Checklist**


duygukalaycioglu:
New setting content feedback:

- [ ] Update "Card Style" to "Card style" - to match the editor labels using lowercase.
- [ ] Add in missing helper text under the new setting "Product, collection, and blog card styles can be updated in theme settings."
- [ ] Default setting should be set to the first option "Use individual card styles" , when I opened the editor it was set to the 2nd option.

duygukalaycioglu (changes requested):
New setting content feedback:

- [x]  Update "Card Style" to "Card style" - to match the editor labels using lowercase.
- [x]  Add in missing helper text under the new setting "Product, collection, and blog card styles can be updated in theme settings."

duygukalaycioglu (commented):
- [x]  Tested card style setting’s default option to be set to “use individual card style” 

- [x]  Tested “use individual card style” setting logic. Media = media global settings, Product cards = product card global settings, Collection cards = collection card global setting
- [x]  Tested “style all as product cards” setting logic. Media and collection cards are using the product card styles.
- [x]  Standalone media - no change to its current setting. (added video section, and tested the global media settings)

Working great! 🎉

duygukalaycioglu (commented):
Is it possible for this branch to use the latest version of Dawn `6.0.2`? 

Some setting labels are old – using the latest version can help avoid _(like Andrew captured [label inconsistency on the previous ticket](https://github.com/Shopify/dawn/pull/1838#issuecomment-1188355030))_ further confusions.

snippets/card-product.liquid
```diff
@@ -30,7 +31,7 @@
       assign ratio = 1
     endif
   -%}
-  <div class="product-card-wrapper card-wrapper underline-links-hover">
+  <div class="card-wrapper underline-links-hover{% if wrapper_class == ''%} product-card-wrapper{% else %} {{ wrapper_class }}{% endif %}">
```
LucasLacerdaUX, line 34: Same here.
```suggestion
  <div class="card-wrapper underline-links-hover {% if wrapper_class == '' %}product-card-wrapper{% else %}{{ wrapper_class }}{% endif %}">
```

sections/collage.liquid
```diff
@@ -213,6 +215,23 @@
       "default": "collage",
       "label": "t:sections.collage.settings.mobile_layout.label"
     },
+    {
+      "type": "select",
+      "id": "card_styles",
+      "options": [
+        {
+          "value": "",
+          "label": "t:sections.collage.settings.card_styles.options__1.label"
```
LucasLacerdaUX, line 224: Do we have instances where we use empty `value` for settings elsewhere?

Wouldn't it be better to use something like `none`, `individual` or something that explicitly tells us what that value is? No strong opinion on this, but it just looked weird to me.

sections/collage.liquid
```diff
@@ -27,7 +27,7 @@
         <div class="collage__item collage__item--{{ block.type }} collage__item--{{ section.settings.desktop_layout }}" {{ block.shopify_attributes }}>
           {%- case block.type -%}
             {%- when 'image' -%}
-              <div class="collage-card color-{{ settings.card_color_scheme }}{% if block.settings.image_padding %} collage-card-spacing{% endif %}">
+              <div class="collage-card{% if section.settings.card_styles == '' %} global-media-settings{% else %} card-wrapper product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```
LucasLacerdaUX, line 30:
```suggestion
              <div class="collage-card {% if section.settings.card_styles == '' %}global-media-settings{% else %}card-wrapper product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```

sections/collage.liquid
```diff
@@ -56,17 +56,19 @@
...
                 card_collection: block.settings.collection,
                 media_aspect_ratio: 'adapt',
                 columns: 2,
-                extend_height: true
+                extend_height: true,
+                wrapper_class: section.settings.card_styles
               %}
             {%- when 'video' -%}
-              <div class="collage-card color-{{ settings.card_color_scheme }}">
+              <div class="collage-card{% if section.settings.card_styles == '' %} global-media-settings{% else %} product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```
LucasLacerdaUX, line 71:
```suggestion
              <div class="collage-card {% if section.settings.card_styles == '' %}global-media-settings{% else %}product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```

sections/collage.liquid
sofiamatulis: Yes we have it in the footer :) https://github.com/Shopify/dawn/blob/main/sections/footer.liquid#L458-L461

I find it better because we dont need to add a class when we dont have to

andrewetchen (approved):
- [x] Card style setting under collage section should be set to "use individual card style" by default
- [x]  When the card style setting set to "use individual card style" media cards should be using the media global settings and all other card types (product and collection) should be using their own respective card type settings. Product cards = product cards global setting, Collection cards: collection cards global setting.
- [x]  When the card style setting set to "style all as product cards" media and collection cards should be using product card global settings and their style, color scheme, and corner style should match the product card within the collage section.

This all looks good to me 🎉 Thanks for adding these changes, Sofia!

snippets/card-product.liquid
```diff
@@ -30,7 +31,7 @@
       assign ratio = 1
     endif
   -%}
-  <div class="product-card-wrapper card-wrapper underline-links-hover">
+  <div class="card-wrapper underline-links-hover {% if wrapper_class == '' %}product-card-wrapper{% else %}{{ wrapper_class }}{% endif %}">
```
LucasLacerdaUX, line 34: This currently breaks product cards everywhere but in the sections which are passing `wraper_class` attribute.
---
If you go to a collection page, for instance, the product cards are not showing any image because they're missing `product-card-wrapper` class:
image

If no `wrapper_class` attribute is passed to this snippet, it won't match `if wrapper_class == ''` condition because `''` isn't the same as "nil" (`''` is a truthy value, btw, everything other than `nil` and `false` is truthy in Liquid). Therefore,  `{{ wrapper_class }}` will render nothing.

Like I mentioned in my previous review, I would suggest revisiting this approach of using `''` as the value of the setting because it can create confusion. The current default state for this snippet is to render `{{ wrapper_class }}` when no `wrapper_class` is passed, and not `product-card-wrapper` which is what we expect.

It's much easier to read something like this:
`{% if wrapper_class != 'none' %}{{ wrapper_class }}{% else %}product-card-wrapper{% endif %}`
and understand that:
- If nothing is passed, it will render `product-card-wrapper` (default attribute value)
- If `none` is passed, it will render `product-card-wrapper` because that is the default style for this component.
- If an override is passed, it will render whatever the value of `wrapper_class` is.

When I see a check for a `''`, I have to actually stop and remember that `''` is truthy in Liquid, but falsy in JavaScript.

I just wrote a lot of information at once so let me know if it's confusing and I can clarify my comments.

sections/collage.liquid
```diff
@@ -27,7 +27,7 @@
         <div class="collage__item collage__item--{{ block.type }} collage__item--{{ section.settings.desktop_layout }}" {{ block.shopify_attributes }}>
           {%- case block.type -%}
             {%- when 'image' -%}
-              <div class="collage-card color-{{ settings.card_color_scheme }}{% if block.settings.image_padding %} collage-card-spacing{% endif %}">
+              <div class="collage-card {% if section.settings.card_styles == '' %}global-media-settings{% else %}card-wrapper product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```
LucasLacerdaUX, line 30: Following my suggestion above, I would do:
```suggestion
              <div class="collage-card {% if section.settings.card_styles == 'none' %}global-media-settings{% else %}card-wrapper product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```
Also, another question: is there any reason for us to output `product-card-wrapper` directly here and not the dynamic value of `{{ section.settings.card_styles }}`? _(check comment below)_

snippets/card-product.liquid
```diff
@@ -216,7 +217,7 @@
     </div>
   </div>
 {%- else -%}
-  <div class="product-card-wrapper card-wrapper underline-links-hover">
+  <div class="card-wrapper underline-links-hover {% if wrapper_class == '' %}product-card-wrapper{% else %}{{ wrapper_class }}{% endif %}">
```
LucasLacerdaUX, line 220: Follow up to the comment above: here, we are outputting the dynamic `wrapper_class` value that matches `{{ section.settings.card_styles }}`.

So for these cards, it's inconsistent with the `image` and `video` ones. I would suggest to either use the future-proof `{{ section.settings.card_styles }}` to render classes, or hardcode `product-card-wrapper` if we think that's the only value we'll have.

sections/collage.liquid
```diff
@@ -56,17 +56,19 @@
...
                 card_collection: block.settings.collection,
                 media_aspect_ratio: 'adapt',
                 columns: 2,
-                extend_height: true
+                extend_height: true,
+                wrapper_class: section.settings.card_styles
               %}
             {%- when 'video' -%}
-              <div class="collage-card color-{{ settings.card_color_scheme }}">
+              <div class="collage-card {% if section.settings.card_styles == '' %}global-media-settings{% else %}product-card-wrapper color-{{ settings.card_color_scheme }}{% endif %}">
```
LucasLacerdaUX, line 71: Even when we have all card types with the same styles (default), changing card styles results in the background color changing for these video blocks.

https://user-images.githubusercontent.com/37168033/181635403-a65a4e8e-95c3-46e3-8991-06962d8a6f71.mp4

Is this intentional? I think it's because one has `color-{{ settings.card_color_scheme }}` while the other doesn't. In the original implementation, `color-{{ settings.card_color_scheme }}` was always there.

LucasLacerdaUX (changes requested):
Great job so far, Sofia! 

Things are looking good but I think I found an issue with the snippet and I would like to suggest revisiting the `''` approach for settings values. I tried to detail in my comments but let me know if it was confusing.

snippets/card-product.liquid
sofiamatulis: I see what you mean 🤔 Good catch, Lucas. I just realized that in the footer example there is no `if` statement so even though it may be passing an empty class (''), it is not an issue since we have no `if` check.

I think instead of passing 'none', which may read as a confusing class we could pass the correct class already so we strip down the logic

Let me know your thoughts
https://github.com/Shopify/dawn/pull/1834/commits/f7be18bb6af07f9c62a97daf2c98ed211600414f
https://github.com/Shopify/dawn/pull/1834/commits/a3b89003e6653b28f93d9f2f613b6f53c62ab7ba

snippets/card-product.liquid
sofiamatulis: Since the card product will always be styled as card product with whichever setting is picked (at least for now), I have stripped the logic from here since I realized it's redundant

sections/collage.liquid
sofiamatulis: If I understand your question correctly. this is expected.

Media doesn't have a color scheme, so it only follows the color scheme once it is selected to follow product cards. Does that make sense? (it follows all settings from prod cards)

https://screenshot.click/29-41-x4t2y-b74jc.mp4

snippets/card-product.liquid
```diff
@@ -11,6 +11,7 @@
     - lazy_load: {Boolean} Image should be lazy loaded. Default: true (optional)
     - show_quick_add: {Boolean} Show the quick add button.
     - section_id: {String} The ID of the section that contains this card.
+    - wrapper_class: {String} Wrapper class for card Default: "" (optional)
```
LucasLacerdaUX, line 14: Are we still using `wrapper_class` on `card-product`? I don't see it being used anywhere.

snippets/card-product.liquid
```diff
@@ -216,7 +217,7 @@
     </div>
   </div>
 {%- else -%}
-  <div class="product-card-wrapper card-wrapper underline-links-hover">
+  <div class="card-wrapper underline-links-hover {{ wrapper_class }}">
```
LucasLacerdaUX, line 220: This is what my product card looks when no product is selected:
image
And this is what it looks like if I select a product:
image

These are the settings:
image

So it looks like we're missing a `product-card-wrapper` class here, because only the color is being applied when using the empty state.

I think we should revert back to what it was before?
```suggestion
  <div class="product-card-wrapper card-wrapper underline-links-hover">
```

snippets/card-collection.liquid
```diff
@@ -21,18 +22,24 @@
...
+  assign card_color_scheme = settings.card_color_scheme
+  assign card_style = settings.card_style
+  if wrapper_class == nil or wrapper_class == 'collection-card-wrapper'
+    assign card_color_scheme = settings.collection_card_color_scheme
+    assign card_style = settings.collection_card_style
+  endif 
 -%}
 
-<div class="collection-card-wrapper card-wrapper animate-arrow">
+<div class="card-wrapper animate-arrow{% if wrapper_class == nil %} collection-card-wrapper{% else %} {{ wrapper_class }}{% endif %}">
```
LucasLacerdaUX, line 33: If we do this:
```suggestion
<div class="card-wrapper animate-arrow {% if wrapper_class and wrapper_class != 'none' %}{{ wrapper_class }}{% else %}collection-card-wrapper{% endif %}">
```

sections/collage.liquid
```diff
@@ -59,14 +59,21 @@
                 extend_height: true
               %}
             {%- when 'collection' -%}
+              {%- liquid
+                  assign wrapper_class = section.settings.card_styles
+                  if section.settings.card_styles == 'none' 
+                    assign wrapper_class = 'collection-card-wrapper'
+                  endif
+               -%}
```
LucasLacerdaUX, line 67: We will be able to remove this:
```suggestion
```

sections/collage.liquid
```diff
@@ -59,14 +59,21 @@
...
               {% render 'card-collection',
                 card_collection: block.settings.collection,
                 media_aspect_ratio: 'adapt',
                 columns: 2,
-                extend_height: true
+                extend_height: true,
+                wrapper_class: wrapper_class
```
LucasLacerdaUX, line 73: If we only have the choice of `product-card-wrapper` and `product-card`, would it make sense for this parameter to be a boolean? If we have plans for more options, just ignore my question :D

Like I said in my previous review, I noticed we're using `product-card-wrapper` hardcoded all over this file except for, where we use the setting value.

If we're okay with having it hardcoded in `card-collection` too, we could do something like this there:
```liquid
{% if override_card_style %}product-card-wrapper{% else %}collection-card-wrapper{% endif %}
```

LucasLacerdaUX (commented):
Looking good. Just a few ideas to refactor the code. 

Most of them are optional, except for the bug in `card-product` which I think we must fix before shipping.

sections/collage.liquid
sofiamatulis: The reason I wanted to write it like this is because eventually we may want to add other card types as options 🤔 Does that make sense? So I can update the hardcoded `product-card-wrapper` class

snippets/card-product.liquid
sofiamatulis: Good catch! I only updated the top one

sections/collage.liquid
LucasLacerdaUX: Ah, that's okay. Let's keep it as is :)

LucasLacerdaUX (approved):
Looks good to me! :D

sections/collage.liquid
```diff
@@ -213,6 +214,23 @@
...
+        {
+          "value": "none",
+          "label": "t:sections.collage.settings.card_styles.options__1.label"
+        },
+        {
+          "value": "product-card-wrapper",
+          "label": "t:sections.collage.settings.card_styles.options__2.label"
+        }
+      ],
+      "default": "none",
```
melissaperreault, line 230: Hey team, the impact on defaulting to `none` here will have visual changes for merchants doing manual updates of their theme when we make the next release, is this what you expected? 

Asking just in case I lost context, I came across this while investigating media ratios. Otherwise we would need to have it set to `product-card-wrapper` for Dawn for a smoother merchants' update from what I understood from our last sync. Let me know if I got this wrong! 🙏 

We still have the time to fix this as it's not in merchants' hand yet!

Something unclear for me is if merchants did change the theme settings values for `Cards` on their Dawn theme, I am not sure if this assumption above is still true. 🤔 But if they made changes to the `Media` theme settings, keeping it to `none` [would introduce this](https://screenshot.click/13-05-14lf3-g10ct.png). 

[This is before on Dawn](https://screenshot.click/13-58-p6nbw-cqrqa.png), where all `Collage` blocks are styled from `Cards`, not `Media`. You can see it as I added styling to it on the section below. 

cc. @martinamarien for following up on more explanation next week when we see fit to continue to spread the theme updates knowledge. 🙏 

cc. @sofiamatulis @duygukalaycioglu

Issue #1752, opened 2022-05-27
[Header] Transparent header


Oliviammarcello:
Add `Transparent overlay header` setting to the header. The settings will include:

- [ ] A checkbox to enable the feature
- [ ] A transparent logo
- [ ] A separate colour picker for the text colour.

When scrolling or on hover the header will no longer be transparent and revert back to the background settings. [See Figma](https://www.figma.com/file/VIshelzL4USycwoA3QU4Ix/Header?node-id=805%3A5399)

xJon:
Looking forward to an official implementation!

stufreen:
Resolved by https://github.com/Shopify/dawn/pull/1839

PR #1839, opened 2022-07-05
[Header] Transparent Header

LucasLacerdaUX:
**PR Summary:** 

Introduces a group of settings to allow the Header section to be displayed as a transparent overlay.

Closes https://github.com/Shopify/dawn/issues/1752

**Important decisions & limitations**
- No "magic": the transparency is set via a setting and won't be restricted to certain types of sections at the top.  
  - We opted for a non-JS based solution to prevent content jumping (cumulative layout shift) on page load
  - There is currently no way to detect the section at the top of the page via Liquid or CSS and apply styles to the header.
- In this initial release, the feature is **restricted to the homepage only**. This will allow merchants to better control its impact and ensure no accessibility issues are caused by the use of this setting.
- The `Transparent logo image` won't be displayed unless a `Logo image` is set in global Theme Settings.
  - This prevents the transition from text to image logo on scroll, which could result in horizontal jumping of menu items as they would have different sizes.

**Testing steps/scenarios**
- [ ] Enable Transparent Header overlay.
  - [ ] Ensure the setting only impacts the homepage
  - [ ] Try different nav settings (drawer, dropdown and megamenu)
  - [ ] Try different breakpoints: mobile+tablet & desktop. Ensure the transparency works consistently across all viewport sizes.
- [ ] `Transparent Logo` setting:
  - [ ] Set a `Transparent Logo` without setting a logo in theme settings. It shouldn't be displayed.
  - [ ] Set a Logo Image in Theme Settings. Set a `Transparent Logo`. Both should be visible depending on the Header state.
- [ ] Set a `Overlay text color`:
  - [ ] Ensure the overlay color is applied to all visible elements in the header
  - [ ] The overlay color shouldn't be applied on `focus-within` state (except for Dropdown nav) nor in sub-menus

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128146014230)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128146014230/editor)

**Checklist**

### DIFF
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1140,7 +1140,7 @@
           "label": "Use transparent header on homepage"
         },
         "transparent_logo": {
-          "label": "Logo for transparent header",
+          "label": "Logo",
           "info": "Only visible when transparent header is displayed and a logo is selected in [theme settings](/editor?context=theme&category=logo)."
         },
         "transparent_text_color": {
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1134,14 +1134,14 @@
           "label": "Enable language selector"
         },
         "transparent_overlay_header": {
-          "content": "Transparent overlay header"
+          "content": "Transparent header"
         },
         "enable_transparent_header": {
-          "label": "Use transparent overlay header on homepage"
+          "label": "Use transparent header on homepage"
         },
         "transparent_logo": {
-          "label": "Logo for transparent overlay",
-          "info": "Visible when transparent overlay header is displayed. For best results, use a transparent PNG image."
+          "label": "Logo for transparent header",
+          "info": "Only visible when transparent header is displayed and a logo is selected in [theme settings](/editor?context=theme&category=logo)."
         },
         "transparent_text_color": {
           "label": "Overlay text color",
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1063,6 +1063,9 @@
     "header": {
       "name": "Header",
       "settings": {
+        "logo": {
+          "content": "Logo"
+        },
         "logo_help": {
           "content": "Edit your logo in [theme settings](/editor?context=theme&category=logo)."
         },
@@ -1082,7 +1085,8 @@
           }
         },
         "menu": {
-          "label": "Menu"
+          "content": "Menu",
+          "label": "Main menu"
         },
         "menu_type_desktop": {
           "label": "Desktop menu type",
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1137,15 +1137,15 @@
           "content": "Transparent overlay header"
         },
         "enable_transparent_header": {
-          "label": "Enable transparent overlay header on homepage"
+          "label": "Use transparent overlay header on homepage"
         },
         "transparent_logo": {
-          "label": "Transparent overlay logo",
+          "label": "Logo for transparent overlay",
           "info": "Visible when transparent overlay header is displayed. For best results, use a transparent PNG image."
         },
         "transparent_text_color": {
           "label": "Overlay text color",
-          "info": "Ensure your text is readable on all templates. Learn more about [text accessibility](https://help.shopify.com/en/manual/online-store/themes/customizing-themes/accessibility#text-accessibility)."
+          "info": "Ensure your text has enough contrast to be readable. Learn more about [text accessibility](https://help.shopify.com/en/manual/online-store/themes/customizing-themes/accessibility#text-accessibility)."
         },
         "margin_bottom": {
           "label": "Bottom margin"
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1133,6 +1133,20 @@
         "enable_language_selector": {
           "label": "Enable language selector"
         },
+        "transparent_overlay_header": {
+          "content": "Transparent overlay header"
+        },
+        "enable_transparent_header": {
+          "label": "Enable transparent overlay header on homepage"
+        },
+        "transparent_logo": {
+          "label": "Transparent overlay logo",
+          "info": "Visible when transparent overlay header is displayed. For best results, use a transparent PNG image."
+        },
+        "transparent_text_color": {
+          "label": "Overlay text color",
+          "info": "Ensure your text is readable on all templates. Learn more about [text accessibility](https://help.shopify.com/en/manual/online-store/themes/customizing-themes/accessibility#text-accessibility)."
+        },
         "margin_bottom": {
           "label": "Bottom margin"
         },
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -51,11 +51,6 @@
     }
 
     @media screen and (min-width: 990px) {
-      .section-header:not(.scrolled-past-header) .header-wrapper:not(.header-wrapper--dropdown):not(:focus-within) .h2,
-      .section-header:not(.scrolled-past-header) .header-wrapper:not(.header-wrapper--dropdown):not(:focus-within) .header__icon,
-      .section-header:not(.scrolled-past-header) .header-wrapper:not(.header-wrapper--dropdown):not(:focus-within) .header__menu-item,
-      .section-header:not(.scrolled-past-header) .header-wrapper:not(.header-wrapper--dropdown):not(:focus-within) .header__active-menu-item,
-      .section-header:not(.scrolled-past-header) .header-wrapper:not(.header-wrapper--dropdown):not(:focus-within) .header__heading-link,
       .section-header:not(.scrolled-past-header) .header-wrapper--dropdown .h2,
       .section-header:not(.scrolled-past-header) .header-wrapper--dropdown .header__icon,
       .section-header:not(.scrolled-past-header) .header-wrapper--dropdown .header__menu-item,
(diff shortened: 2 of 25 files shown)


sections/header.liquid
```diff
@@ -36,6 +36,39 @@
...
+  {% if section.settings.enable_transparent_header_home and request.page_type == 'index'
+     or section.settings.enable_transparent_header_page and request.page_type == 'page'
+     or section.settings.enable_transparent_header_collection and request.page_type == 'collection'
+     or section.settings.enable_transparent_header_blog and request.page_type == 'blog' or request.page_type == 'article'
+  %}
+    .header-wrapper {
+      position: absolute;
+      width: 100%;
+      transition: background var(--duration-short) ease-in, color var(--duration-short) ease-in;
+    }
```
kjellr, line 48: This code suggestion can be cleaned up / relocated, but for illustration purposes: Let's take a closer look at the transitions when the header appears and changes state. The addition here does adds in a few refinements: 

- It slows down the background color + text color transitions very slightly. While we do want these to be fast, the background transition in particular is pretty intense, and can come across as a "flash" if it's too quick. 
- When the nav sticky setting is set to "scroll up", this change gives the nav a subtle "slide in from top" animation.  This makes it feel like the nav isn't just appearing out of thin air, but is sliding in from it's last known location up top — catching up to you as you scroll. 
- When the nav sticky setting is set to "always", and the logo is set to scale, this introduces a scale transformation to match. This way, the logo doesn't just snap into its new appearance, it scales down gracefully. (The video below includes an identical logo that's scaled down, but it also [works nicely](https://screenshot.click/16-56-g84h1-obyr8.mp4) if the logo is different).

Here are these changes in action: 

https://user-images.githubusercontent.com/1202812/219459600-7ba3da27-af44-4be5-b99b-4d3479105b64.mp4

```suggestion
      transition: background var(--duration-short) ease-in, color var(--duration-short) ease-in;
    }
    {% if section.settings.sticky_header_type != 'always' %}
      .section-header.scrolled-past-header .header-wrapper.gradient:not(:hover):not(:focus-within) {
        animation: headerReveal var(--duration-long) ease;
      }

      @keyframes headerReveal {
        0% {
          background-color: transparent;
          transform: translateY(-2em);
        }
        50% {
          transform: translateY(0em);
        }
        100% {
          background-color: var(--gradient-background);
        }
      }
    {% else %}
      .scrolled-past-header .header--reduce-50 .header__heading-logo {
        animation: headerLogo50 var(--duration-long) ease;
      }

      .scrolled-past-header .header--reduce-25 .header__heading-logo {
        animation: headerLogo25 var(--duration-long) ease;
      }

      @media screen and (min-width: 990px) {
        .scrolled-past-header .header--reduce-50 .header__heading-logo,
        .scrolled-past-header .header--reduce-25 .header__heading-logo {
          transform-origin: 0% 50%;
        }
      }

      @keyframes headerLogo50 {
        0% {
          transform: scale(2);
        }
        100% {
          transform: scale(1);
        }
      }

      @keyframes headerLogo25 {
        0% {
          transform: scale(1.333);
        }
        100% {
          transform: scale(1);
        }
      }
    {% endif %}
```

locales/en.default.schema.json
```diff
@@ -1059,6 +1059,20 @@
...
         },
+        "transparent_overlay_header": {
+          "content": "Transparent overlay header"
+        },
+        "enable_transparent_header": {
+          "label": "Enable transparent overlay header on homepage"
+        },
+        "transparent_logo": {
+          "label": "Transparent overlay logo",
+          "info": "Only displaed when transparent overlay header is used."
```
melissaperreault, line 1070:
```suggestion
          "info": "For best results, use a transparent PNG image. Visible when transparent overlay header is displayed."
```
This aligns more with other language used for similar "conditional" setting".

kjellr:
> When transparent Header is enabled, turn the menu type into Drawer. Open the drawer = The Header reverts to its non transparent state. Close the drawer. The drawer do not get back to its transparent state. Could we do that? 

I'm not seeing some of this behavior — when I click out of the mega menu, [it clears for me](https://screenshot.click/30-58-uk0tm-7fob1.mp4). 

> Should we keep the separator line when transparent header is enabled? I think it could be interesting, but only if it inherits the transparent header’s text color.

The separator line is there right now, but it's pretty subtle. I think it's ok for now, and easy to turn off if it doesn't look right. 

A few issues I noticed: 

- When the menu type is set to mega menu, [the background appears in in two steps](https://screenshot.click/30-44-nwmwb-nbu27.mp4): first the background behind the nav, and then the background behind the second+ level. Can we have this fade in quickly all at once instead?
- When sticky header is set to always, the menu should fade in. Right now it just [appears with no animation](https://screenshot.click/30-51-ynm8t-cxycq.mp4). 
- I'm seeing a bug with the drawer nav on the demo site. When I dismiss the menu, [the overlay still remains](https://screenshot.click/30-55-4t3tg-l9dfo.mp4). 🤔 

I'm also think I'm seeing some _very_ slight weirdness with the logo scale animation, but I'll take a closer look tomorrow morning and leave some detailed notes on that.

LucasLacerdaUX:
@YoannJailin Thanks for the feedback! I'll try to add some of the rationale below, but please feel free to add more details if you want to suggest a change to any of our decisions :D

> Should we keep the separator line when transparent header is enabled? I think it could be interesting, but only if it inherits the transparent header’s text color.

image

The transparent line is there, but is very subtle. Should we make it more visible?

> When menu is set to Dropdown, when opened, only the menus are non transparent. Header stays transparent. When menu is set to Megamenu, mega menus + the whole header becomes non-transparent. Same for Drawer ; When opened, header is non transparent 

This was intentional. When we last met to discuss this, @kjellr suggested that for `Dropdown` menus, the header should remain transparent when focused as it was less distracting. For `Drawer` and `Megamenu`, because it can take up to 100% width, we decided that the header background should always be transparent. The reason we opted for `focus-within` instead of `hover` was also to make it less distracting. 

> When transparent Header is enabled, turn the menu type into Drawer. Open the drawer = The Header reverts to its non transparent state. Close the drawer. The drawer do not get back to its transparent state. Could we do that? 

Currently, once the drawer is open the background is set to non-transparent. Once you close the drawer, the header remains non-transparent because the focus is redirected to the hamburger button. If you interact with the keyboard, using ESC to close the drawer, it becomes easier to see why the focus should go back to the hamburger (and therefore, why the header must remain transparent).

> Same when menu is set to Mega menu. When we open one, the header reverts to non transparent state. When clicking out of the mega menu, the header stays non transparent.

This is how it looks for me.

https://user-images.githubusercontent.com/37168033/230264034-098fcc3a-0c22-4d2c-a8d0-f7d423bfe97a.mp4

Are there any steps to reproduce the behaviour you described?

kjellr:
> So, I've investigated a bit and this is apparently a bug with the script not being loaded in the editor as you change the nav type. It happens on main too. If you click Save after changing the nav type, it should then work as expected.

Thanks for figuring that out! I've opened https://github.com/Shopify/dawn/issues/2498 to track it.

kjellr:
Thanks, this is feeling much better! I don't mind the logo animation — it seems mostly ok. It _might_ feel just a little smoother if the logos were fading in/out rather than changing the `display` properties entirely. I also have a vague sense that it'd be more performant if it animated using `transform: scale()` instead of transitioning the `width`. But I doubt it's a drastic difference. 

> Come up with an animation for the Mega Menu (it currently just shows up with no animation) so that this two-step transition feels smoother.

Yeah, I think this is the right approach. Let me know if you could use some help trying things out!

stufreen:
@kjellr @LucasLacerdaUX 

> I noticed that if you drag the announcement bar under the header, [it creates some weirdness](https://screenshot.click/26-11-0gl12-r71qp.png). Is that something we can/should clean up?

This is indeed weird. I can't think of how to fix it though without moving it to be a theme setting. Even then, I'm not sure we'd do it.

stufreen:
> 1. It's the same issue as above, yes. It also happens on dropdown menus - it's just more noticeable on mega nav.

Setting `animation-fill-mode: both;` on `details[open] > .header__submenu` seems to improve it

LucasLacerdaUX:
@eugenekasimov Yes, that's intentional. We actually just added a note to our theme settings to highlight that both logos are required.
image

This is to avoid transitioning from a text to an image when you scroll down, which can cause the menus to shift position.

sections/header.liquid
```diff
@@ -187,14 +230,26 @@
...
+                {%- if section.settings.transparent_logo != blank -%}
+                  {%- assign transparent_logo_height = section.settings.transparent_logo | divided_by: section.settings.transparent_logo.aspect_ratio -%}
+                  {{ section.settings.transparent_logo | image_url: width: 600 | image_tag:
+                    class: 'header__heading-logo header__heading-logo--transparent motion-reduce',
+                    widths: widths,
+                    height: transparent_logo_height,
+                    width: settings.logo_width,
+                    alt: logo_alt,
+                    sizes: sizes,
+                    preload: true
```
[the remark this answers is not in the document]
LucasLacerdaUX: Updated to only preload the transparent logo. Also added a check for `transparent_header` so the transparent logo is only rendered in the homepage now.

eugenekasimov (commented):
Hey Lucas. There is a bug appeared after you pushed the lates commit :( 
https://screenshot.click/28-48-bcxs7-2xfam.mp4

LucasLacerdaUX:
@eugenekasimov I've updated so that if you don't have a Transparent Header logo, it will display the image logo.

This will fix both the alignment issue you mentioned and also most of the cases for the weird bug in your latest comment.

> Hey Lucas. There is a bug appeared after you pushed the latest commit :(
https://screenshot.click/28-48-bcxs7-2xfam.mp4

This should be much better now, but it may still happen if there is a significant difference between the heights of transparent and non-transparent logo. The cause for this is that the `focus-within` is applied BEFORE the click event has finished. So as soon as you press your mouse button, it'll apply the focus-within and display the transparent logo - which may result in a logo shift. I expect this to be an edge case, but I'll see if there's anything we can do to reduce this effect when it happens.

eugenekasimov:
Here is another bug I found
https://screenshot.click/28-49-zrd43-3qx0h.mp4

wilfredojrkuan:
How can I use this on my shopify store that is using Dawn Theme? I was able to make my header transparent by myself coding, but this seems like a more native and clean appraoch

xJon:
Still looking forward for this being included in a dawn release 🚀

LucasLacerdaUX:
Closing this as it's quite out of date and the solution did not meet our standards to be included in a Dawn release.

PR #1840, opened 2022-07-05
[Slideshow] Fix empty text content gap on mobile

allylane:
**PR Summary:** 

Currently if you use a Slideshow section, with no content, it creates a gap underneath the image.
Screen Shot 2022-07-05 at 11 29 16 AM

This PR checks if there is any text content and introduces padding if there is.
No text:
Screen Shot 2022-07-05 at 11 37 28 AM
With text:
Screen Shot 2022-07-05 at 11 37 16 AM

**Testing steps/scenarios**
- [ ] [Open this branch in the editor](https://os2-demo.myshopify.com/admin/themes/128316899350/editor)
- [ ] Add a slidshow
- [ ] Experiment with adding and removing content

### DIFF
--- a/sections/slideshow.liquid
+++ b/sections/slideshow.liquid
@@ -114,7 +114,7 @@
           {%- endif -%}
         </div>
         <div class="slideshow__text-wrapper banner__content banner__content--{{ block.settings.box_align }} page-width{% if block.settings.show_text_box == false %} banner--desktop-transparent{% endif %}">
-          <div class="slideshow__text banner__box content-container content-container--full-width-mobile color-{{ block.settings.color_scheme }} gradient slideshow__text--{{ block.settings.text_alignment }} slideshow__text-mobile--{{ block.settings.text_alignment_mobile }}">
+          <div class="slideshow__text banner__box {% if block.settings.heading != blank or block.settings.subheading != blank or block.settings.button_label != blank %}banner__box-padding {% endif %}content-container content-container--full-width-mobile color-{{ block.settings.color_scheme }} gradient slideshow__text--{{ block.settings.text_alignment }} slideshow__text-mobile--{{ block.settings.text_alignment_mobile }}">
             {%- if block.settings.heading != blank -%}
               <h2 class="banner__heading {{ block.settings.heading_size }}">{{ block.settings.heading | escape }}</h2>
             {%- endif -%}
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -305,7 +305,6 @@
 }
 
 .banner__box {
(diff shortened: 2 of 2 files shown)


sections/slideshow.liquid
```diff
@@ -114,7 +114,7 @@
           {%- endif -%}
         </div>
         <div class="slideshow__text-wrapper banner__content banner__content--{{ block.settings.box_align }} page-width{% if block.settings.show_text_box == false %} banner--desktop-transparent{% endif %}">
-          <div class="slideshow__text banner__box content-container content-container--full-width-mobile color-{{ block.settings.color_scheme }} gradient slideshow__text--{{ block.settings.text_alignment }} slideshow__text-mobile--{{ block.settings.text_alignment_mobile }}">
+          <div class="slideshow__text banner__box {% if block.settings.heading != blank or block.settings.subheading != blank or block.settings.button_label != blank %}banner__box-padding {% endif %}content-container content-container--full-width-mobile color-{{ block.settings.color_scheme }} gradient slideshow__text--{{ block.settings.text_alignment }} slideshow__text-mobile--{{ block.settings.text_alignment_mobile }}">
```
andrewetchen, line 117: In the context of the Slideshow section, taking a look at the current block level setting for the Image slide, **Show container on desktop**. If no textual content exists, the container will still display but we also have the option of unchecking the "Show container on desktop" setting to hide this empty, blank container:



I wonder if it would make more sense to add a new setting of **Show container on mobile** to the Slideshow section. This would remove the blank container that appears when no textual content exists and we'd only need to add one Liquid if condition vs. multiple (less code):



This pattern / setting already exists in the **Image banner** section:



And lastly, for parity, we should also consider adding whatever we decide in this PR to the **Email signup banner** section:

andrewetchen (commented):
Hey @allylane, thanks for making these changes.

I was curious to know if you've ever explored adding a similar setting to that of "Show container on desktop" but, "Show container on mobile"?

See below

sections/slideshow.liquid
ludoboludo, line 117: For the ease of reading we tend to put the conditional statements for classes at the end. So that when reading you can see first the ones that are applied at all times, and then the ones that based on a condition.
```suggestion
          <div class="slideshow__text banner__box content-container content-container--full-width-mobile color-{{ block.settings.color_scheme }} gradient slideshow__text--{{ block.settings.text_alignment }} slideshow__text-mobile--{{ block.settings.text_alignment_mobile }}{% if block.settings.heading != blank or block.settings.subheading != blank or block.settings.button_label != blank %} banner__box-padding{% endif %}">
```

Issue #1709, opened 2022-05-18
Lazy loading issues with section medias above the fold


sofiamatulis:
**Describe the enhancement you'd like**
When the image banner section is placed above the fold by the merchant,  there should be logic to update `loading=lazy` to auto/eager 

This was noticed in the image banner, but we can check other sections with images as well since sections can be reordered 

Original comment by @bacounts 

https://github.com/Shopify/dawn/pull/1701#issuecomment-1129353390

ludoboludo:
@bacounts this is currently an issue for any section. We don't have an easy way to tell which section is placed above the fold. So it's definitely something that'd be great to know and address but tricky as any combination of sections can exist. 

We've tweaked it some places where we could tell that in most cases x or y component is going to sit at the top. 
With the principles we have for the theme we didn't want to use JS to tackle it. I think one way to go about it would be to use an intersection observer on each section and remove/add the attribute where needed. 
There was also some new properties that sounded very useful for those use cases, `content-visibility`: https://web.dev/content-visibility/
But the support for this so far is pretty low.

ludoboludo:
[Here is a PR](https://github.com/Shopify/dawn/pull/1586) that was adding a setting for `image-banner` to decide whether to load the media with `eager` or `lazy`. 
Not the ideal approach but could be an option. Though it would need to be on all sections with medias and with a simpler label probably 🤔

tyleralsbury:
> Shopify would need to add the ability to check if the section is the first one, but as I understood it with previous talk with Thibaut, this would introduce dependencies between sections and would break one big assumption that sections are independant.

That's right.  All of the platform based solutions we've tossed around thus far have had this critical issue.  When a section is updated in the theme editor, only that section is affected.  Any solution that changes the value of something like `is_first_section` or `section_index`, etc... will always have the same problem where another section will be affected by it, meaning the assumption the editor makes that regarding which section could have changes is broken.

---

That said, the PR that was made that Ludo referenced above proposes the use of a setting.  It would put the onus on the merchant to handle it, but does at least give them control.  If we default it to lazy the experience won't get any worse.

Here's my take on what would be required to use it in Dawn:  https://github.com/Shopify/dawn/pull/1586#issuecomment-1147714310

[...]

PR #1844, opened 2022-07-07
Adding lazy loading setting for image optimization

tyleralsbury:
**PR Summary:** 

Added a setting for merchants to be able to indicate when a section is above the fold.  This setting influences the image loading strategy used to improve LCP (Largest Contentful Paint) performance.

I only did it for three sections so far - Image banner, slideshow, and featured collection as a proof of concept.  Each section has its own considerations depending on the provided settings.  For example, Featured Collection allows a merchant to set the number of columns for their grid, so we may want to eager load one or two of the images, depending on that setting.

**Other considerations**

If we decide this is a good approach, we should do this for all of the other sections that have images as well.  Alternatively, we could also pick and choose (not great for parity) based on sections we believe are most likely to be above the fold e.g. hero style sections like image with text.  It feels better to do it for all of them, but maybe there's some nuance there.

**Testing steps/scenarios**

I ran some test using https://web.dev/measure/ to verify the changes to LCP.  We can use https://www.webpagetest.org/ as well, or even just Chrome's dev tools to test the improvements.  Here are some results:

#### 1 image - Lazy
Image banner first, 1 image - lazy

#### 1 image - Lazy LCP

Image banner first, 1 image - lazy LCP

#### 1 image - Lazy reason

Image banner first, 1 image - lazy reason

#### 1 image - Eager

Image banner first, 1 image - eager

#### 1 image - Eager LCP

Image banner first, 1 image - eager LCP

#### 2 images - Lazy
Image banner first, 2 images - lazy

#### 2 images - Lazy LCP
Image banner first, 2 images - lazy LCP

#### 2 images - Lazy reason
Image banner first, 2 images - lazy reason

#### 2 images - Eager
Image banner first, 2 images - eager

#### 2 images - Eager LCP
Image banner first, 2 images - eager LCP

#### After

**Demo links**

- [Store](url)
- [Editor](url)

**Checklist**

### DIFF
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -373,6 +373,13 @@
         "options__4": {
           "label": "Extra large"
         }
+      },
+      "optimization": {
+        "content": "Optimization"
+      },
+      "is_first_section": {
+        "label": "This is the first section on the page",
+        "info": "Optimize images to load faster when they are above the fold."
       }
     },
     "announcement-bar": {
--- a/sections/featured-collection.liquid
+++ b/sections/featured-collection.liquid
@@ -47,21 +47,46 @@
 
 <div class="color-{{ section.settings.color_scheme }} isolate gradient">
   <div class="collection section-{{ section.id }}-padding{% if section.settings.full_width %} collection--full-width{% endif %}">
-      <div class="collection__title title-wrapper title-wrapper--no-top-margin page-width{% if show_mobile_slider %} title-wrapper--self-padded-tablet-down{% endif %}{% if show_desktop_slider %} collection__title--desktop-slider{% endif %}">
-        {%- if section.settings.title != blank -%}
-          <h2 class="title {{ section.settings.heading_size }}">{{ section.settings.title | escape }}</h2>
-        {%- endif -%}
-        {%- if section.settings.description != blank or section.settings.show_description and section.settings.collection.description != empty -%}
-          <div class="collection__description {{ section.settings.description_style }}">{%- if section.settings.show_description -%}{{ section.settings.collection.description }}{%- else -%}{{ section.settings.description }}{% endif %}</div>
-        {%- endif -%}
-      </div>
+    <div class="collection__title title-wrapper title-wrapper--no-top-margin page-width{% if show_mobile_slider %} title-wrapper--self-padded-tablet-down{% endif %}{% if show_desktop_slider %} collection__title--desktop-slider{% endif %}">
+      {%- if section.settings.title != blank -%}
+        <h2 class="title {{ section.settings.heading_size }}">{{ section.settings.title | escape }}</h2>
+      {%- endif -%}
+      {%- if section.settings.description != blank or section.settings.show_description and section.settings.collection.description != empty -%}
+        <div class="collection__description {{ section.settings.description_style }}">
+          {%- if section.settings.show_description -%}
+            {{ section.settings.collection.description }}
+          {%- else -%}
+            {{ section.settings.description -}}
+          {%- endif %}
+        </div>
+      {%- endif -%}
+    </div>
 
     <slider-component class="slider-mobile-gutter{% if section.settings.full_width %} slider-component-full-width{% endif %}{% if show_mobile_slider == false %} page-width{% endif %}{% if show_desktop_slider == false and section.settings.full_width == false %} page-width-desktop{% endif %}{% if show_desktop_slider %} slider-component-desktop{% endif %}">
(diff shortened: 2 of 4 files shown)


bacounts:
@bredowmax - this topic has actually been discussed at length between a few PR's like https://github.com/Shopify/dawn/issues/1709#issuecomment-1148279349. Some of the issue include utilizing a javascript-free solution, that is cross-browser compatible, and keeps Shopify sections independent of each other. While Tylers solution may not be 100% ideal, it is better than not addressing this issue at all.

tyleralsbury:
Thanks for the comments! 

Yeah, it would be absolutely ideal for this to be something that the Shopify platform could figure out itself without any input from either theme developers or merchants.  In a perfect world, we would have a backwards compatible solution that improves image loading for all storefronts without the need for any theme updates.

The implementation in this PR definitely leaves a lot of room for error on the part of merchants - even if we were to give a more fleshed out explanation and they understood what the feature was for, it's very easy to imagine a merchant switching around their layout and not remembering to go back and change these settings.

But, on the upside, at least this optimization could be included in the theme without the need for any platform updates, which historically can take quite a bit longer than theme updates like this.  If the platform did get an update later, we could remove this setting, so it doesn't lock us into anything to go forward with this approach in the short term.

I did test with all of the images on `eager` with three sections on the page and it didn't crush the score too badly - the more sections there are, the worse that would be.  I defaulted the setting to `false` so that if a merchant more or less ignores it, they wouldn't have worse loading behaviour than they do today.

tyleralsbury:
> One rule: first section would generally always be above-the-fold. Second section in 50% of cases as well. Third section only few times.
> 
> Also - same section may end up above-the-fold on one device type but not another

As a note for this - I'm primarily concerned with Mobile in this PR, and I think it's the thing to optimize for in general.  I imagine that if this were done server-side, it might actually be able to account for mobile v.s. desktop (based on user agent), but in terms of where this sort of thing would have the biggest impact, mobile is definitely the priority due to slower internet speeds on average.

andrewetchen:
Thanks Tyler!

> The implementation in this PR definitely leaves a lot of room for error on the part of merchants - even if we were to give a more fleshed out explanation and they understood what the feature was for, it's very easy to imagine a merchant switching around their layout and not remembering to go back and change these settings.
>
>But, on the upside, at least this optimization could be included in the theme without the need for any platform updates, which historically can take quite a bit longer than theme updates like this. If the platform did get an update later, we could remove this setting, so it doesn't lock us into anything to go forward with this approach in the short term.
>
>I did test with all of the images on eager with three sections on the page and it didn't crush the score too badly - the more sections there are, the worse that would be. I defaulted the setting to false so that if a merchant more or less ignores it, they wouldn't have worse loading behaviour than they do today.

I do agree that giving merchants this control is good and how providing a solution is better than no solution -- especially for merchants who are performance minded but as you've implied, I'm a little apprehensive since merchants might forget to uncheck this setting when switching their layouts around -- my experience working in Shopify theme support can substantiate this although your suggestion of providing a more fleshed out explanation might help with this.

I do like the sound of this being a built into the theme editor but since these types of changes typically take longer than theme updates, perhaps providing merchants with a good and feasible solution now could be ideal OR maybe we should hold off 🤔

As you can tell I'm a bit conflicted as I see the positives and negatives here but thought I'd share my perspective.

kjellr:
> But, on the upside, at least this optimization could be included in the theme without the need for any platform updates, which historically can take quite a bit longer than theme updates like this.

I think this is a great proof-of-concept — it's great to see such immediate, obvious results. I don't think it makes sense to ship the setting to merchants though. It puts too much onus on the user to manually fix something that we should really be fixing ourselves automatically (I can imagine users asking: "Why doesn't Shopify already know which section is the first section on my page?" 😅). 

In general, it's a clear workaround for a platform-wide problem. And while deploying it may provide an immediate benefit for some sites, I worry that it would inadvertently put less pressure and immediacy on the need for a true, automatic solution on the platform level.

bakura10:
I am still 100% against having this merged in Dawn, as it will create a precedent for merchants that will expect us (theme developers) to implement this hacky solution that forces merchant to understand something they should not have to care, and maintain this whenever they add, remove or re-order sections.

I have described a completely automatic, backward compatible and JS dependant free and compatible with existing themes approach here: https://github.com/Shopify/dawn/issues/1709#issuecomment-1148096866

As a back end developer originally, I am pretty sure this would be pretty easy to implement from the backend point of view as well. Please do not introduce to Dawn an ugly solution to fix something that should be done at the platform level. If Shopify want to fix this, there is a not so complicated approach that can be done already.

PR #1864, opened 2022-07-20
[Header] New logo position

### LINKED ISSUE #1416: [Header] Logo position enhancements
### Add alternative logo position options on both desktop and mobile to provide additional flexibility.  
Screens on Figma can be [found here.](https://www.figma.com/file/VIshelzL4USycwoA3QU4Ix/Header?node-id=449%3A4122)

Screenshot 2022-02-25 at 2 17 56 PM
Screenshot 2022-02-25 at 2 18 04 PM

image

LucasLacerdaUX:
**PR Summary:** 
Introduces a new theme setting that allows you to control the position of the logo on mobile and adds a new `Middle Center` option to the desktop logo position setting.

Fixes #1416

- **Desktop Logo Position = Middle center -** Used same approach for other position modifications: change the `grid-template-areas` attribute to reorder items, and changed logo position in Liquid by checking`if section.settings.logo_position == 'middle-center'`.
- **Mobile Logo Position = Middle center -** Added a new `header--mobile-{{ position }}` modifier class.

**Other considerations**
- ~**TBD - Accessibility:** Validate that the change to the tab order is valid in this scenario. We've used a similar approach for other `Logo position` options, and this one also keeps the logo `(<h1>)` as the first element on the page, so I need to double check if that is okay.~ Decided to duplicate the logo. It's just better to follow the natural tab order and it's such a simple thing to do.

**Testing steps/scenarios**
- [ ] Change `Mobile logo position` from `Center` to `Left` and test both mobile and tablet viewports.
- [ ] Test all `Desktop logo position` values.
- [ ] Test the new `Middle center` desktop logo position and make sure that it works with:
  - [ ] Small menus
  - [ ] Large menus 
  - [ ] Empty menus
  - [ ] Text logos
  - [ ] Image logos

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128383025174)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128383025174/editor)

**Checklist**
- [ ] Notified the [help.shopify.com](https://help.shopify.com) team about updates to theme settings (current contact: Kimli) ***waiting for UX approval / finalized settings***


Oliviammarcello:
Looking good! 👏 Just one small issue:
- [ ] The logo on mobile changes size when the position changes from center to left. [Video](https://screenshot.click/21-19-jyxxo-glx64.mp4)

sections/header.liquid
```diff
@@ -123,7 +123,7 @@
   </symbol>
 </svg>
 <{% if section.settings.enable_sticky_header %}sticky-header{% else %}div{% endif %} class="header-wrapper color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} header-wrapper--border-bottom{% endif %}">
-  <header class="header header--{{ section.settings.logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
+  <header class="header header--{{ section.settings.logo_position }} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
```
sofiamatulis, line 126:
```suggestion
  <header class="header header--{{ section.settings.logo_position }} header-mobile--{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
```

Because the position is a modifier, does this make more sense? 🤔

LucasLacerdaUX: I personally think because `header` is the element, a `--mobile-center` makes more sense than adding a `header-mobile` class with a `--center` modifier 👀

Would love to get other reviewer's thoughts on this.

locales/en.default.schema.json
```diff
@@ -939,6 +942,18 @@
...
+        "mobile_logo_position": {
+          "label": "Mobile logo position",
+          "options__1": {
+            "label": "Center"
+          },
+          "options__2": {
+            "label": "Left"
+          }
         }
       }
```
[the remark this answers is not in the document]
LucasLacerdaUX: Good catch. 

Removed! :D

sections/header.liquid
sofiamatulis: Fair point 🤔 My only thought is that we have

header--{{ section.settings.logo_position }} 
and
header--mobile-{{ section.settings.mobile_logo_position }}

Where the modifiers should both exist so they are not the same modifier
But they are both modifiers on `header` which I find confusing since they should both exist

Does that make sense?

andrewetchen: I don't have a very strong opinion on this but I'll share my thoughts ❤️ 

My suggestion would be to update the `mobile_logo_position` setting values to `mobile-center` and `mobile-left` then the above would simply become `header--{{ section.settings.mobile_logo_position }}`.

Another option is to add `-mobile` after the modifier: `header--{{ section.settings.mobile_logo_position }}-mobile`. And to Sofia's point, I think the confusion potential is there but at the same time, when I first looked at this, I didn't find it confusing but I'm often in favour in improving readability where we can:

andrewetchen (approved):
- [x] Change `Mobile logo position` from `Center` to `Left` and test both mobile and tablet viewports.
- [x] Test all `Desktop logo position` values.
- [x] Test the new `Middle center` desktop logo position and make sure that it works with:
  - [x] Small menus
  - [x] Large menus
  - [x] Empty menus
  - [x] Text logos
  - [x] Image logos

Everything looks good to me 🎩

Thanks for making these changes, Lucas! And thanks to the UX'ers too! I think a lot of merchant's will be really happy with having a dedicated mobile header/logo layout. I'd also imagine that a lot of folks in support will like this too 😉 

I left some feedback on Sofia's comment here: https://github.com/Shopify/dawn/pull/1864/files?diff=unified&w=0#r927029587

sections/header.liquid
sofiamatulis: I think I would keep header since `mobile-center` would be more of a generic class and not specific to this file. 🤔 Thoughts?

We can keep it like how it is, but thought id share these thoughts 😅

sofiamatulis: @LucasLacerdaUX I dont have any strong feelings either way, but thought i'd bring it up :)

thank you for explaining it 🙌

LucasLacerdaUX:
> The logo on mobile changes size when the position changes from center to left. [Video](https://screenshot.click/21-19-jyxxo-glx64.mp4)

I've been looking at this comment @Oliviammarcello shared for the past hour and I wasn't able to find a good solution. From an implementation point of view, it's working exactly as intended, but the side effect of our approach is that there's more spacing available when the logo is left-aligned so the logo becomes bigger. For it look "centered", we have to force whitespace.

Maybe my brain is too slow today so I'll share an explanation below to see if other devs have input on how to improve this. 

-----
1. This is what a "centered" logo looks like. The space taken by the hamburger menu is exactly what it needs (including some extra to increase the tap area):
image

2. But in order for it to look "centered" relative to the entire bar, we had to add some extra spacing to the hamburger menu so elements on both left and right take the same width. This is how it looks when you selected `Mobile logo position: Center`:
image

3. If we were to keep the same approach and simply update the logo position to left, we wouldn't achieve to the look we want: 
image

4. So in order to push the logo to the left, I had to reduce the hamburger size to the amount we need, like in Image 1, and allow the logo area to fill the rest of the available space:
image

In the screenshots above, the logo is always taking the maximum value set by `Custom logo width`, so that's why they have the same size in `Center` and `Left` configurations. But when we have smaller viewport, it changes from this (142px)
image
to this (173px):
image
simply because we have more space available. 

------

**So, two questions:**
- How big of an issue is this? Is it a blocker? 
  - We're still respecting the maximum logo width defined in `Custom logo width`.
  - The logo won't change from `Center` to `Left` in any buyer-facing situation. The only time we can see the logo fills more space is when switching settings in the Editor.
- Does anyone have an idea on how to fix this issue?

sections/header.liquid
andrewetchen: Thanks for sharing your thoughts, Lucas!

Sorry for the can of 🪱's 😅 

Lucas, since we now have a good discussion around this topic in this thread + since both Sofia and I don't have very strong feelings, I think you can leave it as is 🙌🏼

Oliviammarcello (approved):

sofiamatulis:
@Oliviammarcello @LucasLacerdaUX Not something that needs to be addressed on this PR, but I wonder if we should rename this setting 🤔 If the merchant selects width of 250px on mobile it would be way smaller. We dont want it to be 250 to avoid this: https://screenshot.click/03-25-61qdu-5h30o.mp4

But I wonder if it's confusing for the merchant (maybe not), but figured id raise this 🤔 I wonder if there are two things we could do

- Have the width only affect desktop. Instead of it being called `Custom logo width` it would be `Desktop Custom logo width`.
- Have a different width setting for mobile and change the max size to smaller than 250
- Maybe rename to `Max custom logo width` ?

Just a few thoughts but not something we need to action on :)

Another note is that currently on the password page we have the same setting, but because it's the only thing in the row it is able to get to 250: https://screenshot.click/03-29-dscz6-b7g6x.png So it behaves differently on both templates

sofiamatulis (commented):
Small nitpicks :) Looks great

sections/header.liquid
```diff
@@ -490,6 +492,30 @@
       {%- endif -%}
     {%- endif -%}
 
+    {%- if section.settings.logo_position == 'middle-center' -%}
+      {%- if request.page_type == 'index' -%}
+        <h1 class="header__heading">
+      {%- endif -%}
+          <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
+            {%- if section.settings.logo != blank -%}
+              {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
```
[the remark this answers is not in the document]
LucasLacerdaUX: Not sure if this is what you were referring to, but yes, we do need to assign the variable first because it's being modified (`escape` and `default` filters)

If it was `section.settings.logo.alt`, we could simply pass it as that `alt: section.settings.logo.alt `, and passing it like this: `alt: section.settings.logo.alt | default: shop.name | escape` is invalid.

sofiamatulis (approved):
🎉

PR #1881, opened 2022-07-28
fix: drawer - typing in quantity and pressing enter empties cart

### LINKED ISSUE #1868: Cart drawer - Manually typing in quantity and pressing enter empties cart

When cart is set to Drawer, using the keyboard to enter a quantity and then pressing enter empties the cart. Cart is emptied after entering a quantity using the keyboard + pressing Enter to submit the new quantity

Enter a quantity using the keyboard, cart should update to reflect that change

- Dawn Version: 6.0.2
- Chrome Version 103.0.5060.134
- macOS Version 12.4 

**Possible solution**

treboryx:
**PR Summary:** 

Fixes #1868.

Prevent trash can click event from triggering when entering a manual value

**Other considerations**

**Decision log**

|---|---|---|---|---|
| 1 |   |   |   |   |

**Testing steps/scenarios**

**Demo links**
Before

https://user-images.githubusercontent.com/5408097/181467315-056191b3-eef4-439f-9941-7ca9da9067d3.mov

After

https://user-images.githubusercontent.com/5408097/181467357-74057519-6a40-4ea5-b8b0-26ac7fc359f5.mov

**Checklist**

### DIFF
--- a/assets/cart.js
+++ b/assets/cart.js
@@ -3,8 +3,10 @@ class CartRemoveButton extends HTMLElement {
     super();
     this.addEventListener('click', (event) => {
       event.preventDefault();
-      const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
-      cartItems.updateQuantity(this.dataset.index, 0);
+      if (event.detail !== 0) {
+        const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
+        cartItems.updateQuantity(this.dataset.index, 0);
+      }
     });
   }
 }
@@ -66,7 +68,7 @@ class CartItems extends HTMLElement {
       sections_url: window.location.pathname
     });
 
-    fetch(`${routes.cart_change_url}`, {...fetchConfig(), ...{ body }})
+    fetch(`${routes.cart_change_url}`, { ...fetchConfig(), ...{ body } })
       .then((response) => {
         return response.text();
       })
@@ -87,7 +89,7 @@ class CartItems extends HTMLElement {
         }));
 
         this.updateLiveRegions(line, parsedState.item_count);
-        const lineItem =  document.getElementById(`CartItem-${line}`) || document.getElementById(`CartDrawer-Item-${line}`);
+        const lineItem = document.getElementById(`CartItem-${line}`) || document.getElementById(`CartDrawer-Item-${line}`);
         if (lineItem && lineItem.querySelector(`[name="${name}"]`)) {
           cartDrawerWrapper ? trapFocus(cartDrawerWrapper, lineItem.querySelector(`[name="${name}"]`)) : lineItem.querySelector(`[name="${name}"]`).focus();
         } else if (parsedState.item_count === 0 && cartDrawerWrapper) {
@@ -108,7 +110,6 @@ class CartItems extends HTMLElement {
     if (this.currentItemCount === itemCount) {
       const lineItemError = document.getElementById(`Line-item-error-${line}`) || document.getElementById(`CartDrawer-LineItemError-${line}`);
       const quantityElement = document.getElementById(`Quantity-${line}`) || document.getElementById(`Drawer-quantity-${line}`);
-
       lineItemError
         .querySelector('.cart-item__error-text')
         .innerHTML = window.cartStrings.quantityError.replace(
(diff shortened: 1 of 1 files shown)


treboryx:
I signed it, it's stuck!

andrewetchen:
Hi @treboryx, thanks for your contribution.

Regarding the CLA, based on the following CLA action run: https://github.com/Shopify/dawn/actions/runs/2774103560



https://github.com/Shopify/shopify-cla-action#use-case states the following:

>After signing the CLA, contributors can comment "I've signed the CLA!" on the PR to re-run the test suite.

Please go ahead and add another comment that says, "I've signed the CLA!" or "I have signed the CLA". This will successfully re-trigger the CLA status check on this PR.

I'm here to help so do please let me know if you're still having issues with signing the CLA 👍🏼

ludoboludo:
Hey @treboryx 👋 

Thanks for taking the time to look into the issue and come up with an approach to solve it 💚 
I was having a look at the issue and I was troubleshooting, wondering why on `enter` in the input it would trigger that `remove` button. 

There is I think a simpler approach to fix this which is by adding a `type="button"` attribute to the `<button>` itself. The button element gets by default a type of `"submit"`([MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type)) which why I think it was causing an issue. 

Would you be ok with making that change in this PR ? I can re review when it's done 👍 Otherwise I can go ahead and create the PR myself. Let me know.

treboryx:
> Hey @treboryx 👋
> 
> Thanks for taking the time to look into the issue and come up with an approach to solve it 💚 I was having a look at the issue and I was troubleshooting, wondering why on `enter` in the input it would trigger that `remove` button.
> 
> There is I think a simpler approach to fix this which is by adding a `type="button"` attribute to the `<button>` itself. The button element gets by default a type of `"submit"`([MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type)) which why I think it was causing an issue.
> 
> Would you be ok with making that change in this PR ? I can re review when it's done 👍 Otherwise I can go ahead and create the PR myself. Let me know.

The reason why I didn't do `type="button"` was because when you press enter, it takes you right to the checkout page (pls verify, although I tested multiple times), which is odd, and I remembered that after testing it again.

Could be an entirely different issue but at the time the solution I pushed seemed to be working perfectly and resolving this issue.

ludoboludo:
Ah yes, this is an expected behaviour. It's called [implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) and is the way each browser deals with inputs in a form. 
I don't think we should change this behaviour as it's also something folks will rely on when navigating via keyboard and/or assistive technologies (screen readers, etc). 👍 🙂 
So if we prevented it we would mess with accessibility which we don't want to do.

treboryx:
> Ah yes, this is an expected behaviour. It's called [implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) and is the way each browser deals with inputs in a form. I don't think we should change this behaviour as it's also something folks will rely on when navigating via keyboard and/or assistive technologies (screen readers, etc). 👍 🙂 So if we prevented it we would mess with accessibility which we don't want to do.

gotcha.

assets/cart.js
```diff
@@ -3,8 +3,8 @@ class CartRemoveButton extends HTMLElement {
     super();
     this.addEventListener('click', (event) => {
       event.preventDefault();
-      const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
-      cartItems.updateQuantity(this.dataset.index, 0);
+        const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
+        cartItems.updateQuantity(this.dataset.index, 0);
```
ludoboludo, line 7: quick nitpick, you can revert those indents as well 👍 
```suggestion
      const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
      cartItems.updateQuantity(this.dataset.index, 0);
```

ludoboludo (commented):
Just a little whitespace nitpick otherwise good to go

assets/cart.js
treboryx: good to go

ludoboludo (approved):
Thank you @treboryx

andrewetchen (approved):
This looks good to me 🎉 

@treboryx - thanks for fixing this and for also linting the `cart.js`.

PR #1890, opened 2022-08-03
Fix filter bug to allow filters to be removed

### LINKED ISSUE #1879: Filters not being removed when unchecked

When trying to uncheck filter items, they don't get removed unless it's the last checked filter.

You should be able to uncheck a filter, which would then remove the pill and update the list being shown. 

- Dawn Version: 6.0.2

**Possible solution**

We seem to be keeping track of two filters (mobile version and desktop) and when running our fetch query we're removing the filter from only one of them. We could probably loop through the two array and remove the filter in both instances 🤔 

Looks like it would be related to some changes that were applied in this PR: [#1443](https://github.com/Shopify/dawn/pull/1443/files#diff-9418056d09bded2d61a68790334221e384b6404991d01e3a0f837c75e598fc76)

[video recording of the issue](https://screenshot.click/27-01-c7xv5-2csez.mp4)

sofiamatulis:
**PR Summary:** 

The reason this was happening was due to JS vs no JS.

In the JS experience we have a show more button which means that the full list isnt shown to the buyer at first. 

I discovered this because I noticed that the mobile experience was submitting the correct form: https://screenshot.click/03-21-m5s2g-u34ez.png . You can see here how the input doesnt have checked whereas on desktop youll see that the form was still submitting the input as if it was checked https://screenshot.click/03-47-phe5c-yjg9x.png

So it got me thinking as to why it would go to unchecked on mobile but not desktop, and I realized it is due to us having two lists on desktop (no JS vs JS). That is because we have different designs for JS vs no JS.  And we are hiding the no JS experience with CSS, but Javascript was still picking it up

Which meant that the form was submitting both values and the no JS was overwritting the form on submit: https://screenshot.click/03-51-y2osp-n57n0.png (2 `<ul>` being submitted)

Fixes #1879.

I am removing the no JS list.

**Testing steps/scenarios**
- [ ]  Vertical Filter (JS vs no JS)
- [ ] Drawer Filter (JS vs no JS)
- [ ] Horizontal filter (JS vs no JS)
- [ ] Sorting

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/128443154454/editor)

**Checklist**

### DIFF
--- a/assets/facets.js
+++ b/assets/facets.js
@@ -184,6 +184,8 @@ class FacetFiltersForm extends HTMLElement {
       sortFilterForms.forEach((form) => {
         if (!isMobile) {
           if (form.id === 'FacetSortForm' || form.id === 'FacetFiltersForm' || form.id === 'FacetSortDrawerForm') {
+            const noJsElements = document.querySelectorAll('.no-js-list');
+            noJsElements.forEach((el) => el.remove());
             forms.push(this.createSearchParams(form));
           }
         } else if (form.id === 'FacetFiltersFormMobile') {
--- a/snippets/facets.liquid
+++ b/snippets/facets.liquid
@@ -144,7 +144,7 @@
                         {%- endfor -%}
                       </ul>
                       {% comment %} No show more for no JS {% endcomment %}
-                      <ul class="{% if filter_type != 'vertical' %} facets__list{% endif %} list-unstyled no-js" role="list">
+                      <ul class="{% if filter_type != 'vertical' %} facets__list{% endif %} no-js-list list-unstyled no-js" role="list">
                         {%- for value in filter.values -%}
                           <li class="list-menu__item facets__item">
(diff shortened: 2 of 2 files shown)


dan-menard (approved):
LGTM 👍

metamoni (approved):

assets/facets.js
```diff
@@ -184,6 +184,8 @@ class FacetFiltersForm extends HTMLElement {
       sortFilterForms.forEach((form) => {
         if (!isMobile) {
           if (form.id === 'FacetSortForm' || form.id === 'FacetFiltersForm' || form.id === 'FacetSortDrawerForm') {
+            const noJsElements = document.querySelectorAll('.no-js-list');
```
ludoboludo, line 187: Maybe we could use a selector that isn't adding a new class just to be used/selected in JS 🤔 What do you think about something like: `('.facets-wrap .no-js[role="list"]')` ?

Issue #1403, opened 2022-02-23
Add color scheme to Drawers and apply across components


melissaperreault:
**Describe the expected behavior**

- [ ] Drawers theme setting could benefit [of its own Color scheme setting](https://screenshot.click/22-02-ao8ol-1p7wt.png) so we can support gradient as well. The application of the colors scheme don't make sense:
   - [ ] [Menu navigation](https://screenshot.click/22-02-1h554-6fdw5.png) should inherit `Background 1` and (Background 1 gradient) instead of `Background 2` solid, like Filter drawer [as you can see here](https://screenshot.click/22-02-4xax5-0nxv2.png).
   - [ ] [Filter drawer](https://screenshot.click/22-02-schh0-rgp4m.png) could also benefit from inheriting the `Background 1 gradient` when available, to align how we expect it for `Cart notification`.
   - [ ] Same for the Pickup availability drawer.
   - All of the above could be inherited by the same setting to have harmonious application.

From https://github.com/Shopify/dawn/pull/1304#pullrequestreview-888143249

PR #1907, opened 2022-08-10
[Product information] Add color scheme picker

### LINKED ISSUE #1694: [Product information] Add color picker

eugenekasimov:
**PR Summary:** 

Add `color scheme` and `product information color scheme` pickers to the product content (previously known as product information) settings. Now merchants can easily change color scheme for the product content section.

Fixes #1694.

Add two color pickers to `{% schema %}` in the `main-product.liquid` and add according classes to sections. 

**Other considerations**

**Demo links**

- [Editor](https://os2-demo.myshopify.com/admin/themes/128469893142/editor?previewPath=%2Fproducts%2Fno-featured-image&section=template--15318383722518__main)

**Checklist**


locales/en.default.schema.json
```diff
@@ -1927,7 +1927,8 @@
         },
         "enable_video_looping": {
           "label": "Enable video looping"
-        }
+        },
+        "section_colors_heading": "Colors"
```
andrewetchen, line 1931: From a settings parity perspective, instead of making this apart of `"main-product"`, it would probably make more sense to live under `sections.all.colors` instead.

This way, if we need to add a "Colors" heading to other sections, we don't have to add an additional, `"section_colors_heading": "Colors"` under that section. Then in the section file, we can simply reference this heading the same way we reference the colors, something like this:

```json
{
  "type": "header",
  "content": "t:sections.all.colors.heading"
},
```

The heading can be added here: https://github.com/Shopify/dawn/blob/f538b0a97903173c0298d8b268bcfb130664670a/locales/en.default.schema.json#L368-L386

sections/main-product.liquid
```diff
@@ -1,6 +1,6 @@
 <section
   id="MainProduct-{{ section.id }}"
-  class="page-width section-{{ section.id }}-padding"
+  class="page-width section-{{ section.id }}-padding color-{{section.settings.color_scheme}}"
```
andrewetchen, line 3:
```suggestion
  class="page-width section-{{ section.id }}-padding color-{{ section.settings.color_scheme }}"
```

sections/main-product.liquid
```diff
@@ -1536,6 +1536,66 @@
       "default": false,
       "label": "t:sections.main-product.settings.enable_video_looping.label"
     },
+    {
+      "type": "header",
+      "content": "t:sections.main-product.settings.section_colors_heading"
```
andrewetchen, line 1541: As mentioned earlier, if we add this heading under the `sections.all.colors` [here](https://github.com/Shopify/dawn/blob/f538b0a97903173c0298d8b268bcfb130664670a/locales/en.default.schema.json#L368-L386): then this would need to be updated to reflect that.

andrewetchen (commented):
Hey Eugene, I've added some initial feedback.

Also, do you know if these changes need to be added to the featured collection section as well? Feel free to reach out to UX for clarification.

locales/en.default.schema.json
ludoboludo: We actually had a chat about it and since it's only used in one place for now thought it would make sense for it to live on the main-product section. 

But if we're doing it on featured product as well that would probably be a good idea 🤔

andrewetchen: >But if we're doing it on featured product as well that would probably be a good idea 🤔

I agree 👍🏼 

I was also going to say that the Collapsible content section could also probably use this "Colors" heading:

image

sections/main-product.liquid
```diff
@@ -296,7 +296,7 @@
     <div class="product__info-wrapper grid__item{% if settings.page_width > 1400 and section.settings.media_size == "small" %} product__info-wrapper--extra-padding{% endif %}">
       <div
         id="ProductInfo-{{ section.id }}"
-        class="product__info-container{% if section.settings.enable_sticky_info %} product__info-container--sticky{% endif %}"
+        class="color-{{ section.settings.container_color_scheme }} product__info-container{% if section.settings.enable_sticky_info %} product__info-container--sticky{% endif %}"
```
ludoboludo, line 299: small nitpick, I would put the `color-{{ section.settings.container_color_scheme }}` class at the end of the class list since it's more generic and it follows what you did above 👍

ludoboludo (commented):
Here are a few things that needs to be sorted: 

- what padding value should be applied on the product information container ? [screenshot](https://screenshot.click/10-35-filfq-4grd3.png). Based on the figma linked in the issue it should about `30px` or `3rem` on desktop. When the padding value is applied we can have some logic so that no extra padding is applied if the color scheme for the product page is the same as the one for the product information container. 
- the `color scheme` applied on the product page overall: 
  - should it be full width ? [screenshot](https://screenshot.click/10-37-h3466-e4r77.png) Looks like it based on the figma design linked in the issue

Then finally, we need I think to tweak the settings names. It's very confusing, `container color scheme` ? Which container are we talking about.

locales/en.default.schema.json
eugenekasimov: I like the idea of moving "Colors" heading to `settings.all.colors` there is a chance that we will use it somewhere else.

I talked to @Oliviammarcello and there are a few unclear things about featured product yet. 

I totally agree about the Collapsible content section. I'd prefer to do it in a different PR though.

sections/main-product.liquid
```diff
@@ -1,729 +1,96 @@
...
->
-  {{ 'section-main-product.css' | asset_url | stylesheet_tag }}
-  {{ 'component-accordion.css' | asset_url | stylesheet_tag }}
-  {{ 'component-price.css' | asset_url | stylesheet_tag }}
-  {{ 'component-rte.css' | asset_url | stylesheet_tag }}
-  {{ 'component-slider.css' | asset_url | stylesheet_tag }}
-  {{ 'component-rating.css' | asset_url | stylesheet_tag }}
-  {{ 'component-loading-overlay.css' | asset_url | stylesheet_tag }}
-  {{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
+<div class="color-{{ section.settings.color_scheme }}">
```
ludoboludo, line 1: Im not sure how I feel about having the div wrapping everything 🤔 But due to how we apply the padding on the `<section` element I guess that makes the most sense. Otherwise it could be like this: 



I actually wonder if they should be both `div` element instead 🤔 Cause at the moment we actually have a `section` element within another `section` element.

assets/section-main-product.css
```diff
@@ -298,6 +298,10 @@ fieldset.product-form__input .form__label {
 
 /* Product info */
 
+.different-color-scheme {
+  padding: 2rem;
+}
```
ludoboludo, line 303: I think on mobile we could remove the spacing on the first child of the product info and add a margin-top to the container maybe:

eugenekasimov:
> Then finally, we need I think to tweak the settings names. It's very confusing, `container color scheme` ? Which container are we talking about.

@ludoboludo, I agree with you. It's not that clear. On the other hand, it's something that you can figure out quickly because visual changes are quite obvious.

@Oliviammarcello , what are your thoughts on this?

assets/section-main-product.css
```diff
@@ -298,6 +298,10 @@ fieldset.product-form__input .form__label {
...
+
```
ludoboludo, line 304: Another spacing fix I think to do is removing the top margin on the first element so that there isn't extra spacing here:

sections/main-product.liquid
ludoboludo, line 1: I wonder if drawers triggered from the product information should match the color scheme of it 🤔 [video](https://screenshot.click/12-21-rwzj8-swm88.mp4)

But I think we're wanting to have a specific general color setting for drawers in general in the future. What do you think @Oliviammarcello ?

sections/main-product.liquid
```diff
@@ -1,729 +1,96 @@
...
+  <section
+    id="MainProduct-{{ section.id }}"
+    class="page-width section-{{ section.id }}-padding color-{{ section.settings.color_scheme }}"
```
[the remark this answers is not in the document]
eugenekasimov: Good call, thanks. It seems I forgot remove it from there.

sections/main-product.liquid
Oliviammarcello: Since merchants have the ability to change the drawer colour I wonder if that is good enough. I could see use cases where merchants would want the two to be different since it would apply to all drawers in the theme (cart, filters, etc.)

Oliviammarcello:
> > Then finally, we need I think to tweak the settings names. It's very confusing, `container color scheme` ? Which container are we talking about.
> 
> @ludoboludo, I agree with you. It's not that clear. On the other hand, it's something that you can figure out quickly because visual changes are quite obvious.
> 
> @Oliviammarcello , what are your thoughts on this?

Yeah this one I am hesitant on. We use "container colour scheme" elsewhere in the theme. I'd leave it for now and potentially come back to this for all instances where we use it like the collapsible content section as well.

sections/main-product.liquid
ludoboludo: Yeah Meli had created an issue for it here: https://github.com/Shopify/dawn/issues/1403

assets/section-main-product.css
Oliviammarcello: I agree! Maybe another 1.5rem

ludoboludo:
I think we could potentially change how the section is called. Instead of `Product information` it could be `Product content`. 

As for the `container color scheme`, it could be changed to `Product information color scheme`

As `container color scheme` is too vague. There is a container for the images as well, a container for the heading, description, etc. 🤔

eugenekasimov:
@Oliviammarcello , There is one a minor thing. When Desktop layout set Large, everything looks good to me no matter what color schemes chosen (pic 1 and 2). 

However, when Desktop layout set Medium or Small, the padding is increased from 4rem to 5rem and it quite grabs my attention especially when color schemes are different (pic 3). Is there a reason for this or we can fix it?

Screen Shot 2022-08-12 at 4 23 46 PM (2) 
Screen Shot 2022-08-12 at 4 35 47 PM (2)

Oliviammarcello:
> @Oliviammarcello , There is one minor thing. When Desktop layout set Large, everything looks good to me no matter what color schemes chosen (pic 1 and 2).
> 
> However, when Desktop layout set Medium or Small, the padding is increased from 4rem to 5rem and it quite grabs my attention especially when color schemes are different (pic 3). Is there a reason for this or we can fix it?

Good call out! I think when the screen is set to medium it can be 4rem to match the large setting. However, for the `Small` setting I am hesitant to change it since the extra padding makes the column widths even. When stacked with other sections like ` Image with text` the columns then align perfectly. [Shown here.](https://screenshot.click/22-57-zhqjs-6ecl7.png)

Oliviammarcello (dismissed):
Looks good! 🎉

assets/section-main-product.css
```diff
@@ -62,10 +62,18 @@
...
+  }
+
   .product--large:not(.product--no-media).product--right .product__info-wrapper {
     padding: 0 4rem 0 0;
   }
 
+  .product--medium:not(.product--no-media).product--right .product__info-wrapper {
+    padding: 0 4rem 0 0;
+  }
+
```
ludoboludo, line 76: I wonder if those changes should only be applied when the color scheme are different 🤔 (`color scheme` vs `product information color scheme`).

Or if it's meant to be applied in all scenarios. 

I feel like it could always use 4rem above `750px` when set to medium and when the color scheme and product info color scheme don't match. Right now it's only above a `990px` viewport width.

ludoboludo (commented):
Ill finish reviewing tomorrow 👍 Just a small comment for now

eugenekasimov:
> As @andrewetchen mentioned, what is the plan for the featured product section ? Shouldn't we apply the same styling to it ? Were you thinking of doing it as a follow up PR at this point ?

Last time when I spoke with Olivia she told me that it's not clear yet and asked me to work on Product Information section only.

assets/section-main-product.css
eugenekasimov: 1. I prefer to keep it constant and apply to all scenarios no matter whether color-schemes are different or not.
2. I made it 4rem  only for a viewport wider than 990px on purpose. Olivia wanted it to be 5rem for Desktop media size set to Small. Even when the Desktop media size is set to Medium/Large if your screen is less than 990px the media size looks like a small one. That's why I decided to make it this way.

assets/section-main-product.css
```diff
@@ -370,6 +387,11 @@ a.product__text {
 }
 
 @media screen and (min-width: 750px) {
+  .different-color-scheme {
+    padding: 2rem;
```
andrewetchen, line 391:
```suggestion
```

Like Ludo mentioned, we can omit this since its already applied here:

https://github.com/Shopify/dawn/blob/73991f6a031a4443cf75b7f559e33f13e4af1dd6/assets/section-main-product.css#L309-L312

assets/section-main-product.css
```diff
@@ -383,6 +405,13 @@ a.product__text {
   }
 }
 
+@media screen and (min-width: 990px) {
+  .different-color-scheme {
+    padding: 3rem;
+    margin-top: 0rem;
```
andrewetchen, line 411:
```suggestion
```

We can also omit this as Ludo said since you're already applying it here:

https://github.com/Shopify/dawn/blob/73991f6a031a4443cf75b7f559e33f13e4af1dd6/assets/section-main-product.css#L389-L393

locales/en.default.schema.json
```diff
@@ -382,7 +382,11 @@
           "label": "Inverse"
         },
         "label": "Color scheme",
-        "has_cards_info": "To change the card color scheme, update your theme settings."
+        "has_cards_info": "To change the card color scheme, update your theme settings.",
+        "heading": "Colors",
+        "product_information_color_scheme": {
+          "label": "Product information color scheme"
+        }
```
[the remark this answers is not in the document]
andrewetchen: Hey Eugene, let me know when you reposition this under `main-product` as I'm also seeing this still under `sections.all` 👍🏼

assets/section-main-product.css
```diff
@@ -383,6 +405,13 @@ a.product__text {
   }
 }
 
+@media screen and (min-width: 990px) {
+  .different-color-scheme {
+    padding: 3rem;
```
andrewetchen, line 410: What are your thoughts on using `4rem` which is also being applied to the `product__info-wrapper`? It just looks a little bit off to me but I'm curious to hear your thoughts:

https://github.com/Shopify/dawn/blob/73991f6a031a4443cf75b7f559e33f13e4af1dd6/assets/section-main-product.css#L65-L67

Here's a quick video for comparison:

https://user-images.githubusercontent.com/28404165/186766069-bac5d841-5635-4e61-8a9c-add85db8c33a.mp4

andrewetchen (commented):
I've played around with things and overall, things are looking good 👍🏼 

I've left some feedback.

assets/section-main-product.css
ludoboludo: Just to recap right now here is how the padding logic is working: 
- when `desktop media size` is set to large:
  - between 750px and 990px, padding value is `5rem`
  - above 990px, padding value is `4rem`
- when `desktop media size` is set to medium: 
  - between 750px and 990px, padding value is `5rem`
  - above 990px, padding value is `4rem`

On top of that when the product information is using a color scheme different than the rest of the section, you're losing another `4rem` of width for the text because of [the padding applied](https://screenshot.click/25-51-1mu5u-s6byw.png) which is `2rem` between `750px` and `990px` and `3rem` above `990px`.

I think there is a chance here for us to remove some lines of code, even if it's just a few, and leave that spacing at `4rem` at all time for whatever the media size is. 
Having a larger gap between the media column and product info on smaller screen doesn't make sense to me 🤔  

[Video to help visualize it](https://screenshot.click/25-01-dtiib-4ig4h.mp4)

Im going to cc: @Oliviammarcello to share context with her 🙂

Oliviammarcello: > I think there is a chance here for us to remove some lines of code, even if it's just a few, and leave that spacing at 4rem at all time for whatever the media size is.
Having a larger gap between the media column and product info on smaller screen doesn't make sense to me 

@ludoboludo Thanks for the detailed explanation, you convinced me! Let's make it 4rem regardless of media width.

ludoboludo: And should we do that only when the color scheme is different between the product info and product section 🤔 ? Since product info has some extra padding when it's different.

eugenekasimov: I tried different padding sizes a lot and went to a conclusion that it looks the best for me when we have 3rem padding no matter what media size is but with different color schemes, and 4rem when the color schemes are same.

eugenekasimov: @Oliviammarcello @ludoboludo @andrewetchen 

I made a video explaining my comment above. Also in the video I mentioned the question that Andrew brought up [here](https://github.com/Shopify/dawn/pull/1907#discussion_r955411467). And I forgot to mention in the video that the way I suggest looks more closer to Figma reference. 

Please let me know what you think. Thanks. 

Uploading PAdding size.mov…

assets/section-main-product.css
```diff
@@ -298,6 +306,15 @@ fieldset.product-form__input .form__label {
...
 
+.different-color-scheme {
+  padding: 2rem;
+  margin-top: 1.5rem;
+}
+
+.different-color-scheme.product__info-container > *:first-child {
+  margin-top: 0;
+}
+
```
ludoboludo, line 317: One other thing we will need to adjust is the `top` value for the `product__info-container--sticky` when the color scheme are different. 
Here is an [video explanation](https://screenshot.click/26-27-1tlnp-m6mb8.mp4)

sections/main-product.liquid
ludoboludo, line 1: Another question is, should the product information be receiving the global settings from `content-container` ? 😅 

Something for designers to look into/let us know. 

Multicolumn, the text block from image with text, image banner, rich text, etc, are receiving it. I wonder if it should here as well 🤔

assets/section-main-product.css
eugenekasimov: Thanks Ludo for mention it!

eugenekasimov:
@Oliviammarcello, I dealt with all the concerns we discussed today. 

However, one more concern came up to me when I added the global settings to the product information container. If we add a border and the color schemes are the same, it feels we are missing padding between content and the border.

I wonder if we should add 1rem padding for this particular case? 

Screen Shot 2022-08-29 at 2 42 13 PM

assets/section-main-product.css
```diff
@@ -37,24 +37,24 @@
...
-  }
-
   .product__media-container .slider-buttons {
     display: none;
   }
 }
 
 @media screen and (min-width: 990px) {
+  .product__info-wrapper--extra-padding {
+    padding: 0 0 0 6rem!important;
```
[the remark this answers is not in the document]
eugenekasimov: You're right. I reordered classes and remove it.

Oliviammarcello:
> @Oliviammarcello, I dealt with all the concerns we discussed today so you can check and test them.
> 
> However, one more concern came up to me when I added the global settings to the product information container. If we add a border and the color schemes are the same, it feels we are missing padding between content and the border.
> 
> I wonder if we should add 1rem padding for this particular case? 
> 
> Screen Shot 2022-08-29 at 2 42 13 PM

Is it possible to add extra padding only when the content container thickness is 1px or greater and the opacity is 5% or greater?

eugenekasimov:
> > @Oliviammarcello, I dealt with all the concerns we discussed today so you can check and test them.
> > However, one more concern came up to me when I added the global settings to the product information container. If we add a border and the color schemes are the same, it feels we are missing padding between content and the border.
> > I wonder if we should add 1rem padding for this particular case? 
> > Screen Shot 2022-08-29 at 2 42 13 PM
> 
> Is it possible to add extra padding only when the content container thickness is 1px or greater and the opacity is 5% or greater?

Olivia and Dan came to a decision that we will apply Content Container settings to Product Information container only when color-schemes are different.

assets/section-main-product.css
```diff
@@ -511,6 +533,8 @@ a.product__text {
 
   .product__media-list {
     padding-right: calc(var(--media-shadow-horizontal-offset) * var(--media-shadow-visible));
+    position: relative;
+    z-index: 1;
```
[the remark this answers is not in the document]
eugenekasimov: We added this because for some reason global settings were not applied to media, but it'a not the case anymore. I remove it.

assets/section-main-product.css
```diff
@@ -511,6 +533,8 @@ a.product__text {
 
   .product__media-list {
     padding-right: calc(var(--media-shadow-horizontal-offset) * var(--media-shadow-visible));
```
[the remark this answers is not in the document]
eugenekasimov: thanks Ludo. It helped :)

locales/en.default.schema.json
```diff
@@ -1617,7 +1618,7 @@
       }
     },
     "main-product": {
-      "name": "Product information",
+      "name": "Product content",
```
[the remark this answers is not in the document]
ludoboludo: Created an issue for it here: https://github.com/Shopify/help/issues/33748

assets/section-main-product.css
```diff
@@ -30,19 +30,11 @@
   }
 
   .product__info-wrapper {
-    padding: 0 0 0 5rem;
-  }
-
-  .product__info-wrapper--extra-padding {
-    padding: 0 0 0 8rem;
+    padding: 0 0 0 4rem;
```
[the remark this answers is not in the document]
ludoboludo: For follow things like this, we can create an issue to track it. -> https://github.com/Shopify/dawn/issues/1946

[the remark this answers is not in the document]
ludoboludo: For follow things like this, we can create an issue to track it. -> https://github.com/Shopify/dawn/issues/1946

sections/main-product.liquid
```diff
@@ -906,170 +297,790 @@
...
+      {%- liquid
+        assign differentColorSchemes = true
+        if section.settings.color_scheme == section.settings.product_information_color_scheme
+        assign differentColorSchemes = false
+        endif
+      -%}
+      <div class="product__info-wrapper grid__item{% if settings.page_width > 1400 and section.settings.media_size == "small" %} product__info-wrapper--extra-padding{% endif %}{% if differentColorSchemes %} different-color-scheme{% endif %}">
+        <div
+          id="ProductInfo-{{ section.id }}"
+          class="product__info-container isolate color-{{ section.settings.product_information_color_scheme }}{% if section.settings.enable_sticky_info and section.settings.gallery_layout == 'stacked' %} product__info-container--sticky{% endif %}{% if differentColorSchemes %} content-container{% endif %}"
```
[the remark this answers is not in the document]
ludoboludo: Np 🙂

sections/main-product.liquid
```diff
@@ -1,740 +1,115 @@
...
->
-  {{ 'section-main-product.css' | asset_url | stylesheet_tag }}
-  {{ 'component-accordion.css' | asset_url | stylesheet_tag }}
-  {{ 'component-price.css' | asset_url | stylesheet_tag }}
-  {{ 'component-rte.css' | asset_url | stylesheet_tag }}
-  {{ 'component-slider.css' | asset_url | stylesheet_tag }}
-  {{ 'component-rating.css' | asset_url | stylesheet_tag }}
-  {{ 'component-loading-overlay.css' | asset_url | stylesheet_tag }}
-  {{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
+<div class="isolate color-{{ section.settings.color_scheme }}">
```
ludoboludo, line 1: Maybe let's apply this class ('isolate') instead on the element that also has the class `product__media-wrapper` applied to it. Since this is what is being impacted by the shadows. Rather than applying it to the wrapper of the whole file. 👍 

I think it would be better so it's applied specifically where needed. 🤔

ludoboludo (commented):
It's looking good. A couple small comments that I was thinking about while reviewing.

assets/section-main-product.css
```diff
@@ -369,9 +355,35 @@ a.product__text {
   display: none;
 }
 
+.different-color-scheme {
+  padding: 0;
+}
```
ludoboludo, line 360: There is actually an issue with the quick add modal when the color scheme are the same.

Have a look at [this video](https://screenshot.click/07-37-qte21-z3alg.mp4)

ludoboludo (commented):
Found another issue with the modal 😓 the rest seemed ok

We should try and align with @Oliviammarcello to confirm whether the modal should be pulling in all that styling from the product page or not 🤔

assets/section-main-product.css
eugenekasimov: Okay. Since it's quite urgent now I assume that we need need to pass color scheme styling to quick-add modal. I didn't find any solution to do that except using JS. Now it seems working good and looks good too.

@Oliviammarcello please confirm rather we need this solution or not. thanks.

Oliviammarcello: What if we don't add the colour scheme to the quick add modal and force it to take the colour of the site background. I am not confident that merchants would want an instance where the product information is a different colour.

eugenekasimov: I pushed updates that prevent quick add modal from pulling the styling from PDP.

Oliviammarcello (approved):

Oliviammarcello (approved):

sections/main-product.liquid
```diff
@@ -1,748 +1,126 @@
...
+                elsif section.settings.media_size == 'small'
+                  assign media_width = 0.45
+                endif
+              -%}
               {%- if product.selected_or_first_available_variant.featured_media != null -%}
                 {%- assign featured_media = product.selected_or_first_available_variant.featured_media -%}
                 <li
-                  id="Slide-{{ section.id }}-{{ featured_media.id }}-right"
+                  id="Slide-{{ section.id }}-{{ featured_media.id }}"
                   class="product__media-item grid__item slider__slide is-active{% if single_media_visible %} product__media-item--single{% endif %}{% if featured_media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains featured_media.src %} product__media-item--variant{% endif %}"
```
ludoboludo, line 735: should the `singe_media_visible` condition be removed 🤔

sections/main-product.liquid
```diff
@@ -1,748 +1,126 @@
...
-      id="ModelViewerOverride"
-      rel="stylesheet"
-      href="{{ 'component-model-viewer-ui.css' | asset_url }}"
-      media="print"
-      onload="this.media='all'"
-    >
-  {%- endif -%}
+    <script src="{{ 'product-form.js' | asset_url }}" defer="defer"></script>
 
-  <div class="product product--{{ section.settings.media_size }} product--{{ section.settings.media_position }} product--{{ section.settings.gallery_layout }} product--mobile-{{ section.settings.mobile_thumbnails }} grid grid--1-col {% if product.media.size > 0 %}grid--2-col-tablet{% else %}product--no-media{% endif %}">
```
ludoboludo, line 50: Here we lost the `product--mobile-` class

sections/main-product.liquid
```diff
@@ -1,748 +1,126 @@
...
+                    lazy_load: false
+                  %}
                 </li>
               {%- endif -%}
               {%- for media in product.media -%}
                 {%- unless media.id == product.selected_or_first_available_variant.featured_media.id -%}
                   <li
-                    id="Slide-{{ section.id }}-{{ media.id }}-right"
+                    id="Slide-{{ section.id }}-{{ media.id }}"
                     class="product__media-item grid__item slider__slide{% if single_media_visible %} product__media-item--single{% endif %}{% if product.selected_or_first_available_variant.featured_media == null and forloop.index == 1 %} is-active{% endif %}{% if media.media_type != 'image' %} product__media-item--full{% endif %}{% if section.settings.hide_variants and variant_images contains media.src %} product__media-item--variant{% endif %}"
```
ludoboludo, line 746: same here. that `single_media_visible` shouldn't be removed 👍

sections/main-product.liquid
```diff
@@ -773,40 +175,43 @@
...
           {%- endif -%}
-          {%- if media_count > 1 and section.settings.gallery_layout contains 'thumbnail' or section.settings.mobile_thumbnails == 'show' -%}
+          {%- if media_count > 1
+            and section.settings.gallery_layout contains 'thumbnail'
+            or section.settings.mobile_thumbnails == 'show'
+          -%}
             <slider-component
-              id="GalleryThumbnails-{{ section.id }}-right"
+              id="GalleryThumbnails-{{ section.id }}"
               class="thumbnail-slider slider-mobile-gutter quick-add-hidden{% unless section.settings.gallery_layout contains 'thumbnail' %} medium-hide large-up-hide{% endunless %}{% if section.settings.mobile_thumbnails != 'show' %} small-hide{% endif %}{% if media_count <= 3 %} thumbnail-slider--no-slide{% endif %}"
```
ludoboludo, line 779: Here too, I wonder if that logic should be removed?  Should not be something being touched in this PR i believe.

ludoboludo (commented):
I think we have a few leftover from the rebase still that make some scenario not work (2 column on mobile, 2 columns with the medias to the right)

ludoboludo (commented):
I think that will be my last piece of feedback 👍 

The rest looks good, I didn't find any bugs while testing. Once addressed I'm ready to approve the PR.

ludoboludo (dismissed):
I think we're there 👌 Great job keeping up with the feedback.

galenking:
Is this still being worked on or is it unlikely to be merged in?

PR #1918, opened 2022-08-16
Fetch the blog header (LCP) image at high priority

siakaramalegos:
**PR Summary:** 

When a blog article has a header image, it is usually the LCP (largest contentful paint) element. Unfortunately, several other assets are fetched before the LCP image like non-critical JS. `fetchpriority` is now released on Chromium browsers and can significantly improve the speed of LCP on those browsers (and hopefully more in the future). This PR changes the blog header image to use `fetchpriority=high` to make the LCP metric faster.

To improve loading speed and Core Web Vitals (LCP metric) which also impacts SEO.

I focused only on this one page template since the decision does not require user input. The blog header image is always at the top of the page if it exists. I also tested before and after on our new blog for web performance. On my [test article](https://speedmatters.myshopify.com/blogs/blog/key-web-performance-metrics-in-2022), LCP improved by 0.25-0.5 seconds by setting the high priority.

Here is the webpagetest comparison video: https://www.webpagetest.org/video/view.php?end=visual&tests=220816_BiDc3C_C2T%2C220816_BiDcJJ_C1W&bg=ffffff&text=222222&loc=California%2C+USA+-+EC2+-+Chrome+-+Emulated+Google+Pixel+2+-+4G

In the detailed test history, we can see that the LCP moved up from the 17th resource loaded to the 3rd ([test comparison](https://www.webpagetest.org/video/compare.php?tests=220816_BiDc3C_C2T%2C220816_BiDcJJ_C1W&thumbSize=200&ival=100&end=visual) - adjust filmstrip to 0.1 s to better see). Thus, the priority is improved.

Before testing live, I also ran a WebPageTest experiment - https://www.webpagetest.org/video/compare.php?tests=220816_AiDcW4_C46,220816_AiDc7W_C45

**Other considerations**

It would be good to compare real-user data before and after the change once this is implemented to help drive priorities for expanding similar optimizations in Dawn and on the platform.

**Decision log**

|---|---|---|---|---|
| 1 |   |   |   |   |

**Testing steps/scenarios**

- [ ] Step 1

**Demo links**

- [Store](https://speedmatters.myshopify.com/)
- [Editor](https://speedmatters.myshopify.com/admin/themes/134586040538/editor)

**Checklist**

### DIFF
--- a/sections/main-article.liquid
+++ b/sections/main-article.liquid
@@ -25,6 +25,7 @@
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 }}px, (min-width: 750px) calc(100vw - 10rem), 100vw"
                 src="{{ article.image | image_url: width: 1100 }}"
                 loading="eager"
+                fetchpriority="high"
                 width="{{ article.image.width }}"
                 height="{{ article.image.height }}"
                 alt="{{ article.image.alt | escape }}">


siakaramalegos:
@bacounts I'm confused as these PR's do not touch the same files nor dependencies. It looks to me like they are completely separate.

bacounts:
> @bacounts I'm confused as these PR's do not touch the same files nor dependencies. It looks to me like they are completely separate.

Yes, I comment too soon. Eventually, that PR will include all images throughout the theme including the main article image. When they eventually change this to a image_tag, the fetchpriority will carry over if it is applicable with the image_tag. Sorry for the confusion.

andrewetchen (approved):
This looks good to me! Thanks for adding this, @siakaramalegos 🙌🏼

sections/main-article.liquid
```diff
@@ -25,6 +25,7 @@
                 sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 }}px, (min-width: 750px) calc(100vw - 10rem), 100vw"
                 src="{{ article.image | image_url: width: 1100 }}"
                 loading="eager"
+                fetchpriority="high"
```
ludoboludo, line 28: I like the plus of having this and how beneficial it can be but I'm not too sure where I stand on using experimental technology on our theme. 
This is something we've been trying to avoid where we can to make sure we're limiting our need to add code that will need to be addressed in the future once it's fully supported or if the implementation/specs change. 

This is supported only in the latest version of Chrome and Edge ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority)). 

Do you feel confident that the implementation is going to stay that way ? Experimental sound like it's possible for it to change but maybe in this case it's just because of the low support.

siakaramalegos: Very few features have 100% browser support, and this one has no negative effect on those other browsers. By implementing it, we can make the site significantly faster for all Chrome, Edge, and Opera users which is a pretty large bucket - 67.91% of users globally. If in the rare case it's pulled, then functionally it will not break the code. The browsers that do not recognize it will ignore the attribute.

siakaramalegos: FYI, Safari has it in Technology Preview per caniuse

siakaramalegos: I pinged the Chrome engineer who worked on it to learn more - it's stable in Chrome so really just waiting for another browser to implement to remove the "experimental" flag. The other browsers already participated in the naming. So it's low risk to implement with big gains for Chromium performance:
image

PR #1901, opened 2022-08-08
[Header] Update 'Custom logo width' to 'Desktop custom logo width' and add info text

LucasLacerdaUX:
**PR Summary:** 
Adds more clarity to the setting
**Why are these changes introduced?**
- [x] Update header `Custom logo width` slider to `Desktop custom logo width` 
- [x] Add help text `Logo width is automatically optimized for mobile.`

Fixes #1900 

**Demo links**
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128463306774)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128463306774/editor)

PR #1920, opened 2022-08-16
Fix logo layout shift

siakaramalegos:
**PR Summary:** 

Currently, if the header logo settings make it taller than the other elements in the menu bar, a layout shift will occur to accommodate the image height. This is caused by a cascaded style that does not appropriately set the height so that the browser can correctly reserve space. This PR fixes the CSS so that height and width reserve space for the logo image correctly.

Fixes layout shift due to taller logos.

I saw the layout shift on our test internal site and debugged it to find the cause. Then tested before and after on that same site.

**Other considerations**

 From what I can tell in the section settings, this should not negatively impact other logo sizes since the logo heights and widths are specifically set in the section settings, but I would like someone else to gut check this and suggest maybe more cases to test. 

**Decision log**

|---|---|---|---|---|
| 1 |   |   |   |   |

**Testing steps/scenarios**

Not sure what steps to put here - would be good for more folks to think of edge cases and adjust the header section logos, logo heights, and logo widths. 

For our internal site, here's the before and after video: https://www.webpagetest.org/video/view.php?end=visual&tests=220816_BiDcV2_FAW%2C220816_AiDc42_ETG&bg=ffffff&text=222222&loc=California%2C+USA+-+EC2+-+Chrome+-+Emulated+Google+Pixel+2+-+4G

If you prefer filmstrip view, here's the test comparison (ignore timings as its just network variations): https://www.webpagetest.org/video/compare.php?tests=220816_BiDcV2_FAW%2C220816_AiDc42_ETG&thumbSize=200&ival=100&end=visual# 

**Demo links**

- [Store](https://speedmatters.myshopify.com/)
- [Editor](https://speedmatters.myshopify.com/admin/themes/134586040538/editor)

**Checklist**

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -35,7 +35,7 @@
   }
 
   .header__heading-logo {
-    height: {{ logo_height }};
+    height: {{ logo_height }}px;
     width: {{ section.settings.logo_width }}px;
   }
 
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -26,6 +26,8 @@
 <noscript>{{ 'component-cart-notification.css' | asset_url | stylesheet_tag }}</noscript>
 <noscript>{{ 'component-cart-items.css' | asset_url | stylesheet_tag }}</noscript>
 
+{%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
+
 <style>
   header-drawer {
     justify-self: start;
@@ -33,7 +35,8 @@
   }
 
   .header__heading-logo {
-    max-width: {{ section.settings.logo_width }}px;
+    height: {{ logo_height }};
+    width: {{ section.settings.logo_width }}px;
   }
 
   @media screen and (min-width: 990px) {
@@ -376,7 +379,6 @@
           <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
             {%- if section.settings.logo != blank -%}
               {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
-              {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
               {{ section.settings.logo | image_url: width: section.settings.width | image_tag:
                 class: 'header__heading-logo',
                 widths: '50, 100, 150, 200, 250, 300, 400, 500',


sections/header.liquid
```diff
@@ -26,14 +26,17 @@
...
+
 <style>
   header-drawer {
     justify-self: start;
     margin-left: -1.2rem;
   }
 
   .header__heading-logo {
-    max-width: {{ section.settings.logo_width }}px;
+    height: {{ logo_height }};
```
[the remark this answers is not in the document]
siakaramalegos: Thanks for the catch! So weird that visually it still worked in the browser at least for Chrome. Just pushed the fix

sections/header.liquid
```diff
@@ -26,14 +26,17 @@
...
 <style>
   header-drawer {
     justify-self: start;
     margin-left: -1.2rem;
   }
 
   .header__heading-logo {
-    max-width: {{ section.settings.logo_width }}px;
+    height: {{ logo_height }}px;
+    width: {{ section.settings.logo_width }}px;
```
sofiamatulis, line 39: I think the tricky part of making this `width` and not `max-width` is because of this:https://screenshot.click/24-05-4ysyc-w33ie.mp4

You can see in the video where on mobile it was being optimized before (beginning of the video) but now we have a way larger size.

We actually also have a PR to update language: https://github.com/Shopify/dawn/pull/1901

But by introducing this change, if the merchant selects 250, theyd end up with a really big logo on mobile

https://screenshot.click/24-07-sv528-5m2pi.png

cc @LucasLacerdaUX @Oliviammarcello

sofiamatulis (commented):
Thanks for creating this PR! This is great for layout shift. Just have a note on the effects this will have on mobile. cc @Oliviammarcello @LucasLacerdaUX

sections/header.liquid
```diff
@@ -376,7 +379,6 @@
           <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
             {%- if section.settings.logo != blank -%}
               {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
-              {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
```
sofiamatulis, line 379: Also if we make the change here, we should probably update the password page too: https://github.com/Shopify/dawn/blob/main/sections/main-password-header.liquid#L3

sections/header.liquid
siakaramalegos: I just tested making a change to the logo width in the theme settings and 250px looks equally bad on desktop. I don't think we should override what a merchant puts in the settings though. I might be misunderstanding your point so please let me know.

FYI, I put in a square logo. So by default, it's taller than the other menu elements. I didn't change the setting for logo width. So there is layout shift by default at the moment. I have a feeling that's a large use case so we need a way to prevent layout shift for taller logos.

PR #1929, opened 2022-08-24
Add a max width to iframes within the product info container

ludoboludo:
### PR Summary: 

Here we're adding a simple CSS declaration to prevent iframes within the `product__info-container` from being bigger than the container element and creating some horizontal scrolling issues. 

Fixes an issue where video would be bigger than the container they're in. 

Simply added a `max-width: 100%;`.

### Other considerations

Could have used the `aspect-ratio` property to keep a `16/9` aspect ratio but it's not supported on older browser versions which we still support (Safari 14.1). And it looks odd as we also need to setup a `height: auto;` so it ends up looking like this: 

### Decision log

|---|---|---|---|---|
| 1 | Not use `aspect-ratio` for compatibility purposes  | Only rely on `max-width` for now  | Poor browser support (Safari 14.1) | We're not forcing an aspect ratio so sometimes the container can be 1:1 on some viewport  |

The only impact should be fixing the iframes added via the admin within product descriptions. 

If a merchant had some custom code/apps adding iframes within the `product__info-container` they could be impacted but `max-width: 100%;` shouldn't really break anything. 

### Testing steps/scenarios

- [ ] You can test the `puppy` product on which I added two videos from vimeo and youtube. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=131455909910)
- [Editor](https://os2-demo.myshopify.com/admin/themes/131455909910/editor)

### Checklist

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -302,6 +302,10 @@ fieldset.product-form__input .form__label {
   margin: 1.5rem 0;
 }
 
+.product__info-container iframe {
+  max-width: 100%;
+}
+
 .product__info-container .product-form,
 .product__info-container .product__description {
   margin: 2.5rem 0;


andrewetchen (approved):
LGTM! Thanks for fixing this, Ludo! 🎉 

The empty space above + below the Vimeo video threw me off a little but then I saw [this](https://user-images.githubusercontent.com/28404165/186473066-117b345d-5e14-4c23-bbbd-db6792c8828c.png)

PR #1933, opened 2022-08-24
[Product page] "Icon with text" block

### LINKED ISSUE #1719: [Product page] New "Icon with text" block

eugenekasimov:
### PR Summary: 

Create a new `Icon with text` block. This block allows users to add up to three icons/images with headings to the Product page. There are two different layouts. If a user uploads an image it overrides the icon selection. If there are no icon and image selected the heading is not displayed. 

[Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=265%3A32841)

Fixes #1719 .

- Create a logic to apply different css classes depending on selected layout.
- Add css classes that allow to change icon/image aligning depending on how many icons/images are displayed.
- To implement the ability of rendering a Heading only if there is at least an icon or image selected I check pair of icon/image first. Then I assign a result to a variable and use it in the final condition to check if any of the pairs is false.

### Testing steps/scenarios
- [ ] Delete one of the icons and leave only two any icons.
- [ ] Delete any two of the icons and leave only one.
- [ ] Delete icon and image of any ones among three.
- [ ] Upload an image for an existing icon.
- [ ] Change layout from Horizontal to Vertical.

### Screenshots

<details>
<summary><strong>Horizontal Layout - three icon/images</strong></summary>
Screen Shot 2022-08-24 at 10 30 30 AM
</details>
<details>
<summary><strong>Horizontal Layout - two icon/images</strong></summary>
Screen Shot 2022-08-24 at 10 30 40 AM
</details>
<details>
<summary><strong>Horizontal Layout - one icon/images</strong></summary>
Screen Shot 2022-08-24 at 10 30 55 AM
</details>
<details>
<summary><strong>Vertical Layout - three icon/images</strong></summary>
Screen Shot 2022-08-24 at 10 30 03 AM
</details>
<details>
<summary><strong>Vertical Layout - two icon/images</strong></summary>
Screen Shot 2022-08-24 at 10 30 15 AM
</details>

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/128406454294/editor?previewPath=%2Fproducts%2Fno-featured-image&section=template--15848218165270__main&block=template--15848218165270__main%2Fd014c413-880c-49b5-be6c-536e7970ba39)

### Checklist

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -1203,3 +1203,56 @@ a.product__text {
   margin-left: 1.2rem;
   flex-shrink: 0;
 }
+
+/* Icon with text */
+.icon-large .icon {
+  width: 3rem;
+  height: 3rem;
+  margin-bottom: 1rem;
+}
+
+.image-large {
+  height: 3rem;
+  margin-bottom: 1rem;
+}
+
+.icon-large {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.icon-small .icon {
+  min-width: 2rem;
+  min-height: 2rem;
+  margin-right: 1rem;
+}
+
+.icon-small {
+  display: flex;
+  align-items: center;
+  margin: 0 0 1rem 0;
+}
+
+.image-small {
+  width: 2rem;
+  margin-right: 1rem;
+}
+
+.icon-container-vertical {
+  padding-top: 1.5rem;
+}
+
+.icon-container-horizontal--space-between {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding-top: 2rem;
+}
+
+.icon-container-horizontal--space-evenly {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+  padding-top: 2rem;
+}
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1868,7 +1868,53 @@
               "content": "To display a rating, add a product rating app. [Learn more](https://help.shopify.com/manual/online-store/themes/theme-structure/page-types#product-rating-block)"
             }
           }
-        }
+        },
+
+        "icon_with_text": {
+            "name": "Icon with text",
+            "settings": {
+              "layout": {
+                "label": "Layout",
+                "options__1": {
+                  "label": "Horizontal"
+                },
+                "options__2": {
+                  "label": "Vertical"
+                }
+              },
+              "content": {
+                "label": "CONTENT",
+                "info": "Choose an icon or add an image for each column or row."
+              },
+              "first_icon": {
+                "label": "First icon"
+              },
+              "first_image": {
+                "label": "First image"
+              },
+              "first_heading": {
+                "label": "First heading"
+              },
+              "second_icon": {
+                "label": "Second icon"
+              },
+              "second_image": {
+                "label": "Second image"
+              },
+              "second_heading": {
+                "label": "Second heading"
+              },
+              "third_icon": {
+                "label": "Third icon"
+              },
+              "third_image": {
+                "label": "Third image"
+              },
+              "third_heading": {
+                "label": "Third heading"
+              }
+            }
+          }
       },
       "settings": {
         "header": {
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -1496,7 +1496,7 @@
     },
     {
       "type": "icon-with-text",
-      "name": "Icon with text",
+      "name": "t:sections.main-product.blocks.icon_with_text.name",
       "settings": [
         {
           "type": "select",
@@ -1504,20 +1504,20 @@
           "options": [
             {
               "value": "horizontal",
-              "label": "Horizontal"
+              "label": "t:sections.main-product.blocks.icon_with_text.settings.layout.options__1.label"
             },
             {
               "value": "vertical",
-              "label": "Vertical"
+              "label": "t:sections.main-product.blocks.icon_with_text.settings.layout.options__2.label"
             }
           ],
           "default": "horizontal",
-          "label": "Layout"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.layout.label"
         },
         {
           "type": "header",
-          "content": "CONTENT",
-          "info": "Choose an icon or add an image for each column or row."
+          "content": "t:sections.main-product.blocks.icon_with_text.settings.content.label",
+          "info": "t:sections.main-product.blocks.icon_with_text.settings.content.info"
         },
         {
           "type": "select",
@@ -1701,18 +1701,18 @@
             }
           ],
           "default": "heart",
-          "label": "First icon"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.first_icon.label"
         },
         {
           "type": "image_picker",
           "id": "first_image",
-          "label": "First image"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.first_image.label"
         },
         {
           "type": "text",
           "id": "first_heading",
           "default": "Heading",
-          "label": "First heading"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.first_heading.label"
         },
         {
           "type": "select",
@@ -1896,18 +1896,18 @@
             }
           ],
           "default": "leaf",
-          "label": "Second icon"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.second_icon.label"
         },
         {
           "type": "image_picker",
           "id": "second_image",
-          "label": "Second image"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.second_image.label"
         },
         {
           "type": "text",
           "id": "second_heading",
           "default": "Heading",
-          "label": "Second heading"
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.second_heading.label"
         },
         {
           "type": "select",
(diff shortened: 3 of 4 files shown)


Oliviammarcello:
Looking great! Just have a few notes:

- [ ] The text should be at 100% opacity similarly to the accordion titles. [Screenshot](https://screenshot.click/25-22-fo5v5-tyw1t.png)
- [ ] The icons should change colour with the _Icon color scheme_ picker in the theme settings. [Screenshot](https://screenshot.click/25-24-cln2f-fay9e.png)
- [ ] Currently when you add the block a leaf is used as the second default icon. Could we update that to be the return icon? A leaf feels like to niche of a use case. [Screenshot](https://screenshot.click/25-27-2ibv3-io9kh.png)
- [ ] When the image size is set to Medium and Small, the space between each icon becomes too large. [Screenshot](https://screenshot.click/25-31-xisje-5s6bs.png) Can we force a fixed 3rem space between each and center align them in the column? 
- [ ] When the icon heading is left blank can we hide everything and adapt the layout to 2 or 1 column(s). [Screenshot.](https://screenshot.click/25-38-5khdo-pkbgx.png) Also, add help text under the heading like we do for buttons. [Screenshot.](https://screenshot.click/25-42-3lwhp-h44a1.png) `"Leave the heading label blank to hide the icon column."`

assets/section-main-product.css
```diff
@@ -1203,3 +1203,56 @@ a.product__text {
...
+  justify-content: space-between;
+  padding-top: 2rem;
+}
+
+.icon-container-horizontal--space-evenly {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+  padding-top: 2rem;
+}
```
kmeleta, line 1258: I think you could simplify the styles/classes here a bit by using a`.icon-container-horizontal` class with all the shared styles and a default `justify-content` value. Then you can use the `--space-evenly` modifier class to override only what's needed. This should allow you to remove an `else` case from the liquid where you conditionally render `evenly` or `between`

```suggestion
.icon-container-horizontal {
  display: flex;
  flex-direction: row;
  justify-content: space-between;
  padding-top: 2rem;
}

.icon-container-horizontal--space-evenly {
  justify-content: space-evenly;
}
```

sections/main-product.liquid
```diff
@@ -1436,6 +1504,618 @@
...
+        },
+        {
+          "type": "text",
+          "id": "first_heading",
+          "default": "Heading",
+          "label": "t:sections.main-product.blocks.icon_with_text.settings.first_heading.label"
+        },
+        {
+          "type": "select",
+          "id": "second_icon",
```
kmeleta, line 1730: I'm not sure we've used words (first, second, third) instead of numbers for setting ids like this (correct me if you know of a case) Probably doesn't comes up a lot but I think we would usually call this `image_1`, `icon_1`, `heading_1`. Image banner is at least one place off the top of my head where I know this occurs (uses `image_1` and `image_2`) and I'd slightly prefer we keep consistent with that convention. I also just think the numbers are more concise and a little easier to parse when used in liquid conditionals and such.

kmeleta (commented):
Did a first pass. Looking good so far, Eugene. Couple (UX?) questions and some code suggestions for you. Let me know if you want to discuss any.

sections/main-product.liquid
```diff
@@ -689,10 +689,78 @@
...
+                        height="auto"
+                        width="auto"
+                        loading="lazy"
+                      >
+                    {%- endif -%}
+                    <span>{{- block.settings.second_heading -}}</span>
+                  </div>
+                {%- endunless -%}
+                {%- unless isIconImage_3 == false -%}
+                  <div class="{% if block.settings.layout == 'horizontal' %}icon-large{%else%}icon-small{% endif %}">
```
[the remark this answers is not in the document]
eugenekasimov: I added `column-gap: 3rem` and set `width 33%` for each column.

sections/main-product.liquid
```diff
@@ -689,10 +689,78 @@
...
+                 else
+                assign isIconImage_2 = true
+                endif
+                 if block.settings.third_icon == 'none' and block.settings.third_image == nil
+                 assign isIconImage_3 = false
+                 else
+                  assign isIconImage_3 = true
+                endif
+              -%}
+              <div class="{% if block.settings.layout == 'horizontal' %}icon-container-horizontal--space-{% if isIconImage_1 == false or isIconImage_2 == false or isIconImage_3 == false %}evenly{% else %}between{% endif %}{% else %}icon-container-vertical{% endif %}">
```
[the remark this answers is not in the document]
eugenekasimov: This one is not actual anymore because I changed the logic a bit.

snippets/icon-accordion.liquid
```diff
@@ -1,5 +1,14 @@
 {%- if icon != 'none' -%}
-  <svg class="icon icon-accordion color-foreground-{{ settings.accent_icons }}" aria-hidden="true" focusable="false" role="presentation" xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
+  <svg
+    class="icon icon-accordion color-foreground-{{ settings.accent_icons }}"
+    aria-hidden="true"
```
metamoni, line 4: `aria-hidden` is not necessary here. Using `role="presentation"` already hides the icon from screen readers.

metamoni (commented):
Went through the testing steps and this looks/works great. The only thing I'm not sure about is the "Add alt text" option when you add an image instead of an icon.

These images will probably be decorative, just like icons, so we should hide them from screen readers rather than encourage people to add alternative text. I'm assuming you've used an existing component for the image uploader, but it would be nice to hide the "Add alt text" in this case.

snippets/icon-accordion.liquid
kmeleta: My understanding is that if we're using `aria-hidden="true"`, the additional `role="presentation"` wouldn't have any effect at all, so I would agree we don't need both. However I'm less sure about which one of those is more desirable for this use case. They're _technically_ a little different akaik.

That said, all our icon snippets (like icon-accordion) and many other instances of svg icons in the theme are already using both. If we wanted to update one instance, we should update all of them. And I wouldn't think that should be done as part of this commit.

eugenekasimov:
> Went through the testing steps and this looks/works great. The only thing I'm not sure about is the "Add alt text" option when you add an image instead of an icon.
> 
> These images will probably be decorative, just like icons, so we should hide them from screen readers rather than encourage people to add alternative text. I'm assuming you've used an existing component for the image uploader, but it would be nice to hide the "Add alt text" in this case.

Good call @metamoni! Thanks. After a discussion we agreed that the best solution will be add `role=presentation` if alt text is empty otherwise add the alt text.

sections/main-product.liquid
```diff
@@ -697,91 +713,175 @@
...
+                  assign heading2Empty = true
+                endif
+
+                if block.settings.heading_3 == empty
+                  assign heading3Empty = true
+                endif
+              -%}
+              <div class="icon-container--{{ block.settings.layout }}">
+                {%- unless heading1Empty -%}
+                  <div class="icon--{{ block.settings.layout }}">
```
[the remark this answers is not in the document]
metamoni: That sounds like a good idea, thanks for calling that out @kmeleta 👍

assets/section-main-product.css
```diff
@@ -1244,3 +1244,80 @@ a.product__text {
...
+  margin-bottom: var(--icon-spacing);
+}
+
+.icon-with-text--vertical img {
+  margin-right: var(--icon-spacing);
+}
+
+.icon-with-text .h4 {
+  text-align: center;
+}
```
kmeleta, line 1299: I think we actually only wanted this to be for the horizontal variant, right? Guess I wasn't wasn't using long enough test labels.

locales/en.default.schema.json
```diff
@@ -1912,6 +1912,54 @@
...
+              "label": "Layout",
+              "options__1": {
+                "label": "Horizontal"
+              },
+              "options__2": {
+                "label": "Vertical"
+              }
+            },
+            "content": {
+              "label": "CONTENT",
```
kmeleta, line 1929: I don't think this needs to be all caps, does it? I feel like that's a style that's up to the editor.

assets/section-main-product.css
```diff
@@ -1244,3 +1244,80 @@ a.product__text {
...
+
+.icon-with-text .icon {
+  fill: rgb(var(--color-foreground));
+  height: var(--icon-size);
+  width: var(--icon-size);
+}
+
+.icon-with-text--horizontal .icon {
+  margin-bottom: var(--icon-spacing);
+}
```
kmeleta, line 1275: Looks like we can consolidate a couple other selectors into this rule.
`.icon-with-text--horizontal img` below on line 1289 and `.icon-with-text--vertical .icon-with-text__item` on line 1321
```
.icon-with-text--horizontal .icon,
.icon-with-text--horizontal img,
.icon-with-text--vertical .icon-with-text__item {
  margin-bottom: var(--icon-spacing);
}
```

assets/section-main-product.css
eugenekasimov: good catch, thanks.

metamoni (dismissed):
Looks good to me! 👍 Thanks for taking the time to carefully consider all my feedback, even the stuff that's not related to this PR!

assets/section-main-product.css
kmeleta: Yep I may have been crazy, disregard 🙂

locales/en.default.schema.json
kmeleta: I believe the editor adds the all caps styling automatically to content labels. Meaning I'd expect you could edit this and the result in the editor will look the same. In general, source text content should rarely enforce any styling decisions, like all caps. That should usually be left to the CSS/UI to apply as needed.

Compare the rendered result of this


To this

metamoni (approved):
As long as the other comments have been resolved, this looks good to me! 👍

metamoni (approved):

metamoni (approved):

PR #1938, opened 2022-08-31
Complementary products

ludoboludo:
**PR Summary:** 

Adding complimentary products block to the main-product page. 
[Figma file](https://www.figma.com/file/zfCoe8cJdGjcftBSon99v4/Complementary-products?node-id=1628%3A137697)

Adds a complementary product block on the product page

**Decision log**
We've decided to remove the `Adapt to image` option from the image ratio setting for a couple reasons: 
- when it's on, it messes with the height of the slides. Where a slide might have a much bigger height creating a lot of empty white space on other slides. 
- we also thought that the idea behind the complimentary products is to show a minimized version of a product card and not take up too much space. We have more control when keeping the sizes `square` or `portrait` so it seems nicer to make sure most of the experiences should be great (even though the same can happen with long product titles). 

**Other considerations**

**Testing steps/scenarios**
- [ ] You will need to use the Search & Discoverability app.There is currently one product (Puppy) setup with some complementary products, the rest should only show the skeleton state and then nothing. [Here is more info about setting up complementary products](https://videobin.shopify.io/v/bdr2BZ)
- [ ] Test with different product cards global settings (shadows, border, etc)
- [ ] Test slow connections
- [ ] Test accessibility
- [ ] Test other functionalities, like quick add to make sure it still works as expected

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=131484254230)
- [Editor](https://os2-demo.myshopify.com/admin/themes/131484254230/editor)

**Checklist**
- [ ] Notified the [help.shopify.com](https://help.shopify.com) team about updates to theme settings


kmeleta (commented):
Still reviewing, but figured I'd leave some partial feedback in case any of it is actionable. Looking really slick though.

sections/main-product.liquid
```diff
@@ -693,6 +693,100 @@
                   </span>
                 </p>
               {%- endif -%}
+
+              {%- when 'complementary' -%}
```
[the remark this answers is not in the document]
ludoboludo: I'm hoping that ideally merchant wouldn't do 10 products and 1 per slides 😅 But you it's doable 🤷 
I can bring it up tomorrow during a chat I'm having with Scott about accessibility.

assets/component-complementary-products.css
```diff
@@ -0,0 +1,114 @@
...
+  --shadow-padding-sides-negative: calc((var(--shadow-horizontal-offset) * -1 + var(--shadow-blur-radius)) * var(--shadow-visible));
+}
+
+.complementary-slide.grid__item,
+.complementary-card-container .skeleton {
+  width: 100%;
+  padding-top:  max(var(--focus-outline-padding), var(--shadow-padding-top));
+  padding-bottom: max(var(--focus-outline-padding), var(--shadow-padding-bottom));
+  padding-right: max(var(--focus-outline-padding), var(--shadow-padding-sides), var(--shadow-padding-sides-negative));
+  padding-left: max(var(--focus-outline-padding), var(--shadow-padding-sides), var(--shadow-padding-sides-negative));
```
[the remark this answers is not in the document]
ludoboludo: Ah yeah I don't why I added the too many options within the `max()`. Good catch 🙂

sections/main-product.liquid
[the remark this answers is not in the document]
kmeleta: I mean the 10 slides doesn't help, but it's a fair amount of tabs anyway with 10 products, even if its 4 slides or whatever. Especially if quick add is enabled too. Interested in how much we assume merchants will configure 10. Seems like a lot to allow to me, even if the component handles that many well (which it does). But yeah might as well run it by Scott at least.

baldurh:
This is great! Can’t wait to use this. I had started working on a custom solution using metafields and was planning to cross-sell in the cart drawer and cart page. Can someone explain to me what is the difference between the `related` and `complementary` intents in the Product Recommendation API? It doesn’t seem to be documented (yet).

metamoni:
When the row is collapsed, the arrow icon isn't aligned to the center

Screenshot 2022-09-06 at 17 29 10

metamoni:
On mobile screens, the button isn't aligned with the image at the bottom, like the header.
Screenshot 2022-09-06 at 17 33 32

sections/product-recommendations.liquid
```diff
@@ -20,8 +20,22 @@
   }
 {%- endstyle -%}
 
-<div class="color-{{ section.settings.color_scheme }} gradient">
+<div class="color-{{ section.settings.color_scheme }} gradient no-js-hidden">
```
metamoni, line 23: This feels like it would be a good fit for a landmark. You could make it an `aside` with an `aria-label`, this way assistive tech users would be able to navigate to it easily.

ludoboludo: This is for the section that shows up below. But you mean for the complementary block right ?

metamoni: I mean the complementary block, yes 👍

sections/main-product.liquid
```diff
@@ -693,6 +693,100 @@
...
+                            <div class="skeleton__content card__content">
+                              <div><span></span></div>
+                              <div><span></span></div>
+                            </div>
+                          </div>
+                        {%- endfor -%}
+                      </div>
+                      {%- if recommendations.performed and recommendations.products_count > 0 -%}
+                        {%- if block.settings.make_collapsible_row -%}
+                          <details id="Details-{{ block.id }}-{{ section.id }}" open>
```
ludoboludo, line 715: @metamoni when I check it with voice over it does work and announces it properly for me: [video](https://screenshot.click/07-02-9byc6-ynev1.mp4)

I wonder if there is an issue triggered due to the previous collapsible rows no containing anything and messing with the component code to update the attributes properly. But as you can see in the video when you get to the complementary product as a collapsible row, the state is conveyed properly.

metamoni: Interesting, it's still not working for me. But it's not a component you wrote from scratch, so it's not a blocking comment
https://user-images.githubusercontent.com/22390758/188917148-388e5165-191c-4ad4-90a8-55855f7facca.mov

sections/product-recommendations.liquid
ludoboludo: Interesting 🤔 changed the wrapper to `aside` and it announces automatically as `complimentary` based on the specs. I think I can keep it as is.

One thing though, not related to this change, when using voice over on Safari for some reasons it traps the focus on this block when it's not set to a `collapsible row`. And I'm not sure why, when I try voice over on chrome it works fine. when I try tabbing only it's not trapped and on TalkBack on Android I also don't get trapped on the block.

assets/component-complementary-products.css
```diff
@@ -0,0 +1,113 @@
...
+  margin-top: 0;
+}
+
+.complementary-products__container .card--card .card__inner .card__media {
+  border-radius: calc(var(--corner-radius) - var(--border-width) - var(--image-padding));
+}
+
+.complementary-products__container .card--horizontal .quick-add {
+  margin: 1rem 0 0;
+  max-width: 200px;
```
metamoni, line 93: What's the reason for using `px` instead of `rem` or `em` here? Maybe we should stick to the same unit for consistency.

assets/global.js
```diff
@@ -719,13 +719,17 @@ class SlideshowComponent extends SliderComponent {
 
   setSlideVisibility() {
     this.sliderItemsToShow.forEach((item, index) => {
-      const button = item.querySelector('a');
+      const linkElements = item.querySelectorAll('a');
       if (index === this.currentPage - 1) {
-        if (button) button.removeAttribute('tabindex');
+        if (linkElements.length ) linkElements.forEach(button => {
```
metamoni, line 724: Nit: there's a space before the closing bracket. Same on line 730

kmeleta (commented):
Just a few more mostly minor notes and questions. Also, might want to modify the PR title to exclude mention of the skeleton state

assets/component-card.css
```diff
@@ -107,6 +121,10 @@
   padding: 1.3rem 1rem;
 }
 
+.card--horizontal .card__information {
+  grid-row-start: 1;
+}
```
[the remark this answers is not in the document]
ludoboludo: Might have been more useful in the initial layouts 🤔 I tested with quick add and for products with sales badge and it seemed ok. I will remove it 👍

sections/main-product.liquid
```diff
@@ -693,6 +693,88 @@
                   </span>
                 </p>
               {%- endif -%}
+
+              {%- when 'complementary' -%}
+                <product-recommendations class="complementary-products quick-add-hidden no-js-hidden{% if block.settings.make_collapsible_row %} is-accordion{% endif %}" data-url="{{ routes.product_recommendations_url }}?section_id={{ section.id }}&product_id={{ product.id }}&limit={{ block.settings.product_list_limit }}&intent=complementary">
+                  {%- if recommendations.performed and recommendations.products_count > 0 -%}
+                    <aside {{ block.shopify_attributes }}{% if block.settings.make_collapsible_row %} class="product__accordion accordion"{% endif %}>
```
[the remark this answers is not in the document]
ludoboludo: I think for now it's ok as is 🤔 Since you get the heading that comes up right after. And I'm guessing this type of announcing might be expected by users that are familiar with it.

sections/product-recommendations.liquid
metamoni: It announces as complementary because of `aside`, which has that role built in. However, if you add an `aria-label`, then that helps give more context and differentiate from other `complementary` sections (although we don't have more of these at the moment).

Here's what it looks like with an `aria-label` (I added it in as "Product recommendations", but it would have to be translated):

https://user-images.githubusercontent.com/22390758/189345513-de49ffb9-b739-4f3f-ac95-c35b55744b5a.mp4

snippets/card-product.liquid
```diff
@@ -39,8 +41,9 @@
         {% if settings.card_style == 'card' %} color-{{ settings.card_color_scheme }} gradient{% endif %}
         {% if extend_height %} card--extend-height{% endif %}
         {% if card_product.featured_media == nil and settings.card_style == 'card' %} ratio{% endif %}
+        {{ extra_class }}
       "
-      style="--ratio-percent: {{ 1 | divided_by: ratio | times: 100 }}%;"
+      {% unless remove_ratio %} style="--ratio-percent: {{ 1 | divided_by: ratio | times: 100 }}%;"{% endunless %}
```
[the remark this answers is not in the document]
kmeleta: Ahh without image. Excellent. My next feedback was definitely going to handle that case some other way 🙂

metamoni (approved):
Great job! 👌

edmund-teh (approved):
I think it's finally ready 🚀 !

Issue #1939, opened 2022-08-31
Featured product video isn't letting you access the video controls


ludoboludo:
**Describe the current behavior**

When you have a feat.  product section with a featured media shown that is a video, you can click on it to play the video but it won't let you us the controls due to a `z-index` value on an element that is coming up on top of the video controls.

**Describe the expected behavior**

You should be able to us the video controls to pause/play, etc.

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 6.0.2

**Possible solution**

Remove the this declaration: 
```css
.featured-product .global-media-settings:after {
  z-index: 0;
}
```

**Additional context/screenshots**
[Video](https://screenshot.click/31-29-oeemd-amqn2.mp4)

PR #1940, opened 2022-08-31
Fix video controls on the feat product section

ludoboludo:
### PR Summary: 

This is a fix for the feat. product section's video. There is an issue where you can't click on the control due to some elements overlapping each other.

Fixes https://github.com/Shopify/dawn/issues/1939

Added a z-index of 1 on the iframe and video element so that they come up on top of the :after pseudo element.

### Other considerations

The z-index of 0 on the pseudo element is there for focus ring related reasons I believe so I decided not to touch it.

No visual impact just functionality. Controls are usable again.

### Testing steps/scenarios

- [ ] Add a feature product section, 
- [ ] set the product to be the `Test 3D models, video` one which has a video as a featured media
- [ ] Play the video and make sure you can pause/play, etc. 
- [ ] Then you can go in the admin and move around the medias so the featured one can be a youtube or vimeo video ([video on how to do it for that product](https://screenshot.click/31-56-yi1mq-pqlcj.mp4)). 
- [ ] Test as well with image but that should not have an impact on it. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=131486973974)
- [Editor](https://os2-demo.myshopify.com/admin/themes/131486973974/editor)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2850,6 +2850,7 @@ details-disclosure > details {
   border-radius: var(--media-radius);
   box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-shadow), var(--media-shadow-opacity));
   z-index: -1;
+  pointer-events: none;
 }
 
 .global-media-settings--no-shadow {
--- a/assets/section-featured-product.css
+++ b/assets/section-featured-product.css
@@ -4,10 +4,6 @@
   padding-bottom: 0;
 }
 
-.featured-product .global-media-settings:after {
-  z-index: 0;
-}
-
 .featured-product .product__modal-opener {
   margin-bottom: var(--media-shadow-vertical-offset);
 }
--- a/assets/section-featured-product.css
+++ b/assets/section-featured-product.css
@@ -8,6 +8,10 @@
   z-index: 0;
 }
 
+.featured-product .product__media-item iframe, .featured-product .product__media-item video {
+  z-index: 1;
+}
+
 .featured-product .product__modal-opener {
   margin-bottom: var(--media-shadow-vertical-offset);
 }
--- a/assets/section-featured-product.css
+++ b/assets/section-featured-product.css
@@ -8,10 +8,6 @@
   z-index: 0;
 }
 
-.featured-product .product__media-item iframe, .featured-product .product__media-item video {
-  z-index: 1;
-}
-
 .featured-product .product__modal-opener {
   margin-bottom: var(--media-shadow-vertical-offset);
 }


assets/section-featured-product.css
```diff
@@ -8,6 +8,10 @@
   z-index: 0;
 }
 
+.featured-product .product__media-item iframe, .featured-product .product__media-item video {
+  z-index: 1;
+}
+
```
kmeleta, line 14: So the purpose of the z-index: 0 isn't super clear. There's no real context on that line in the [original commit](https://github.com/Shopify/dawn/pull/1124/commits/56080b13c8080e62fd93eaacaf515bf7543555d9) 
Side note: This is why I want to use code comments more 🙂 

But my theory right now is that it was an attempt to prevent the media shadow from being hidden behind the featured product background, before we more globally implemented the `.isolate` z-index fix on sections. In such a case I'd think we can remove that rule altogether. Testing, of course, that media drop shadows don't disappear behind from this section, with and without secondary background enabled.

Regarding the effect on focus outlines.. The white line is actually more expected currently (see main product media). But I also think the z-index: 0 currently makes the shadow overlap the border which is not correct https://screenshot.click/31-21-4lkep-j2msq.mp4

kmeleta: We also have the option to add `pointer-events: none` on the pseudo element to solve the issue. I'm not fully in this context any more, but off-hand I'm not sure why that isn't already implemented for the shadow elements.

kmeleta (commented):
This might be less of a review and more.. me thinking out loud 😅  Let me know what you think.

assets/section-featured-product.css
ludoboludo: Oh true, `pointer-events` could be a nice alternative. 🤔

kmeleta: Cool, I messed with it quite a bit on my end and feel pretty good that it doesn't break anything, but I'll give it a final look later

eugenekasimov: After removing `z-index: 0` the issue with video has gone. Why do we still need `pointer-event`?

ludoboludo: It just seemed like a good thing to have on it to future proof things as we never need to click the shadows 🤔 So that in case anything similar ever happens it won't be a problem.

assets/base.css
```diff
@@ -2850,6 +2850,7 @@ details-disclosure > details {
   border-radius: var(--media-radius);
   box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius) rgba(var(--color-shadow), var(--media-shadow-opacity));
   z-index: -1;
+  pointer-events: none;
```
eugenekasimov, line 2853: Do we still need this `z-index: -1` after we added `pointer-events: none;`?

ludoboludo: Yeah so it can come up below other elements 🤔 👍

kmeleta: For context Eugene, the negative z-index is a necessary evil we implemented to allow various configurations of cards, media, etc to have shadows that don't overlap the adjacent siblings up and to the left of them, which is the natural outcome in CSS

assets/section-featured-product.css
kmeleta: Yeah I don't believe I'm aware of a case where this would be affecting anything currently, but I don't think it's wrong to include it. Theoretically could be added to the other global element types (cards and content containers) but I'd have to retest that.

kmeleta (approved):
Looks good. No regressions noticed with the removal of the featured product z-index

eugenekasimov (approved):
🚀

Issue #1921, opened 2022-08-17
Login button doesn't display on Mobile


rachellej:
**Describe the current behavior**
1. User goes to website with Dawn theme
2. User clicks hamburger icon to open navigation
3. Login button is not visible
---
4. User scrolls down the page
5. User opens navigation
6. Login button is now visible 

**Describe the expected behavior**
1. User goes to website with Dawn theme
2. User clicks hamburger icon to open navigation
3. Login button is visible

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 4.0.0
- Chrome Version 104.0.5112.69
- Android 12

**Possible solution**

The css is not accounting for the space taken up by the browser url. This is a common problem with using 100 `vh` units. 

There's a fix in the code for iphone: https://github.com/Shopify/dawn/blob/main/assets/global.js#L292

By removing the line above and adding the following, I was able to fix the issue for android chrome. I have NOT tested this on every device type, which is why this is an issue & not a PR :) 

Fix suggested [here](https://observablehq.com/@gianordoli/css-variables-2-tips-and-tricks)

```
  const windowWidth = document.documentElement.clientWidth;
  const windowHeight = document.documentElement.clientHeight;
  
  if (windowWidth < 740) {
    document.documentElement.style.setProperty('--viewport-height', `${windowHeight}px`); 
  }
```

And here's the suggested fix overall: 

**Additional context/screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.
![Screenshot_20220817-100801](https://user-images.githubusercontent.com/6774640/185201026-ecce0061-91e8-444d-8b1e-023882b16c8e.png)
![Screenshot_20220817-100815](https://user-images.githubusercontent.com/6774640/185201065-975f321d-6234-489c-a8e0-32de2632818c.png)

PR #1945, opened 2022-09-02
Android menu drawer account fix

ludoboludo:
### PR Summary: 

This PR is shipping a fix for Android devices where the menu drawer wasn't set with a proper height and customers weren't able to login into their account if necessary. 

Fixes https://github.com/Shopify/dawn/issues/1921

I used something similar than what was mentioned in the issue. Seems like Chrome is dealing with `vh` units the same way as Safari where the address bar on mobile is part of the size.

Added a bit of JS to add the right value and moved the code so that that height is updated everything you open the menu. This way when you rotate your device and open the menu again it uses the right value. 

### Decision log

|---|---|---|---|---|
| 1 | Moved the code setting the value of `--viewport-height` into the function dealing with the menu drawer being opened  | keep in the menu drawer constructor and the value is only updated on page reload  | By updating the value on when the menu is opened instead, we're account for more situation. Like a device rotation or even just when scrolling up, the address bar shows up again  | 🤷   |

Shows the account link in the menu drawer properly

### Testing steps/scenarios

- [ ] Using your own devices or browserstack, test both the last 2 version of chrome and safari on Android and iOS devices
- [ ] Test and make sure the account link is showing properly
- [ ] Remove the need from having an account on the store (from the admin) and make sure it's still looks good. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=131496247318)
- [Editor](https://os2-demo.myshopify.com/admin/themes/131496247318/editor)

### Checklist


assets/global.js
```diff
@@ -325,6 +323,10 @@ class MenuDrawer extends HTMLElement {
     if (detailsElement === this.mainDetailsToggle) {
       if(isOpen) event.preventDefault();
       isOpen ? this.closeMenuDrawer(event, summaryElement) : this.openMenuDrawer(summaryElement);
+
+      if (window.matchMedia('(max-width: 990px)')) {
```
dan-menard, line 327: This will apply to non-mobile browsers too, right? How significant a change is this for desktop browsers or tablets?

ludoboludo: Yes. So I used `990px` as it's the threshold where we're triggering the menu drawer. So it made sense to apply this fix on anything under 990 👍

dan-menard (approved):

Issue #870, opened 2021-11-10
[Product] Display variant inventory quantity


andrewetchen:
### Describe the enhancement you'd like

Add an inventory block on the product page and featured product section to display the inventory quantity on variant selection.

https://www.nngroup.com/articles/ecommerce-product-pages
>Customers need to have enough information to make an informed purchase decision. The product page, or product-detail page, is where users decide whether and what to buy. The page must include complete product information, educating and informing the user about the product in a straightforward way.

Displaying the product inventory count (_or stock level_) can help customers make informed purchasing decisions. For stores with limited product quantities, an inventory count can help create a positive sense of urgency and excitement during the buyer journey. And with [Black Friday Cyber Monday](https://www.shopify.ca/bfcm) and the holidays soon approaching, there could be other use cases.

**Potential drawbacks**
- Stores that allow customers to purchase the item when it's [out of stock](https://help.shopify.com/en/manual/products/inventory/track_inventory#selling-when-out-of-stock).
- Stores that [sell services or digital products](https://help.shopify.com/en/manual/products/digital-service-product).

A quick solution would be to make sure we don't enable the inventory block by default on the product page and featured product section.

###  Additional context

The Shopify Help Center currently lists the inventory feature as a supported customization for Dawn:
- [Show variant inventory on the product page](https://help.shopify.com/en/manual/online-store/themes/os20/themes-by-shopify/dawn/customize#show-variant-inventory-on-the-product-page)
- [Eligible](https://help.shopify.com/en/manual/online-store/themes/theme-support#shopify-design-policy) merchants can submit a support request to have this added to their Dawn theme.

**Here's a video showing the inventory on the product page**

https://user-images.githubusercontent.com/28404165/141182865-e0124baa-7dc1-40d6-b1ec-da6e51a49654.mp4

### Next steps

Submit a PR to add an inventory block on the product page and featured product section.

gregjotau:
![screenshot](https://user-images.githubusercontent.com/13047497/157449435-b1c90bc1-3c76-40f6-8b99-da4225980e66.png)

```
The problem is that the inventory quantity value is not included in the variant javascript object, so we can't check it from there, and this theme doesn't have a master variants select to which we could have added thw inventory quantity values for each select option.

```

Getting some feedback on inventory quantity for a customization we want done. Is there a plan to make inventory quantity more "available" in the theme as a whole @ludoboludo ?

ludoboludo:
[...]

PR #1979, opened 2022-09-20
Inventory status block

eugenekasimov:
### PR Summary: 

This PR adds an Inventory block to the main-product section. The block allows you to display/hide an amount of items in stock. You can  also set a threshold to display "In stock" or "Low stock" status. If stock is 0 then "Out of stock" is displayed. For unavailable variants nothing will be displayed. 

[Internal Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=208%3A23596)

Fixes https://github.com/Shopify/dawn/issues/870

- I utilized the `renderProductInfo ()` and ` setUnavailable` functions to update the inventory stock upon variant selection or hide the block if the varian is not available or is sold out.
- I included the inventory in the [product form `<noscript>` wrapper] to ensure the inventory quantity is displayed when JS is disabled.
- I added `.product--no-media .product__inventory` as an additional declaration in the [`section-main-product.css`] to center the inventory block for products with no media/images.

### Decision log

|---|---|---|---|---|
| 1 | When the variant is set to continue selling we're showing `In stock`  | We could show a different string like `backorder`  | There are too many different scenario as for why a merchant might set their product to `continue selling` so we don't want to take a guess this way  | Saying `in stock` can be odd as products with an inventory above `0` and lower than the threshold set in the editor setting will show as `Low stock`. Going from one to the other seem weird  |

### Testing steps/scenarios
- [ ] When you test please make sure you choose a product where at least one variant has a few items in stock, one variant has stock 0 and one of the variants is not available. A good example will be [Louise Slide Sandal](https://os2-demo.myshopify.com/admin/themes/131581935638/editor?previewPath=%2Fproducts%2Flouise-slide-sandal&section=template--15870572757014__main).
- [ ] Test at least one product with variant that has 0 in stock but with check-box on `Continue selling when out of stock` in Admin.  A good example will be [Louise Slide Sandal](https://os2-demo.myshopify.com/admin/themes/131581935638/editor?previewPath=%2Fproducts%2Flouise-slide-sandal&section=template--15870572757014__main). Size 36 and 41. It's supposed to show **In stock**.  
- [ ] When you test, switch randomly variants to make sure that Inventory status changes icon status/quantity or disappears if there is no stock or the variant is not available. Don't forget to test with `Show Inventory count` setting off.
- [ ] Test a product as a featured product.
- [ ] Test a product through quick-add-module.
- [ ] Make sure that text style applies when you change it in the block settings. 
- [ ] Turn off JS in a browser and make sure that you still can see quantity of each variant.

### Demo links

[Editor](https://os2-demo.myshopify.com/admin/themes/131581935638/editor?previewPath=%2Fproducts%2Flouise-slide-sandal&section=template--15870572757014__main&block=template--15870572757014__main%2F9fadf303-6422-4139-92a1-22ee9f560ebc)


YoannJailin:
Hi Eugene! Here are the updates i told you about : [Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=2915%3A109294)

My recommendation would be :
- Use a range slider instead of the counter 0 to 100 (step : 1, default 10)
- 0 = always « In stock »
- If stock is superior to threshold, « In stock »
- If stock = threshold, « Low stock »
- If stock is inferior to threshold, « Low stock »
- Since the main request is to show the inventory count (and not the status) I would check « Show inventory » by default.
- We did some updates on the settings order and content.

Also, we did a quick updates on the orangerand green [status UI](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=2909%3A109470)!

eugenekasimov:
> Not sure what's causing this, but I'm seeing different things displayed for two different products 🤔
> 
> The stock indicator is hidden for some variants/sizes of the Louise Slide Sandal, which is what we expect. But for some variants/sizes, the price gets hidden completely. Also, for Shoe Test variant product I'm seeing "not quantity" for variants that aren't in stock.
> 
>  30-31-ww779-r16tt.mp4

Thanks for paying attention to this. This "issue" not related to this PR, but I have already brought it up to designers so they know about it. When some existing variant has stock 0 then you see `sold out`, but sometimes there is a variant which doesn't exist for selected option (but you still see it because it exist for another option) and then you see just empty space without price.

eugenekasimov:
I have discovered an edge case that is not covered here. What do we suppose to display in case if there is 0 in stock but a merchants chose _continue selling when out of stock_? Now in this case we see nothing and this is quite confusing because you still can buy a product. @YoannJailin @ludoboludo 
Screen Shot 2022-10-05 at 11 47 43 AM

locales/en.default.schema.json
```diff
@@ -773,6 +773,30 @@
...
+              "options__2": {
+                "label": "Subtitle"
+              },
+              "options__3": {
+                "label": "Uppercase"
+              }
+            },
+            "inventory_threshold": {
+              "label": "Low inventory threshold",
+              "info": "Choose 0 to always show in stock if available."
```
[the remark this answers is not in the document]
ludoboludo: I just find it a bit confusing. I think the wording could be better. But maybe it's only confusing to me. cc: @YoannJailin

sections/featured-product.liquid
```diff
@@ -134,6 +155,44 @@
                     {%- endform -%}
                   </div>
                 {%- endif -%}
+              {%- when 'inventory' -%}
+                {%- if product.selected_or_first_available_variant.inventory_quantity != 0 -%}
+                  <p
+                    class="product__inventory no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}"
```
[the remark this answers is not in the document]
YoannJailin: @ludoboludo yes, fonts sizes are consistant with the text block style options!

metamoni (changes requested):
@eugenekasimov I noticed an issue with the featured product. In my editor, I'm able to add a featured product, add the stock size, and save it. But when I go to the store, I'm only seeing the Example product. Not sure what's causing this, but it doesn't seem to happen on `main`. Here's a video of the current behaviour:

https://user-images.githubusercontent.com/22390758/194276903-f790aebd-2d51-4d8f-a630-809a8ff157df.mp4

sections/featured-product.liquid
```diff
@@ -134,6 +155,45 @@
...
+                    id="Inventory-{{ section.id }}"
+                    role="status"
+                    {{ block.shopify_attributes }}
+                  >
+                    {%- if product.selected_or_first_available_variant.inventory_quantity
+                        <= block.settings.inventory_threshold
+                    -%}
+                      <svg width="15" height="15">
```
metamoni, line 169: This is a decorative image, so we should hide it from screen readers with `role="presentation"`.

eugenekasimov: Good call, Monica. Thanks.

assets/global.js
```diff
@@ -846,11 +846,19 @@ class VariantSelects extends HTMLElement {
...
+
         if (source && destination) destination.innerHTML = source.innerHTML;
+        if (inventorySource && inventoryDestination) inventoryDestination.innerHTML = inventorySource.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
+        const inventory = document.getElementById(`inventory-${this.dataset.section}`);
 
         if (price) price.classList.remove('visibility-hidden');
+        if (inventory && inventorySource == null) inventory.classList.add('hidden');
+        if (inventory && inventorySource !== null) inventory.classList.remove('hidden');
```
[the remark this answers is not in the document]
YoannJailin: Hey, i think it would be better to remove all the white space when a block disappears (even for price) but i agree we can just keep the current logic for now (keeping the white space like for prices). Let's just fix that issue later, for every block.

YoannJailin:
Hi @eugenekasimov @ludoboludo 

- For the message under the threshold : 
- 

> I have discovered an edge case that is not covered here. What do we suppose to display in case if there is 0 in stock but a merchant chose _continue selling when out of stock_? Now in this case we see nothing and this is quite confusing because you still can buy a product. @YoannJailin @ludoboludo Screen Shot 2022-10-05 at 11 47 43 AM

@eugenekasimov I recommend we just display the "In stock" status with no inventory count, even if the option is checked. I don't think it's relevant to add any information message about that, because we are going to add a lot of extra informations that will make the editor more confusing. I think we should keep it simple at this point.

sections/featured-product.liquid
```diff
@@ -134,6 +155,45 @@
...
                 {%- endif -%}
+              {%- when 'inventory' -%}
+                {%- if product.selected_or_first_available_variant.inventory_quantity != 0 -%}
+                  <p
+                    class="product__inventory no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}"
+                    id="Inventory-{{ section.id }}"
+                    role="status"
+                    {{ block.shopify_attributes }}
+                  >
+                    {%- if product.selected_or_first_available_variant.inventory_quantity
```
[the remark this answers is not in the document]
eugenekasimov: Fair enough. I think it's an edge case. We have this logic for other blocks as well. I think we should consider it as a different issue.

eugenekasimov:
Let me sum up. 
- To be consistent I kept whitespace and hide the element from screen readers. When the decision is made we will change it for all blocks in a different PR.
- For variants that have stock 0 we display "In stock" status no mater if `Show inventory count` off/on.
- There is still a concern about `Choose 0 to always show in stock if available`. Maybe we can change it latter to avoid a delay.

sections/featured-product.liquid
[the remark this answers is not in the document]
metamoni: That depends on what the other blocks do. Setting a threshold for low stock is linked to data (the number of items left available), so it can have different consequences compared to just a visual difference. Indicating "low stock" vs "in stock" in two different places can suggest to buyers that there's an issue somewhere. I think that kind of data should be consistent across a shopping site. @ludoboludo, @lucianosousa, @YoannJailin what do you think?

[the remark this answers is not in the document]
eugenekasimov: I remember originally Dan and Yoann mentioned that they would like to have a threshold setting in the admin. Until it happens I more lean towards  keeping the feature for both sections but add additional info for the block in Featured product saying something like "Pls make sure to set threshold exactly the same as you set it in Product page"

assets/global.js
```diff
@@ -878,9 +886,13 @@ class VariantSelects extends HTMLElement {
     const addButton = button.querySelector('[name="add"]');
     const addButtonText = button.querySelector('[name="add"] > span');
     const price = document.getElementById(`price-${this.dataset.section}`);
+    const inventory = document.getElementById(`Inventory-${this.dataset.section}`);
+
     if (!addButton) return;
     addButtonText.textContent = window.variantStrings.unavailable;
     if (price) price.classList.add('visibility-hidden');
+    if (inventory) inventory.classList.add('visibility-hidden'), inventory.setAttribute('aria-hidden', 'true');
```
ludoboludo, line 894: I don't believe the `aria-hidden` part here is actually needed. Because it's running for unavailable variants, it never re renders anything. And that means no change to what's bing announced. So in this scenario we don't need it.

assets/section-main-product.css
```diff
@@ -1245,6 +1245,26 @@ a.product__text {
...
+  justify-content: center;
+}
+
+p.visibility-hidden:empty {
+  display: block;
+}
+
+p.visibility-hidden:empty::after {
+  content: "#";
+}
```
ludoboludo, line 1266: I think those declarations are too generic. They could be impacting other `p` elements in different part of the theme so they should be more specific to the inventory feature. Using `.product__inventory` instead of `p`.

locales/en.default.schema.json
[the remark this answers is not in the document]
ludoboludo: I even wonder if it needs help text ? Or maybe something to explain how this value they're using is used. 
`Anything below the threshold set will show a `low stock` status unless it's 0.`

sections/featured-product.liquid
```diff
@@ -325,8 +372,15 @@
...
                           >
                             {{ variant.title }}
-                            {%- if variant.available == false %} - {{ 'products.product.sold_out' | t }}{% endif %}
                             - {{ variant.price | money | strip_html }}
+                            {%- if variant.available == false %} - {{ 'products.product.sold_out' | t }}{% endif %}
+                            {% if variant.available and variant.inventory_management == 'shopify' -%}
+                              -
+                              {{
+                                'products.product.inventory_in_stock_show_count'
+                                | t: quantity: variant.inventory_quantity
```
ludoboludo, line 381: This should be matching the block setting. Or since we can't make it match since it's based on a setting from a different block we should default to not showing the inventory quantity. Just `in stock` if above `0`.

ludoboludo: In fact you can probably do a else rather than a second if statement: 
```html
{%- if variant.available == false %} - {{ 'products.product.sold_out' | t }}{% elsif variant.available and variant.inventory_management == 'shopify' -%}
                              -
                              {{
                                'products.product.inventory_in_stock'
                                | t

File: sections/main-product.liquid [hunk 9]
```diff
@@ -361,6 +361,45 @@
...
+                    <svg width="15" height="15" role="presentation">
+                      <circle cx="7.5" cy="7.5" r="7.5" fill="rgb(238,148,65, 0.3)"/>
+                      <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb(238,148,65)"/>
+                    </svg>
+                    {%- if block.settings.show_inventory_quantity -%}
+                      {{- 'products.product.inventory_low_stock_show_count' | t: quantity: product.selected_or_first_available_variant.inventory_quantity -}}
+                      {%- else -%}
+                        {{- 'products.product.inventory_low_stock' | t -}}
+                    {%- endif -%}
+                    {%- else -%}
```
ludoboludo, line 382: There is some formatting/indentation issues here

sections/main-product.liquid
```diff
@@ -361,6 +361,45 @@
...
+                        <circle cx="7.5" cy="7.5" r="7.5" fill="rgb(62,214,96, 0.3)"/>
+                        <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb(62,214,96)"/>
+                      </svg>
+                        {%- if block.settings.show_inventory_quantity -%}
+                          {{- 'products.product.inventory_in_stock_show_count' | t: quantity: product.selected_or_first_available_variant.inventory_quantity -}}
+                          {%- else -%}
+                            {{- 'products.product.inventory_in_stock' | t -}}
+                        {%- endif -%}
+                  {%- endif -%}
+                    {%- else -%}
```
ludoboludo, line 393: Same here. Indentation/formatting issues

sections/main-product.liquid
```diff
@@ -578,8 +617,9 @@
                           value="{{ variant.id }}"
                         >
                           {{ variant.title }}
-                          {%- if variant.available == false %} - {{ 'products.product.sold_out' | t }}{% endif %}
                           - {{ variant.price | money | strip_html }}
+                          {%- if variant.available == false %} - {{ 'products.product.sold_out' | t }}{% endif %}
+                          {% if variant.available and variant.inventory_management == 'shopify' -%} - {{ 'products.product.inventory_in_stock_show_count' | t: quantity: variant.inventory_quantity }}{% endif %}
```
ludoboludo, line 622: Same here as for the feat. product section. We shouldn't show the inventory quantity by default. And we could have an if/elsif rather than 2 if statements 👍

assets/section-main-product.css
eugenekasimov: good call, thanks Ludo!

sections/featured-product.liquid
eugenekasimov: Thanks for catching this. I added a full logic to cover all three cases when "Show count" on and off, and when it's set to continue selling even if it's sold out. It may be too much code and unnecessary for no JS case though. What do you think @ludoboludo . Should I keep it or make it simple and show just a status In Stock/Low Stock?

metamoni (dismissed):

locales/en.default.json
```diff
@@ -108,6 +108,10 @@
       "choose_options": "Choose options",
       "choose_product_options": "Choose options for {{ product_name }}",
       "description": "Description",
+      "inventory_in_stock": "In stock",
+      "inventory_in_stock_show_count": "{{ quantity }} in stock",
+      "inventory_low_stock": "Low stock",
+      "inventory_low_stock_show_count": "Low stock: {{ quantity }} left",
```
ludoboludo, line 114: I wonder if it could be useful for a merchant to only display this block when the amount in stock is below the threshold. So only showing `Low stock: X left` as a way to push customers to action their purchase without showing the `In stock` version. 

It might be not great in some scenarios with some variants in stock and others with low stock but it could be something none the less. Some stores with single variant products, etc, could have a use for it. 

Just wanted to bring it up, not to change anything in this PR but to check if we thought about this specific use case. 

cc: @YoannJailin

sections/main-product.liquid
```diff
@@ -106,6 +106,45 @@
...
+                    {%- endif -%}
+                  {%- endif -%}
+                {%- else -%}
+                  {%- if product.selected_or_first_available_variant.inventory_policy == 'continue' -%}
+                    <svg width="15" height="15" role="presentation">
+                      <circle cx="7.5" cy="7.5" r="7.5" fill="rgb(62,214,96, 0.3)"/>
+                      <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb(62,214,96)"/>
+                    </svg>
+                    {{- 'products.product.inventory_in_stock' | t -}}
+                  {%- endif -%}
```
ludoboludo, line 145: Even though we already have a sold out badge, I'd prefer if we had a sold out/out of stock message rather than a blank. 
Ideally, i think the blank would only come up for unavailable variants: [video](https://screenshot.click/18-07-xdlac-4bivj.mp4)

cc: @YoannJailin

sections/main-product.liquid
```diff
@@ -310,8 +349,34 @@
...
+                                    {%- endif -%}
+                                  {%- endif -%}
+                                {%- else -%}
+                                  {%- if variant.inventory_policy == 'continue' -%}
+                                    {{- 'products.product.inventory_in_stock' | t -}}
+                                  {%- endif -%}
+                                {%- endif -%}
+                              {%- endif -%}
+                            {%- endif -%}
+                          {%- endfor -%}
```
ludoboludo, line 379: Yeah I think this is probably overkill. I'd say if it's no JS then we just don't show anything to avoid having to loop through blocks and having to re run so much logic. 

What do you think @metamoni ?

ludoboludo (commented):
Left a few small comments. It's very close to be good, I just want to question a couple design decision. Let's see what Yoann thinks 👍

locales/en.default.schema.json
```diff
@@ -1645,6 +1645,30 @@
...
+            },
+            "show_inventory_quantity": {
+              "label": "Show inventory count"
+            }
+          }
+        },
```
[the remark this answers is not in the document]
eugenekasimov: Thanks Ludo, I've done it many times. I just prefer to do it when I'm sure we are not going to change anything in locales.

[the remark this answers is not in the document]
ludoboludo: Sounds good 👍 You can probably request them now since Yoann seemed to be ok we the copy and Monica and I haven't had new comments about it 🤔 They can take a few days to come back so to be safe, might as well do it now.

locales/en.default.json
YoannJailin: Hey @ludoboludo, that's a good point. I considered it as well in the beginning, but after trying that kind of block on our partner themes, most of them use that "In Stock" state by default. Made more sense while configuring the settings and especially the threshold.

sections/main-product.liquid
ludoboludo: Nice for designs 👌 And yeah I feel like it'd be nicer with it. If the block is supposed to show me the inventory status then show it to me in all scenarios, when it's out of stock, low stock and lots of stock. If I don't see it for an unavailable variant, I'm ok because it's expected as it doesn't exist.

eugenekasimov: @YoannJailin it should be an easy thing. I'll update it soon.

metamoni (approved):

YoannJailin (approved):

sections/main-product.liquid
```diff
@@ -106,6 +106,51 @@
...
+                    {{- 'products.product.inventory_in_stock' | t -}}
+                  {%- else -%}
+                    <svg width="15" height="15" role="presentation">
+                      <circle cx="7.5" cy="7.5" r="7.5" fill="rgb(200,200,200, 0.3)"/>
+                      <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb(200,200,200)"/>
+                    </svg>
+                    {{- 'products.product.inventory_out_of_stock' | t -}}
+                  {%- endif -%}
+                {%- endif -%}
+              </p>
```
baldurh, line 153: I feel like this code block could be cleaned up to reduce it’s cognitive complexity. First of all it seems this should only render anything when `product.selected_or_first_available_variant.inventory_management == 'shopify'` so that could just wrap the whole block to simplify the inner conditions. Also, it might be better to have the base case come first (`inventory_quantity == 0`)
Secondly, it might clean things up a bit to have a liquid block at the top of this and assign some of the values and booleans to variables to use inside here.

I personally feel like these changes would make this block way more readable, and therefore maintainable

ludoboludo: Yeah there is quite a few checks. Not sure it was worth the assigns. But maybe it would make everything easier to read.

sections/main-product.liquid
```diff
@@ -106,6 +106,51 @@
                   {{ form | payment_terms }}
                 {%- endform -%}
               </div>
+            {%- when 'inventory' -%}
+              <p
+                class="product__inventory no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if product.selected_or_first_available_variant.inventory_management != 'shopify'%} visibility-hidden{% endif %}"
+                {{ block.shopify_attributes }}
+                id="Inventory-{{ section.id }}"
+                role="status"
```
ludoboludo, line 114: I wonder if we should actually apply this role in all situation 🤔 The reason I say that is because we're already conveying the sold out and unavailable state when selecting different variant options. So it doubles the announcement about it. 
Maybe we can conditionally apply it when the inventory quantity is above `0` or `when it's 0 and inventory_policy == 'continue'`.

ludoboludo (commented):
Good a few more comments in a couple extra scenarios. Let me know if you want to pair to discuss them "in person"

sections/main-product.liquid
```diff
@@ -106,6 +106,51 @@
                   {{ form | payment_terms }}
                 {%- endform -%}
               </div>
+            {%- when 'inventory' -%}
+              <p
+                class="product__inventory no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if product.selected_or_first_available_variant.inventory_management != 'shopify'%} visibility-hidden{% endif %}"
```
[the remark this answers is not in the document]
eugenekasimov: We don't want to put any condition before <p> because we can re-render things only inside <p>.

sections/main-product.liquid
```diff
@@ -106,6 +106,51 @@
...
+                {{ block.shopify_attributes }}
+                id="Inventory-{{ section.id }}"
+                role="status"
+              >
+                {%- if product.selected_or_first_available_variant.inventory_quantity != 0 and product.selected_or_first_available_variant.inventory_management == 'shopify' -%}
```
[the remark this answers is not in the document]
eugenekasimov: As we discussed, let's do this improvement latter after we hear opinion on this from different devs. Right now I'd prefer to move forward, since it's not a technical issue.

sections/main-product.liquid
ludoboludo: For this I have created a follow up issue for us to find a better solution to announcing the information without repeating ourselves too much: https://github.com/Shopify/dawn/issues/2061

sections/main-product.liquid
```diff
@@ -106,6 +106,53 @@
...
+                      {{- 'products.product.inventory_in_stock' | t -}}
+                    {%- else -%}
+                      <svg width="15" height="15" role="presentation">
+                        <circle cx="7.5" cy="7.5" r="7.5" fill="rgb(200,200,200, 0.3)"/>
+                        <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb(200,200,200)"/>
+                      </svg>
+                      {{- 'products.product.inventory_out_of_stock' | t -}}
+                    {%- endif -%}
+                  {%- endif -%}
+                {%- endif -%}
```
baldurh, line 154: Continuing with my earlier comment and looking at the code a bit more I realise there is a lot of repeated code here. Mainly the svg code but this in general hard to read and understand. I also think there is a bug in the current implementation. If you have a product with negative inventory you will still go into the in-stock or low-stock condition, because`inventory_quantity != 0` will match negative numbers and you’ll never reach the ‘continue selling’ condition (in stock), unless the inventory is at exactly 0.

Below is an example of how this _could_ look like if we move the conditional logic into a liquid tag:
<sub>Note: I understand this is few lines longer, but personally I feel this is more readable. Also, I have not run this</sub>
```liquid
{%- if product.selected_or_first_available_variant.inventory_management == 'shopify' -%}
  {%- liquid
    assing variant = product.selected_or_first_available_variant
    assign inventory_quantity = variant.inventory_quantity
    assign show_inventory_quantity = block.settings.show_inventory_quantity
    assign inventory_threshold = block.settings.inventory_threshold
    assign in_stock = false
    assign low_stock = false
    assign continue_selling = false
    assign color_in_stock_rgb = '62, 214, 96'
    assign color_low_stock_rgb = '238, 148, 65'
    assign color_sold_out_rgb = '200, 200, 200'
    assign color = ''
    assign text = ''
    
    if inventory_quantity > 0
      assign in_stock = true
      if inventory_quantity <= inventory_threshold
        assign low_stock = true
      endif
    endif

    if variant.inventory_policy == 'continue'
      assign continue_selling = true
    endif

    if in_stock
      assign color = color_in_stock_rgb
      assign text = 'products.product.inventory_in_stock' | t
      if low_stock
        assign color = color_low_stock_rgb
        if show_inventory_quantity
          assign text = 'products.product.inventory_low_stock_show_count' | t: quantity: inventory_quantity
        else
          assign text = 'products.product.inventory_low_stock' | t
        endif
      elsif show_inventory_quantity
        assign text = 'products.product.inventory_in_stock_show_count' | t: quantity: inventory_quantity
      endif
    elsif continue_selling
      assign color = color_in_stock_rgb
      assign text = 'products.product.inventory_in_stock' | t
    else
      assign color = '200, 200, 200'
      assign text = 'products.product.inventory_out_of_stock' | t
    endif
  -%}
  <svg width="15" height="15" role="presentation">
    <circle cx="7.5" cy="7.5" r="7.5" fill="rgb({{- color -}}, 0.3)"/>
    <circle cx="7.5" cy="7.5" r="5" stroke="rgb(255, 255, 255)" stroke-width="1" fill="rgb({{- color -}})"/>
  </svg>
  {{- text -}}
{%- endif -%}
```

baldurh: Sounds good 😊 glad to help!

ludoboludo (dismissed):
Works well 👌 Didn't notice any bugs. 

I have one small suggestion and I can re approve once it's been implemented, if decided to be necessary.

metamoni (approved):
I've tested this and it to work well for me 👍

sections/main-product.liquid
```diff
@@ -106,6 +106,53 @@
...
+              <p
+                class="product__inventory no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if product.selected_or_first_available_variant.inventory_management != 'shopify' %} visibility-hidden{% endif %}"
+                {{ block.shopify_attributes }}
+                id="Inventory-{{ section.id }}"
+                role="status"
+              >
+                {%- if product.selected_or_first_available_variant.inventory_management == 'shopify' -%}
+                  {%- if product.selected_or_first_available_variant.inventory_quantity > 0 -%}
+                    {%- if product.selected_or_first_available_variant.inventory_quantity <= block.settings.inventory_threshold -%}
+                      <svg width="15" height="15" role="presentation">
```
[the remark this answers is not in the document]
ludoboludo: We should try and add this change before we merge 👍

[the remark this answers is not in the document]
eugenekasimov: Sure, I'll do it now. Thanks Ludo.

Issue #465, opened 2021-08-23
[Zoom] Product modal opening position


melissaperreault:
Let's try to have the position center/center when opening the modal on the product template. This is to give the chance for customers to see details of the media. 

It is more likely that media details will appear in the middle of the media as oppose to relying at center/top where this only benefit vertical tall products.

bredowmax:
The modal is the biggest single weakness of Dawn so far, rendering it unusable for most merchant if no custom code is added. I added screenshots in #355

Pease find another way for shoppers to view these product images. These modals just don't work.

PR #1983, opened 2022-09-22
Add zoom flexibility

### LINKED ISSUE #1704: [Product page] Add image zoom flexibility
Update product information to include an image zoom selector. Provide options to `open lightbox`, `zoom on hover`, and `No zoom`. [Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=483%3A28704)

@metamoni, we can split this into two parts. The latter is a nice-to-have, but there seems to be some open PRs with potential solutions that could be a quick win. Would you please start exploring what would be required to do both (disable and zoom on hover)

**Version 1**
- [ ] Add checkbox to disable zoom

**Version 2**
- [ ] Add option to zoom on hover

metamoni:
### PR Summary: 
Add a new "Image zoom" selector in the "Product information" section, providing three options:
- Open lightbox (currently available/default option)
- Zoom on hover (new)
- No zoom (new)

Fixes #1704

#### No zoom
When this option is selected in the editor, the zoom icon is hidden from the top left corner. Clicking the image or hovering over it doesn't do anything.

#### Zoom on hover
Having looked at a few different options, including whether this can be solved with CSS alone, I have chosen the following approach:
- when an image is clicked, we create an overlay that takes the full-size image as its background
- when hovering over the full-size image, we use the position of the mouse to determine how to move the image around
- when the full-size image is clicked, the overlay is removed and the original image is displayed

### Other considerations
The zoom ratio on hover is currently set to 3, but this can be easily adjusted.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

Merchants will now see an image "Image zoom" option. The default option will be "Open lightbox", which is the option we currently have in place.

### Testing steps/scenarios
- [] Go to the [theme editor](https://os2-demo.myshopify.com/admin/themes/131508699158/editor) and select a product
- [] Click the image to bring up the "Product information" section
- [] You should be able to see a new "Image zoom" selector with the following options: "Open lightbox", "Zoom on hover", "No zoom"
- [] Select "Open lightbox" and verify that a zoom icon appears in the top left corner of the image
- [] Click the image and verify that the full-size image opens up in a modal. Close the modal
- [] Select "No zoom" and verify that the zoom icon is not present in the top left corner of the image
- [] Click the image and verify that nothing happens
- [] Select "Zoom on hover" and verify that the zoom icon is not present in the top left corner of the image
- [] Move your mouse over the image and verify that the mouse cursor looks like a magnifier with the "+" symbol
- [] Click the image and verify that the image is magnified
- [] While the image is enlarged, verify that the cursor looks like a magnifier with the "-" symbol
- [] Move your mouse around and verify that the image moves around accordingly
- [] Click the enlarged image and verify that the image and cursor have gone back to the original state
- [] With "Zoom on hover" selected, save your changes and go to the [store](https://os2-demo.myshopify.com/?preview_theme_id=131508699158). Verify that the setting has saved
- [] Go back to the editor and select "No zoom". Save your changes. Head to the [store](https://os2-demo.myshopify.com/?preview_theme_id=131508699158) and verify that the setting has saved

### Demo links
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=131508699158)
- [Editor](https://os2-demo.myshopify.com/admin/themes/131508699158/editor)

### Checklist

### DIFF
--- a/assets/magnify.js
+++ b/assets/magnify.js
@@ -1,0 +1,65 @@
+const images = document.querySelectorAll('.product__media-image');
+
+const magnify = (image, scale) => {
+  let glass, w, h, bw;
+
+  glass = document.createElement("div");
+  glass.setAttribute("class", "img-magnifier-glass");
+  image.parentElement.insertBefore(glass, image);
+
+  glass.addEventListener('click', () => {
+    glass.remove();
+  });
+
+  /* Set background properties for the magnifier glass: */
+  glass.style.backgroundImage = "url('" + image.src + "')";
+  glass.style.backgroundRepeat = "no-repeat";
+  glass.style.backgroundSize = (image.width * scale) + "px " + (image.height * scale) + "px";
+  bw = 3;
+  w = glass.offsetWidth / 2;
+  h = glass.offsetHeight / 2;
+
+  /* Execute a function when someone moves the magnifier glass over the image: */
+  glass.addEventListener("mousemove", moveMagnifier);
+  image.addEventListener("mousemove", moveMagnifier);
+
+  function moveMagnifier(e) {
+    var pos, x, y;
+    /* Prevent any other actions that may occur when moving over the image */
+    e.preventDefault();
+    /* Get the cursor's x and y positions: */
+    pos = getCursorPos(e);
+    x = pos.x;
+    y = pos.y;
+    /* Prevent the magnifier glass from being positioned outside the image: */
+    if (x > image.width - (w / scale)) {x = image.width - (w / scale);}
+    if (x < w / scale) {x = w / scale;}
+    if (y > image.height - (h / scale)) {y = image.height - (h / scale);}
+    if (y < h / scale) {y = h / scale;}
+    /* Set the position of the magnifier glass: */
+    glass.style.left = (x - w) + "px";
+    glass.style.top = (y - h) + "px";
+    /* Display what the magnifier glass "sees": */
+    glass.style.backgroundPosition = "-" + ((x * scale) - w + bw) + "px -" + ((y * scale) - h + bw) + "px";
+  }
+
+  function getCursorPos(e) {
+    var a, x = 0, y = 0;
+    e = e || window.event;
+    /* Get the x and y positions of the image: */
+    a = image.getBoundingClientRect();
+    /* Calculate the cursor's x and y coordinates, relative to the image: */
+    x = e.pageX - a.left;
+    y = e.pageY - a.top;
+    /* Consider any page scrolling: */
+    x = x - window.pageXOffset;
+    y = y - window.pageYOffset;
+    return {x : x, y : y};
+  }
+}
+
+images.forEach(image => {
+  image.addEventListener('click', () => {
+    magnify(image, 2);
+  });
+});
--- a/assets/magnify.js
+++ b/assets/magnify.js
@@ -1,65 +1,37 @@
-const images = document.querySelectorAll('.product__media-image');
+// create a container and set the full-size image as its background
+const createOverlay = (image) => {
+  overlay = document.createElement('div');
+  overlay.setAttribute('class', 'image--full-size');
+  overlay.setAttribute('aria-hidden', 'true');
+  overlay.style.backgroundImage = `url('${image.src}')`;
+  return overlay;
+};
 
-const magnify = (image, scale) => {
-  let glass, w, h, bw;
+const moveWithHover = (image, event, zoomRatio) => {
+  // calculate mouse position
+  const ratio = image.height / image.width;
+  const container = event.target.getBoundingClientRect();
+  const xPosition = event.clientX - container.left;
+  const yPosition = event.clientY - container.top;
+  const xPercent = `${xPosition / (overlay.clientWidth / 100)}%`;
+  const yPercent = `${yPosition / ((overlay.clientWidth * ratio) / 100)}%`;
 
-  glass = document.createElement("div");
-  glass.setAttribute("class", "img-magnifier-glass");
-  image.parentElement.insertBefore(glass, image);
+  // determine what to show in the frame
+  overlay.style.backgroundPosition = `${xPercent} ${yPercent}`;
+  overlay.style.backgroundSize = `${image.width * zoomRatio}px`;
+};
 
-  glass.addEventListener('click', () => {
-    glass.remove();
-  });
+const magnify = (image, zoomRatio) => {
+  // add full-size image on top of original
+  const overlay = createOverlay(image);
+  image.parentElement.insertBefore(overlay, image);
 
-  /* Set background properties for the magnifier glass: */
-  glass.style.backgroundImage = "url('" + image.src + "')";
-  glass.style.backgroundRepeat = "no-repeat";
-  glass.style.backgroundSize = (image.width * scale) + "px " + (image.height * scale) + "px";
-  bw = 3;
-  w = glass.offsetWidth / 2;
-  h = glass.offsetHeight / 2;
-
-  /* Execute a function when someone moves the magnifier glass over the image: */
-  glass.addEventListener("mousemove", moveMagnifier);
-  image.addEventListener("mousemove", moveMagnifier);
-
-  function moveMagnifier(e) {
-    var pos, x, y;
-    /* Prevent any other actions that may occur when moving over the image */
-    e.preventDefault();
-    /* Get the cursor's x and y positions: */
-    pos = getCursorPos(e);
-    x = pos.x;
-    y = pos.y;
-    /* Prevent the magnifier glass from being positioned outside the image: */
-    if (x > image.width - (w / scale)) {x = image.width - (w / scale);}
-    if (x < w / scale) {x = w / scale;}
-    if (y > image.height - (h / scale)) {y = image.height - (h / scale);}
-    if (y < h / scale) {y = h / scale;}
-    /* Set the position of the magnifier glass: */
-    glass.style.left = (x - w) + "px";
-    glass.style.top = (y - h) + "px";
-    /* Display what the magnifier glass "sees": */
-    glass.style.backgroundPosition = "-" + ((x * scale) - w + bw) + "px -" + ((y * scale) - h + bw) + "px";
-  }
-
-  function getCursorPos(e) {
-    var a, x = 0, y = 0;
-    e = e || window.event;
-    /* Get the x and y positions of the image: */
-    a = image.getBoundingClientRect();
-    /* Calculate the cursor's x and y coordinates, relative to the image: */
-    x = e.pageX - a.left;
-    y = e.pageY - a.top;
-    /* Consider any page scrolling: */
-    x = x - window.pageXOffset;
-    y = y - window.pageYOffset;
-    return {x : x, y : y};
-  }
+  overlay.onclick = () => overlay.remove();
+  overlay.onmousemove = (event) => moveWithHover(image, event, zoomRatio);
 }
 
+const images = document.querySelectorAll('.image--original');
+
 images.forEach(image => {
-  image.addEventListener('click', () => {
-    magnify(image, 2);
-  });
+  image.onclick = () => magnify(image, 3);
 });
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -1142,6 +1142,21 @@ a.product__text {
   border-color: rgb(var(--color-foreground));
 }
 
+.img-magnifier-glass {
+  position: absolute;
+  cursor: zoom-out;
+  z-index: 1;
+  border: 1px solid black;
+}
+
+.image--zoom {
+  cursor: zoom-in;
+}
+
+.product__media-container {
+  overflow: hidden !important;
+}
+
 .thumbnail[aria-current]:focus-visible {
   box-shadow: 0 0 0 0.3rem rgb(var(--color-background)),0 0 0rem 0.5rem rgba(var(--color-foreground), 0.5);
 }
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -1142,14 +1142,12 @@ a.product__text {
   border-color: rgb(var(--color-foreground));
 }
 
-.img-magnifier-glass {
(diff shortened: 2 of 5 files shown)


bacounts:
Could this issue (enhancement) ticket be added to this PR? https://github.com/Shopify/dawn/issues/465

metamoni:
> Could this issue (enhancement) ticket be added to this PR? https://github.com/Shopify/dawn/issues/465

@bacounts This PR introduces quite a few new options, so I'm going to focus on the requirements described in #1704 for now.

assets/magnify.js
```diff
@@ -0,0 +1,43 @@
...
+  const images = document.querySelectorAll('.image--hover');
+  images && images.forEach(image => {
+    image.onclick = () => magnify(image, 2);
+  });
+}
+
+enableZoomOnHover();
+
+// also enable in theme editor
+document.addEventListener('shopify:section:load', () => enableZoomOnHover());
```
metamoni, line 43: Now sure if there's a better way to do this. Without it, the script only loads once in the editor, so when we change the setting from whatever is currently saved to "hover", we can't preview it.

assets/magnify.js
```diff
@@ -0,0 +1,43 @@
...
+  overlay.style.backgroundSize = `${image.width * zoomRatio}px`;
+};
+
+const magnify = (image, zoomRatio) => {
+  // add full-size image on top of original
+  const overlay = createOverlay(image);
+  image.parentElement.insertBefore(overlay, image);
+
+  overlay.onclick = () => overlay.remove();
+  overlay.onmousemove = (event) => moveWithHover(image, event, zoomRatio);
```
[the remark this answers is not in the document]
metamoni: We already have an `onclick` event on it, though, as we need to remove the zoom when the user clicks the image again: `overlay.onclick = () => overlay.remove();`

assets/magnify.js
metamoni: If I add the `shopify:section:load` event listener to `theme-editor.js`, how do I import the `magnify` function? I haven't seen a pattern for working with modules in our codebase.

Or do you mean move all of the js code in `magnify.js` into the `main-product` section with `{% javascript %}`, and using`request.design_mode` in there?

locales/en.default.schema.json
```diff
@@ -1950,6 +1950,19 @@
             "label": "Large"
           }
         },
+        "image_zoom": {
+          "label": "Image zoom",
+          "info": "Zoom on hover defaults to open lightbox on mobile.",
+          "options__1": {
+            "label": "Open lightbox"
```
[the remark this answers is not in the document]
metamoni: Let's get @Oliviammarcello's input on this

locales/en.default.schema.json
```diff
@@ -1950,6 +1950,19 @@
...
+          },
+          "options__2": {
+            "label": "Zoom on hover"
```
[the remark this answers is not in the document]
Oliviammarcello: I don't think we need to be this specific. The goal of the setting is to zoom in on hover, and since the cursor changes on hover it should be okay 🤔

locales/en.default.schema.json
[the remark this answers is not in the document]
ludoboludo: Sweet, thanks for checking!

assets/section-main-product.css
```diff
@@ -1142,6 +1142,38 @@ a.product__text {
   border-color: rgb(var(--color-foreground));
 }
 
+.image--full-size {
+  cursor: zoom-out;
+  z-index: 1;
+  margin: 0;
+}
+
+.image--hover {
```
kmeleta, line 1151: Could we consider naming the classes related to the zoom functionality slightly less generically? 
Something like `.image-magnify` and `.image-magnify__full-size` or `.image-hover-zoom` and `.image-hover-zoom__full-size`. I mention the "magnify" term because matching the target classnames to the js terminology could make the relationship a little clearer.

I also don't think we have a base `.image` class being used so the `--full-size` and `--hover` modifiers don't align exactly with our intended (but probably inconsistent) naming practices.

assets/magnify.js
```diff
@@ -0,0 +1,41 @@
...
+  overlay.onmouseleave = () => overlay.remove();
+}
+
+function enableZoomOnHover() {
+  const images = document.querySelectorAll('.image--hover');
+  images && images.forEach(image => {
+    image.onclick = () => magnify(image, 2);
```
kmeleta, line 37: Very minor, but could we capture the 2 in a constant or perhaps as an argument of `enableZoomOnHover`? Just so it's a little more self documenting? I wonder if there would ever be a use case where this value varies between different instances.

kmeleta (commented):
Leaving a couple notes for now. I need to spend more time testing and considering the js loading aspects, but I know you've been waiting on feedback for a bit.

Oliviammarcello:
This is looking great, I just see one issue:

- [ ] On mobile, the click and hover is still working. It obviously doesn't let me hover, but when I click the image the lightbox doesn't open, the image just zooms in.

assets/section-main-product.css
```diff
@@ -1142,6 +1142,38 @@ a.product__text {
...
+  .lightbox-button--hover {
+    display: none;
+  }
+
+  .product-modal--hover {
+    display: none;
+  }
+}
+
+.product__media-container {
```
ludoboludo, line 1173: I see another styling declaration with this class on line `53` of the same file. I think we can probably remove it as there won't be any `slider-buttons` within this media container. 👍

sections/featured-product.liquid
```diff
@@ -681,6 +684,12 @@
   </script>
   <script src="{{ 'product-model.js' | asset_url }}" defer></script>
 {%- endif -%}
+{% if section.settings.image_zoom == 'hover' %}
+  <script id="enableZoomOnHover-featured" src="{{ 'magnify.js' | asset_url }}" defer="defer"></script>
+{% endif %}
```
ludoboludo, line 689: I also wonder... if I visit the store on mobile only I'm getting a network request for nothing right ? I wonder what's best between: 
- having the JS in `global.js` and have a `matchMedia()` check to run the function based on `hover` as well (`const canHover = window.matchMedia('(hover: hover)').matches; //true or false`)
- keep it as is and have an network request that isn't necessary as well as the code within the file for devices that don't support `hover`.

The difference is probably minimal 🤷

ludoboludo (commented):
It's looking good. I think the main thing is what Ken brought up about the position not being quite accurate when you first click.

sections/featured-product.liquid
metamoni: I previously considered [using js to determine whether we're on a mobile device](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent), but during a pairing session decided to go with CSS at the time. I'm starting to think that approach is causing some inconsistencies, and it might be better to just use javascript for this.

assets/section-main-product.css
kmeleta: I agree that the other instance isn't being used for anything and we should remove it if we're keeping this class in this PR (if not it's not necessarily this pr's responsibility). However I'm wondering why we need this at all. 

Having `product__media` and `product__media-container` on the same element seems confusing. Could we not just target the existing classes instead? My other motivation for not adding this class is I'm working on a PR that's hoping to use a similar "product media container or wrapper" class in the thumbnail snippet.

metamoni: Initially, the `product__media` class doesn't have `overflow: hidden` on it. If I target `product__media` instead of creating a new class for these images alone, then we'd be hiding the overflow for other elements as well, and that might break the current design.

We're currently applying this class in our `noscript` code. If you think it's safe to force `overflow: hidden` on these too, then I can remove the `product__media-container` class. Alternatively, I can rename it to `product__media-image`, as it's only applied to images.

kmeleta: So looking down at your changes on main-product.liquid. We're always applying the `product__media-container`, but we're _not_ including that in the noscript instances of the `product__media` class, right?

Could we target that particular instance in another way without a new class? `.product__modal-opener .product__media` or if my assumption above is correct, maybe even `.js .product__media`

Oliviammarcello (approved):

assets/magnify.js
```diff
@@ -0,0 +1,43 @@
...
+  const images = document.querySelectorAll('.image-magnify__hover');
+  images.forEach(image => {
+    image.onclick = (event) => {
+      magnify(image, zoomRatio);
+      moveWithHover(image, event, zoomRatio);
+    };
+  });
+}
+
+enableZoomOnHover(2);
```
ludoboludo, line 43: That could be an interesting value to tie to a setting in the editor. Can probably be done in the future if we see a lot of requests to change the zoom value.

ludoboludo (commented):
A few minor comments. Otherwise the PR looks great and the functionality is very nice add to our current themes 👌 🎉

assets/magnify.js
```diff
@@ -0,0 +1,43 @@
+// create a container and set the full-size image as its background
+function createOverlay(image) {
+  overlay = document.createElement('div');
+  overlay.setAttribute('class', 'image-magnify__full-size');
+  overlay.setAttribute('aria-hidden', 'true');
+  overlay.style.backgroundImage = `url('${image.src}')`;
```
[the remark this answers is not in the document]
metamoni: Good shout! We can keep that in mind if we see any performance issues pop up 👍

assets/section-main-product.css
```diff
@@ -1148,6 +1148,45 @@ a.product__text {
   border-color: rgb(var(--color-foreground));
 }
 
+.image-magnify-full-size {
+  cursor: zoom-out;
+  z-index: 1;
+  margin: 0;
```
ludoboludo, line 1154: Here we're going to need to add some styling for the border radius I believe. I had not tested before with `global settings` and when [I go here to tweak the appearance](https://screenshot.click/20-59-lvnk9-q0xxm.mp4), it shows that the image is outside of it's containing box due to the border radius not being applied.

So that would need to be added.

ludoboludo (commented):
My last piece of feedback before approval ? Who knows 😅 
Sorry for the constant edit requests.

assets/section-main-product.css
metamoni: Great catch, thanks! I've added that in 👍

assets/section-main-product.css
```diff
@@ -1148,6 +1148,46 @@ a.product__text {
...
+  .product__media-zoom-hover {
+    display: none;
+  }
+
+  .product__media-icon--hover {
+    display: none;
+  }
+}
+
+@media screen and (max-width: 23.4375rem) {
```
ludoboludo, line 1177: Why was this the value of the `max-width` here ?

assets/section-main-product.css
```diff
@@ -1148,6 +1148,46 @@ a.product__text {
...
+  .product__media-zoom-hover {
+    display: flex;
+  }
+
+  .product__media-icon--hover {
+    display: flex;
+  }
```
ludoboludo, line 1184:
```suggestion
  .product__media-zoom-hover,
  .product__media-icon--hover {
    display: flex;
  }
```

assets/magnify.js
```diff
@@ -0,0 +1,43 @@
...
+  const xPercent = `${xPosition / (overlay.clientWidth / 100)}%`;
+  const yPercent = `${yPosition / ((overlay.clientWidth * ratio) / 100)}%`;
+
+  // determine what to show in the frame
+  overlay.style.backgroundPosition = `${xPercent} ${yPercent}`;
+  overlay.style.backgroundSize = `${image.width * zoomRatio}px`;
+};
+
+function magnify(image, zoomRatio) {
+  // add full-size image on top of original
```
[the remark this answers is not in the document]
ludoboludo: I'm not even sure we need the comment to be honest 🤔

assets/section-main-product.css
ludoboludo: Hmm we're usually relying on 2 sizes, `750px` (or `749` when using `max-width`) and `990px` (or `989` when using `max-width`).
I'd default to `px` values for media queries, as `rem` values will be different based on the `font size scale` picked ([screenshot](https://screenshot.click/24-04-itc5a-p9m1z.png)). 
Do you forsee an issue with switching to `749px` instead ?

assets/section-main-product.css
```diff
@@ -1148,6 +1148,40 @@ a.product__text {
...
+.product__media-icon--none {
+  display: none;
+}
+
+@media (hover: hover) {
+  .product__media-zoom-hover,
+  .product__media-icon--hover {
+    display: none;
+  }
+}
```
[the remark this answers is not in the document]
kmeleta: I wonder if giving the hover case its own element to target and hide as needed would give us more flexibility here 🤔

ludoboludo (approved):
Looking good to me. I could not replicate Ken's issue but it might be worth looking into.

assets/section-main-product.css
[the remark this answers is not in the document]
metamoni: I couldn't replicate this either

kmeleta (approved):
Tested about every scenario I could think of and haven't managed to break it. I think we're good to go. Thanks Monica! :shipit:

assets/magnify.js
```diff
@@ -0,0 +1,42 @@
+// create a container and set the full-size image as its background
+function createOverlay(image) {
+  overlay = document.createElement('div');
+  overlay.setAttribute('class', 'image-magnify-full-size');
+  overlay.setAttribute('aria-hidden', 'true');
+  overlay.style.backgroundImage = `url('${image.src}')`;
```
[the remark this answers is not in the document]
metamoni: I tried this with an `img` element instead of a `div`, but the overlay image remains constrained in its container instead of being loaded at full size. Not sure what causes it or how to test whether it would load faster.

https://user-images.githubusercontent.com/22390758/197939841-0b16d25a-b2ce-474f-bf26-fd1a5ead2b9b.mp4

Issue #1748, opened 2022-05-27
[SPIKE] Allow pulling variant level dynamic sources on the product page


nicklepine:
Currently, we can't pull dynamic sources at the variant level (like SKU, or variant-level metafields) on the product page. 

As a buyer, as a select different variants, I should be able to see variant-specific information displayed on the product page. 

----
**More context:**
https://shopify.slack.com/archives/C03STBS00DP/p1662485121404529
https://shopify.slack.com/archives/C03STBS00DP/p1662494988426059

PR #1987, opened 2022-09-26
[PDP] Add SKU block

### LINKED ISSUE #1986: [Product] Add SKU block to product page
Add SKU block to Product Page

[Figma](https://www.figma.com/file/mYV4VA1cld0UYcAIMe2Y8N/Product-Page?node-id=202%3A15976)

eugenekasimov:
### PR Summary: 

This feature allows merchants to add SKU block to Product Page and choose text style for it.

Fixes #1986.

- Add a new block for SKU
- 

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](url)

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -849,10 +849,20 @@ class VariantSelects extends HTMLElement {
         if (source && destination) destination.innerHTML = source.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
+        const sku = document.getElementById(`sku-${this.dataset.section}`);
 
         if (price) price.classList.remove('visibility-hidden');
+        if (sku) this.updateSKU(html);
         this.toggleAddButton(!this.currentVariant.available, window.variantStrings.soldOut);
       });
+  }
+
+  updateSKU(html) {
+    const id = `sku-${this.dataset.section}`;
+    const destination = document.getElementById(id);
+    const source = html.getElementById(id);
+
+    if (source && destination) destination.innerHTML = source.innerHTML;
   }
 
   toggleAddButton(disable = true, text, modifyClass = true) {
@@ -878,6 +888,7 @@ class VariantSelects extends HTMLElement {
     const addButton = button.querySelector('[name="add"]');
     const addButtonText = button.querySelector('[name="add"] > span');
     const price = document.getElementById(`price-${this.dataset.section}`);
+
     if (!addButton) return;
     addButtonText.textContent = window.variantStrings.unavailable;
     if (price) price.classList.add('visibility-hidden');
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -400,6 +400,7 @@ a.product__text {
 .product--no-media .product__text,
 .product--no-media noscript .product-form__input,
 .product--no-media .product__tax,
+.product--no-media .product__sku,
 .product--no-media shopify-payment-terms {
   text-align: center;
 }
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -362,11 +362,11 @@
             {%- when 'sku' -%}
               {%- assign current_variant = product.selected_or_first_available_variant -%}
               <p
-                class="product__text--sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}"
+                class="product__sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}"
                 id="sku-{{ section.id }}"
                 {{ block.shopify_attributes }}
               >
-                {{ current_variant.sku }}
+                {{- current_variant.sku -}}
               </p>
             {%- when 'custom_liquid' -%}
               {{ block.settings.custom_liquid }}
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -359,6 +359,15 @@
                   {{ product.description }}
                 </div>
               {%- endif -%}
+            {%- when 'sku' -%}
+              {%- assign current_variant = product.selected_or_first_available_variant -%}
+              <p
+                class="product__text--sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}"
+                id="sku-{{ section.id }}"
+                {{ block.shopify_attributes }}
+              >
+                {{ current_variant.sku }}
+              </p>
             {%- when 'custom_liquid' -%}
               {{ block.settings.custom_liquid }}
             {%- when 'collapsible_tab' -%}
@@ -1261,6 +1270,33 @@
       "type": "description",
       "name": "t:sections.main-product.blocks.description.name",
       "limit": 1
+    },
+    {
+      "type": "sku",
+      "name": "t:sections.main-product.blocks.sku.name",
+      "limit": 1,
+      "settings": [
+        {
+          "type": "select",
+          "id": "text_style",
+          "options": [
+            {
+              "value": "body",
+              "label": "t:sections.main-product.blocks.sku.settings.text_style.options__1.label"
+            },
+            {
+              "value": "subtitle",
+              "label": "t:sections.main-product.blocks.sku.settings.text_style.options__2.label"
+            },
+            {
+              "value": "uppercase",
+              "label": "t:sections.main-product.blocks.sku.settings.text_style.options__3.label"
+            }
+          ],
+          "default": "body",
+          "label": "t:sections.main-product.blocks.sku.settings.text_style.label"
+        }
+      ]
     },
     {
       "type": "share",
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1677,6 +1677,23 @@
         "description": {
           "name": "Description"
         },
+        "sku": {
+          "name": "SKU",
+          "settings": {
+            "text_style": {
+              "label": "Text style",
+              "options__1": {
+                "label": "Body"
+              },
+              "options__2": {
+                "label": "Subtitle"
+              },
+              "options__3": {
+                "label": "Uppercase"
+              }
+            }
+          }
+        },
         "share": {
           "name": "Share",
           "settings": {


bacounts:
@eugenekasimov — This may need to be a bigger effort. Please refer to issue: https://github.com/Shopify/dawn/issues/1748 originally raised by @nicklepine.

assets/global.js
```diff
@@ -846,15 +846,31 @@ class VariantSelects extends HTMLElement {
...
+  updateSKU(html) {
+    const id = `sku-${this.dataset.section}`;
+    const destination = document.getElementById(id);
+    const source = html.getElementById(id);
+
+    if (source && destination) destination.innerHTML = source.innerHTML;
+    if (destination.innerHTML !== '#') {
+      destination.classList.remove('visually-hidden');
+    } else {
+      destination.classList.add('visually-hidden');
```
metamoni, line 870: Adding this class will hide it from sighted users but keep it in the DOM, so this will be read out by screen readers. To completely hide it, you need to either remove the paragraph from the DOM or give it an additional `aria-hidden="true"` attribute.

https://user-images.githubusercontent.com/22390758/192733908-d0a1dcbd-f47b-41e0-9fed-b6d07742a055.mov

sections/main-product.liquid
```diff
@@ -359,6 +359,15 @@
                   {{ product.description }}
                 </div>
               {%- endif -%}
+            {%- when 'sku' -%}
```
metamoni, line 362: Since we're surfacing new information when a variant is selected, we should probably announce that to screen readers too. We do this with the product's price, which has a `role="status"`. We can use a similar approach here.

assets/global.js
eugenekasimov: @metamoni thanks for this opinion. I had spent a while thinking about it. I'm open to discuss and investigate it deeper. The reason why I decided not to remove it completely from DOM is to let screen readers know that there is supposed to be SKU even if it's empty now. I still assume that let's say first variant might not have SKU but the second one have it and I'd like to let users know about it from a first stem without switching to the next variant.

sections/main-product.liquid
```diff
@@ -359,6 +359,15 @@
...
+              {%- assign current_variant = product.selected_or_first_available_variant -%}
+              <p
+                class="product__sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if current_variant.sku.size == 0 %} visually-hidden{% endif %}"
+                id="sku-{{ section.id }}"
+                {{ block.shopify_attributes }}
+              >
+                {{- '#' | append: current_variant.sku -}}
```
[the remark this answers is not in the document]
eugenekasimov: That's a good point. I asked about it @YoannJailin and he told me that he would have it permanently for default that's why I added it this way. But maybe we have to reconfirm it with @YoannJailin.

I didn't get your point about translation, would you pls clarify what you meant? Thanks.

assets/global.js
metamoni: There should be a way to remove/hide it when it's empty, and add/show it when we have a SKU. Having an empty `#` is irrelevant to both sighted and screen reader users, particularly as we're not clearly saying that this is a the SKU - it's just either a number or a symbol.

bacounts:
@eugenekasimov - Could we surface both SKU and barcode in this block? While barcode is less used, there are use cases for some customers especially B2B.

You could use checkbox settings and allow the block to be used more than once in case someone wants to surface the SKU apart from the barcode (less likely, but will allow for the user to pick placement of each).

YoannJailin (approved):

sections/main-product.liquid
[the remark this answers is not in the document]
eugenekasimov: I talked to Yoann and he confirms that we can remove "#". Thank you Monica for a really good point on this.

eugenekasimov:
Hey @andrewetchen, thanks for spending time on this and for a good advice on tests. I'll add some test steps soon.

>Previous variant SKU still remains

This was a nice catch, I definitely missed it. I changed my approach in JS that allowed me to better reuse existing functions. It also allowed me to fix the issue you mentioned. Thanks again. 

>Unavailable / non-SKU variant whitespace

As you right noticed, in case when variant is unavailable we have whitespace dedicated for price block. In my opinion, this is not related to current PR and I think it shouldn't be a blocker now. However, I have already brought this question up to UX folks because this bothers me too. As far as I know, originally it was built that way to avoid content shifting. But since we add new blocks that potentially might be empty in the same cases, we will have content shifting anyway. Otherwise we will have to keep too much whitespace. We can always delete this whitespace in a different PR if designers confirm it. @danielvan.

YoannJailin:
@andrewetchen @eugenekasimov for the whitespace issue, i'd say it's better to reduce it when a block is empty. We can fix that on the price block later.

eugenekasimov:
In my latest commit I made minor changes:
1) Capitalized ID to be consistent with most of the blocks
2) Instead of removing the block's HTML element I kept the whitespace to be consistent with the Price block (we will care about it later in a different PR when the final decision is made)

metamoni (approved):

assets/section-main-product.css
```diff
@@ -1322,6 +1323,16 @@ a.product__text {
...
 
+/* SKU block */
+
+p.visibility-hidden:empty {
+  display: block;
+}
+
+p.visibility-hidden:empty::after {
+  content: "#";
+}
```
eugenekasimov, line 1334: I think these better to be changed to something more specific for SKU.
```suggestion
product__sku.visibility-hidden:empty {
  display: block;
}

product__sku.visibility-hidden:empty::after {
  content: "#";
}
```

assets/global.js
```diff
@@ -889,9 +896,12 @@ class VariantSelects extends HTMLElement {
     const addButton = button.querySelector('[name="add"]');
     const addButtonText = button.querySelector('[name="add"] > span');
     const price = document.getElementById(`price-${this.dataset.section}`);
+    const sku = document.getElementById(`Sku-${this.dataset.section}`);
+
     if (!addButton) return;
     addButtonText.textContent = window.variantStrings.unavailable;
     if (price) price.classList.add('visibility-hidden');
+    if (sku) sku.classList.add('visibility-hidden'), sku.setAttribute('aria-hidden', 'true');
```
LucasLacerdaUX, line 904:
```suggestion
    if (sku) sku.classList.add('visibility-hidden'), sku.setAttribute('aria-hidden', 'true');
```
You should only need the `visibility: hidden` here.

assets/global.js
```diff
@@ -851,15 +851,22 @@ class VariantSelects extends HTMLElement {
...
         if (source && destination) destination.innerHTML = source.innerHTML;
         if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
+        if (skuSource && skuDestination) skuDestination.innerHTML = skuSource.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
+        const sku = document.getElementById(`Sku-${this.dataset.section}`);
 
         if (price) price.classList.remove('visibility-hidden');
+        if (sku && skuSource.innerHTML.length === 0) sku.classList.add('visibility-hidden'), sku.setAttribute('aria-hidden', 'true');
+        if (sku && skuSource.innerHTML.length !== 0) sku.classList.remove('visibility-hidden'), sku.setAttribute('aria-hidden', 'false');
```
LucasLacerdaUX, line 868: You should only need to use the `visibility-hidden` class here. Then, you can use `toggle` to combine the statements :D
```suggestion
        if (sku && skuSource) sku.classList.toggle('visibility-hidden', skuSource.innerHTML.length)
```

assets/global.js
```diff
@@ -851,15 +851,22 @@ class VariantSelects extends HTMLElement {
...
+        const skuDestination = document.getElementById(`Sku-${this.dataset.section}`);
         const variantPickerDestination = document.getElementById(`variant-radios-${this.dataset.section}`) || document.getElementById(`variant-selects-${this.dataset.section}`);
         const variantPickerSource = html.getElementById(`variant-radios-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`) || html.getElementById(`variant-selects-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`);
 
         if (source && destination) destination.innerHTML = source.innerHTML;
         if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
+        if (skuSource && skuDestination) skuDestination.innerHTML = skuSource.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
+        const sku = document.getElementById(`Sku-${this.dataset.section}`);
```
LucasLacerdaUX, line 864: Is this variable different from `skuDestination`?

sections/main-product.liquid
```diff
@@ -112,6 +112,16 @@
...
               {%- endif -%}
+            {%- when 'sku' -%}
+              {%- assign current_variant = product.selected_or_first_available_variant -%}
+              <p
+                class="product__sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if current_variant.sku.size == 0 %} visibility-hidden{% endif %}"
+                id="Sku-{{ section.id }}"
+                role="status"
+                {{ block.shopify_attributes }}
+              >
+                {{- current_variant.sku -}}
```
LucasLacerdaUX, line 123: I think we should have a visually hidden `SKU: ` preceding the code or assign an `aria-label="SKU: {{ current_variant.sku }}"` to this element.

Usually, a SKU is an eight-digits code and simply announcing that set of alphanumeric characters without any context doesn't seem like the right way to do it.

sections/main-product.liquid
```diff
@@ -112,6 +112,16 @@
                   {{ product.description }}
                 </div>
               {%- endif -%}
+            {%- when 'sku' -%}
+              {%- assign current_variant = product.selected_or_first_available_variant -%}
+              <p
+                class="product__sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if current_variant.sku.size == 0 %} visibility-hidden{% endif %}"
+                id="Sku-{{ section.id }}"
+                role="status"
```
LucasLacerdaUX, line 120: I wonder if we really need to announce this at all. The price is critical for the buyer to make a decision to add it to cart or not, but I'm not sure if the SKU is as critical information to be announced every time the user switches the variant.

I say this as someone with little to no context on how that code is used outside of an inventory control context, though. Can you folks please share some thoughts on this decision and help me understand it a bit more? 👀

LucasLacerdaUX (changes requested):
Looking good! I like the approach you took and it works as expected :D

I have some suggestions re: a11y and some cleanup opportunities on the code. No big changes requested.

assets/section-main-product.css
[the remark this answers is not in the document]
LucasLacerdaUX: You updated this already, so I'll resolve it :P

sections/main-product.liquid
eugenekasimov: I was thinking about that too, otherwise it may not be clear what exactly that number/text is.

sections/main-product.liquid
LucasLacerdaUX: @YoannJailin I was actually referring to announcing it to screen readers via the `role="status"` attribute every time the variant is changed. Is this what you mean when you said "dislpay"?

As much as we want assistive technologies users to get the same information, announcing too many things or things that are not as relevant to their experience can lead to a distracting UX for these users. That's why I was wondering if we really need to announce an 8-digit code when we update the variant.

LucasLacerdaUX (approved):
Just some minor improvements and we should be good! :D

EDIT: I've accidentally approved 🤦‍♂️

assets/section-main-product.css
```diff
@@ -1322,6 +1323,16 @@ a.product__text {
   margin-bottom: var(--icon-size);
 }
 
+/* SKU block */
+
+.product__sku.visibility-hidden {
+  display: block;
+}
+
```
LucasLacerdaUX, line 1331:
```suggestion
```

Do we need this? I think `p` elements have `display: block` by default, but maybe there's a use case I haven't considered.

LucasLacerdaUX (commented):
👁️👄👁️

sections/featured-product.liquid
```diff
@@ -155,6 +155,16 @@
...
                   </div>
                 {%- endif -%}
+              {%- when 'sku' -%}
+                <p
+                  class="product__sku no-js-hidden{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if product.selected_or_first_available_variant.sku.size == 0 %} visibility-hidden{% endif %}"
+                  id="Sku-{{ section.id }}"
+                  role="status"
+                  {{ block.shopify_attributes }}
+                >
+                  <span class="visually-hidden">SKU:</span>
```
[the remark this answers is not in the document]
eugenekasimov: Good call Lucas. Thanks.

assets/global.js
```diff
@@ -851,15 +851,20 @@ class VariantSelects extends HTMLElement {
...
         const variantPickerSource = html.getElementById(`variant-radios-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`) || html.getElementById(`variant-selects-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`);
 
         if (source && destination) destination.innerHTML = source.innerHTML;
         if (variantPickerSource && variantPickerDestination) variantPickerDestination.innerHTML = variantPickerSource.innerHTML;
+        if (skuSource && skuDestination) skuDestination.innerHTML = skuSource.innerHTML;
 
         const price = document.getElementById(`price-${this.dataset.section}`);
 
         if (price) price.classList.remove('visibility-hidden');
+        if (skuDestination && skuSource) skuDestination.classList.toggle('visibility-hidden', skuSource.textContent.trim() === skuSource.querySelector('span').textContent.trim());
```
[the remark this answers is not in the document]
eugenekasimov: Seems like you are right. Nice to have a fresh eye on it :) Thanks.

assets/section-main-product.css
eugenekasimov: Actually you are right. I think we are good to remove it.

sections/main-product.liquid
metamoni: @eugenekasimov as previously discussed, I'm not sure how important the SKU is either. If it's important enough to display to sighted users, why would it be irrelevant to screen reader users? It's certainly not as important as the price, I agree with that. And we shouldn't announce too many things because that could get annoying. But, ideally, when the content on a page changes, those changes should be reflected to assistive tech users too.

I agree it shouldn't be a blocker. It's up to @YoannJailin and you to decide whether you want to leave it in.

eugenekasimov: Okay, let me make this decision. For now let's keep it as is. We can discuss it and change later as a follow up.

metamoni (approved):

LucasLacerdaUX (commented):
Just a minor thing before I approve it :D

Everything else looks perfect.

LucasLacerdaUX (dismissed):
Looks good to me! Thanks for making all the changes :DD

I'll approve it again once translations are done.

metamoni (approved):

PR #1988, opened 2022-09-27
Enable language/country selector by default

slucaskim:
### PR Summary: 

We recently launched [Translate & Adapt](https://apps.shopify.com/translate-and-adapt) as a 1P app, and have started receiving lots of merchant questions and feedback on how to enable the language selector on their theme. If they had a Dawn variant theme, in most cases it was a matter of showing them the setting in the theme and tell them to check the box. Since these selectors wouldn't show up anyways if the merchant doesn't have additional languages or sells to only one country, defaulting these settings to TRUE is not harmful. It will be beneficial for net-new merchants who have an international strategy to see the language/country selector show up naturally.

In the theme editor, these checkboxes would be enabled by default on a fresh Dawn install:
image

Which means that if the shop sells to more than one country, or they have more than one language, they'll see the following dropdowns at the footer of the online store:
image

### Checklist


lgunn4 (approved):

slucaskim:
If you could just merge this in once you deem it ready, that would be great. Thank you!

melissaperreault:
> Since these selectors wouldn't show up anyways if the merchant doesn't have additional languages or sells to only one country, defaulting these settings to TRUE is not harmful


@slucaskim In the case for stores that only sell in one currency but ship worldwide, is there a benefit to having all of the choices display in the dropdown list? Unless the merchant actually made changes to the prices in different regions, we're introducing a long list in the dropdown that seemingly doesn't do anything.  Video reference: https://screenshot.click/04-54-6mkv1-zklwh.mp4 

I understand merchants might be confused about how to enable all these location within stores, but can the inverse of having merchants confused as to why all these different locations are displayed by default be true too?

cc. @robertdsharpe

danielvan:
I agree with @melissaperreault. For languages, as I can see, you only see the languages you've added, but for currencies/region you see all countries in one go unless you pro-actively reduce the list. 

@melissaperreault what is your suggestion here? Keep both on for now and tackle your concern about the list of the countries later?

The only thing I can see is that the list may help a buyer understand if that shop ships to their country (even if they're paying in a foreign currency).

PR #2052, opened 2022-10-19
Add constrain media and media fit settings to product page

### LINKED ISSUE #2004: [Media size setting] Product information template
**PR SUMMARY**
**Problem:** Product images expanding below the viewport, resulting in undesirable product image representation for our merchants.
**Solution:** The problem will be solved with two new settings. We will be offering a viewport height checkbox: "Constrain media to screen size" and "media fit" setting. Media fit setting issue: [2016](https://github.com/Shopify/dawn/issues/2016)

**UPDATES**
- [ ] **New checkbox:** Add a new checkbox under the media section of product information template. Checkbox: Constrain media to screen size
- [ ] **Update to Desktop media size setting:** Change "desktop media size" label to "desktop media width" and explicitly display % values.
- [ ] **Default settings:** Constrain media to screen size toggle: off, Desktop media width: Large (65%) – Don't adapt to image.
Screen Shot 2022-10-27 at 11 45 41 AM

- [ ] **Border logic update:** The borders will be around the images instead of containers for the product images. Won't impact the thumbnails. 
border logic

- [ ] **Spacing between mobile images:** Should work the same as the current implementation, always have a margin to avoid images clashing to each other if the merchant interacts with the global layout/grid/horizontal settings.
- [ ] **Image zoom:** The new media settings should work well with the upcoming image zoom settings. Issue: [1983](https://github.com/Shopify/dawn/pull/1983) Zoom icon should always positioned on the image. If the image is cropped or scaled down as an outcome of new media settings, the zoom effect should consider the full image, not part of the image. 
- [ ] **Stop scaling down the mobile images at 600px height:** When the new checkbox is on, the minimum height for the images to stop scaling is at 600px to avoid increased spacing between the images and images becoming very small. 

**CONSIDERATIONS FOR REVIEWERS**

- **Testing ground:** Desktop: 1360x768px, Mobile: 390x844px
- **Vertical padding values between image x content block:** No change. 
- **Design requirements:** 
**Desktop:** When checkbox is on, if the desktop layout is set to thumbnails/thumbnail carousel – show part of the thumbnails within the viewport. For stacked layout: show part of the images on the 2nd row. 
**Mobile:** When the checkbox is on if the mobile layout is set to thumbnails – show part of the thumbnails within the viewport. If the thumbnails are turned off, show part of the product information (product title and price)
desk-mob northstars

**DESIGN DOCUMENTS:**
- [Figma file](https://www.figma.com/file/FXgF4YTsvGmHEY6hD0KuTd/Image-sizing?node-id=2343%3A22035)
- [Design testing considerations](https://docs.google.com/document/d/1k69x7rn_Be8GJNTmttxhcsJzM2UIWrhzUTwpBoxXJlg/edit#)

### LINKED ISSUE #2016: [Media size setting] Media fit
**PR SUMMARY**
**Problem:** Product images expanding below the viewport, resulting in undesirable product image representation for our merchants.
**Solution:** We will be adding "media fit" setting as an addition to the new media settings for product information template. [Issue: 2004](https://github.com/Shopify/dawn/issues/2004)

**UPDATES**
- [ ] **New setting:** Add "Media fit" setting under media section with two options for the merchant to pick from: Original, fill.
- [ ] **Default setting:** Media fit: Original
Screen Shot 2022-10-27 at 11 47 36 AM

- [ ] **Border logic per media fit setting:** When the media is set to “fill”, the image should cover the full image column width. Container size is determined by the column. 
media fit borders

- [ ] **Video thumbnails:** There are two ways that the video thumbnails can be pulled. a) if the merchant uploads an image from the admin. If not, the thumbnail will be pulled from the original source. (eg. Youtube) In both cases, if the merchant chooses "fill" from the media fit setting, the video thumbnail image should cover the full image column width. (this will be not-interacted state, once the buyer interacts with the video (clicks play) – the video goes back to its original aspect ratio)
media fit for videos

- [ ] **3D models:** Similar to the video thumbnails, the 3D model thumbnails can be uploaded by the merchant or if there is not thumbnail, the thumbnail is automatically uploaded in square aspect ratio. If the merchant uploads a video thumbnail in a portrait aspect ratio, the logic should work the same for media fit: fill – the image should cover the full image column width. (this will be not-interacted state, once the buyer interacts with the 3D model – the model goes back to square aspect ratio)
3d models

- [ ] **Stacked layout x media setting: fill:** Always scale up shorter images to fill the space created by the tallest image in the same row. On mobile this would mean the tallest image in the whole set.

**DESIGN DOCUMENTS:**

- [Figma file](https://www.figma.com/file/FXgF4YTsvGmHEY6hD0KuTd/Image-sizing?node-id=2343%3A22035)
- [Design testing considerations](https://docs.google.com/document/d/1k69x7rn_Be8GJNTmttxhcsJzM2UIWrhzUTwpBoxXJlg/edit#)

kmeleta:
### PR Summary: 

Added two new settings to improve flexibility around sizing media on the product page.

Fixes #2004
Fixes #2016 

**Adds a `Constrain media to screen size` setting**
Taller product images often display too large on the product page to view the whole image without scrolling. This is exacerbated when using the "Large" desktop media size setting. This new boolean setting effectively sets a max media height and scales down or "constrains" the media based on the height of the viewport.

 

**Adds a `Media fit` setting**
When images are constrained or a product has a variety of aspect ratios, whitespace is created in certain layouts. This new setting gives the option to `Fill` whatever space is created by layout. Think of fill as `object-fit: cover` and the default/original setting as more of `object-fit: contain`.

 

<details>
<summary>Determining the constrained height</summary>
The main priority is to constrain the image enough to where it can entirely fit in the viewport. This is not <i>necessarily</i> the same as fitting within the fold, as there are too many variables that would need to be dynamically factored in to <i>always</i> ensure this is the case (padding and margin, logo size, announcement bar, etc). I'm using a simple a `100vh` minus an arbitrary amount of offset. This offset varies for mobile and desktop breakpoints and was chosen to meet the design needs of this feature in a majority of theme configurations.
</details>

<details>
<summary>Determining the constrained width</summary>
In order to play nice with the way our product media items are built (and prevent cumulative layout shift), I decided to explicitly calculate the proportional scaled width as a percent using the image's aspect ratio. This width is applied to the `.product media-container` element, which is now the parent element of everything in `product-thumbnail.liquid` (media, deferred media, modal opener, badges, etc). 
<br><br>
Any global media styles (like border and shadow) are now also applied to the same element so these styles stay tied to the proper scaled image dimensions. For context, the work I did in #2022 was to better facilitate this approach.
</details>

<details>
<summary>Applying the settings via classes</summary>
The constrain height setting defaults to off, but when true, is applied via a `constrain-height` class in the `product-thumbnail` snippet. Along with that there is `media-fit-contain` or `media-fit-cover` class to support the value of the media fit setting. I'm basing these names/terms on the object-fit CSS property rather than the end user facing language shown in the editor. "contain" represents the "original" media fit behavior and "cover" represents the "fill" behavior.
</details>

<details>
<summary>Using css variables to modify ratio context</summary>
I made some modifications to the way aspect ratios are passed to the CSS in order to account for the differences in mobile and desktop deferred-media behavior and the effect of the media fit options. 
<br><br>
You'll notice <i>all</i> ratio-related variable context is now defined on the root product-media-container element so that changes in those values are easy to track and make sense of. I did need to add a `media-type-{{media.type}}` class on this element to facilitate this and avoid making `modal-opener--{{media-type}}` responsible for modifying variable context in one case, but I see that as more logical place for that type of modifier class anyway.
</details>

<details>
<summary>Applying the cover/fill media fit</summary>
CSS does have the object-fit property which is applied to most media images by default. So most of the work here is around ensuring the height or width of the media containers are actually taking 100% of their allocated space. The approach to this varies a bit by case.
</details>

### Other considerations

**General behavior**
If images cannot fit vertically in the viewport, they will be scaled down and centered in the available column space.
https://screenshot.click/19-13-3re96-lqsv7.mp4

**Effect on horizontal spacing**
The layout spacing between images on the same row will appear to be disregarded if images the images need to scale. This is most prominent on mobile where images are always on the same row
https://screenshot.click/19-06-0vcj3-8ln4p.mp4

**Portrait image considerations**
In some layouts where images are in a 2 column configuration, constraining of these images is still a possibility if the images are tall enough, even with a maximum constrain value in place.
https://screenshot.click/19-15-cbr8t-erodp.mp4

On the tablet breakpoint, it is also likely to see multiple [portrait] images being constrained
https://screenshot.click/19-20-1z06k-8xvsw.mp4

**Effect on deferred media**
Videos/3d models on PDP will also be affected by the constrain setting (most commonly on mobile) and the media fit setting. It's worth noting that _on desktop_, the aspect ratio used for the media item is that of the actual video or model source. Not the preview/poster image. If a poster of a different aspect ratio is used, there will be white space around the image. Using media fit: fill, provides an option to mitigate that, but ideally merchants provide a properly sized poster image. We can always change this behavior, but that would be another discussion.
https://screenshot.click/26-03-ozqxw-e27m0.mp4

Constrain media: Product images could scale with new constrain setting on. No visual impact with the setting off. No visual impact for images that already fit within the viewport. Portrait images will be the most effected.

Media fit fill: Images that are scaled by the constrain setting and images [horizontally] adjacent to taller images will be scaled up to fill the whitespace left and those images may be partially cropped. With the constrain setting off or inapplicable, the media fit setting should have no noticeable effect on products using images with the same aspect ratio.

### Testing steps/scenarios
- Test a variety of product media, single media, video, 3d models, portrait, square, landscape
- Test with multiple desktop media size and layout options (stacked, carousel, thumbnails)
- Test mobile, tablet, desktop
- Test combinations of constrain and media fit
- Test with and without media borders and shadows
- Test a variety of viewport heights
- Test with different page width settings up to 1600px
- Test with different page layout setting values
- Test with image zoom

### Demo links
- [Editor](https://os2-demo.myshopify.com/admin/themes/131728900118/editor)

### Checklist


assets/section-main-product.css
```diff
@@ -770,8 +770,8 @@ a.product__text {
 }
```
duygukalaycioglu, line 771: Match default settings to Dawn’s current default, currently the toggle is on and the mob thumbs are set to show thumbs. 
VH toggle: Off, 
Desk width: Large, 
Desk layout: Stacked 
Mob layout: Hide thumbs

duygukalaycioglu (changes requested):

assets/section-main-product.css
kmeleta: The defaults should be that way. The demo shop was probably just saved with different settings. I've reverted them for now but they could be resaved again at some point.

duygukalaycioglu:
The constrain media setting is looking good! 🎉 
I'll wait for the media fit setting and image zoom to be added to continue the design review.

assets/section-main-product.css
```diff
@@ -1325,14 +1325,46 @@ a.product__text {
...
     max-width: 100%;
   }
+
+  .product-media-container.constrain-height {
+    --viewport-offset: 170px;
+    --contrained-min-height: 500px;
+  }
+}
+
+.product-media-container.constrain-height.media-fit-contain {
```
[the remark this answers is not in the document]
kmeleta: I intend for `media-fit-contain/cover` and `constrain-height` to become the common naming conventions for any section to offer either of these features, so cutting out `.product-media-container` would probably be shortsighted even if it's not required presently.

I know you didn't have the full picture with the media fit setting when you first looked, but for context, these other classes and the specific combinations of them are deliberate and shouldn't be removed for specificity. I did think about using `:where()` for maintaining minimal specificity in these cases, but decided against it because browser support is still relatively recent, just barely being within our requirements.

kmeleta:
> I gotta say I'm not a huge fan of the side effect creating more "gap" between images, specifically on mobile 

It's just an option. Like any of our settings, it might work for some shops better than others. For some, leaving constrain off and instead using shorter images or the small/medium media width options may be the better approach. Not to say we shouldn't look for opportunities to tweak it, but the general approach and its caveats have been considered and discussed extensively. @ludoboludo

duygukalaycioglu (dismissed):
Looking good! :tada:

Tested:
- [x] Desktop, tablet, mobile views
- [x] New setting labels and order
- [x] Desktop and mobile layouts (stacked, thumbs, thumb carousel, hide thumbs)
- [x] Breakpoints
- [x] Minimum image heights for desktop and mobile to stop scaling down
- [x] Combination of new settings: media fit, constrain media toggle with desktop media width options
- [x] Default settings for Dawn
- [x] All media types with multiple medias (3D images, videos)
- [x] Different aspect ratios (square, portrait, landscape)
- [x] Borders and global settings 
- [x] Media fit options on desktop and mobile
- [x] Image zoom

sections/main-product.liquid
```diff
@@ -2074,20 +2074,46 @@
     },
     {
       "type": "select",
-      "id": "media_position",
```
ludoboludo, line 2077: Media position being a bit of more general setting for that column of the section, wouldn't we want to keep it at the top ? 

I don't have strong feelings just sharing some thoughts as I go through the review 😅

snippets/product-thumbnail.liquid
```diff
@@ -51,8 +49,8 @@
 {%- endcapture -%}
 
 <div
-  class="product-media-container gradient global-media-settings"
-  style="--ratio-percent: {{ preview_media_ratio }}%;"
+  class="product-media-container media-type-{{ media.media_type }} media-fit-{{ media_fit }}{% if constrain_to_viewport %} constrain-height{% endif %} gradient global-media-settings"
+  style="--ratio: {{ media.aspect_ratio | default: 1.0 }}; --preview-ratio: {{ media.preview_image.aspect_ratio | default: 1.0 }}"
```
ludoboludo, line 53:
```suggestion
  style="--ratio: {{ media.aspect_ratio | default: 1.0 }}; --preview-ratio: {{ media.preview_image.aspect_ratio | default: 1.0 }};"
```

snippets/product-media-gallery.liquid
```diff
@@ -72,7 +72,20 @@
           data-media-id="{{ section.id }}-{{ featured_media.id }}"
         >
           {%- assign media_position = 1 -%}
-          {% render 'product-thumbnail', media: featured_media, media_count: media_count, position: media_position, desktop_layout: section.settings.gallery_layout, mobile_layout: section.settings.mobile_thumbnails, loop: section.settings.enable_video_looping, modal_id: section.id, xr_button: true, media_width: media_width, lazy_load: false %}
+          {% render 'product-thumbnail',
+            media: featured_media,
+            media_count: media_count,
+            position: media_position,
+            desktop_layout: section.settings.gallery_layout,
```
ludoboludo, line 79: When both `constrain media to screen size` is `true` and `media fit` set to `fill`, the main first image when the layout is anything by two columns can get quite zoomed in. 
<details>
<summary>screenshot</summary>


</details>

But I guess they more than enough options to work around it.

snippets/product-media-gallery.liquid
```diff
@@ -72,7 +72,20 @@
...
+            mobile_layout: section.settings.mobile_thumbnails,
+            loop: section.settings.enable_video_looping,
+            modal_id: section.id,
+            xr_button: true,
+            media_width: media_width,
+            media_fit: section.settings.media_fit,
+            constrain_to_viewport: section.settings.constrain_to_viewport,
```
ludoboludo, line 86: The misalignment can a look a little funky but I don't think there is really anything to do about it 🤷

assets/section-main-product.css
```diff
@@ -1366,14 +1366,79 @@ a.product__text {
...
+  }
+
+  .product-media-container.constrain-height {
+    --viewport-offset: 170px;
+    --contrained-min-height: 500px;
+  }
+
+  .product-media-container.media-fit-cover,
+  .product-media-container.media-fit-cover .product__modal-opener,
+  .product-media-container.media-fit-cover .media {
```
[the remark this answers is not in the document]
kmeleta: I've felt the same way about :is that I do for :where (because they're the same except for the compounding specificity), I don't think the support is _that_ good to use it often (just within latest 2 versions).

However, we could give some thought to using these for styles that wouldn't cause _catastrophic_ layout issues in older browsers. In any case though, to save a line or two I'm not sure it's worth ruining the feature for some users. I think in cases where we'd be avoiding a significant number of repetitive selectors we'd have more reason to use these.

sections/main-product.liquid
kmeleta: The short answer is I don't disagree with you and we've already made some adjustments based on feedback I had on the natural order these, but to be consistent across all sections we want to keep media related settings together and generally above other categories of settings, like layout (Image with text is a good example with similar settings).

"Media position" despite having the word "media" in it, is more of a general layout setting, in that the language could just as easily be inversed to read "Product info position". It doesn't explicitly effect the display of the image itself.

locales/en.default.schema.json
```diff
@@ -2055,6 +2058,15 @@
             "label": "Right"
           }
         },
+        "media_fit": {
+          "label": "Media fit",
+          "options__1": {
+            "label": "Original"
```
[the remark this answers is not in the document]
kmeleta: Not sure I see a conflict here, personally, and the language has been Katy-approved, but worth calling out to make sure we've considered that. I'd probably frame the question as, if we _were_ ever to add an aspect ratio setting to PDP, and "original" was the default option, do we have a problem with that, given the default for media fit? @duygukalaycioglu

assets/section-main-product.css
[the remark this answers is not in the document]
ludoboludo: yeah I agree. Maybe it can be something we look into once it has a bit more support than the last 2 versions of Safari. 👍

sections/main-product.liquid
ludoboludo: True 👍 Sounds good

snippets/product-media-gallery.liquid
ludoboludo: Yeah it can look like a thumbnail in this case. But maybe it was deemed to "magical" to treat it this way ? 🤷

ludoboludo (dismissed):
I couldn't find a way to break tbh. So it's looking good to me. I might have more scenarios that come to mind later.

One thing Im wondering is if we have considered having a separation of those behaviour based on breakpoint like we do for the desktop and mobile layout. 
I'm wondering if it's something you'd want to apply on desktop only and use a different combination on mobile. 
So for example, you could mitigate the spacing between images in the slider on mobile by using `fill` along with `Constrain media to screen size` while on desktop you might not want to use the `fill` 🤔 
Also this is the type of thing that could be looked into in the future once we get merchant feedback on the existing functionality.

locales/en.default.schema.json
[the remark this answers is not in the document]
duygukalaycioglu: Thanks for bringing this up @ludoboludo! I think it is a good thing we use it before – merchants will be familiar with the concept. The reason we didn't use a checkbox was now we have 2 options under media fit. Media fit: Original and fill. @kmeleta we aware of this one, no problem.

snippets/product-thumbnail.liquid
```diff
@@ -51,8 +50,8 @@
 {%- endcapture -%}
 
 <div
-  class="product-media-container gradient global-media-settings"
-  style="--ratio-percent: {{ preview_media_ratio }}%;"
+  class="product-media-container media-type-{{ media.media_type }} media-fit-{{ media_fit }}{% if constrain_to_viewport %} constrain-height{% endif %} gradient global-media-settings"
```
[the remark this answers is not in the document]
LucasLacerdaUX: +1 :D

locales/en.default.schema.json
```diff
@@ -2019,17 +2019,20 @@
             "label": "Thumbnail carousel"
           }
         },
+        "constrain_to_viewport": {
+          "label": "Constrain media to screen size"
```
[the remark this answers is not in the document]
kmeleta: Worth mentioning. I actually don't have the context of that specific language choice. @duygukalaycioglu could you speak to why we didn't go with "screen height"?

assets/section-main-product.css
```diff
@@ -1366,14 +1366,79 @@ a.product__text {
...
   max-width: calc(100% - calc(var(--media-border-width) * 2));
 }
 
+.product-media-container.constrain-height {
+  /* amount to subtract from viewport height when calculating media height */
+  --viewport-offset: 400px;
+  /* smallest height the image is allowed to scale down to */
+  --constrained-min-height: 300px;
+  /* the larger of either the calculated viewport height or the fixed minimum */
+  --constrained-height: max(var(--constrained-min-height), calc(100vh - var(--viewport-offset)));
```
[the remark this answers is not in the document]
kmeleta: I was leaning that way because there's not too much downside to keeping them around, but fully assumed there was a chance I'd get pushback in review. I tried to name the vars clearly but I also assumed I was too close to it to adequately judge if everything would be easy to decipher for others (or future me). I can remove them if we think the vars are clear enough as is. @ludoboludo @eugenekasimov any opinions?

snippets/product-thumbnail.liquid
```diff
@@ -111,9 +110,9 @@
 
   {%- if media.media_type != 'image' -%}
     {%- if media.media_type == 'model' -%}
-      <product-model class="deferred-media media media--transparent no-js-hidden" style="--ratio-percent: 100%;" data-media-id="{{ media.id }}">
+      <product-model class="deferred-media media media--transparent no-js-hidden" data-media-id="{{ media.id }}">
```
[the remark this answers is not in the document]
kmeleta: By default we preserve the existing 1:1 ratios for models, but you are correct, when using media fit: fill there are scenarios we they can diverge from that. I believe these cases a fairly minimal as I understand, but I'll outline them below for clarity. Let me know if I'm missing any.

1. In cases of `media_fit: fill` and `desktop_layout: 2 columns` and model shares a row with a _portrait_ image. If the other image on the same row is a landscape or also square, the model media shouldn't crop. Particularly if 1:1 is the most common product image ratio, this should be an infrequent combination.


2. In cases of `media_fit: fill` and `constrain: true` and viewport is small enough to constrain even square images. This can happen on tablet/desktop whenever models render fullwidth, where they can feasibly be large enough to be constrained as 1:1 images. I don't think this feel this will be _super_ common either, but I do I think this is more likely to occur than the other scenario, especially if we were to default the constrain setting to true/on.


> If we were to force 3D models to be square would add a layer of "magic" to the setting and make it less clear

Just want to mention this is an option of course, particularly for cases when the models are fullwidth (anything but desktop layout: 2 columns), but it would indeed be an unpredictable exception to the setting, which as I understand is our main issue with it.

snippets/product-thumbnail.liquid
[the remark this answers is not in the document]
kmeleta: We never aligned on this formally as a team. I don't feel too strongly about pushing back in this particular case, so I can adjust, but I'm not always sold on basing the order of classes because of the arbitrary decision to include an inline conditional or not. `gradient` for a example is a much more generic global class and we usually like to order those after more specific classes. Again it's a bit of a toss up here if gradient and media-settings are more generic than constrain-height, but that's how I think about it.

```class=”grid grid--2-col has-border show-heading-or-something slider {% if show_slider_desktop %} slider–desktop{% endif %}{% if some_setting %} grid--1-col{% endif %}”```

Also consider this ^ example, `grid--1-col` gets moved to the end and is now harder to associate with other relevant related classes. I don't like that and that doesn't make this code any easier to understand at a glance.

File: assets/section-main-product.css (hunk 11 above)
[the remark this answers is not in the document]
Comment (kmeleta) [reply assets/section-main-product.css]: I ended up removing _most_. I left a couple in elsewhere that I felt warranted special mention

**Comment by kmeleta** [comment]
I've made all the actionable updates requested, mostly updates to settings language and minor code formatting changes.

**Review by duygukalaycioglu — DISMISSED** [review]

File: assets/section-main-product.css [hunk 13]
```diff
@@ -1366,14 +1366,71 @@ a.product__text {
...
 .product-media-container {
+  --aspect-ratio: var(--preview-ratio);
+  --ratio-percent: calc(1 / var(--aspect-ratio) * 100%);
+  position: relative;
   width: 100%;
   max-width: calc(100% - calc(var(--media-border-width) * 2));
 }
 
+.product-media-container.constrain-height {
+  --viewport-offset: 400px;
```
ludoboludo, line 1377: I was thinking actually that because this is a value we chose "randomly" that we could leave a comment at least just here to mention that. 
So that if we ever need to revisit it, we don't need to go anywhere else to look for answers. 

Something like: `amount chosen and decided after general testing`

ludoboludo: Just thought it'd be good thing to keep doing going forward. We've found ourselves sometimes not being sure how some approaches/values were decided. 
That's the only thing left I'd like to see in the PR yeah 🤔 👍 Not a huge thing but I feel like let's start doing it now.

kmeleta: https://github.com/Shopify/dawn/pull/2052/commits/54bd15dd8edca2ae329c570de608e56e26e631d6

ludoboludo: Much appreciated 🤗

PR #2083, opened 2022-11-01
[Header] Move 'Logo' to Global Theme settings

LucasLacerdaUX:
### PR Summary: 
Important: We moved “Logo” and “Logo width” settings from the header and password header sections to the global theme settings under the heading “Logo”. This update may cause a visual change to your logo.

Moves logo to so it can be configured in a single place and used in multiple parts of the theme.

The `header` and `password-header` sections will no longer have different logo pickers. A single logo is now used across both section, with the same width setting. This may require merchants to make some adjustments to either their logo image or to the settings in these sections.

### Testing steps/scenarios
- [ ] Add a logo under Global Theme Settings
- [ ] Update the maximum logo width on Global Theme Settings
- [ ] Check if the changes are reflected on the Header section on the homepage
- [ ] Check if the changes are reflected on the Header section on the password page

### Demo links
- [Store](url)
- [Editor](url)

### Checklist
- [ x Linted with [Theme Check](https://github.com/Shopify/theme-check)

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -8,12 +8,22 @@
     "theme_support_url": "https://support.shopify.com/"
   },
   {
-    "name": "Logo",
+    "name": "t:settings_schema.logo.name",
     "settings": [
       {
         "type": "image_picker",
         "id": "logo",
-        "label": "Logo"
+        "label": "t:settings_schema.logo.settings.logo_image.label"
+      },
+      {
+        "type": "range",
+        "id": "logo_width",
+        "min": 50,
+        "max": 250,
+        "step": 10,
+        "default": 100,
+        "unit": "t:settings_schema.logo.settings.logo_width.unit",
+        "label": "t:settings_schema.logo.settings.logo_width.label"
       }
     ]
   },
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -8,6 +8,17 @@
     "theme_support_url": "https://support.shopify.com/"
   },
   {
+    "name": "Logo",
+    "settings": [
+      {
+        "type": "image_picker",
+        "id": "logo",
+        "label": "Logo",
+        "info": "t:settings_schema.favicon.settings.favicon.info"
+      }
+    ]
+  },
+  {
     "name": "t:settings_schema.colors.name",
     "settings": [
       {
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -13,8 +13,7 @@
       {
         "type": "image_picker",
         "id": "logo",
-        "label": "Logo",
-        "info": "t:settings_schema.favicon.settings.favicon.info"
+        "label": "Logo"
       }
     ]
   },
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -962,6 +962,9 @@
         "logo": {
           "label": "Logo image"
         },
+        "logo_help": {
+          "content": "Logo image can be updated in Theme Settings"
+        },
         "logo_width": {
           "unit": "px",
           "label": "Custom logo width"
@@ -1626,6 +1629,9 @@
       "settings": {
         "logo": {
           "label": "Logo image"
+        },
+        "logo_help": {
+          "content": "Logo image can be updated in Theme Settings"
         },
         "logo_max_width": {
           "label": "Custom logo width",
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1636,7 +1636,7 @@
       "name": "Password header",
       "settings": {
         "logo_header": {
-          "label": "Logo"
+          "content": "Logo"
         },
         "logo_help": {
           "content": "Edit your logo in theme settings."
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -159,6 +159,18 @@
         }
       }
     },
+    "logo": {
+      "name": "Logo",
+      "settings": {
+        "logo_image": {
+          "label": "Logo image"
+        },
+        "logo_width": {
+          "unit": "px",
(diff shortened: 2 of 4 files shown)


danielvan:
I have a few questions here around how this behaves. Will this new 'Global logo' impact only the password page header logo and the homepage header logo? In what other places, sections and blocks is this logo displayed and for those, would the merchant ever want two colors? (e.g. the logo in color, and the logo in reverse colors)

LucasLacerdaUX:
@danielvan It should only impact the `header` and `password-header` for now. Eventually, we may expand this setting to work on the gift-card template - which currently does not support the use of logos.

A downside of this decision, like you mentioned, is that the image now must have enough contrast with the color schemes being applied to both these sections. Our current use of logos is restricted to nearly identical contexts (headers) so we felt confident that this would not result in many issues for merchants.

For existing merchants, we will include a note in the release notes that they should review the password-header & header sections before publishing the updated theme, as only the main `header` logo will be carried forward.

melissaperreault:
Hey! In the case that the logo lives in theme settings there is a new situation to prevent confusion for.
- [ ] Let's discuss how we can improve the experience for when merchants are editing their Checkout. We might need some helper text to inform the logo is only displaying on their header and password page but not Checkout. We can have a situation where merchants wonder where is their logo since the setting lives in the side panel now among the Checkout preview as you navigate to that step.  [Video reference](https://screenshot.click/04-50-vuqeu-qasa2.mp4)

melissaperreault:
> Should we consider it for all these settings in this case, or do you think the Logo is more likely to cause confusion?

It would be ideal in general that if we could present the Checkout settings panel in view/in context of what you can edit so you don't have a big learning curve that the logo, typography and colors of Checkout is made in a separate category in the panel. Might not be worth solving or adding extra content for now and monitor if merchants are more confused. It's true that it is already the case with colors and that we would need to look at this from a more global robust angle.

nicklepine:
Other small comments: 

-[ ] Could we label the logo picker "Logo" and not "Logo Image"? (Feedback from SLT) 
- [ ] Could we label the category "Logo" and not "Logo and Favicon"
**
image
**

sections/header.liquid
```diff
@@ -33,7 +33,7 @@
   }
```
sofiamatulis, line 33: I am getting this error on `middle center` https://screenshot.click/10-07-2ads6-1mdfs.mp4

locales/en.default.schema.json
```diff
@@ -159,6 +159,21 @@
...
       }
     },
+    "logo": {
+      "name": "Logo",
+      "settings": {
+        "logo_image": {
+          "label": "Logo"
+        },
+        "logo_width": {
+          "label": "Custom logo width"
```
sofiamatulis, line 169: I wonder about this label because in some scenarios the width is smaller than what we have in the setting. In this case you have 250 in the setting and 193 in the actual size. thoughts? 👀 
Screen Shot 2022-11-10 at 1 13 06 PM

sofiamatulis (commented):
Looks great 🎉 Added a few comments

sections/header.liquid
```diff
@@ -497,14 +497,14 @@
         <h1 class="header__heading">
       {%- endif -%}
           <a href="{{ routes.root_url }}" class="header__heading-link link link--text focus-inset">
-            {%- if section.settings.logo != blank -%}
-              {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
-              {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
+            {%- if settings.logo != blank -%}
+              {%- assign logo_alt = settings.logo.alt | default: shop.name | escape -%}
+              {%- assign logo_height = settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
```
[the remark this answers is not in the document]
LucasLacerdaUX: My bad. Forgot to update the references. Good catch. :D

sections/header.liquid
```diff
@@ -497,14 +497,14 @@
...
               {{ section.settings.logo | image_url: width: section.settings.width | image_tag:
```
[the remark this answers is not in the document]
LucasLacerdaUX: Good catch! Fixed :D

sections/header.liquid
LucasLacerdaUX: Good catch! Fixed. Thanks Sofia :D

allylane:
Screenshot 2022-11-14 at 10 55 55 AM
Some settings get cut off when you have a portrait image uploaded.

locales/en.default.schema.json
allylane: Yea, I think it is a more accurate description so we should change it.

config/settings_schema.json
```diff
@@ -7,6 +7,32 @@
     "theme_documentation_url": "https://help.shopify.com/manual/online-store/themes",
```
sofiamatulis, line 7: Is this expected? (when it's at the largest size)

https://screenshot.click/08-49-udirq-zfrjx.mp4

sofiamatulis (commented):
Looks great 🎉 Works as expected

Added one question

config/settings_schema.json
```diff
@@ -7,6 +7,32 @@
...
+      {
+        "type": "image_picker",
+        "id": "logo",
+        "label": "t:settings_schema.logo.settings.logo_image.label"
+      },
+      {
+        "type": "range",
+        "id": "logo_width",
+        "min": 50,
+        "max": 250,
```
LucasLacerdaUX, line 22:
```suggestion
        "max": 300,
```

eugenekasimov (approved):

Issue #1893, opened 2022-08-05
Broken Liquid when you use images wider than 5760px in places


willbroderick:
This occurs in a few places. This example shows the problem on the article template.

**To recreate**

Upload an 8000px wide image to a blog post's featured image.
Go to that blog post in Dawn (any version that passes this through the `image_url` filter) and view source.
See error in source:
![image](https://user-images.githubusercontent.com/1590737/183026625-8d3a37cf-f86f-4acb-8c4e-d350906cdf01.png)

Caused by this line: 
https://github.com/Shopify/dawn/blob/87b7f1edaa6859fb36623d262d4044bcd6c73441/sections/main-article.liquid#L265

**Possible solution**

The `image_url` filter should automatically apply `min(param, 5760)` to the width parameter.

It seems like a strange choice (to me) to error when passing a large width into this filter. I can't think of a case where this is the preferred behaviour over the suggestion above.

At the moment, every time `image_url: image.width` is used, the code needs to include an extra line:
```
{% assign image_width = image.width | at_most: 5760 %}
{{ image_url: width: image_width }}
```
This is unnecessarily verbose, slower to process, more prone to user error, and duplicates an arbitrary platform-limit in theme code - a limit that could change.

richardmonette:
If only width (or height) is specified then perhaps it could be min(param, 5760)

What would you expect the behaviour to be if both width _and_ height as specified? An aspect ratio preserving resize of the params? e.g.

```
longest_dimension = max(width, height)
scale_factor = 5760 / longest_dimension
width *= scale_factor
height *= scale_factor
```

bakura10:
Hi @richardmonette ,

Thanks for having a look at this. If only one dimension is set, easy, just clamp to whatever the max is supported by Shopify. If both are specified, yes, clamp based on the max dimension and preserve the ratio. This would solve all our trouble for those extreme use case where merchants upload image directly coming from their camera :D

bakura10:
I always wonder as well why the width is needed at all 😂. If I remember, the width allow to generate the URL being used as the src attribute. However with srcset the src is never used. Maybe Shopify could use the largest possible size as the src attribute if no width is explicitly set.

ludoboludo:
>I do see that the docs https://shopify.dev/api/liquid/filters#image_url warn "You need to specify either a width or height parameter. If neither are specified, then an error is returned." however when I tested this it didn't actually appear to be necessary.
cc: @richardmonette 

[...]

PR #2087, opened 2022-11-01
Add fixed sizes for some images to prevent errors

ludoboludo:
### PR Summary: 

In a few places we're passing in the width of our images as an argument for the `image_url` filter. This can create issues as there is a limit being checked and images can't be wider than `5760`. 

So I'm adding a fixed width to those so we're not running into an issue. 

Fixes https://github.com/Shopify/dawn/issues/1893

Added a fixed size, Google states: `Images must be at least 160×90 pixels and at most 1920×1080 pixels.` 

### Other considerations

On the issue there was some chat about the filter itself applying a limit to size that can be passed. 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

No visual impact expected. 

### Testing steps/scenarios

I haven't been able to replicate the issue brought up as I wasn't able to upload an image that was wider than `5760`. I posted on the issue to see if I'm missing something in my replication steps. 
- [ ] Step 1

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/136849326102/editor)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -729,7 +729,7 @@
     "@type": "Organization",
     "name": {{ shop.name | json }},
     {% if section.settings.logo %}
-      "logo": {{ section.settings.logo | image_url: width: 700 | prepend: "https:" | json }},
+      "logo": {{ section.settings.logo | image_url: width: 500 | prepend: "https:" | json }},
     {% endif %}
     "sameAs": [
       {{ settings.social_twitter_link | json }},
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -729,7 +729,7 @@
     "@type": "Organization",
     "name": {{ shop.name | json }},
     {% if section.settings.logo %}
-      "logo": {{ section.settings.logo | image_url: width: section.settings.logo.width | prepend: "https:" | json }},
+      "logo": {{ section.settings.logo | image_url: width: 700 | prepend: "https:" | json }},
     {% endif %}
     "sameAs": [
       {{ settings.social_twitter_link | json }},
--- a/sections/main-article.liquid
+++ b/sections/main-article.liquid
@@ -278,7 +278,7 @@
         "logo": {
           "@type": "ImageObject",
           "height": {{ settings.share_image.height | json }},
-          "url": {{ settings.share_image | image_url: width: settings.share_image.width | prepend: "https:" | json }},
+          "url": {{ settings.share_image | image_url | prepend: "https:" | json }},
           "width": {{ settings.share_image.width | json }}
         },
       {% endif %}
--- a/sections/main-article.liquid
+++ b/sections/main-article.liquid
@@ -263,7 +263,7 @@
     {% endif %}
     {% if article.image %}
       "image": [
-        {{ article | image_url: width: article.image.width | prepend: "https:" | json }}
+        {{ article | image_url | prepend: "https:" | json }}
       ],
     {% endif %}
     "datePublished": {{ article.published_at | date: '%Y-%m-%dT%H:%M:%SZ' | json }},
--- a/sections/main-article.liquid
+++ b/sections/main-article.liquid
@@ -263,7 +263,7 @@
     {% endif %}
     {% if article.image %}
       "image": [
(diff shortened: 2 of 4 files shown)


bakura10:
Has the change to make the width argument optional been merged already ?

ludoboludo:
@bakura10 From the sound of it, it always was the case but I'm a bit skeptical. Which is why I had not asked for reviews yet. I should change the status to `draft` for now as I would ideally want to have our docs updated to reflect the change as well. So we don't mention its requirement.

bakura10:
100% sure it was not the case :D. The image_url used to fail if you did not pass a width attribute so it would be nice to have a confirmation indeed :)

ludoboludo:
Yeah you're right it's still failing. I'm getting `Liquid error (sections/main-product.liquid line 58): image_url must define width or height
`.

sections/main-article.liquid
```diff
@@ -278,7 +278,7 @@
         "logo": {
           "@type": "ImageObject",
           "height": {{ settings.share_image.height | json }},
-          "url": {{ settings.share_image | image_url: width: settings.share_image.width | prepend: "https:" | json }},
+          "url": {{ settings.share_image | image_url: width: 1920 | prepend: "https:" | json }},
           "width": {{ settings.share_image.width | json }}
```
eugenekasimov, line 282: I'm not sure for 100% but I can assume that `"height":` and `"width":` may lead to an error. I would remove them, or leave only one of them with fixed value. What do you think?

eugenekasimov (commented):
Hey Ludo. I think your approach fixes the current issue. I left a minor comment about my concern.

Based on what I have learned about Schema.org it may be a good idea to have a few alternative versions of images (different resolutions and/or aspect ratios). It can improve a search result for pages. It can be as a follow up PR though. Let me know what you think about it. Thanks.

sections/main-article.liquid
ludoboludo: I'm not sure I follow. 
>or leave only one of them with fixed value

Isn't it what I'm doing in the PR already ^ ?

As for adding multiple images, yeah I think it could be done as a follow up if we see it's necessary 👍

eugenekasimov: I'm sorry Ludo, it seems like I'm bad at explaining. My concern is next:

- line 281 - you provide `url` with fixed `width: 1920`.
- line 282 - you provide `width` with value **`settings.share_image.width`** that may pass an original size of an image and it can be 5000px or bigger.

So, there shouldn't be an error in Liquid as it was before this PR but I just wanna be sure that there is no conflict or potential issue for merchants for their search optimization. Because I don't know what happens in case when the actual image is pulled through the url has **`width 1920px`**, but at the same time we are telling Schema.org to use `width: settings.share_image.width` which can be 5000px for example.

ludoboludo: Oh I see what you mean 👍 good point, thanks for bringing it up 🙂

sections/main-article.liquid
```diff
@@ -274,14 +274,6 @@
...
     "publisher": {
       "@type": "Organization",
-      {% if settings.share_image %}
-        "logo": {
-          "@type": "ImageObject",
-          "height": {{ settings.share_image.height | json }},
-          "url": {{ settings.share_image | image_url: width: settings.share_image.width | prepend: "https:" | json }},
-          "width": {{ settings.share_image.width | json }}
-        },
-      {% endif %}
```
ludoboludo, line 284: Removed this as it's not outputting ever as a `share_image` setting doesn't exists.

sections/header.liquid
```diff
@@ -729,7 +729,7 @@
     "@type": "Organization",
     "name": {{ shop.name | json }},
     {% if section.settings.logo %}
-      "logo": {{ section.settings.logo | image_url: width: section.settings.logo.width | prepend: "https:" | json }},
+      "logo": {{ section.settings.logo | image_url: width: 700 | prepend: "https:" | json }},
```
metamoni, line 732: Just out of curiosity, where is this requirement for the logo width coming from?

eugenekasimov (approved):

metamoni (approved):

PR #2106, opened 2022-11-07
Select text input content on trapFocus

FCalabria:
### PR Summary: 

Select text input's content when automatically focusing it

This affects the search input on the header, that gets focused automatically when the user clicks on the "search" icon. The behaviour mimics the browser default when tabbing into an input, but I restricted it to text types to avoid causing UX problems or bugs on other kinds of inputs (passwords, dates, etc).

Use the native `setSelectionRange` function

<details>
<summary>Gif preview</summary>

</details>

### Demo links
- [Store](https://pakichu.myshopify.com/?_ab=0&_fd=0&_sc=1)
- [Editor](https://admin.shopify.com/store/pakichu/themes/129287651384/editor)

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -68,6 +68,13 @@ function trapFocus(container, elementToFocus = container) {
   document.addEventListener('focusin', trapFocusHandlers.focusin);
 
   elementToFocus.focus();
+
+  if (elementToFocus.tagName === 'INPUT' &&
+    ['search', 'text', 'email', 'url'].includes(elementToFocus.type) &&
+    elementToFocus.value) {
+    const end = elementToFocus.value.length;
+    elementToFocus.setSelectionRange(0, end);
+  }
 }
 
 // Here run the querySelector to figure out if the browser supports :focus-visible or not and run code based on it.


assets/global.js
```diff
@@ -68,6 +68,13 @@ function trapFocus(container, elementToFocus = container) {
   document.addEventListener('focusin', trapFocusHandlers.focusin);
 
   elementToFocus.focus();
+
+  if (elementToFocus.tagName === 'INPUT' &&
+    ['search', 'text', 'email', 'url'].includes(elementToFocus.type) &&
+    elementToFocus.value) {
+    const end = elementToFocus.value.length;
+    elementToFocus.setSelectionRange(0, end);
+  }
```
ludoboludo, line 77: since this is in the `trapFocus()` function this will only run when we're trapping the focus. But won't happen anywhere else on inputs. So for example on the `about us` page for the `os2-demo` store, it won't do anything on the contact us form fields.

We currently run `trapFocus()` on: 
- the cart drawer
- the cart notification
- modals using the `detais-modal` custom HTML element (search, password page modal to enter password)
- the menu drawer 
- the `modal-dialog` custom HTML element
- the pickup availability drawer

In all those cases, I think the search is the only scenario where the element receiving focus due to our `trapFocus` function is an `input` element. 

I'm not sure if introducing this behaviour only for that one specific part of the theme makes sense when the rest of the input don't get the same behaviour on focus 🤔 

But then having a listener for `focusin` on all inputs of the theme might be a bit much. Not sure. 

What do you think ? 🤔

ludoboludo: >Adding this on every focusin event will select the value also when the user focuses the input by clicking on it, which I don't think is a good UX since they might be clicking exactly in the point were they want their cursor.
💯 

I'll approve the PR 👍 Though we should wait a bit before merging (once you get the second approval).

assets/global.js
```diff
@@ -68,6 +68,13 @@ function trapFocus(container, elementToFocus = container) {
   document.addEventListener('focusin', trapFocusHandlers.focusin);
 
   elementToFocus.focus();
```
sofiamatulis, line 70: Search only focuses when I close and re-enter the field:https://screenshot.click/08-12-xeqqs-dai70.mp4

That doesnt happen for the other inputs: https://screenshot.click/08-15-5ykcg-12f5p.mp4

Is that expected? 👀

assets/global.js
```diff
@@ -68,6 +68,13 @@ function trapFocus(container, elementToFocus = container) {
...
+
+  if (elementToFocus.tagName === 'INPUT' &&
+    ['search', 'text', 'email', 'url'].includes(elementToFocus.type) &&
+    elementToFocus.value) {
+    const end = elementToFocus.value.length;
+    elementToFocus.setSelectionRange(0, end);
```
sofiamatulis, line 76: Do we need to create the const if it's only used once?
```suggestion
    elementToFocus.setSelectionRange(0, elementToFocus.value.length);
```

sofiamatulis (commented):
Looks good 👍 Thanks for making the change. Added some small nitpicks

assets/global.js
sofiamatulis: I am using chrome and just tabbing 🤔 Am I missing something?

https://screenshot.click/10-34-j4hjv-6b43g.mp4

sofiamatulis: Interesting 🤔 Then how come Paqui is able to do it in her video? https://screenshot.click/10-59-0u5in-jynpn.mp4

FCalabria: That should be standard browser behaviour, it's not due to any change i have done

sofiamatulis: What I mean is that the text input's content is selected when automatically focusing it on your video but not in mine

PR #2110, opened 2022-11-08
Tweak the image_url size to be the max value where necessary

ludoboludo:
### PR Summary: 

This change affect the logo in the header (in the password page template and home page) as well as for the image banner images. 
We were limiting the images' size wrongfully. This should output images as initially intended. 

Image in image banner section can get a little blurry even though the original is big enough to be sharp.

In a few places we were setting `image_url: width: xxx` to the wrong size. It should be using the maximum size we're setting in the `srcset` attribute.

### Other considerations

N/A

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

Very small, if anything, sharper images.

### Testing steps/scenarios

- [ ] Add an image for the image banner section and check out the network tab to see which version of the image is loaded. It used to go up to a max of `1500` when it should be a max of `3840`. [Video explanation](https://screenshot.click/08-53-tg2o2-t6owh.mp4)

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/137431154710/editor)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -377,7 +377,7 @@
             {%- if section.settings.logo != blank -%}
               {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
               {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
-              {{ section.settings.logo | image_url: width: section.settings.width | image_tag:
+              {{ section.settings.logo | image_url: width: 500 | image_tag:
                 class: 'header__heading-logo',
                 widths: '50, 100, 150, 200, 250, 300, 400, 500',
                 height: logo_height,
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -500,7 +500,7 @@
             {%- if section.settings.logo != blank -%}
               {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
               {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
-              {{ section.settings.logo | image_url: width: section.settings.width | image_tag:
+              {{ section.settings.logo | image_url: width: 500 | image_tag:
                 class: 'header__heading-logo',
                 widths: '50, 100, 150, 200, 250, 300, 400, 500',
                 height: logo_height,
--- a/sections/image-banner.liquid
+++ b/sections/image-banner.liquid
@@ -45,7 +45,7 @@
           assign sizes = "100vw"
         endif
       -%}
-      {{ section.settings.image | image_url: width: 1500 | image_tag:
+      {{ section.settings.image | image_url: width: 3840 | image_tag:
         loading: 'lazy',
         width: section.settings.image.width,
         height: image_height,
@@ -75,7 +75,7 @@
           assign sizes = "100vw"
         endif
       -%}
-      {{ section.settings.image_2 | image_url: width: 1500 | image_tag:
+      {{ section.settings.image_2 | image_url: width: 3840 | image_tag:
         loading: 'lazy',
         width: section.settings.image_2.width,
         height: image_height_2,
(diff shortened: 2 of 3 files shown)


sections/header.liquid
```diff
@@ -500,7 +500,7 @@
             {%- if section.settings.logo != blank -%}
               {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
               {%- assign logo_height = section.settings.logo_width | divided_by: section.settings.logo.aspect_ratio -%}
-              {{ section.settings.logo | image_url: width: section.settings.width | image_tag:
+              {{ section.settings.logo | image_url: width: 500 | image_tag:
```
ludoboludo, line 503: Here I'm changing it as the width is a max of `250` which means we're never giving the logo a chance to be a `500` wide file.

sections/main-password-header.liquid
```diff
@@ -9,7 +9,7 @@
     {%- if section.settings.logo != blank -%}
       {%- assign logo_alt = section.settings.logo.alt | default: shop.name | escape -%}
       {%- assign logo_height = section.settings.logo_max_width | divided_by: section.settings.logo.aspect_ratio -%}
-      {{ section.settings.logo | image_url: width: section.settings.logo.width | image_tag:
+      {{ section.settings.logo | image_url: width: 500 | image_tag:
```
ludoboludo, line 12: Here it's a bit different, since the logo uploaded could be much wider, we don't really need to have something that big. So I default to the max value it's going to use instead.

sections/image-banner.liquid
```diff
@@ -45,7 +45,7 @@
           assign sizes = "100vw"
         endif
       -%}
-      {{ section.settings.image | image_url: width: 1500 | image_tag:
+      {{ section.settings.image | image_url: width: 3840 | image_tag:
```
eugenekasimov, line 48: Here I believe we can keep 3840 as a biggest value among all available widths (line 54). Or we could replace it with `section.settings.image.width ` which gives us the same result. I agree that we don't want to have a limit of 1500 since most uploaded images have bigger resolution which caused a blurry effect.

sections/image-banner.liquid
```diff
@@ -75,7 +75,7 @@
           assign sizes = "100vw"
         endif
       -%}
-      {{ section.settings.image_2 | image_url: width: 1500 | image_tag:
+      {{ section.settings.image_2 | image_url: width: 3840 | image_tag:
```
eugenekasimov, line 78: Same thing is here.

sections/main-password-header.liquid
eugenekasimov: I think here is the same thing that I mentioned in the first comment about logo in header.liquid. I don't see what changed. It seems even with previous setting `width: section.settings.logo.width` it doesn't go higher than a biggest value from  `widths: '50, 100, 150, 200, 250, 300, 400, 500',` (line 14).

ludoboludo: This is to prevent anything too big from being loaded. If we know the max value we're outputting is `500` no need to load an image bigger than that.

sections/image-banner.liquid
ludoboludo: The reasoning here is that when using this way of outputting images we tend to mention sizes in two different spots. 

- Once in the `image_url: width: XXX`, which is a fallback in case the browser doesn't support the `srcset` attribute. And it also serves as the image we will be querying/resizing in `srcset`.
- Once in the `srcset` attribute. This will use the image mentioned in the point above and its size to output the different sizes. So that's why we always need to make sure that the size is as big as our biggest output value in the `srcset`.

kmeleta (approved):
Fixes the original issue and aligned the approach 👍  

Thought I noticed another sizes issue while testing but I'm struggling to reproduce it. I'm thinking now it was just an editor effect https://screenshot.click/10-42-ksbkc-nhrk4.mp4

PR #2112, opened 2022-11-09
[Footer] Brand information block

### LINKED ISSUE #2092: 🌐 Brand Information Settings
- [ ] Introduces brand information settings in global theme settings
- [ ] new footer block
- [ ] remove footer white space if empty

LucasLacerdaUX:
### PR Summary: 
We added a new Brand Information block to the Footer section. The content can be configured on Theme Settings.

Introduces a new Brand Information block to the Footer.
Fixes #2092.

- The `Brand Information` content lives in Theme Settings. This follows a similar pattern to our Social Media icons, and we intentionally designed it this way (future project requirement).
- Re-used classes from other footer blocks
- Moved Social Media icons into a `social-icons` snippet.
- Social Media icons and settings have been re-ordered to follow popularity of each platform
- We've increased the maximum size in the `Image` block. 

### Testing steps/scenarios
- [ ] Add a new Brand Information block to the Footer
- [ ] If the Brand Information is the sole block on the Footer, content should be centered. Otherwise, left-aligned.
- [ ] Add Social Media
- [ ] Tweak image size

### Demo links
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139158257686)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139158257686/editor)

### Checklist


allylane:
Screenshot 2022-12-02 at 2 45 44 PM
If there's no content in this section, it shouldn't have padding.

allylane:
Screenshot 2022-12-02 at 2 53 20 PM
Ideally we'll have a border top on this section. 
We could add a checkbox to "Footer" to show or hide it so it's not on every theme.

allylane:
As we chatted about, let's put the footer links on a new line.
Screenshot 2022-12-02 at 3 05 52 PM

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
...
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
                       {%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
                       <img
                         srcset= "{{ block.settings.image | image_url: width: block.settings.image_width }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
-                        src="{{ block.settings.image | image_url: width: 400 }}"
+                        src="{{ block.settings.image | image_url: width: 760 }}"
                         alt="{{ block.settings.image.alt | escape }}"
+                        class="global-media-settings"
```
ludoboludo, line 105: Pseudo elements don't work on `image` elements. So only the border radius and border work. Is that expected or should the shadow be ideally coming up ?
<details>
<summary>Screenshot</summary>

[docs](https://developer.mozilla.org/en-US/docs/Web/CSS/::after)

</details>

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'brand_information' -%}
+                  <div class="footer-block-brand-information">
+                    {%- if settings.brand_image != blank -%}
+                      {%- assign image_size_2x = settings.brand_image.width | times: 2 | at_most: 5760 -%}
```
ludoboludo, line 79: I think here, since we know what the max value can be, `550px` then we should use that. Otherwise if a merchant use a large image we're going to be loading an image super heavy without any real benefit but a hit on performance. 
<details>
<summary>Screenshot</summary>


</details>

And when it uses `at_most: 550` then the weight is 10 times lighter and still look good 👍

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
...
+                      <img
+                        srcset= "{{ settings.brand_image | image_url: width: settings.brand_image.width }}, {{ settings.brand_image | image_url: width: image_size_2x }} 2x"
+                        src="{{ settings.brand_image | image_url: width: 760 }}"
+                        alt="{{ settings.brand_image.alt | escape }}"
```
ludoboludo, line 83: I've noticed in another PR that we tend to use the `| escape` filter often for alt text but I'm not sure why 🤔

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
...
+                        class="global-media-settings"
+                        loading="lazy"
+                        width="{{ settings.brand_image.width }}"
+                        height="{{ settings.brand_image.height }}"
+                        style="max-width: min(100%, {{ settings.brand_image_width }}px);"
+                      >
```
ludoboludo, line 89: should we leverage the image_tag filter instead ?

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'brand_information' -%}
+                  <div class="footer-block-brand-information">
+                    {%- if settings.brand_image != blank -%}
+                      {%- assign image_size_2x = settings.brand_image.width | times: 2 | at_most: 5760 -%}
+                      <img
+                        srcset= "{{ settings.brand_image | image_url: width: settings.brand_image.width }}, {{ settings.brand_image | image_url: width: image_size_2x }} 2x"
```
ludoboludo, line 81:
```suggestion
                        srcset= "{{ settings.brand_image | image_url: width: settings.brand_image_width }}, {{ settings.brand_image | image_url: width: image_size_2x }} 2x"
```

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
...
+                        alt="{{ settings.brand_image.alt | escape }}"
+                        class="global-media-settings"
+                        loading="lazy"
+                        width="{{ settings.brand_image.width }}"
+                        height="{{ settings.brand_image.height }}"
+                        style="max-width: min(100%, {{ settings.brand_image_width }}px);"
+                      >
+                    {%- endif -%}
+                    <h2 class="footer-block__heading">{{ settings.brand_headline }}</h2>
+                    <p>{{ settings.brand_description }}</p>
```
ludoboludo, line 92: We ok with rendering those elements when not content is added in the settings for them ?

sections/footer.liquid
```diff
@@ -73,14 +73,36 @@
...
+                    {%- if block.settings.show_social -%}
+                      {%- render 'social-icons' -%}
+                    {%- endif -%}
+                  </div>
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
                       {%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
```
ludoboludo, line 100: I guess my comment would apply here though it's been like that for a while.

config/settings_schema.json
```diff
@@ -33,6 +33,38 @@
       }
     ]
   },
+  {
+    "name": "t:settings_schema.brand_information.name",
+    "settings": [
+      {
+        "type": "richtext",
+        "id": "brand_headline",
+        "default": "<p></p>",
```
ludoboludo, line 42: I wonder if we should have something here 🤔 I guess we don't want to create work for people to go and delete it. 
<details>
<summary>Screenshot</summary>


</details>

assets/section-footer.css
```diff
@@ -479,11 +479,30 @@ noscript .localization-selector.link {
...
+
+  .footer-block:not(:only-child) .footer-block-brand-information .footer__list-social.list-social {
+    justify-content: flex-start;
+    margin-left: -1.3rem;
+  }
+}
+
 .footer-block__details-content .placeholder-svg {
   max-width: 20rem;
 }
```
ludoboludo, line 508: Not related to any of your changes on the PR but the default column gap value looks tight on those footer columns: 
<details>
<summary>Screenshot</summary>


</details>

config/settings_schema.json
metamoni: We shouldn't use `<p></p>` as the default value here, as this results in invalid HTML. A `p` element can't be a child of a heading.
invalid

locales/en.default.schema.json
```diff
@@ -904,6 +921,21 @@
...
+          "settings": {
+            "paragraph": {
+              "content": "This block will display your brand information. [Edit brand information.](/editor?context=theme&category=brand%20information)"
+            },
+            "header__1": {
+              "content": "Social media icons"
+            },
+            "show_social": {
+              "label": "Show social media icons",
+              "info": "To display your social media accounts, link them in your theme settings."
```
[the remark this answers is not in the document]
LucasLacerdaUX: Good catch. I've updated the translation files to link to the social media tab.

sections/footer.liquid
LucasLacerdaUX: Updated to add a check for emptiness.

sections/footer.liquid
LucasLacerdaUX: Updated to use `image_tag`

config/settings_schema.json
```diff
@@ -33,6 +33,38 @@
...
+        "label": "t:settings_schema.brand_information.settings.brand_headline.label"
+      },
+      {
+        "type": "richtext",
+        "id": "brand_description",
+        "default": "<p></p>",
```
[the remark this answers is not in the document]
metamoni: Hmm, I'm still seeing this on the page though 🤔
Screenshot 2022-12-08 at 09 38 04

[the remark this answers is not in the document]
metamoni: I'm still seeing it in the header, and the logo alignment seems to have changed as well

https://user-images.githubusercontent.com/22390758/206412866-461ddf77-c1fc-49d7-9d85-284eae948c9e.mp4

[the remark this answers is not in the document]
LucasLacerdaUX: @metamoni That was a sync error on the demo link, the type was changed to `inline_richtext` which does not have wrapping `<p>` tags on [this commit](https://github.com/Shopify/dawn/pull/2112/commits/aee2a28a7a349ce8a43e5a5b3e0a0a3cadc186cc).

image

I just re-connected the theme and it should now be gone. Let me know if you find any additional issues. I'll continue addressing the other feedback :D

LucasLacerdaUX:
@allylane 

> If there's no content in this section, it shouldn't have padding.

That's a good catch. This will require additional work as it's something that is already present with current blocks: if you add an empty image block, it'll have padding. I'll try to address this in a separate PR.

> When I upload a logo, the brand information section gets cut off.

That's a Theme Editor issue and not related to the theme. I can flag this to the team. Thanks.

> As we chatted about, let's put the footer links on a new line.

Already done in a separate PR (#2084).

> Everything in the design is a bit smaller. Can we bring the size down and bit, and increase the opacity of the body copy:

I'm using the values in our design system. This ensures consistency with other blocks such as the `Text block`:

image

I would suggest to not change opacity of the body copy - or at least not as part of this PR without proper review of everything else - it's a global property on the theme and may affect a lot more than we expect to.

The text size is also following the defaults in the rest of the theme. Are we sure we want to make a special case for this block that will look different from everything else?

As for the social icons, I can easily update their size to be slightly smaller. Would it be okay to change it for both `Brand info` and `Footer`, though?

LucasLacerdaUX:
Updated PR to:
- Fix spacing between image and Headline
- Updated to support RTE link styling
- Updated content copyright font size to 11px instead of 10px.

assets/section-footer.css
```diff
@@ -463,16 +464,54 @@ noscript .localization-selector.link {
   }
 }
 
-.footer-block-image > img {
+.brand-information__image-wrapper {
+  box-shadow: var(--media-shadow-horizontal-offset)
+    var(--media-shadow-vertical-offset)
+    var(--media-shadow-blur-radius)
+    rgba(var(--color-shadow), var(--media-shadow-opacity));
+  margin-bottom: calc(2rem + var(--media-shadow-vertical-offset) * var(--media-shadow-visible));
```
ludoboludo, line 472: this doesn't work well when the vertical offset is a negative value. 
I think on the product page we did it using `padding` so that when it's a negative value, nothing happens 🤔 
But we could also use a `max()`. `max((var(--media-shadow-vertical-offset) * var(--media-shadow-visible)), 0rem)`

assets/section-footer.css
```diff
@@ -463,16 +464,54 @@ noscript .localization-selector.link {
...
+  margin-right: auto;
+}
+
+.footer-block-image > img,
+.footer-block-brand-information > img {
   height: auto;
 }
 
+.footer-block:only-child .footer-block-brand-information .footer__list-social.list-social {
+  justify-content: center;
```
ludoboludo, line 501: I think here we would need to also remove the negative margin left to make sure it's properly centered. Or add a negative `margin-right`
[video](https://screenshot.click/09-30-64en6-l8g6r.mp4)

sections/footer.liquid
```diff
@@ -73,14 +73,43 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'brand_information' -%}
+                  <div class="footer-block-brand-information">
```
ludoboludo, line 77: Nitpick to match the other blocks.
```suggestion
                  <div class="footer-block__brand-information">
```

sections/footer.liquid
```diff
@@ -73,14 +73,43 @@
...
+                    {%- if settings.brand_image != blank -%}
+                      {%- assign brand_image_height = settings.brand_image_width | divided_by: settings.brand_image.aspect_ratio -%}
+                      <style>
+                        .brand-information__image-wrapper {
+                          max-width: min(100%, {{ settings.brand_image_width }}px);
+                        }
+                      </style>
+                      <div class="brand-information__image-wrapper global-media-settings">
```
ludoboludo, line 85: This class doesn't have a parent `brand-information` one to justify the `__image-wrapper` so maybe it could just be `brand-information-image-wrapper` ? 🤷

sections/footer.liquid
```diff
@@ -73,14 +73,43 @@
...
                 {%- when 'image' -%}
                   <div class="footer-block__details-content footer-block-image {{ block.settings.alignment }}">
                     {%- if block.settings.image != blank -%}
                       {%- assign image_size_2x = block.settings.image_width | times: 2 | at_most: 5760 -%}
                       <img
                         srcset= "{{ block.settings.image | image_url: width: block.settings.image_width }}, {{ block.settings.image | image_url: width: image_size_2x }} 2x"
-                        src="{{ block.settings.image | image_url: width: 400 }}"
+                        src="{{ block.settings.image | image_url: width: 760 }}"
                         alt="{{ block.settings.image.alt | escape }}"
+                        class="global-media-settings"
```
ludoboludo, line 112: This image would also need a wrapping div with that class like you did for the branding block. It might have got lost in the rebases.
<details>
<summary>Screenshot</summary>


</details>

LucasLacerdaUX:
@ludoboludo @metamoni Addressed your comments! Let me know if I missed anything! :D

sections/footer.liquid
LucasLacerdaUX: Updated :D

assets/section-footer.css
LucasLacerdaUX: Updated with `margin-right`. Removing `margin-left` would break the layout when icons are left aligned :D

metamoni:
@LucasLacerdaUX when you remove both the Quick links menu and the email signup form there's a lot of whitespace at the bottom of the Brand info block compared to the top. Should these be equal?

Also not seeing the social media icons, for some reason, even though they're enabled. Maybe rebasing the branch might fix it?
spacing

LucasLacerdaUX:
@metamoni The social icons are working fine for me on the demo store in the PR description. Did you add the social media to the Theme Settings?

I've rebased the branch to the latest main - with no conflicts.

As for the additional spacing, this is already in main and happening to other blocks like Menu. I'll tag @allylane to check if this is intentional and maybe we can fix in the follow up PR.
image

assets/section-footer.css
```diff
@@ -457,22 +458,80 @@ noscript .localization-selector.link {
...
 @media screen and (max-width: 749px) {
-  .footer-block-image {
-    text-align: center;
+  .footer-block-image,
+  .footer-block-image.left,
+  .footer-block-image.center,
+  .footer-block-image.right {
+    justify-content: center;
   }
 }
```
ludoboludo, line 484: This seem a bit opposite to our usual approach where we tend to have default values on mobile and overwrites in media queries for tablet/desktop, no ?

sections/footer.liquid
```diff
@@ -75,19 +75,43 @@
                       {%- endfor -%}
                     </ul>
                   {%- endif -%}
+                {%- when 'brand_information' -%}
```
ludoboludo, line 78: I wonder if we should/could re organize the general settings to match the output order: 
<details>
<summary>Screenshot</summary>


</details>

Seems like brand info is only used in the footer so far 🤔 So it should be ok

ludoboludo (approved):
Looking good. Added some little nitpicks but nothing blocking 👍

metamoni (approved):

assets/section-footer.css
LucasLacerdaUX: Good point. That's how it was originally designed so I didn't change, but I can refactor this in the follow up PR 👀

sections/footer.liquid
LucasLacerdaUX: cc @allylane

sections/footer.liquid
```diff
@@ -75,19 +75,43 @@
...
+                          height: brand_image_height,
+                          width: settings.brand_image_width
+                        }}
+                      </div>
+                    {%- endif -%}
+                    {%- if settings.brand_headline != blank -%}
+                      <h2 class="footer-block__heading rte">{{ settings.brand_headline }}</h2>
+                    {%- endif -%}
+                    {%- if settings.brand_description != blank -%}
+                      <div class="rte">{{ settings.brand_description }}</div>
```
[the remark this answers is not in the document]
LucasLacerdaUX: Will also include in the follow up PR :D

PR #2127, opened 2022-11-15
Search ux improvements

FCalabria:
### PR Summary: 

UX improvements on search:
- Added clear button to search input
- The search term is kept in sync between the main search and the search form in the header
- When focusing the main search in mobile, it automatically scrolls to the top of the page

The improvements were suggested by @al-mcquiston in the context of a new feature (that will be on a different PR). For reference, here is the [figma](https://www.figma.com/file/HFKhDU5jNbnpR7U0JnsAur/Query-Suggestions-in-Storefront-Search?node-id=93%3A2521&t=AHxRw6HVpelQWflc-0) on that feature that includes the improvements on this PR.

I created a common `search-form` component to avoid duplicating code between the header and the main search. The input sync functionality is developed on the main search because the only use case where more than one search input is visible would be with that one present.

### Other considerations

You might want to review per commit, since I did each improvement on a different one.

New clear button on search inputs:

Hidden when the input is empty

<details>
<summary>Auto scroll when focusing main search</summary>

</details>

### Demo links

- [Store](https://mcicq3997fukw1yo-45840990230.shopifypreview.com)
- [Editor](https://os2-demo.myshopify.com/admin/themes/137451733014/editor)

### Checklist


assets/component-search.css
```diff
@@ -1,32 +1,71 @@
...
+  content: "";
+  position: absolute;
+  right: 0;
 }
 
-.search__button:focus {
+.search__button:focus-visible,
+.reset__button:focus-visible,
+.search__button:focus,
+.reset__button:focus {
```
ludoboludo, line 28: One thing that comes to mind at quick glance is that we keep those separated `:focus-visible` declarations and `:focus` ones as a browser that doesn't support `focus-visible` like older versions of Safari won't read the whole thing if it doesn't support one selector. 
```suggestion
.search__button:focus-visible,
.reset__button:focus-visible {
  ...
}

.search__button:focus,
.reset__button:focus {
 ...
 }
```

sofiamatulis (commented):
Nice improvements 🎉 Did a first pass

assets/component-search.css
FCalabria: TIL

locales/en.default.json
```diff
@@ -36,7 +36,8 @@
       "previous": "Previous page"
     },
     "search": {
-      "search": "Search"
+      "search": "Search",
+      "reset": "Clear search term"
```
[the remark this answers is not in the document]
ludoboludo: I think that sounds good 👍 I was trying to see what screen readers announce when you use a search input that has the `clear` button by default but it doesn't even announce it. [Video example](https://screenshot.click/18-24-57k3b-snsfz.mp4)

assets/search-form.js
```diff
@@ -0,0 +1,23 @@
...
+    // Prevent default so the form reset doesn't set the value gotten from the url on page load
+    event.preventDefault();
+    // Don't reset if the user has selected an element on the predictive search dropdown
+    if (!document.querySelector('a[aria-selected="true"')) {
+      this.input.value = '';
+    }
+  }
+}
+
+customElements.define('search-form', SearchForm);
```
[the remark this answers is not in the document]
FCalabria: The `extend` is a very nice idea, I'll refactor that.

What do you think about the scope of each file, any concern there? I made them assuming the only moment that there is more than one search input is when the main search is visible, that's why most of the logic is on that file. But if we don't assume that, then most of the logic will move to `search-form.js`. That means loading a bigger file with the header (with code that is not really needed) but also we will have a more resilient code in case that assumption changes in the future. What do yo think?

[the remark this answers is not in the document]
ludoboludo: That's a good point. Maybe it is better for now to keep it as you have it 🤔 Only load that script when actually needed. Since it's on the search page anyway it shouldn't be too impactful. 👍

assets/main-search.js
```diff
@@ -0,0 +1,47 @@
...
+    this.allSearchInputs = document.querySelectorAll('input[type="search"]');
+    this.setupEventListeners();
+  }
+
+  setupEventListeners () {
+    let allSearchForms = [];
+    this.allSearchInputs.forEach(input => allSearchForms.push(input.form))
+    if (allSearchForms.length > 1) {
+      allSearchForms.forEach(form =>
+        form.addEventListener('reset', this.onFormReset.bind(this))
```
ludoboludo, line 13: Here, I think instead of adding another `reset` listener, we could add functionality to our original function `onSearchFormReset(event){}` from the `SearchForm` class using `super` (you can see it used in a few places already in our codebase)

I think this way should work: 
```javascript
onSearchFormReset(event) {
  super.onSearchFormReset(event);
  this.keepInSync('');
}
```

ludoboludo (commented):
Hey 👋 a few more comments/ideas 👍

assets/main-search.js
ludoboludo: Ah right 👍 Makes sense.

assets/search-form.js
```diff
@@ -0,0 +1,20 @@
...
+
+    const form = this.querySelector('form.search');
+    form.addEventListener('reset', this.onFormReset.bind(this));
+  }
+
+  onFormReset(event) {
+    // Prevent default so the form reset doesn't set the value gotten from the url on page load
+    event.preventDefault();
+    // Don't reset if the user has selected an element on the predictive search dropdown
+    if (!document.querySelector('a[aria-selected="true"')) {
```
ludoboludo, line 14: I think this is meant to be `li[aria-selected="true"` 🤔 When I check it's the `li` element that is getting the `aria-selected` attribute updated. 

Though I'm not sure the check is working. Or maybe I'm not testing it properly. When I select one item in the dropdown using my keyboard then click on the reset button `X` it still empties the input value. 
[Video recording](https://screenshot.click/22-32-eld40-kvp2w.mp4)

assets/main-search.js
```diff
@@ -5,37 +5,39 @@ class MainSearch extends SearchForm {
...
     let allSearchForms = [];
     this.allSearchInputs.forEach(input => allSearchForms.push(input.form))
-    if (allSearchForms.length > 1) {
-      allSearchForms.forEach(form =>
-        form.addEventListener('reset', this.onFormReset.bind(this))
-      );
+    this.input.addEventListener("focus", this.onInputFocus.bind(this));
+    if (allSearchForms.length < 2) {
+      return;
     }
```
ludoboludo, line 14:
```suggestion
    if (allSearchForms.length < 2) return;
```

assets/search-form.js
```diff
@@ -3,15 +3,20 @@ class SearchForm extends HTMLElement {
...
-    const form = this.querySelector('form.search');
-    form.addEventListener('reset', this.onFormReset.bind(this));
+    if (this.input) {
+      this.input.form.addEventListener('reset', this.onFormReset.bind(this));
+    }
+  }
+
+  shouldResetForm() {
+    return !document.querySelector('[aria-selected="true"] a')
   }
```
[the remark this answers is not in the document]
FCalabria: I followed the pattern for the search button, that also does not submit when an element in the predictive search is active. Maybe @melissaperreault knows why this solution was adopted.
I don't know if it will be a lot of changes for one PR, but I could remove the `active` state and reset the list when the user tabs around, although that does not solve all the use cases because they could tab first and use the cursor later 🤔

sections/header.liquid
```diff
@@ -318,8 +318,9 @@
             <div class="search-modal__content{% if settings.inputs_shadow_vertical_offset != 0 and settings.inputs_shadow_vertical_offset < 0 %} search-modal__content-top{% else %} search-modal__content-bottom{% endif %}" tabindex="-1">
               {%- if settings.predictive_search_enabled -%}
                 <predictive-search class="search-modal__form" data-loading-text="{{ 'accessibility.loading' | t }}">
-              {%- endif -%}
+              {%- else -%}
                 <search-form>
```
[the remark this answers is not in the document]
FCalabria: Nice catch! Although it was easier to fix: I was missing a class and an unclosed html tag 🤦‍♀️

melissaperreault (changes requested):
Hey team! Thanks for the hard work! I think we have the opportunity to improve the buyer experience still before shipping this change!

- [x] When clearing the query, we should remain in the "actively typing" more where the cursor is still in action and waiting for the next letter to be pressed. This would avoid buyer to continuously needing to reveal the keyboard every time they made a mistake or change their query term. [Video reference](https://screenshot.click/25-30-g0kfc-vc5g6.mp4)
- [x] The feedback as come in a little later but we are looking at differentiating the clear icon from the close icon. Stay tuned!

I'll continue to review and reply where you need more attention, thanks!

assets/component-search.css
```diff
@@ -1,32 +1,70 @@
 .search__input.field__input {
```
sofiamatulis, line 1: When tabbing the outline is cut off at the bottom at these 2 places:

https://screenshot.click/06-01-yjs3x-tc9kq.png

https://screenshot.click/06-01-3n5yb-rul2p.png

assets/component-search.css
```diff
@@ -1,32 +1,70 @@
...
-  padding-right: 5rem;
+  padding-right: 9.8rem;
 }
 
 .search__button {
```
sofiamatulis, line 5: After clearing the search term, the search results dont stop and keep loading:

https://screenshot.click/06-03-ykpwy-cj92h.mp4

Is this expected? Shouldnt it be paused?

config/settings_data.json
```diff
@@ -133,9 +133,7 @@
               }
             }
           },
-          "block_order": [
```
sofiamatulis, line 136: Were these changes intended? 👀 Can we revert them?

sofiamatulis (commented):
This looks great :) Added a few small comments and will do a final pass in the code

assets/component-search.css
FCalabria: Nice suggestion!

config/settings_data.json
```diff
@@ -168,11 +170,15 @@
               "type": "text",
               "settings": {
                 "heading": "Our mission",
-                "subtext": "<p>Share contact information, store details, and brand content with your customers.</p>"
+                "subtext": "<p>Share contact information, store details, and brand content with your customers.<\/p>"
```
ludoboludo, line 173: we should revert this. You might have a formatter applied on save  🤔

sections/header.liquid
```diff
@@ -312,51 +319,60 @@
...
+                          aria-autocomplete="list"
+                          autocorrect="off"
+                          autocomplete="off"
+                          autocapitalize="off"
+                          spellcheck="false"
+                        {%- endif -%}
+                      >
+                      <label class="field__label" for="Search-In-Modal-1">{{ 'general.search.search' | t }}</label>
+                      <input type="hidden" name="options[prefix]" value="last">
+                      <button type="reset" class="reset__button field__button {% if search.terms == blank %} hidden{% endif %}" aria-label="{{ 'general.search.reset' | t }}">
```
ludoboludo, line 348:
```suggestion
                      <button type="reset" class="reset__button field__button{% if search.terms == blank %} hidden{% endif %}" aria-label="{{ 'general.search.reset' | t }}">
```

sections/header.liquid
```diff
@@ -534,52 +550,61 @@
...
+                          aria-autocomplete="list"
+                          autocorrect="off"
+                          autocomplete="off"
+                          autocapitalize="off"
+                          spellcheck="false"
+                        {%- endif -%}
+                      >
+                      <label class="field__label" for="Search-In-Modal">{{ 'general.search.search' | t }}</label>
+                      <input type="hidden" name="options[prefix]" value="last">
+                      <button type="reset" class="reset__button field__button {% if search.terms == blank %} hidden{% endif %}" aria-label="{{ 'general.search.reset' | t }}">
```
ludoboludo, line 579:
```suggestion
                      <button type="reset" class="reset__button field__button{% if search.terms == blank %} hidden{% endif %}" aria-label="{{ 'general.search.reset' | t }}">
```

ludoboludo (commented):
Looking good 👍 ill test it on different browser and devices later today 👍

melissaperreault (dismissed):
Thank you Paqui for the iteration! 🎉 

I am testing out and seems to work well for me!
I am just experience some difference for that following behaviour:

_When focusing the main search in mobile, it automatically scrolls to the top of the page_

- [ ] It's not always scrolling up for me on my Android Galaxy Samsung A70. (_Chrome version 108.0.5359.79, I updated my version and it's still happening_ ) There is an attempt at scrolling I can perceive but it seems like there is something preventing from scrolling further to the top. [Visual reference](https://screenshot.click/07-44-dgz9t-t77lg.png) of where it stops as I interact with the component 

Although, I am still able to search so for me it's more to look out for refinement for various device type. Otherwise UX wise it works as expected for the rest, great work team! 🎉

assets/main-search.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  keepInSync(value, target) {
+    this.allSearchInputs.forEach(input => {
+      if (input !== target) {
+        input.value = value;
+      }
+    });
+  }
```
ludoboludo, line 46: One small thing with the sync but I don't think of it as a blocker: [video](https://screenshot.click/12-01-i30oe-hbv29.mp4)

ludoboludo (dismissed):
Just shared something I noticed but otherwise works as I'd expect.

assets/main-search.js
FCalabria: I did some changes to fix that, take a look at my last commit and tell me what do you think!

sofiamatulis (approved):
Looks good 🎉 Thank you for these improvements 🚢

Issue #2138, opened 2022-11-21
Product form error with specific variant label


kmeleta:
There appears to be some issue when submitting the product form's formdata (upon adding to cart) when a product variant option using the label "format". The response errors and noticeably prevents the cart notification from displaying and the cart icon from updating. The error does not appear to prevent the item from actually being added to the cart however.

"Format", "formatt", and other lowercase labels do not cause the issue. I couldn't tell exactly what the issue is, but I've got to assume it has something to do with a reserved term.

Was able to reproduce in the test store https://screenshot.click/format-variant-error.mp4

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 7.0.0

PR #2139, opened 2022-11-22
Move buy-buttons into snippet.

### LINKED ISSUE #2043: Move product blocks into snippets
Most product section blocks are at least reused in featured product. It would be ideal if these components all shared a single source of truth. This ensures anti-fragility and allow changes to these features to automatically propagate to the other instances. Moving more of these out of the section code will also generally make that code easier to digest and work in, speeding up future iterations.

Media gallery and modal have already been split out into snippets, but other smaller components could be considered for this treatment as well. Some make a stronger case than others. For example, product description is used in main and featured product but is only a few lines long. The share button popup is used in main product, featured product, and on article pages. It's not a massive amount of code but it's not completely insignificant.

**Components included**
- ~Media gallery~
- ~Media modal~
- Variant pickers
- Buy buttons
- Share button
- Icon with text

**Other considerations**
- If needed `section.settings` should be accessible within a snippet file, but if reasonable to do so, it might be better to pass the setting value to the snippet as a prop
- Ensure `block.shopify_attributes` and `block.settings` get passed to the snippets
  - We can consider having `block` be a prop of the snippet but, like `section.settings` it might be a better pattern to pass these setting values as individual props of the snippet
  - Share button example: `{{ label }}` instead of `{{ block.settings.share_label }}`
- Ensure snippets covers any exceptions between instances of the duplicated block components =
  - As long as _most_ of the code is the same we should be able to account for any differences by using props
  - Share button example: The urls generated for articles are different than products so it might be cleaner to generate the share link urls outside the snippet and pass it in as a prop, rather than creating very specific conditional decisions within the snippet.

eugenekasimov:
### PR Summary: 

I moved buy-buttons block into snippet. 

Fixes #2043.

I moved the entire buy-button's code into its own file to use it as a snippet in different places rather than duplicate code. Since we don't want to render the `pick-up availability` on the Featured-product page I created a logic that depends on the value of the optional prop `show-pickup-availability`.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

There is no visual impact. 

### Testing steps/scenarios

- [ ] Go to the main-product page and make sure the buy-buttons work as expected
- [ ] Add the featured-product section and make sure the buy-buttons works as expected. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/137467953174/editor)

### Checklist


snippets/buy-buttons.liquid
```diff
@@ -0,0 +1,122 @@
...
+              {{ 'products.product.pickup_availability.refresh' | t }}
+            </button>
+          </div>
+        </pickup-availability-preview>
+      </template>
+    </pickup-availability>
+  {%- endif -%}
+</div>
+
+<script src="{{ 'pickup-availability.js' | asset_url }}" defer="defer"></script>
```
ludoboludo, line 122: This should be in the `{%- if show_pickup_availability -%}` otherwise we're loading it even when it's not needed.

sections/main-product.liquid
```diff
@@ -382,101 +382,7 @@
...
-                          {{ 'products.product.pickup_availability.refresh' | t }}
-                        </button>
-                      </div>
-                    </pickup-availability-preview>
-                  </template>
-                </pickup-availability>
-              </div>
-
-              <script src="{{ 'pickup-availability.js' | asset_url }}" defer="defer"></script>
+              {%- render 'buy-buttons', block: block, product: product, product_form_id: product_form_id, show_pickup_availability: true -%}
```
ludoboludo, line 385: It's not working for me on the `Puppy` product. I haven't seen why so far but trying to check. Ill update the comment when I do 👍

ludoboludo: This is the error by the way. I've seen it this week for another reason somewhere 🤔 But can't remember where. 
<details>
<summary>Screenshot</summary>


</details>

ludoboludo: Oh I remember why 😅 It's not related to your PR but the varian option name `format` which cause issue and was brought up in [another issue](https://github.com/Shopify/dawn/issues/2138) 👍

snippets/buy-buttons.liquid
eugenekasimov: good catch Ludo, thanks.

kmeleta (commented):
I think this looks good as is. Tentatively approved. I just want to take a little more time to test some other dynamic checkout button cases and things to make sure everything still behaves as expected.

> I had to add a minor logic to the snippet because it's used slightly different on different pages.

It'd be helpful if you outlined what these differences were in the PR description so we know what to pay closer attention to.

PR #1912, opened 2022-08-12
Fix popup cart notification overlap in the latest Safari version

ludoboludo:
**PR Summary:** 

Fixing an issue where the popup notification for the cart, once you've added an item to your cart was showing up behind the product image and product information. 

**Why are these changes introduced?**

Fixes https://github.com/Shopify/dawn/issues/1911

**What approach did you take?**

Removed the negative `z-index` on the `.cart-notification` element and added a `z-index: 1;` on the `.header-wrapper`. 

**Other considerations**

**Testing steps/scenarios**

- [ ] Test in Safari on version 15.6 and also older version
- [ ] Make sure it still looks fine in other browsers
- [ ] Test with different header layout (mega menu, drawer on mobile, etc)
- [ ] Test with global settings applied
- [ ] Test how the popup notification overlaps with other sections

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=128483754006)
- [Editor](https://os2-demo.myshopify.com/admin/themes/128483754006/editor)

**Checklist**
- [ ] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [ ] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2147, opened 2022-11-28
Fix mega-menu vertical shadow issue

### LINKED ISSUE #2067: [Header] Mega menu shadow on top of header

If you're setting a vertical top shadow on the popup and have mega menu enabled, the shadow will be on top of the header

The vertical top shadow should be going under the header/hidden

- Dawn Version: 7.0.0

**Possible solution**
Look into using some pseudo elements like done for other components

eugenekasimov:
### PR Summary: 

This PR fixes a bug when shadow of mega-menu overlaps the main menu.

Fixes #2067 .

1. I rolled back some changes that were made in the PR #1912 

- retuned back `z-index: -1` to `.cart-notification`
- removed `z-index: 1` from `.header-wrapper`

2.  I added `z-index: -1` to `.global-settings-popup` which is basically canceling the changes in the PR #2034 

3. The previous changes fix the issue of when shadow of mega-menu overlaps the main menu  👍 but add back issues in Safari when sticky media/product info overlaps mega-menu or cart notification. I added `position: sticky;` to `.section-header` and it fixed the Safari issues.  (tested in Safari 13.1, 14.1, 15.6 and 16.1)

### Other considerations

More tests need  to be done to make sure that the current decision is safe.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=137486893078)
- [Editor](https://os2-demo.myshopify.com/admin/themes/137486893078/editor)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -89,7 +89,7 @@
   }
 
   .section-header {
-    transform: translate3d(0,0,0); /* This is for fixing a Safari z-index issue. PR #2147 */
+    position: sticky; /* This is for fixing a Safari z-index issue. PR #2147 */
     margin-bottom: {{ section.settings.margin_bottom | times: 0.75 | round: 0 }}px;
   }
 


assets/base.css
```diff
@@ -2216,6 +2216,7 @@ input[type='checkbox'] {
 
 /* section-header */
 #shopify-section-header {
+  position: sticky;
```
ludoboludo, line 2219: I think it might be better to apply it on `.section-header` 🤔 so it's over specific. But otherwise seems to do the trick 👍

ludoboludo: Because it's one of the first thing to load on the page. It's considered critical styling so it's faster this way. Since it's right there vs doing a network request for a `.css` file and waiting for it to be downloaded and parsed which can depend on the quality of your connection. ⚡

ludoboludo (approved):

kmeleta (commented):
This does seem to work. Great job digging for a solution 👍  I'd just like to understand the safari "logic" going on here more before I approve. I assume this works by normalizing (safarizing?) the header stacking with the main-product sticky column stacking.

I notice we also need to have the position: sticky set on .section-header even when the sticky header is inactive, which makes me uncomfortable, but I haven't noticed any unintended side effects from it. I just want to be certain we're not introducing any bugs here so I want to take a little more time to explore before approving.

sections/header.liquid
```diff
@@ -89,6 +89,7 @@
   }
 
   .section-header {
+    position: sticky;
```
kmeleta, line 92: Sorry Eugene I haven't had to time to fully test/think on this just yet, but wanted to make sure I remembered to mention that we should definitely add a comment to indicate this is for fixing a safari z-index issue. Maybe include the PR number #2147 in the comment as well

sections/header.liquid
```diff
@@ -89,6 +89,7 @@
   }
 
   .section-header {
+    position: sticky; /* This is for fixing a Safari z-index issue. PR #2147 */
```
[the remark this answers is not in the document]
kmeleta: > I was applying too specifically to elements vs the general header container

^ exactly. Got me too. But because Eugene found that putting the position sticky there worked, it got me trying this on the same element.

sections/header.liquid
```diff
@@ -89,6 +89,7 @@
   }
 
   .section-header {
+    transform: translate3d(0,0,0); /* This is for fixing a Safari z-index issue. PR #2147 */
```
ludoboludo, line 92: This actually brings in a new issue in Safari 😓 
So it seems like the original approach might be better 🤔 
[video](https://screenshot.click/15-39-nzdoj-o1bu3.mp4)

ludoboludo (changes requested):

sections/header.liquid
eugenekasimov: Yeah, it's interesting I don't see this issue on my end but there is another one you didn't mentioned that I can see, the sticky header now works always (on scroll up and down) rather than works on scroll up only 😅

Huge thanks to Ludo for catching it!

PR #514, opened 2021-08-30
Add featured predictive search

martinamarien:
**Why are these changes introduced?**
This PR adds the featured product feature. 

Fixes #39 

**What approach did you take?**
This approach uses the new [Predictive search endpoint](https://shopify.dev/api/ajax/reference/predictive-search#get-search-suggest) to request a predictive search section using the [Predictive search Liquid object](https://shopify.dev/api/liquid/objects/predictive-search). This allows us to avoid JS templating.

**Other considerations**
N/A

**Testing** 
- [ ] Header: Index, Customer, Search , Product templates 
- [ ] Search Page: Search input 
- [ ] Arrows to change selection 
- [ ] Enter 
  * No item selected via arrows --> Search page 
  * Item selected via arrows --> Product page 
  * "Search for" selected via arrows --> Search page 
- [ ] Focus trapped 
- [ ] Sticky header

**Demo links**

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=126297047062)
- [Editor](https://os2-demo.myshopify.com/admin/themes/126297047062/editor)

**Checklist**
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [x] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [x] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [x] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2192, opened 2023-01-04
Adapt predictive search layout to new api

FCalabria:
### PR Summary: 

Update the predictive search component to take advantage of the new default values in the API and support all resource types

https://github.com/Shopify/discovery/issues/6681

New default values on resource[types] https://github.com/Shopify/discovery/issues/6617

The two column structure has been done with flex, using the `flex-direction` property to change between the mobile and desktop structure, with the appropiate changes in `flex-grow` for the children to fill the space when needed.

The trickiest part is the "Pages" section, that is placed in a different column depending on the screen size. I solved this duplicating the html, adding some conditional classes to the wrapper and using them with media queries to hide the page section that was not needed.

Very simplified, the final html structure looks like this:

```html
<div id="predictive-search-results">
  <div class="predictive-search__results-groups-wrapper"> <!-- This div gets conditional .predictive-search__results-groups-wrapper--no-products and .predictive-search__results-groups-wrapper--no-suggestions -->
    <div class="predictive-search__result-group"> <!-- This div is not present if it has no content (it could be present but with hidden pages on mobile) -->
      <div> <!-- Not present if it has no content -->
        <h2 id="predictive-search-queries" class="predictive-search__heading">
          Suggestions
        </h2>
        <ul
          id="predictive-search-results-queries-list"
          class="predictive-search__results-list"
        >
          <li class="predictive-search__list-item" aria-selected="false">
            <a class="predictive-search__item"> </a>
          </li>
        </ul>
      </div>
      <div class="predictive-search__pages-wrapper"> <!-- Not present if it has no content (it could be present but with hidden pages on mobile)-->
        <h2 id="predictive-search-pages" class="predictive-search__heading">
          Pages
        </h2>
        <!-- Similar ul > li, changing the ids -->
      </div>
    </div>
    <div class="predictive-search__result-group"> <!-- This div is not present if it has no content (it could be present but with hidden pages on desktop) -->
      <div> <!-- Not present if it has no content -->
        <h2 id="predictive-search-products" class="predictive-search__heading">
          Products
        </h2>
        <!-- Similar ul > li, changing the ids -->
      </div>
      <div class="predictive-search__pages-wrapper"> <!-- Not present if it has no content (it could be present but with hidden pages on desktop)-->
        <h2 id="predictive-search-pages" class="predictive-search__heading">
          Pages
        </h2>
        <!-- Similar ul > li, changing the ids -->
      </div>
    </div>
  </div>
  <div class="predictive-search__loading-state" aria-hidden="true"></div> <!-- Loading spinner -->

  <div id="predictive-search-option-search-keywords">
    <button class="predictive-search__item predictive-search__item--term">
      Search for “c”
    </button>
  </div>
</div>
```

Users will start to get more type of results on the predictive search. They will also see the max number of results increased to 10

### Testing steps/scenarios

[Figma](https://www.figma.com/file/HFKhDU5jNbnpR7U0JnsAur/Query-Suggestions-in-Storefront-Search?node-id=1169%3A3838&t=OEHoeSfG1tREoxFm-0) for reference

- [ ] The predictive search dropdown can show all possible resource types (even the ones that are not queried by default)
- [ ] The predictive search dropdown looks as expected even when some of the resource types are not populated
- [ ] The predictive search dropdown is still accessible

### Demo links

- [Store](https://2dpfl6jn8thn0b8m-45840990230.shopifypreview.com)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139215274006/editor)

### Checklist


sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+        <div>
+          <h2 id="predictive-search-queries" class="predictive-search__heading text-body caption-with-letter-spacing">
+            {{ 'templates.search.suggestions' | t }}
+            <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+              <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+            </svg>
+          </h2>
+          <ul id="predictive-search-results-queries-list" class="predictive-search__results-list list-unstyled" role="listbox" aria-labelledby="predictive-search-queries">
+            {%- for query in predictive_search.resources.queries -%}
+              <li id="predictive-search-option-A{{ forloop.index }}" class="predictive-search__list-item">
```
NathanPJF, line 25: You can get duplicated `id`'s with this as you use the same formula for the Pages section below (`predictive_search.resources.pages`). You could avoid this by appending the resource name with the "A" and "B"

```
{%- for query in predictive_search.resources.queries -%}
  <li id="predictive-search-option-{{ forloop.index }}-queries-A">
{% endfor %}
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                <a href="{{ query.url }}" class="predictive-search__item link link--text" tabindex="-1" role="option" aria-selected="false">
+                  <div class="predictive-search__item-content predictive-search__item-content--centered">
+                    <h3 class="predictive-search__item-heading predictive-search__item-query-result h5">{{ query.styled_text }}</h3>
+                  </div>
+                </a>
+              </li>
+            {%- endfor -%}
+            {%- for collection in predictive_search.resources.collections -%}
+              <li id="predictive-search-option-A{{ forloop.index | plus: predictive_search.resources.queries.size }}" class="predictive-search__list-item">
```
NathanPJF, line 34: Related to my comment above about the duplicated IDs: you could avoid the `plus:` here by adding the resource name. I don't think you actually need `plus` as you are relying on that number in your JS.

```suggestion
              <li id="predictive-search-option-{{ forloop.index }}-collection-A" class="predictive-search__list-item">
```

NathanPJF: Same comment everywhere else `predictive-search-option-A{{ forloop.index` is used

NathanPJF (changes requested):

NathanPJF (changes requested):
Requesting changes for the aria-hidden text.

FYI: I made a codepen where I used `grid-template-area` to manage the display in order to avoid duplicating the HTML. However, I realized it would make the JS a lot more complicated because the order of the DOM nodes couldn't be relied upon to match the visual order. Similar to focus order issues with Grid talked about [here on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/CSS_Grid_Layout_and_Accessibility). So ultimately I think the repeated HTML is fine here.

NathanPJF:
@FCalabria Following up on our chat: Instead of having a new `listbox` on every `<ul>`, I think the "grouped options" pattern described here is a better fit. https://www.w3.org/WAI/ARIA/apg/example-index/listbox/listbox-grouped.html

sections/predictive-search.liquid
```diff
@@ -78,16 +156,41 @@
         <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
       </svg>
     </div>
+
+    <div id="predictive-search-option-search-keywords">
+      <button
```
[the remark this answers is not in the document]
FCalabria: I added the `role` to match the W3C pattern properly, but for the record: what was missing by the screen reader was an `id` 🤦‍♀️

assets/component-predictive-search.css
```diff
@@ -50,8 +65,40 @@ predictive-search[loading] .predictive-search {
   display: block;
 }
 
+.predictive-search__result-group {
+  display: flex;
+  flex-direction: column;
+  flex: 1 1 auto;
+  gap: 3rem;
```
melissaperreault, line 72:
```suggestion
  gap: 2rem;
```
I find the 3rem gap to be very large. Suggesting 2rem to balance white space. Thoughts @al-mcquiston ?

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+        <div>
+          <h2 id="predictive-search-queries" class="predictive-search__heading text-body caption-with-letter-spacing">
+            {{ 'templates.search.suggestions' | t }}
+            <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+              <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+            </svg>
+          </h2>
+          <ul id="predictive-search-results-queries-list" class="predictive-search__results-list list-unstyled" role="group" aria-labelledby="predictive-search-queries">
+            {%- for query in predictive_search.resources.queries -%}
+              <li id="predictive-search-option-query-{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
```
NathanPJF, line 25:
```suggestion
              <li id="predictive-search-option-query-{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false" aria-label="{{ query.text }}">
```

I was testing in Safari + VoiceOver and Mobile Safari + VoiceOver and realized that it was interpreting the styled text as two separate enteries. It would read `p<mark>rime</mark>` as "P Rime" instead of "Prime".

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                <a href="{{ query.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                  <div class="predictive-search__item-content predictive-search__item-content--centered">
+                    <h3 class="predictive-search__item-heading predictive-search__item-query-result h5">{{ query.styled_text }}</h3>
```
NathanPJF, line 28:
```suggestion
                    <p class="predictive-search__item-heading predictive-search__item-query-result h5">{{ query.styled_text }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                  </div>
+                </a>
+              </li>
+            {%- endfor -%}
+            {%- for collection in predictive_search.resources.collections -%}
+              <li id="predictive-search-option-collection-{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
+                <a href="{{ collection.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                  <div class="predictive-search__item-content predictive-search__item-content--centered">
+                    <h3 class="predictive-search__item-heading h5">{{ collection.title }}</h3>
```
NathanPJF, line 37:
```suggestion
                    <p class="predictive-search__item-heading h5">{{ collection.title }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+            <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+              <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+            </svg>
+          </h2>
+          <ul id="predictive-search-results-pages-list" class="predictive-search__results-list list-unstyled" role="group" aria-labelledby="predictive-search-pages">
+            {%- for page in predictive_search.resources.pages -%}
+              <li id="predictive-search-option-page-A{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
+                <a href="{{ page.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                  <div class="predictive-search__item-content predictive-search__item-content--centered">
+                    <h3 class="predictive-search__item-heading h5">{{ page.title }}</h3>
```
NathanPJF, line 58:
```suggestion
                    <p class="predictive-search__item-heading h5">{{ page.title }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                  </div>
+                </a>
+              </li>
+            {%- endfor -%}
+            {%- for article in predictive_search.resources.articles -%}
+              <li id="predictive-search-option-article-A{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
+                <a href="{{ article.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                  <div class="predictive-search__item-content predictive-search__item-content--centered">
+                    <h3 class="predictive-search__item-heading h5">{{ article.title }}</h3>
```
NathanPJF, line 67:
```suggestion
                    <p class="predictive-search__item-heading h5">{{ article.title }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                      {%- if product.featured_media != blank -%}
+                        <img class="predictive-search__image" src="{{ product.featured_media | image_url: width: 150 }}" alt="{{ product.featured_media.alt }}" width="50" height="{{ 50 | divided_by: product.featured_media.preview_image.aspect_ratio }}">
+                      {%- endif -%}
+                      <div class="predictive-search__item-content{% unless settings.predictive_search_show_vendor or settings.predictive_search_show_price %} predictive-search__item-content--centered{% endunless %}">
+                        {%- if settings.predictive_search_show_vendor -%}
+                          <span class="visually-hidden">{{ 'accessibility.vendor' | t }}</span>
+                          <div class="predictive-search__item-vendor caption-with-letter-spacing">{{ product.vendor }}</div>
+                        {%- endif -%}
+                        <h3 class="predictive-search__item-heading h5">{{ product.title }}
+                        </h3>
```
NathanPJF, line 101:
```suggestion
                        <p class="predictive-search__item-heading h5">{{ product.title }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                <svg aria-hidden="true" focusable="false" role="presentation" class="spinner" viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg">
+                  <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+                </svg>
+              </h2>
+              <ul id="predictive-search-results-pages-list" class="predictive-search__results-list list-unstyled" role="group" aria-labelledby="predictive-search-pages">
+                {%- for page in predictive_search.resources.pages -%}
+                  <li id="predictive-search-option-page-B{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
+                    <a href="{{ page.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                      <div class="predictive-search__item-content predictive-search__item-content--centered">
+                        <h3 class="predictive-search__item-heading h5">{{ page.title }}</h3>
```
NathanPJF, line 125:
```suggestion
                        <p class="predictive-search__item-heading h5">{{ page.title }}</p>
```

sections/predictive-search.liquid
```diff
@@ -1,71 +1,149 @@
...
+                      </div>
+                    </a>
+                  </li>
+                {%- endfor -%}
+                {%- for article in predictive_search.resources.articles -%}
+                  <li id="predictive-search-option-article-B{{ forloop.index }}" class="predictive-search__list-item" role="option" aria-selected="false">
+                    <a href="{{ article.url }}" class="predictive-search__item link link--text" tabindex="-1">
+                      <div class="predictive-search__item-content predictive-search__item-content--centered">
+                        <h3 class="predictive-search__item-heading h5">{{ article.title }}</h3>
```
NathanPJF, line 134:
```suggestion
                        <p class="predictive-search__item-heading h5">{{ article.title }}</p>
```

NathanPJF (approved):
Nit about the use of `h3` for the list options. These can be safely changed to `p` tags as the `.h5` class overwrites all the styling. 

Tested on VoiceOver in Mobile Safari, Safari, and Chrome. And JAWS in Edge.  The arrow keys navigation isn't picked up in Safari unfortunately, but I can walkthrough the page correctly with a virtual cursor (hence the addition of an `aria-label` on the queries list)

NathanPJF:
@melissaperreault you wrote

> not being convinced we need 3 times the loading spinner as the query is typed

I had noticed the pattern for loading is a bit strange as well. Though something to note is that there isn't always 3 loaders. The number of loaders depends on how many heading (suggestions, pages, products) were on the page *before* the next search. 

📹 [In this video](https://screenshot.click/05-42-wanc9-5avxp.mp4), I slow down the network connection, and the following can be seen:
- I type "L" and there is a loader just in the middle of the page. I get suggestions for both suggestions and products
- I type "Z" and now there are 2 loaders... but I only get products back
- I type "V", I only get 1 loader, but then 3 resource types back (suggestions, pages, products)

The loading indication could be changed. This approach may have made sense when 'Products' was the only possible heading you'd see, but now the number of resource types you get back is dynamic. Perhaps the loaders next to the resources headings should be removed completely - so no indication that anything is happening, the next content will just pop into existence. Alternatively: maybe the spinner get removed, but all the content gets dimmed while we wait for the new results.

bakura10:
Maestrooo developers here. Just found this. Do you have any documentation? As this feature may have visual implication on our themes I would like to start working on this as soon as possible. I have tried to add a new "query" resource type but I have a 422 error.

As always, we would appreciate if we (theme dev) could be also added in the loop for important upgrades like this, it would also eventually allow us to provide feedback.

sections/predictive-search.liquid
kmeleta: I wouldn't say I disagree with this suggestion, but just FYI it looks like during a11y review of the original implementation of predictive search it was specifically [recommended](https://github.com/Shopify/dawn/pull/514#discussion_r698783581) that these product titles use h3. Though I could maybe see that being more relevant if the vendor or price is also showing. If we're confident about this change though, feel free.

sections/predictive-search.liquid
NathanPJF: Good point 👍🏻

FCalabria:
> Related, but separate issue to this PR: there's a larger issue with how loading is already being done IMO. The contents of the listbox aren't cleared between searches. This makes sense for as you're typing, but not so much when the search box is emptied; as a result, you will briefly see the results of the _previous_ search when you start a new search. [Video example](https://screenshot.click/05-59-sfyuo-dx0mx.mp4) where I show typing "V", clearing results, then type "T". I see the results for "V" as my new results are being loaded. This issue exists in the search box that's in production.

@NathanPJF I have created [this issue](https://github.com/Shopify/dawn/issues/2199) to fix this problem later, since it is a current bug and not caused by my changes. Take a look in case you want to suggest any changes.

About the loading, I talked with @al-mcquiston and we decided to remove the loading icons on the headers, since that seem to be a common pattern in other websites.

kmeleta (commented):
I'm still doing some testing (a11y in particular) but overall I think this is looking good. I'm raising a couple UX questions and some code formatting stuff, but I haven't come across anything too alarming or blocking yet.

sections/predictive-search.liquid
svinkle: > do you recall why you recommended an h3 in this past commit?

I do not. As mentioned, the heading structure is not conveyed here. Swap for a generic container.

sections/header.liquid
```diff
@@ -571,8 +571,8 @@
                         {%- if settings.predictive_search_enabled -%}
```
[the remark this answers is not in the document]
FCalabria: That's actually one of the reasons I changed the fetch query to not set any parameter that is not mandatory (and we made some of them optional with the API changes). I'm not sure about dates and priorities, but the idea is that in the future merchants could decide what resource types they want to show if the theme is not specifying any.

assets/component-predictive-search.css
al-mcquiston: @melissaperreault totally agree. myself and @FCalabria chatted about this and this should be updated to your suggestion

assets/component-predictive-search.css
```diff
@@ -50,8 +65,40 @@ predictive-search[loading] .predictive-search {
...
+  .predictive-search__results-groups-wrapper--no-products
+    .predictive-search__result-group:first-child,
+  .predictive-search__result-group:only-child {
+    flex-grow: 1;
+  }
+}
+
 .predictive-search__heading {
-  border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.08);
+  border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.2);
```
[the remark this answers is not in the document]
melissaperreault: +1 The opacity should not change here and remain at 0.08.

- [ ] Another point I just noticed; I am not sure what happened but [we lost the type specific dividers here](https://screenshot.click/10-30-78wss-mjawd.png). When I initially reviewed, [they were in place properly](https://screenshot.click/10-54-0urbl-uzviv.png), now they just blend from left to right. The initial layout is best to maintain hierarchy between the results and the CTA at the bottom of the box.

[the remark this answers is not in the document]
kmeleta: Looks like the borders are extending to the left of the headings now. I don't believe I've see any discussion about this here or in the figma so I assume it's unintentional.

melissaperreault:
Adding last piece of observations!

### Transition
_*This feedback is not launch blocking but something I'd like us to improve._

> Perhaps the loaders next to the resources headings should be removed completely - so no indication that anything is happening, the next content will just pop into existence.

Yeah, this is the first impression that I would hope we could spend more time in a follow up task to improve the perceived performance. I am going to throw a few buzz word with hope that it helps convey in which direction we can explore: it could be smoother, snappier; buyers don't need to hold on their toes to only find out there is none of X result type. I am not sure what is the common pattern but I think we can explore further.

Would a fade in animation instead help? Is 3 relevant because things are very likely to load at different times? Could customers have the perception that things all load together, all load separate, populates as it loads (e.g. skeleton) . I just wonder if we can build an opinion of what is the most ideal experience for the buyers.


> About the loading, I talked with @al-mcquiston and we decided to remove the loading icons on the headers, since that seem to be a common pattern in other websites

We need to keep at least 1 spinner though. This will help the customer waiting as they type their query on slower connection to not be holding their breath with an empty screen.

sections/main-search.liquid
```diff
@@ -20,7 +20,7 @@
 
   .template-search__search {
```
kmeleta, line 21: Group headings are centered on the search page.

al-mcquiston:
>

sections/header.liquid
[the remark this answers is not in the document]
kmeleta: Gotcha, thanks for the context 👍

assets/component-predictive-search.css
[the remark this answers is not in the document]
NathanPJF: Regression fixed in latest commit

sections/predictive-search.liquid
```diff
@@ -15,7 +15,7 @@
       {%- if predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 -%}
         <div>
           <h2 id="predictive-search-queries" class="predictive-search__heading text-body caption-with-letter-spacing">
-            {{ 'templates.search.suggestions' | t }}
+            {{- 'templates.search.suggestions' | t -}}
```
NathanPJF, line 18: While reviewing the output, I noticed that the whitespaces were being rendered in the inspector. This doesn't have an impact on the UI really, but it makes the output more inline with what we want.

NathanPJF (approved):
My comments about the output in VoiceOver and JAWS have been addressed, and the feature works as expected. Nice work @FCalabria !

melissaperreault (approved):
Thank you! Looks good to me!

Tiny detail for when we tackle loading states, we need to consider this use case too for when no results are found and are only left with the CTA. ([video example](https://screenshot.click/11-17-8jsjy-261ex.mp4))

PR #2195, opened 2023-01-04
Add drawer menu desktop

### LINKED ISSUE #1455: [Header] Collapse links into a hamburger menu on desktop

eugenekasimov:
### PR Summary: 

This PR adds a new desktop menu type "Drawer".

Key updates:
 - the drawer menu has a full width of the page. The left and right paddings are `32px`. 
 - the header's paddings for tablet have been decreased from `50px` to `32px`. The header is not perfectly aligned with a content anymore. 
 - the `closing X` button, `nav` and `log in` icon have been aligned for desktop and tablet layouts. 
 - the `nav` and` log in` icon have been aligned for mobile layout. 

Fixes #1455.

### Decision log

|---|---|---|---|---|
| 1 |  Reuse existing drawer-menu that is used for mobile/tablet layouts.  |  Create a new one.  |  I don't want to add the same code that already exists.  |  The current changes in code to reuse the existing drawer-menu may look not very clear because the initial approach doesn't seem to count it could be extended.  |
| 2 |  There was `margin-left: -0.75rem` for `header__heading-link` and it's applied all the time no matter where the `header__heading-link` was positioned . I made it more targeted for left positions only.  |  Leave it as is.  |  It helps avoid a minor shifting of `header__heading-link` when it's positioned in the center.  |  I don't see any of it.  |
|  3  |  I added an event listener to the window for the open drawer to update `width` and `height` in real time.  |  -  |  The size of the semi-transparent background which is present while the drawer is open is calculated based on the current `height` and `width` of the window. I had to deal with resizing a browser.   |  We need to listen for any changes of `height` but for the `width` we need to know when it passes a threshold of `990px` back-and-forward only. Since I couldn't find any good way to listen for `height` changes only the event listener listens for changes in both directions. But I believe that it's a very edge case itself and it should not cost a problem.  |

<details>
<summary>Before</summary>
Screenshot 2023-01-05 at 2 30 37 PM
</details>
<details>
<summary>After</summary>
Screenshot 2023-01-31 at 2 38 13 PM
Screenshot 2023-01-31 at 2 38 23 PM
</details>

### Testing steps/scenarios

- [ ] Choose a new Desktop menu type - **Drawer**. Save the settings. Test it.
- [ ] Test it with different Desktop logo positions. Each time when you set a new logo position don't forget to save settings. Also keep in mind that the Desktop logo position `Top left` and `Middle left` are supposed to look the same.
- [ ] While you test previous steps try to change the width of your browser, shrinking it to tablet and mobile sizes. Make sure everything works correctly. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139400445974)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139400445974/editor)

### Checklist


duygukalaycioglu (dismissed):

stufreen (changes requested):
I found a couple small issues and also have a few ideas/questions/suggestions/ Take them all with a grain of salt though as I'm new to the codebase! Let me know if any of them don't make sense or you want to chat about them.

assets/base.css
```diff
@@ -2415,23 +2430,44 @@ input[type='checkbox'] {
...
+    display: block;
+    position: absolute;
+    background: rgba(var(--color-foreground), 0.5);
+  }
+}
+
+@media screen and (min-width: 990px) {
+  summary.header__icon--menu.drawer-open {
+    position: absolute;
+    left: 2rem;
```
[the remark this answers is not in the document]
duygukalaycioglu: Hi @stufreen thanks for noticing that! The latest designs are on the Figma file which has the header not covered, there will be no change to the current drawer menu functionality on tablet.

assets/base.css
```diff
@@ -2415,23 +2430,44 @@ input[type='checkbox'] {
...
-  }
-}
-
 .header__icon--menu[aria-expanded="true"]::before {
   content: "";
-  top: 100%;
+  bottom: 0;
   left: 0;
-  height: calc(var(--viewport-height, 100vh) - (var(--header-bottom-position, 100%)));
+  height: calc(var(--viewport-height, 100vh) - (var(--header-bottom-position, 100%)) - 1px);
```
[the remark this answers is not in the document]
duygukalaycioglu: @eugenekasimov I thought that this was happening on the editor + fixed, let's make sure it is not happening on the store.

assets/base.css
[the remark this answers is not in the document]
duygukalaycioglu: Thank you, I think between these sizes - it still works like desktop, where the X aligns with the nav links instead of the hamburger icon. I think that's acceptable, however we can keep the logo to stay on the left instead of jumping to the middle/center when the desktop logo layout is set to `middle left`. 

https://user-images.githubusercontent.com/88160784/211859297-1f25a412-d122-4539-8f51-f9d20f6aeb70.mov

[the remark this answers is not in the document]
duygukalaycioglu: @eugenekasimov for context – @stufreen and I aligned on leaving this as is. No change needed for the padding for now.

sections/header.liquid
```diff
@@ -106,7 +108,8 @@
   }
 
   @media screen and (min-width: 990px) {
-    .header {
+    .header,
+    summary.header__icon--menu.drawer-open {
```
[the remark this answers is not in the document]
eugenekasimov: Fair enough. It doesn't need it. I'll change it here and in the base.css as well.

sections/header.liquid
```diff
@@ -468,50 +471,52 @@
...
-                    <span {%- if link.current %} class="header__active-menu-item"{% endif %}>{{ link.title | escape }}</span>
-                  </a>
-                {%- endif -%}
-              </li>
-            {%- endfor -%}
-          </ul>
-        </nav>
+        {%- if section.settings.menu_type_desktop != 'drawer' -%}
+          <nav class="header__inline-menu">
+            <ul class="list-menu list-menu--inline" role="list">
```
metamoni, line 476: I know you probably got this from already existing code, but we don't need `role="list"` here, as `ul` already has that [implicit ARIA role](https://www.w3.org/TR/html-aria/), so it's redundant. We have this kind of stuff in the code already, but we shouldn't continue to do it because it's [not good practice](https://www.w3.org/TR/html-aria/#rules-wd).

sections/header.liquid
```diff
@@ -468,50 +471,52 @@
...
+              {%- for link in section.settings.menu.links -%}
+                <li>
+                  {%- if link.links != blank -%}
+                    <header-menu>
+                      <details id="Details-HeaderMenu-{{ forloop.index }}" class="mega-menu">
```
metamoni, line 481: Again, I know is code you've borrowed, but I don't think this is the right usage of the `summary` and `details` tags. These elements are used to [show/hide details about something](https://www.w3schools.com/tags/tag_summary.asp), not for collapsible navigation. This seems to me like an overly-complicated implementation of what should be a nested menu. See a semantic implementation under ["Indicate submenus"](https://www.w3.org/WAI/tutorials/menus/flyout/). With nested navigation the parent can be a button that expands/collapses a _new_ list of link items.

assets/global.js
```diff
@@ -430,13 +435,18 @@ class HeaderDrawer extends MenuDrawer {
...
     summaryElement.setAttribute('aria-expanded', true);
+    window.addEventListener('resize', this.onResize);
     trapFocus(this.mainDetailsToggle, summaryElement);
     document.body.classList.add(`overflow-hidden-${this.dataset.breakpoint}`);
+    summaryElement.classList.add(`drawer-open`);
   }
 
   closeMenuDrawer(event, elementToFocus) {
     super.closeMenuDrawer(event, elementToFocus);
+    console.log(elementToFocus);
```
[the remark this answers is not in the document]
eugenekasimov: oh, good catch Monica 😅

sections/header.liquid
metamoni: Yes, you can do that for your PR if you like, and that'd be great! But we can also deal with it as part of [this issue](https://github.com/Shopify/dawn/issues/2212).

assets/global.js
```diff
@@ -417,6 +417,11 @@ customElements.define('menu-drawer', MenuDrawer);
 class HeaderDrawer extends MenuDrawer {
   constructor() {
     super();
+
+    this.onResize = () => {
+      document.documentElement.style.setProperty('--header-bottom-position', `${parseInt(this.header.getBoundingClientRect().bottom - this.borderOffset)}px`);
+      document.documentElement.style.setProperty('--viewport-height', `${window.innerHeight}px`);
+    };
```
stufreen, line 424: I can see that what we did yesterday didn't work (`onResize() {`) because `this` wasn't bound. However I think it's cleaner to use a class method with an arrow function, which is also valid and will automatically bind `this`:
```js
onResize = () => {
  ...
}
```

assets/global.js
```diff
@@ -430,13 +435,18 @@ class HeaderDrawer extends MenuDrawer {
     });
 
     summaryElement.setAttribute('aria-expanded', true);
+    window.addEventListener('resize', this.onResize);
```
stufreen, line 438: Just summarizing a conversation we had yesterday:
1) `resize` listener can be a bit heavy since it's called repeatedly when you resize a window; _**however**_
2) it's appropriate here because the overlay needs to be resized along with the browser

As you pointed out, there's a bug in production where the overlay gets cut off if the user enlarges the browser window vertically. This will fix it.

assets/global.js
```diff
@@ -417,6 +417,11 @@ customElements.define('menu-drawer', MenuDrawer);
 class HeaderDrawer extends MenuDrawer {
   constructor() {
     super();
+
+    this.onResize = () => {
+      document.documentElement.style.setProperty('--header-bottom-position', `${parseInt(this.header.getBoundingClientRect().bottom - this.borderOffset)}px`);
```
stufreen, line 422: Small suggestion: Check if `this.header` is defined before calling this:
```js
this.header && document.documentElement.style.setProperty('--header-bottom-position', `${parseInt(this.header.getBoundingClientRect().bottom - this.borderOffset)}px`);
```

stufreen (approved):
LGTM with a couple small changes made ✨

assets/global.js
eugenekasimov: Yeah, and I tried even .bind(this) and it didn't work. But I didn't try to use arrow function instead of a normal one. Now this works and look better I agree. Thanks.

stufreen (approved):

melissaperreault (dismissed):
Thank you Eugene! When I save everything works as expected! 🚢 🎉

assets/base.css
```diff
@@ -2432,6 +2454,12 @@ input[type='checkbox'] {
   background: rgba(var(--color-foreground), 0.5);
 }
 
+@media screen and (max-width: 989px) {
```
stufreen, line 2457: Do we have any conventions around mixing max-width and min-width media queries? Generally I would prefer something like this to avoid having to think about media queries in both directions:
```css
menu-drawer ~ .header__icons .header__icon--account {
  display: none;
}

  menu-drawer ~ .header__icons .header__icon--account {
    display: block;
  }
}
```
In other words, use min-width or max-width, but not both.

stufreen (approved):
LGTM with one nit address 🚀

assets/base.css
eugenekasimov: I agree with you. However, unfortunately, we have a 50/50 split of using both min and max -width for media queries in base.css.
In other words there is no sense to change it in this PR. Maybe we should do it within the next tech debt or when we have time for this.

vfonic:
One oddity I noticed about this drawer is that `#menu-drawer` (the drawer itself) is nested within:
header section => `.header-wrapper` => `header` => `nav` => `ul` => `li` => `header-drawer` => `details`

This is causing issues in my case because I want to blur the rest of the page when the drawer is open.

<details>
<summary>Like this</summary>

Screenshot 2023-02-10 at 4 34 20 pm

</details>


Having drawer element nested makes it much more difficult to manage its stacking context (z-index). I see that `cart-drawer` and `quick-add-to-cart` elements are "extracted" to be direct children of `body` tag. I'm not sure how easy would it be to do the same here, but it would allow me to simply add `filter: blur(4px)` just how I did when `cart-drawer` is open. I'm not sure there's a good reason to have two (or more) drawers work in a slightly different manner. I've seen this throughout the theme in several places (close icon, drawers, pub/sub vs directly calling methods, etc.). I've even created a discussion topic about this: https://github.com/Shopify/dawn/discussions/2285

assets/global.js
```diff
@@ -430,14 +430,22 @@ class HeaderDrawer extends MenuDrawer {
     });
 
     summaryElement.setAttribute('aria-expanded', true);
+    window.addEventListener('resize', this.onResize);
     trapFocus(this.mainDetailsToggle, summaryElement);
     document.body.classList.add(`overflow-hidden-${this.dataset.breakpoint}`);
   }
 
   closeMenuDrawer(event, elementToFocus) {
+    if (!elementToFocus) return;
```
ludoboludo, line 439: Curious to know why this was added 🤔

sections/header.liquid
```diff
@@ -137,9 +139,9 @@
   </symbol>
 </svg>
 <{% if section.settings.sticky_header_type != 'none' %}sticky-header data-sticky-type="{{ section.settings.sticky_header_type }}"{% else %}div{% endif %} class="header-wrapper color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} header-wrapper--border-bottom{% endif %}">
-  <header class="header header--{{ section.settings.logo_position }} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
+  <header class="header header--{{ section.settings.logo_position }} {% if section.settings.menu_type_desktop == 'drawer' %}drawer-menu{% endif %} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %} ">
```
ludoboludo, line 142: Nitpick for white space management:
```suggestion
  <header class="header header--{{ section.settings.logo_position }}{% if section.settings.menu_type_desktop == 'drawer' %} drawer-menu{% endif %} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
```

sections/header.liquid
```diff
@@ -137,9 +139,9 @@
...
     {%- if section.settings.menu != blank -%}
-      <header-drawer data-breakpoint="tablet">
+      <header-drawer data-breakpoint="{% if section.settings.menu_type_desktop == 'drawer' %}desktop{% else %}tablet{% endif %}">
```
ludoboludo, line 144: So here to go back to my previous comment about `overflow-hidden-desktop` maybe you can change here the naming to something that doesn't sound specific to desktop only. 
Maybe `all-breakpoints`

sections/header.liquid
```diff
@@ -468,50 +470,52 @@
           </ul>
         </nav>
       {%- else -%}
```
ludoboludo, line 472: Here instead of having another `if` nested, you can change the `else` to an `elsif`
```suggestion
      {%- elsif section.settings.menu_type_desktop != 'drawer' -%}
```

assets/base.css
```diff
@@ -2432,6 +2454,12 @@ input[type='checkbox'] {
...
+  menu-drawer ~ .header__icons .header__icon--account {
+    display: none;
+  }
+}
+
 /* Search */
 menu-drawer + .header__search {
```
ludoboludo, line 2464: I noticed something when the header is set to the `top center` layout. The search icon needs a negative margin like the icon to be vertically aligned with the rest of the content: 
<details>
<summary>Screenshot</summary>


</details>

But that's not something touched in this PR anyway. Ill create an issue for it 👍 
Issue -> https://github.com/Shopify/dawn/issues/2300

assets/base.css
```diff
@@ -244,11 +244,14 @@ html.no-js .no-js-hidden {
 }
 
 .page-width {
-  max-width: var(--page-width);
   margin: 0 auto;
   padding: 0 1.5rem;
 }
 
+.page-width:not(.drawer-menu) {
```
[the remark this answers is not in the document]
ludoboludo: Actually I wonder if this should live in the `header.liquid` file. 
So you could keep it as it was and only remove the page-width's `max-width` when the menu is set to be a drawer AND there is a menu attached 🤔 

Cause if `section.settings.menu == blank` then the header is not limited to the page with: 
<details>
<summary>Screenshot</summary>


</details>

It's a niche case for which the merchant could change their layout style not limit the header to the page width. Though that isn't really clear for them to understand.

ludoboludo (commented):
Looking good 👍 

I have a few nitpick comments. 

Could you rebase from main when you get a chance too, to see how it plays with group section :)

assets/base.css
```diff
@@ -259,6 +262,11 @@ html.no-js .no-js-hidden {
     padding: 0 5rem;
   }
 
+  .header.page-width {
+    padding-left: 3.2rem;
+    padding-right: 3.2rem;
+  }
```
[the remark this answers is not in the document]
eugenekasimov: It was a UI decision and the main point was that we want to keep the hamburger/closing X button aligned with navigation links (which has 3.2rem padding) in the drawer. And we are okay about that not perfect alignment for the tablet that you mentioned.

assets/base.css
```diff
@@ -273,6 +281,11 @@ html.no-js .no-js-hidden {
 }
 
 @media screen and (min-width: 990px) {
+  .header:not(.drawer-menu).page-width {
+    padding-left: 5rem;
+    padding-right: 5rem;
+  }
+
```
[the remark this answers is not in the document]
eugenekasimov: We want to have this 3.2rem padding for tablet as well, that's why I keep it there under `@media screen and (min-width: 750px)`

assets/base.css
```diff
@@ -2432,6 +2454,12 @@ input[type='checkbox'] {
   background: rgba(var(--color-foreground), 0.5);
 }
 
+@media screen and (max-width: 989px) {
+  menu-drawer ~ .header__icons .header__icon--account {
+    display: none;
+  }
+}
```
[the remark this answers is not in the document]
eugenekasimov: That's a good question Ludo. Thanks for brining it up.

@melissaperreault please let us know what you think.

assets/base.css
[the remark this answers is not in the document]
eugenekasimov: Regarding your second comment I think we don't want to let merchants have a full with header if there is no menu selected. We only want to have a full width header when the menu type set to drawer for desktop.

assets/base.css
```diff
@@ -2285,6 +2306,12 @@ input[type='checkbox'] {
       'navigation navigation navigation';
   }
 
+  .header--top-center.drawer-menu {
+    grid-template-areas:
+      '. left-icon heading icons';
```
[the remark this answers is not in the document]
ludoboludo: I don't think so because the `.` is a column itself in which the hamburger menu icon is. And only the search icon is associated to the `grid-area: left-icon;`
<details>
<summary>Screenshot</summary>


</details>

assets/base.css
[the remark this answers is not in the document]
melissaperreault: Oh good catch! 🔥 
Please remove the login in the drawer and keep the login in the header until we hide for mobile. If we do have the space in the header, it's better to prioritize this one on larger screens. 🙏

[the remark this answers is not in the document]
eugenekasimov: @melissaperreault there is one more thing I'd like to clarify. We are removing the login in from the drawer for tablet and desktop, is it right?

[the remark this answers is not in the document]
melissaperreault: Yes! When the `Desktop menu type` = `Drawer` + Customer account is enabled

- Desktop: Account icon is in the header
- Tablet: Account icon is in the header until it reaches the mobile breakpoint
- Mobile: Account icon with Login label is available from the Drawer only

sections/header.liquid
melissaperreault: > but I don't think this is the correct usage of the summary and details tags. These elements are used to [show/hide details about something](https://www.w3schools.com/tags/tag_summary.asp), not for collapsible navigation.

@metamoni I would love to learn more if you could share with the wider Theme devs audience, I recall we purposefully decided to use this but I don't recall why. 🙏

melissaperreault (approved):
Thank you Eugene for all the hard work!

Looks good to me! 🚢

stufreen (dismissed):
🚀

assets/component-menu-drawer.css
```diff
@@ -224,11 +224,20 @@ details[open].menu-opening > .menu-drawer__submenu {
...
+@media screen and (min-width: 750px) {
+  .menu-drawer__account {
+    display: none;
+  }
+
+  .menu-drawer__utility-links:has(ul:empty) {
+    display: none;
+  }
+}
+
```
ludoboludo, line 240: I think we could remove the margin about the social media list if it's an `:only-child` in the drawer: 
<details>
<summary>Screenshot</summary>


</details>

assets/component-menu-drawer.css
```diff
@@ -224,11 +224,20 @@ details[open].menu-opening > .menu-drawer__submenu {
...
   font-size: 1.4rem;
   color: rgb(var(--color-foreground));
 }
 
+@media screen and (min-width: 750px) {
+  .menu-drawer__account {
+    display: none;
+  }
+
+  .menu-drawer__utility-links:has(ul:empty) {
```
ludoboludo, line 236: `:has` isn't supported very well so we haven't been using it so far in the theme. 
Maybe it's ok here since for older browser version or firefox it will just show a little grey box which isn't super problematic. 
<details>
<summary>Screenshot</summary>


</details>

assets/base.css
```diff
@@ -2463,13 +2485,19 @@ product-info .loading-overlay:not(.hidden) ~ *,
   background: rgba(var(--color-foreground), 0.5);
 }
 
+@media screen and (max-width: 989px) {
+  menu-drawer ~ .header__icons .header__icon--account {
+    display: none;
+  }
+}
```
ludoboludo, line 2492: I don't think this ever does anything. The only time we use `menu-drawer` is for filters on collection pages. 
So this will never be a sibling of the `header__icons` element.

metamoni:
@eugenekasimov after testing this with VoiceOver on Chrome, I noticed the new drawer menu doesn't show up as a landmark for screen readers, the way the old mega menu did. This is a level A accessibility rule, which means it needs to be met as the lowest level of requirement. You can read more about it [here](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0#qr-navigation-mechanisms-skip). I think if the old menu offered that context, the new option needs to do the same.

To reproduce, follow these steps with each setting enabled:
- [ ] enable VoiceOver with `command` + `F5`
- [ ] visit the store and hit `control` + `option` + `U` to open the rotor menu
- [ ] use left-right arrows to navigate between different options
- [ ] under landmarks, look for `navigation`
- [ ] disable VoiceOver with `command` + `F5`

https://user-images.githubusercontent.com/22390758/219710355-fcd93bd4-48a8-4b29-8178-6592627b303a.mp4

assets/component-menu-drawer.css
ludoboludo: ah yeah I realize now 👍

assets/base.css
eugenekasimov: Honestly, I did not touch this code, it was here before. But it seems you are right. Let's remove it but keep on eye on it.

assets/component-menu-drawer.css
eugenekasimov: The only workaround I could find is to use `:has`, move margin from `.list-social` to .menu-drawer__account and change it to margin-bottom.

I saw you feedback about `:has` but I think it's okay and at least we could improve it for those cases when `:has` is supported.

eugenekasimov:
> @eugenekasimov after testing this with VoiceOver on Chrome, I noticed the new drawer menu doesn't show up as a landmark for screen readers, the way the old mega menu did. This is a level A accessibility rule, which means it needs to be met as the lowest level of requirement. You can read more about it [here](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0#qr-navigation-mechanisms-skip). I think if the old menu offered that context, the new option needs to do the same.
> 

@metamoni thanks for that. It's really good catch. Well, it happened that we never had this role for our drawer menu on mobile or tablet.

I added a role 'navigation' and now it seems good. As an alternative I thing I could wrap the current drawer menu with <nav> and it will make it works as well. Let me know what you think. Thanks.

metamoni:
> As an alternative I thing I could wrap the current drawer menu with <nav> and it will make it works as well.

@eugenekasimov thank you for updating this! Let's go with `nav`, based on this [recommendation from MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#prefer_html).

eugenekasimov:
> > As an alternative I thing I could wrap the current drawer menu with  and it will make it works as well.
> 
> @eugenekasimov thank you for updating this! Let's go with `nav`, based on this [recommendation from MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#prefer_html).

@metamoni sure, I changed role=navigation to `<nav> `tag.

metamoni:
@eugenekasimov thank you for making changes to help screen readers identify the menu. Unfortunately, something is still not quire right, as I'm now seeing two `navigation` landmarks when the drawer is open, and one of them is actually the Close button (see video below).

https://user-images.githubusercontent.com/22390758/220343989-9e4e03fe-e8ae-4f1f-82a0-38afb8f148fa.mp4

It looks like there are more issues with the drawer menu and it might make it more confusing for screen reader users if we only half-fix them, so let's tackle them in a separate issue. You can either remove the `nav` element for now and I'll follow-up with an issue that looks into adding it back properly, or leave it in and we can fix the open menu. Let me know what you think and I'll create the new issue.

sections/header.liquid
metamoni: @melissaperreault will ask the wider team, I'm sure there was a good reason behind it.

eugenekasimov:
> It looks like there are more issues with the drawer menu and it might make it more confusing for screen reader users if we only half-fix them, so let's tackle them in a separate issue. You can either remove the `nav` element for now and I'll follow-up with an issue that looks into adding it back properly, or leave it in and we can fix the open menu. Let me know what you think and I'll create the new issue.

@metamoni I agree it's better to fix it as a different issue because this problem was here before.

metamoni (approved):

assets/component-menu-drawer.css
stufreen: I guess the alternative would be to do something like @LucasLacerdaUX did here where we check in Liquid if the social icons are set and if so omit the whole area: https://github.com/Shopify/dawn/pull/2226

Personally I am OK with a progressive enhancement here. It will work for almost all buyers and not look terribly broken for the rest.

stufreen (approved):

assets/base.css
```diff
@@ -2312,10 +2333,16 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
-      'left-icon heading icons'
+      'search heading icons'
       'navigation navigation navigation';
   }
 
+  .header--top-center.drawer-menu {
+    grid-template-areas:
+      '. search heading icons';
+    grid-template-columns: auto auto 1fr auto;
+  }
```
ludoboludo, line 2344: There is an issue with this setup. Setting some columns to auto means that anytime the amount of content will change in those columns, it will impact the other ones and misalign the logo. 

Eg: apps in an upcoming PR and also if a store isn't enabling accounts

<details>
<summary>Screenshot</summary>


</details>

sections/header.liquid
```diff
@@ -137,175 +139,175 @@
   </symbol>
 </svg>
 <{% if section.settings.sticky_header_type != 'none' %}sticky-header data-sticky-type="{{ section.settings.sticky_header_type }}"{% else %}div{% endif %} class="header-wrapper color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} header-wrapper--border-bottom{% endif %}">
-  <header class="header header--{{ section.settings.logo_position }} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %}">
+  <header class="header header--{{ section.settings.logo_position }}{% if section.settings.menu_type_desktop == 'drawer' %} drawer-menu{% endif %} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu != blank %} header--has-menu{% endif %} ">
```
ludoboludo, line 142: Ideally group de conditional statements so it's easier to read the classes + extra white space at the end: 
```suggestion
  <header class="header header--{{ section.settings.logo_position }} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu_type_desktop == 'drawer' %} drawer-menu{% endif %}{% if section.settings.menu != blank %} header--has-menu{% endif %}">
```

sections/header.liquid
```diff
@@ -137,175 +139,175 @@
...
-                                  {%- for childlink in link.links -%}
-                                    <li>
-                                      {%- if childlink.links == blank -%}
-                                        <a href="{{ childlink.url }}" class="menu-drawer__menu-item link link--text list-menu__item focus-inset{% if childlink.current %} menu-drawer__menu-item--active{% endif %}"{% if childlink.current %} aria-current="page"{% endif %}>
-                                          {{ childlink.title | escape }}
-                                        </a>
-                                      {%- else -%}
-                                        <details id="Details-menu-drawer-submenu-{{ forloop.index }}">
-                                          <summary class="menu-drawer__menu-item link link--text list-menu__item focus-inset">
+        <header-drawer data-breakpoint="{% if section.settings.menu_type_desktop == 'drawer' %}desktop{% else %}tablet{% endif %}">
```
ludoboludo, line 144: Nitpick - formatting/indentation here.

assets/base.css
ludoboludo: You could maybe do: 
`left-icons heading icons` - `1fr auto 1fr`

Then assign both the `.header__search` & `header-drawer` to that `grid-area: left-icons;` and finally add a `margin-left: 3.2rem;` on the `header__search` which is the result of the hamburger icon width `4.4rem` minus its margin-left value `-1.2rem`

ludoboludo (changes requested):
Just setting it to `request changes` so we can fix some of the issues I found

assets/base.css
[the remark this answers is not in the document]
vfonic: Suggest: `position: fixed; right: 0` on cart notification _could_ fix this.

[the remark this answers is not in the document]
melissaperreault: Gooood find! The popup notification position helps convey the item was added to the cart so its proximity with the cart icon is needed. Thanks team!

assets/base.css
eugenekasimov: Good catch @ludoboludo 👍.

ludoboludo: The change I was mentioning was just needed for that one instance I believe. I see some other grid declaration were edited and I don't believe it's needed 🤔 

I'll comment more in my upcoming review.

assets/base.css
```diff
@@ -2280,8 +2293,8 @@ product-info .loading-overlay:not(.hidden) ~ *,
 
 .header {
   display: grid;
-  grid-template-areas: 'left-icon heading icons';
-  grid-template-columns: 1fr 2fr 1fr;
+  grid-template-areas: 'left-icons heading icons';
+  grid-template-columns: 1fr auto 1fr;
```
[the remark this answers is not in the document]
melissaperreault: Good find!

assets/component-cart-notification.css
```diff
@@ -40,6 +40,12 @@
   .cart-notification {
     border-width: 0 var(--popup-border-width) var(--popup-border-width);
     max-width: 36.8rem;
+    right: 2.2rem;
```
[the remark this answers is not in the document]
melissaperreault: +1 aligned with this feedback! Thanks Ludo!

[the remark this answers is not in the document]
melissaperreault: Actually, I think I misunderstood the recommendation, but what I see on the live demo is good. The alignment is the same for all experiences. 

[Reference](https://screenshot.click/24-33-r08jb-h15pd.png)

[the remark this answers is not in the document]
eugenekasimov: @ludoboludo thanks for this feedback. I got what you mean, however, we have the same alignments currently in dawn.  As Melissa said let's keep it as is for now.

assets/base.css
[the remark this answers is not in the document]
eugenekasimov: I agree about changing `auto` to `2fr` back as it used to be. I'm going to leave `left-icons` to avoid adding more logic because in some cases it comes together with hamburger icon.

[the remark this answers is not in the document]
ludoboludo: But on mobile do we ever have 2 icons ? I don't believe so. 
So there shouldn't be a need to change the area name 🤷

assets/base.css
```diff
@@ -2469,10 +2495,22 @@ menu-drawer + .header__search {
...
 }
 
+.header:not(.drawer-menu) > .header__search {
+  margin-left: 0;
+}
+
+header-drawer {
+  grid-area: left-icons;
+  justify-self: start;
+}
```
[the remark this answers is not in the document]
eugenekasimov: I still need this declaration

```
header-drawer {
  grid-area: left-icons;
}
```

Otherwise it looks like this
Screenshot 2023-02-24 at 12 46 18 PM

assets/base.css
```diff
@@ -2469,10 +2495,22 @@ menu-drawer + .header__search {
...
 .header > .header__search {
-  grid-area: left-icon;
+  grid-area: left-icons;
   justify-self: start;
+  margin-left: 3.2rem;
 }
 
+.header:not(.drawer-menu) > .header__search {
+  margin-left: 0;
+}
```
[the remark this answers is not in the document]
eugenekasimov: I guess you meant the class `.header__search` and not a tag `header__search`

assets/base.css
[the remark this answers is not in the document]
eugenekasimov: Let's replace it with 

```
.header--top-center header-drawer {
  grid-area: left-icons;
}
```

assets/base.css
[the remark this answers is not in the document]
eugenekasimov: We agreed to keep it as`left-icons`

metamoni (approved):

ludoboludo (approved):
🚀

melissaperreault:
@eugenekasimov I found a bug that I think is related to that merge. Page section has lost its width reference, becoming too wide for readable content. [Video reference](https://screenshot.click/03-31-q8cbs-2dvut.mp4)

Do you prefer I open a new issue and take it from there?

PR #2201, opened 2023-01-10
Fix gradient bug in header.

### LINKED ISSUE #2033: [Header + Gradients] Gradient displays fully for a moment when the sticky nav is entering the viewport
When I have a gradient background attached to my header and scroll down, and up again (making the sticky nav appear), for a moment the full gradient is displayed in one go causing some undesired visual effects.

I would expect the gradient to always display as it does at the start of the journey, or when the sticky behaviour motion ends.

- Dawn Version: 7.0.0

**Possible solution**

[Video](https://screenshot.click/13-56-chor7-pxmvd.mp4)

eugenekasimov:
### PR Summary: 

This PR fixes the bug that appears when gradient background set for sticky-header. 

Fixes #2033 .

[Demo video. Before fix.](https://screenshot.click/10-01-yg1bc-8f27f.mp4)

[Demo video. After fix.](https://screenshot.click/10-59-4yf7f-dwpmz.mp4)

It's a known issue and it occurs when we apply `background-attachment: fixed` to the element we apply `transform:`. In our case `transform:` is applied to the parent element but it's still the case. Basically it works that way that `background-attachment: fixed` is not applied until `transform:` is done.

### Decision log
|---|---|---|---|---|
| 1 |  I changed the way we animate the sticky-header. Instead of using `transform:` I use `top:`.  |  I couldn't find alternatives.  |  It seems this is the only way to fix the issue.  |  I can see a minor speed-drop in rendering. However, different tests I did show that the difference is significantly small to account for.  |
| 2 |  I added `--header-height` variable in css and update it using JS when the header is rendered for the first time and each time when the viewport crosses a width threshold `990px` back-and-forward.  |  Listen for any changes of the viewport  |  The header height changes only when the viewport crosses a width threshold `990px`  |    |

### Testing steps/scenarios

- [ ] Go to the Theme settings -- Colors. Set gradient color for a color scheme. For better visibility choose radial gradient. Go to the Sections -- Header. Choose a Color Scheme you set the gradient for.
- [ ] Set the Sticky header `On scroll up`. Test the header. 
- [ ] Test the header on mobile and different browsers. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139238309910)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139238309910/editor)
### Checklist

### DIFF
--- a/sections/multirow.liquid
+++ b/sections/multirow.liquid
@@ -112,6 +112,9 @@
 {
   "name": "t:sections.multirow.name",
   "class": "section",
+  "disabled_on": {
+    "groups": ["header", "footer"]
+  },
   "settings": [
     {
       "type": "select",


assets/base.css
```diff
@@ -2225,15 +2225,15 @@ input[type='checkbox'] {
 }
 
 .shopify-section-header-hidden {
-  transform: translateY(-100%);
+  top: calc(-1 * var(--header-height));
```
ludoboludo, line 2228: Ideally for animation we want to keep using `transform` as it's much more performant and doesn't trigger repaints from the browser. 
This is an interesting article about animation: https://web.dev/animations-guide/

But I wonder what other approach we could have to fix the issue 🤔

eugenekasimov: I'm open to any suggestions but unfortunately I couldn't find any other workarounds. Also, as far I have investigated this topic, using both `transform` and `background-attachment` will always lead to a conflict we have.

ludoboludo: Yeah I'm not seeing any other way to fix it 🤔 
I kind of wonder if we should fix it 😅 Since we have different header styles now. People can work around it by setting the sticky header to always on or not sticky. 

Ideally we keep the best approach in place. As this change applies to all merchants when it's probably a smaller amount using gradients on their theme. 

For once, it's not an issue on Safari (desktop) but it is on Chrome and Firefox.

eugenekasimov: I just tested it in Safari and the issue is there. In my opinion, the bug is pretty visible. I decided to spend a bit more time on finding another workaround. 

As an argument on your opinion I can say that, on one hand, you are absolutely right about the best practice but, on the other hand, this animation itself is quite simple and small to worry about what approach we take.

eugenekasimov: Are you sure that you are using main and not this PR for testing? Because for me it's still happening in Safari ([video](https://screenshot.click/13-50-i65xc-725nw.mp4)).

KaichenWang: Ludo's article shows some tips on how to measure the impact of an animation on rendering using [Chrome devtools](https://web.dev/animations-guide/#layout-chrome). I think that we can start by comparing the use of transform vs top in Dawn's header to understand what the cost is (how much more time in ms is the browser spending on re-rendering)? If we observe >1000ms rendering for the animation to complete, then I'd say we need to look into more performant options.

eugenekasimov: I did tests and it looks like the difference is fairly small, it's just around 16ms.  

<details><summary>Using `transform`</summary>
Screenshot 2023-01-16 at 4 09 56 PM
</details>

<details><summary>Using `top`</summary>
Screenshot 2023-01-16 at 4 11 22 PM
</details>

ludoboludo: ill try and take some time to test those changes on my old Pixel 2 to see if anything is noticeable.

KaichenWang: @eugenekasimov I tried to replicate your results and indeed it seems both animations complete in under ~150ms which is what we set the animation duration to. So maybe looking at total animation duration like i suggested isn't the most useful 🤔 I looked at some underlying numbers (Scripting, Rendering, Painting, System) and it does appear that the "top" animation requires 2-3x the time for those. I think that does indicate a drop in performance, but I don't think it's bad enough to block this PR.

Screenshot 2023-01-17 at 4 13 14 PM2

eugenekasimov: @KaichenWang thank you for the opinion. 

Btw, it seems you did not apply gradient for the header when you tested `transform` which might affect numbers :)

danielvan:
From a design perspective this looks good to me. I looked from two perspectives:
1. The new solution keeps the default state the same as before – it seems to do.
2. The new solution doesn't change the colour of the sticky header depending on how much you scroll down (some sections have the gradient scroll due to the background-attachment properties).

Both scenarios worked well for me. Let me know if I should consider anything else, otherwise happy to ✅

assets/base.css
KaichenWang: Good point! Just re-ran the test. In general, the numbers can fluctuate a bit, but overall I'm observing the same trend.

Screenshot 2023-01-18 at 10 14 28 AM 2

sections/header.liquid
```diff
@@ -738,6 +738,7 @@
     hide() {
       if (this.headerIsAlwaysSticky) return;
       this.header.classList.add('shopify-section-header-hidden', 'shopify-section-header-sticky');
+      document.documentElement.style.setProperty('--header-height', `${this.header.offsetHeight}px`);
```
ludoboludo, line 741: Would it make sense to move this into `connectedCallback()` ? This way it's setting the variable to the right value without waiting for `hide()` to run 🤔

ludoboludo (commented):
Looking good 👍 Thanks for testing performance and the impact of the change. 

I have one small comment but otherwise ready to approve.

ludoboludo (dismissed):
👍

ludoboludo (approved):

sections/header.liquid
```diff
@@ -673,6 +673,7 @@
       this.header = document.querySelector('.section-header');
       this.headerIsAlwaysSticky = this.getAttribute('data-sticky-type') === 'always' || this.getAttribute('data-sticky-type') === 'reduce-logo-size';
       this.headerBounds = {};
+      document.documentElement.style.setProperty('--header-height', `${this.header.offsetHeight}px`);
```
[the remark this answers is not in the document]
eugenekasimov: Good catch, Kai. I'll explore it. Thanks.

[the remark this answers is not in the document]
eugenekasimov: I explored the issue and couldn't find a ccs-only approach to fix it. Please let me know if you have something in mind. 

I added a bit of JS code to update the value of `--header-height `when the header changes its size (more details in Decision log, p2).

ludoboludo (commented):
It's looking pretty good 👌 Good catch by Kai on those needed tweaks 🙌 

I can't help but notice a bit of different feeling/lagginess when I check on my phone. 
I've tried capturing it on video but it doesn't really capture it well. It does seem a bit choppier in the animation compared to the `transform`. 

I think it must be specific to phones with higher refresh rates. Give it a shot on your phone and check if you notice any difference. cc: @KaichenWang

sections/header.liquid
```diff
@@ -691,6 +695,10 @@
       this.createObserver();
     }
 
+    setHeaderHeight() {
+      document.documentElement.style.setProperty('--header-height', `${this.header.offsetHeight}px`);
```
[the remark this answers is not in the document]
eugenekasimov: I feel like my approach is more clear and reliable.

ludoboludo (dismissed):

sections/header.liquid
```diff
@@ -674,6 +674,10 @@
       this.headerIsAlwaysSticky = this.getAttribute('data-sticky-type') === 'always' || this.getAttribute('data-sticky-type') === 'reduce-logo-size';
       this.headerBounds = {};
 
+      this.setHeaderHeight.bind(this)();
```
KaichenWang, line 677: I don't think you need `.bind` here
```suggestion
      this.setHeaderHeight();
```

KaichenWang (dismissed):
Looks good @eugenekasimov. Left a code suggestion.

I also tested on my phone with 60Hz and 90Hz refresh rate and did not notice lag, but if @ludoboludo is seeing a performance drop, let's keep and eye on this after shipping.

ludoboludo (approved):

KaichenWang (approved):
Looks good! 🚀

Just noting a related issue: gradients still seems to "shift" when opening the search. Might need a similar fix in the future.

https://user-images.githubusercontent.com/3710548/214376893-37ebd5cb-65f4-4cfa-8efc-c60105a180bd.mp4

PR #2211, opened 2023-01-11
Feat/follow on shop

schaab:
### PR Summary: 

Add the _Follow on Shop_ setting. Having this enabled creates a way for a merchants customers to _follow_ their store in the Shop App. This allows the customer to get personalized product recommendations and notifications about the store.

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


metamoni:
Running `shopify theme check` should help with the failing CI checks 🙂

LucasLacerdaUX (dismissed):
Looks good to me! :D

schaab:
> Running `shopify theme check` should help with the failing CI checks 🙂

I did run it and the "error" is related to a new `liquid` filter that we created. I documented it in this https://github.com/Shopify/storefront-renderer/pull/16606. As I understood it this documentation fed the `theme-check`, however, what is opaque to me, is how long it takes to propagate there.

LucasLacerdaUX (dismissed):

eugenekasimov (commented):
For some reason FOLLOW ON SHOP is not displayed without SOCIAL MEDIA ICONS or EMAIL SIGNUP.

[Video](https://screenshot.click/13-49-30som-w8m8r.mp4).

LucasLacerdaUX:
Good catch, @eugenekasimov. I think we need to updte line 40 of footer.liquid:
```liquid
  {%- if section.blocks.size > 0 or section.settings.newsletter_enable or section.settings.show_social -%}
```
should probably become:
```liquid
  {%- if section.blocks.size > 0 or section.settings.newsletter_enable or section.settings.show_social or section.settings.enable_follow_on_shop -%}
```

assets/section-footer.css
```diff
@@ -160,46 +155,94 @@
   }
 }
 
+.footer__follow-on-shop:empty,
```
ludoboludo, line 158: I don't think we need this since we're already checking before we output this element: 
```liquid
{%- if shop.features.follow_on_shop? and section.settings.enable_follow_on_shop -%}
```
But maybe there scenarios in which nothing could be outputted in the `div` ?

assets/section-footer.css
```diff
@@ -121,20 +121,15 @@
 
 .footer-block--newsletter {
```
ludoboludo, line 122: nothing to change here. But it's weird that we call this block like this when there isn't a `footer-block` class applied to that element in the first pace. 

And it's not actually a block but a block wrapper that can contain a newsletter block but not in all cases. 🤔 
I can create an issue for a small refactor on this.

assets/section-footer.css
```diff
@@ -121,20 +121,15 @@
...
   display: flex;
-  flex-wrap: wrap;
-  justify-content: center;
   align-items: flex-end;
   margin-top: 3rem;
+  gap: 1rem;
```
ludoboludo, line 126: Here I'm wondering if we should use the variables we already have instead of a hard coded value. The blocks coming up in the footer are using `column-gap: var(--grid-desktop-horizontal-spacing);` and `column-gap: var(--grid-mobile-horizontal-spacing);`.

Not a blocker, just a thought so everything is using a similar spacing.

assets/section-footer.css
```diff
@@ -160,46 +155,94 @@
...
     justify-content: center;
   }
+
+  .footer-block__newsletter {
+    flex-direction: column;
+  }
+
+  .footer__follow-on-shop {
+    text-align: center;
+  }
```
ludoboludo, line 245: this isn't needed as it's on line `166` already 👍

assets/section-footer.css
```diff
@@ -160,46 +155,94 @@
...
-.footer-block__newsletter + .footer__list-social {
-  margin-top: 3rem;
+  .footer__follow-on-shop {
+    margin-bottom: 0.4rem;
+  }
+  /* Follow on shop is the first button but it has siblings*/
+  .footer__follow-on-shop:first-child:not(:last-child) {
+    justify-content: flex-start;
+    margin-right: auto;
+    text-align: left;
```
ludoboludo, line 206: Curious to see in which scenarios the text align applies 🤔 I feel like it doesn't much impact in what I tested.

ludoboludo: Same for the justify-content. It seems to be working with the `margin-right: auto;` fine and maybe not needing text align and justify-content 🤷 
But I might be missing something 😅

assets/section-footer.css
```diff
@@ -160,46 +155,94 @@
...
+  }
+
+  /*
+    All three components are present, email, Follow on Shop, and social icons.
+    Moves the FoS button next to the social icons so they appear grouped together
+  */
+  .footer__follow-on-shop:not(:first-child):not(:last-child) {
+    justify-content: flex-end;
+    text-align: right;
```
ludoboludo, line 215: Same here. Since email has `flex-grow: 1;` it's preventing the `follow on shop` block to take more space than it needs and pushes it in the right spot already.

ludoboludo (commented):
Im not sure if we would re work the focus outline on the button since it's not pulling the styles from the theme. But it looks like that on mobile: 
<details>
<summary>Screenshot</summary>


</details>

<details>
<summary>Screenshot 2 with different outline color and spacing</summary>


</details>

schaab:
> 

We have a PR open to try and solve this one https://github.com/Shopify/shop-js/pull/752.

assets/section-footer.css
LucasLacerdaUX: I thought this was to prevent the button showing on `no-js`, but you're right that it still outputs something.

So maybe I'm missing another scenario or this isn't needed.

schaab: I just saw this pattern with the other two components and just followed it. Assumed it was there for a reason.

assets/section-footer.css
schaab: There is some small spacing issues we have for the button. The TLDR; is that the transitioned to width is larger than the original width. If we don't have this there is actually a small gap which causes some misalignment which is really visible in mobile view.

assets/section-footer.css
LucasLacerdaUX:
```suggestion
```

assets/section-footer.css
ludoboludo: Oh maybe it's due to the `flex-direction` on mobile ? Since it's column 🤔

assets/section-footer.css
ludoboludo: Sounds good 👍 thanks for the context 🙂

LucasLacerdaUX (approved):
LGTM. Thanks for making the changes 🐕 👍

PR #2239, opened 2023-01-20
Header app block

ludoboludo:
testing app block next to icons with currency picker


bakura10:
I like the idea and considered this in our themes as well. The issue though is that most app blocks are definitely not made to be inserted inside  the header, and I am afraid that merchants may misunderstood that and confuse it with section blocks.

I think that Shopify should introduce a new kind of app blocks, so that when app developers register an app block they can indicate "header" type for instance, where one of the constraint of header app block would be that they won't be more than an icon or a text. Then, theme developer could restrict the app blocks:

```
{
   "type": "@app",
   "app_block_type": "header"
}
```

And merchants would only see app blocks that have been thought from the start to be inserted here.

ludoboludo:
Yeah I agree with some of those pain points. We will most likely start approaching it this way and tweak as we go. No apps/theme extension has really been created for the header since there was no support for app blocks (at least in our themes). So it will be a new opportunity for apps to create header/footer content and interesting to see the use cases. 

Ideally there would be an option to choose where it should come up as an option, similar to how you can choose on which template your app/extension appears. We will most likely look into this as a follow up.

PR #2249, opened 2023-01-23
Fix old results shown on predictive search

### LINKED ISSUE #2199: [Predictive search] Old results are shown when the query is cleared and rewriten

1. Write some query on the search input (With predictive search activated)
2. Remove the input, with the clear button or with backspace. The results are hidden.
3. Start writing another word: the old results appear.

Current behaviour that shouldn't be changed:
- When I add new characters to the current query, the old results are shown since they might still be relevant

Ideal behaviour:
- When I change the query (remove some characters and write new different ones) the old results are not shown since they are no longer relevant

Compromise solution if the ideal behaviour is not possible:
- When I clear the query (manually or with the clear button) and then start writing again, the old results should not be shown.

- Dawn Version: 7.0.1
- macOS Version 13.1

[Video showing the bug](https://screenshot.click/05-59-sfyuo-dx0mx.mp4)

FCalabria:
### PR Summary: 

Hide stale results on the predictive search when writting a new query and refresh the "search for" button with the new term as soon as possible.

Fixes #2199 and improves the user experience updating the "search for" button while the user types, instead of waiting for the predictive search query to resolve.

To fix the bug, delete the results and show the loading spinner when the new query is not an addition over the old one.

For the improvement, I used a search and replace technique modifying the html on the fly with JS. I could not figure out a way to identify the search term that wasn't just looking for the string itself (depending on the language the quote symbol and the term position changes). This can cause a problem when the user looks for exactly the label on the search button ("Search for...") which i avoided not replacing anything if there is more than one match.

### Testing steps/scenarios

- [ ] When writing a query, the search term in the button updates before the predictive search results
- [ ] When a query is total or partially deleted and then a different term is written, the old predictive search results are not shown, just a loading state.

### Demo links

- [Store](https://afpsbbuvix8y3fsp-45840990230.shopifypreview.com?preview_theme_id)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139285364758/editor)

### Checklist

### DIFF
--- a/assets/predictive-search.js
+++ b/assets/predictive-search.js
@@ -110,16 +110,17 @@ class PredictiveSearch extends SearchForm {
   }
 
   updateSearchForTerm(previousTerm, newTerm) {
-    const searchForButton = this.querySelector("#search-for-button");
-    const currentButtonText = searchForButton?.innerText;
+    const searchForTextElement = this.querySelector(
+      "[data-predictive-search-search-for-text]"
+    );
+    const currentButtonText = searchForTextElement?.innerText;
     if (currentButtonText) {
-      const buttonIcon = searchForButton.querySelector("svg").outerHTML;
       if (currentButtonText.match(new RegExp(previousTerm, "g")).length > 1) {
         // The new term matches part of the button text and not just the search term, do not replace to avoid mistakes
         return;
       }
       const newButtonText = currentButtonText.replace(previousTerm, newTerm);
-      searchForButton.innerHTML = newButtonText + buttonIcon;
+      searchForTextElement.innerText = newButtonText;
     }
   }
 
--- a/sections/predictive-search.liquid
+++ b/sections/predictive-search.liquid
@@ -147,13 +147,12 @@
 
     <div id="predictive-search-option-search-keywords" class="predictive-search__search-for-button">
       <button
-        id="search-for-button"
         class="predictive-search__item predictive-search__item--term link link--text h5 animate-arrow"
         tabindex="-1"
         role="option"
         aria-selected="false"
       >
-        {{ 'templates.search.search_for' | t: terms: predictive_search.terms }}
+        <span data-predictive-search-search-for-text>{{ 'templates.search.search_for' | t: terms: predictive_search.terms }}</span>
         {% render 'icon-arrow' %}
       </button>
     </div>


assets/predictive-search.js
```diff
@@ -263,6 +289,10 @@ class PredictiveSearch extends SearchForm {
     this.resultsMaxHeight = false
     this.predictiveSearchResults.removeAttribute('style');
   }
+
+  clearResources() {
```
NathanPJF, line 293: Can you use the existing `closeResults` method for this instead? Is this additional method necessary?

sections/predictive-search.liquid
```diff
@@ -7,7 +7,7 @@
 %}
   <div id="predictive-search-results" role="listbox">
     {%- if first_column_results_size > 0 or predictive_search.resources.products.size > 0 -%}
-      <div class="predictive-search__results-groups-wrapper{% unless predictive_search.resources.products.size > 0 %} predictive-search__results-groups-wrapper--no-products{% endunless %}{% unless predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 %} predictive-search__results-groups-wrapper--no-suggestions{% endunless %}">
+      <div id="predictive-search__results-groups-wrapper" class="predictive-search__results-groups-wrapper{% unless predictive_search.resources.products.size > 0 %} predictive-search__results-groups-wrapper--no-products{% endunless %}{% unless predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 %} predictive-search__results-groups-wrapper--no-suggestions{% endunless %}">
```
ludoboludo, line 10: I can see why you'd have the `__` underscores there. But since we didn't use that same pattern for `predictive-search-option-search-keywords` I think we should approach it the same way. With a `-` instead.

ludoboludo (commented):
Looking good. Ill test a bit tomorrow some scenarios. 
But no bugs/issue noticed 👍

assets/predictive-search.js
```diff
@@ -99,6 +108,23 @@ class PredictiveSearch extends SearchForm {
...
   }
 
+  updateSearchForTerm(previousTerm, newTerm) {
+    const searchForButton = this.querySelector("#search-for-button");
+    const currentButtonText = searchForButton?.innerText;
+    if (currentButtonText) {
+      const buttonIcon = searchForButton.innerHTML.replace(
+        currentButtonText,
+        ""
+      );
```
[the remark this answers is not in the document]
ludoboludo: I think I'd suggest more grabbing the icon via it's class or element name 🤔 `searchForButton.querySelector('svg')`
It would be easier to read.

assets/predictive-search.js
```diff
@@ -27,7 +27,16 @@ class PredictiveSearch extends SearchForm {
 
   onChange() {
     super.onChange();
-    this.searchTerm = this.getQuery();
+    const newSearchTerm = this.getQuery();
+    if (!this.searchTerm || !newSearchTerm.startsWith(this.searchTerm)) {
+      // Hide the results when they are no longer relevant for the new search term
+      this.closeResults(false);
```
ludoboludo, line 33: This does what the function you had before did 🤔 
Here are two video comparisons: [using clearResources()](https://screenshot.click/26-37-z1bah-zoj8s.mp4) & [using closeResults(false)](https://screenshot.click/26-41-laoke-eu6yn.mp4)

FCalabria: I tested and it seemed to be fine, but maybe my internet was too fast to appreciate the difference. Thanks for the thorough testing!

sections/predictive-search.liquid
```diff
@@ -7,7 +7,7 @@
 %}
   <div id="predictive-search-results" role="listbox">
     {%- if first_column_results_size > 0 or predictive_search.resources.products.size > 0 -%}
-      <div class="predictive-search__results-groups-wrapper{% unless predictive_search.resources.products.size > 0 %} predictive-search__results-groups-wrapper--no-products{% endunless %}{% unless predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 %} predictive-search__results-groups-wrapper--no-suggestions{% endunless %}">
+      <div id="predictive-search-results-groups-wrapper"class="predictive-search__results-groups-wrapper{% unless predictive_search.resources.products.size > 0 %} predictive-search__results-groups-wrapper--no-products{% endunless %}{% unless predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 %} predictive-search__results-groups-wrapper--no-suggestions{% endunless %}">
```
ludoboludo, line 10:
```suggestion
      <div id="predictive-search-results-groups-wrapper" class="predictive-search__results-groups-wrapper{% unless predictive_search.resources.products.size > 0 %} predictive-search__results-groups-wrapper--no-products{% endunless %}{% unless predictive_search.resources.queries.size > 0 or predictive_search.resources.collections.size > 0 %} predictive-search__results-groups-wrapper--no-suggestions{% endunless %}">
```

NathanPJF (approved):
Thanks for owning this @FCalabria 

It doesn't seem like the failing theme check step will be a blocker to merge. I see that the recent PRs have merged without this passing ([example](https://github.com/Shopify/dawn/pull/2253#event-8347446695)). But don't forget to rebase this first 🔂 🚢

FCalabria:
> I think it's looking good 👍 I can see how the proposed change by Nathan would make sense too 👍 About wrapping the text in a span. Not sure if we need the data attribute since we can select the `span` element itself.

I decided to use the data attribute because with the span i don't need the reference to the full button anymore, so i could delete that id.

> It doesn't seem like the failing theme check step will be a blocker to merge. I see that the recent PRs have merged without this passing (https://github.com/Shopify/dawn/pull/2253#event-8347446695). But don't forget to rebase this first 🔂 🚢

Sure! I was waiting for you to finish reviewing to not mess up anything

NathanPJF (approved):
🎩 I just re-tophatted https://os2-demo.myshopify.com/?preview_theme_id=139285364758

ludoboludo (approved):
🎉

PR #2264, opened 2023-01-27
Edit the layout and info to match the emails customers receive

ludoboludo:
Edit the layout and info to match the email notification merchants' customers receive.

### DIFF
--- a/assets/customer.css
+++ b/assets/customer.css
@@ -196,13 +196,13 @@
   }
 }
 
-.customer tfoot tr:last-of-type td {
+.customer .order-details__totals td {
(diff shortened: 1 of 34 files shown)


ludoboludo:
>Would it be possible for shopify to try to add some official documentation about the order that thèmes should follow to provide a consistent experience across all Shopify ?

Comment from the partner dev channel in slack.

melissaperreault:
**Some notes**

I am unsure of the new items order and why we don't subtract the refund from the total.

danielvan:
Is there a need for Total after Refund?

PR #2266, opened 2023-01-30
Update default value of constrain media setting

kmeleta:
### PR Summary: 

Changes the default value of the `Constrain media to screen height` setting.

Important: Could cause a visual change in existing themes for some users depending on product image sizes.

Relates to https://github.com/Shopify/dawn/pull/2052

The constrain setting was originally included with the default set to `false`. Which would prevent any visual changes. However, we intend for the setting to be `true` moving forward so this PR just changes that default value.

We originally considered a softer rollout of this change, where we update the setting default in a subsequent release, but adding the desired end state now as a breaking change for a major release comes with less logistical overhead.

Updated default setting in the main-product and featured-product section schemas and also added the missing default in the product.json template.

This is a visually breaking change. 

Depending on the selected media width setting, the aspect ratio of the product images, and the end user's viewport dimensions, it is possible some product images will be scaled down to fit within the viewport. This will be most common with taller portrait images on relatively shorter viewports. Otherwise the effect of this change might not be immediately noticeable.

Before:

After:

### Testing steps/scenarios
- Ensure constrain setting is checked and applied by default on the product page
- Ensure constrain setting is checked and applied by default when adding new featured product sections

### Demo links
- [Editor](https://os2-demo.myshopify.com/admin/themes/139313250326/editor)

### Checklist

### DIFF
--- a/templates/product.json
+++ b/templates/product.json
@@ -103,6 +103,7 @@
         "enable_sticky_info": true,
         "gallery_layout": "stacked",
         "media_size": "large",
+        "constrain_to_viewport": true,
         "mobile_thumbnails": "hide",
         "hide_variants": true,
         "enable_video_looping": false,


melissaperreault (approved):
Thanks Ken! Looks good!
It's definitely an improvement for keeping tall images shorter on larger screens and for normalizing the height when all media have different ratio. 🚢 🚢 When they do manual update, merchants can always uncheck this to revert to what they had.

ludoboludo (approved):

PR #2270, opened 2023-02-01
Remove async CSS pattern where it may introduce layout shifts

krzksz:
### PR Summary: 

This PR removes the usage of asynchronous CSS pattern from stylesheets that may actually be required to render the initial screen. Having those stylesheets as not render blocking leads to unstyled content being shown and shifting when they are downloaded and applied.

Marking some of the stylesheets as targeted for printing with `media="print"` and then switching them to apply for all devices with `onload="this.media='all'"` is a common pattern that lets developers download additional CSS with low priority and without blocking the initial render. As a rule, this technique should only be applied for styles that do not influence the look of anything that users may see without interacting with the page.

Unfortunately, it seems like this solution ended up in too many places when it comes to Dawn. Below you can see the [experiment that I run using WebPageTest](https://www.webpagetest.org/video/compare.php?tests=230127_AiDc15_7MH,230127_AiDcXY_7MG) where I remove the pattern for some of the stylesheets.

experiment

You may notice the whole `Obsessive Attention. Intelligent Effort.` shifting to the right in the control run because of asynchronous `component-rte.css`. This is not the case after applying changes from this PR in the experiment run, improving Cumulative Layout Shift score by 0.065 (95%). This is especially important, because sites need to be below 0.1 to be considered as "Good" within this metric.

I analysed all of the occurrences of this pattern and checked if a certain CSS is required to render anything that may be visible for users before they interact with the page.

### Other considerations

Initially, I thought that the pattern would have to be removed from everywhere. This could slow down the initial render as pointed out by @siakaramalegos as the browser would have to download more files. Fortunately, there's only a bunch of stylesheets that needed adjusting. I also made multiple tests across different Dawn-based stores and all rendering metrics land within the margin of error.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

Parts of the page that relied on styles from asynchronous stylesheets will no longer appear initially unstyled, especially on first visits and/or slower connections.

### Testing steps/scenarios

- [ ] Setup page with sections or features that incorrectly used asynchronous styles.
- [ ] Disable caching in the browser, you can also slow down the internet connection using DevTools.
- [ ] Open the page and confirm that with this change above sections and features no longer render partially unstyled.

### Demo links

- [Store](https://mateusz-krzeszowiak.myshopify.com/) - There are sections on homepage + Github integration setup for preview.

### Checklist

### DIFF
--- a/sections/featured-blog.liquid
+++ b/sections/featured-blog.liquid
@@ -1,12 +1,6 @@
-<link rel="stylesheet" href="{{ 'component-slider.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-card.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-article-card.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'section-featured-blog.css' | asset_url }}" media="print" onload="this.media='all'">
-
-<noscript>{{ 'component-slider.css' | asset_url | stylesheet_tag }}</noscript>
-<noscript>{{ 'component-card.css' | asset_url | stylesheet_tag }}</noscript>
-<noscript>{{ 'component-article-card.css' | asset_url | stylesheet_tag }}</noscript>
-
+{{ 'component-slider.css' | asset_url | stylesheet_tag }}
+{{ 'component-card.css' | asset_url | stylesheet_tag }}
+{{ 'component-article-card.css' | asset_url | stylesheet_tag }}
 {{ 'section-featured-blog.css' | asset_url | stylesheet_tag }}
 
 {%- style -%}
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -1,16 +1,10 @@
 {% comment %}theme-check-disable UndefinedObject{% endcomment %}
 {{ 'section-footer.css' | asset_url | stylesheet_tag }}
-<link rel="stylesheet" href="{{ 'component-newsletter.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-list-menu.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-list-payment.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-list-social.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'disclosure.css' | asset_url }}" media="print" onload="this.media='all'">
-
-<noscript>{{ 'component-newsletter.css' | asset_url | stylesheet_tag }}</noscript>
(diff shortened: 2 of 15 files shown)


stufreen:
I double-checked how this affects the page performance and I agree that it is a net-positive and we should ship it. As you say, it brings up the CSS files earlier in the network waterfall and reduces CLS. It does seem to increase blocking time but maybe that is to be expected. These lighthouse reports are the clearest examples:

**Before**
lighthouse - main

**After**
lighthouse - branch

Related note: I wonder if our pattern of having many small CSS files is hurting us and we should put them in one medium-sized CSS file that can easily be cached. But that's out of scope for this PR.

I will wait for one of the other devs with more context in the codebase to review for whether there are any other CSS files that should/should not be loaded synchronously instead of async.

stufreen (dismissed):

ludoboludo (commented):
Looking good 👍 Thanks for doing this. 

I think we just need a rebase maybe and removing any calls to `component-rte.css`.

sections/footer.liquid
```diff
@@ -1,18 +1,11 @@
...
-<noscript>{{ 'component-list-payment.css' | asset_url | stylesheet_tag }}</noscript>
-<noscript>{{ 'component-list-social.css' | asset_url | stylesheet_tag }}</noscript>
-<noscript>{{ 'component-rte.css' | asset_url | stylesheet_tag }}</noscript>
-<noscript>{{ 'disclosure.css' | asset_url | stylesheet_tag }}</noscript>
+{{ 'component-newsletter.css' | asset_url | stylesheet_tag }}
+{{ 'component-list-menu.css' | asset_url | stylesheet_tag }}
+{{ 'component-list-payment.css' | asset_url | stylesheet_tag }}
+{{ 'component-list-social.css' | asset_url | stylesheet_tag }}
+{{ 'component-rte.css' | asset_url | stylesheet_tag }}
+{{ 'disclosure.css' | asset_url | stylesheet_tag }}
```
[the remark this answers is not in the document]
krzksz: It depends, there are pages with short content where footer is visible right away on desktop. Example: https://theme-dawn-demo.myshopify.com/pages/shipping

sections/featured-blog.liquid
```diff
@@ -1,12 +1,6 @@
-<link rel="stylesheet" href="{{ 'component-slider.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-card.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'component-article-card.css' | asset_url }}" media="print" onload="this.media='all'">
-<link rel="stylesheet" href="{{ 'section-featured-blog.css' | asset_url }}" media="print" onload="this.media='all'">
```
stufreen, line 4: Non blocking: Were we loading this twice before?

stufreen (approved):

sections/featured-blog.liquid
krzksz: Defined yes, fortunately it wasn't requested twice thanks to caching.

stufreen (approved):

Issue #763, opened 2021-10-11
Draft product previews do not change price on variant selection and price defaults to 19.99


mikevich:
**Describe the current behavior**
When previewing a Draft product from the admin in Dawn, only the first available variant's price is shown. Selecting another variant does not change the price. 

**Describe the expected behavior**
Price should change to match the selected variant. This occurs for draft products on our other themes. 

**Version information (Dawn, browsers and operating systems)**
- Dawn Version: 2.2.0
- Chrome Version 94.0.4606.81 
- macOS Version 11.6

**Possible solution**
No proposed solution currently.

**Additional context/screenshots**
To replicate this problem, create a draft product with multiple variants, each with a different price. Then, click the "Preview" button at the top of the product page. The page will load in correctly with the price of the first available variant, but using the variant selector to change variants will not affect the price.

However, other variant changes occur: the associated variant image changes, and clicking "Add to Cart" will add the selected variant to the cart correctly.

Example draft product with differently-priced variants:

![image](https://user-images.githubusercontent.com/26332238/136821770-43943e8a-d006-4c1b-bc2a-37f223e47dbb.png)

[Video showing price change issue](https://screenshot.click/11-11-44pwa-aarrc.mp4)

For comparison, here is a video of the same product previewed in Debut:

[Video showing price change in Debut](https://screenshot.click/11-15-1esyw-egd9r.mp4)

okaykomputer:
Same thing is happening in the ticket that I'm working on that klawrz89 pointed out.

ryanaldred11:
[Problem ticket create here](https://shopify.zendesk.com/agent/tickets/29619941)

#29619941 [Dawn] Product price breaks on variant change of draft products

@Shopify/theme-support

rustypaul:
This happens for us using Empire too.

chloemcgov:
Hey folks, just sharing for visibility that this has been reported by a Merchant on the Ride theme also. 

Example test store replication here if interested:
* [Internal](https://app.shopify.com/services/internal/shops/45632946232)
* [Video example](https://screenshot.click/27-27-91423-77682.webm)

dontgiveajack:
Hey team, I have another store with the same issue: https://app.shopify.com/services/internal/shops/61675897017

Draft products with variants are not changing prices correctly on the preview.

Setting the product to "Active" fixes this.

Issue #532, opened 2021-08-31
Sizes attribute for images in collection views assumes at least 4 products


jsunsawyer:
The sizes logic for the grid of images in collection views assumes that there are at least 4 products. If a collection has fewer products, cards are displayed larger, but their images don't serve the larger sizes to reflect that. Instead, they're blurry/low quality.

Possible solutions:
1. Instead of allowing collections with fewer than 4 items display larger, render all content at the same size (4 column at largest size).
2. Use [lazySizes](https://github.com/aFarkas/lazysizes#automatically-setting-the-sizes-attribute) to automatically set the sizes attribute

**Dawn Version: 2.0.0**

PR #2277, opened 2023-02-02
Fix blurry product/collection grid images

marija-scuric:
### PR Summary:

Fixes blurry card images by including dynamic intrinsic size calculation for the rendered images in grids.

Fixes [#763](https://github.com/Shopify/dawn-internal/issues/763) | [#532](https://github.com/Shopify/dawn/issues/532)

This change introduces a variable to store the column values based on the theme editor settings selection for the number of columns for desktop and mobile rather than the current fixed value in the calculation, which can cause card images to be displayed in lower quality. 

For example, the featured collection section has the card image size calculated based on 4 product cards being displayed, however a row can contain anywhere between 1 and 5 products, which results in lower quality images when there are fewer than 4 products.

This way, the calculation for intrinsic size on card files (product, collection) is based on dynamic input, so the images are always displayed with a slightly higher intrinsic size than the rendered size. There's a reference to the column values when rendering snippets in section files (collection, collection list, search, product recommendations).

### Other considerations

In some cases where uploaded images are very large, the load time may be slightly increased to accommodate the larger image size.

For some merchants, rendered card images will be in higher quality.

### Testing steps/scenarios

- [ ] Step 1 - Home page: Collection-list.liquid section, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection
- [ ] Step 2 - Home page: Featured-collection.liquid section, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection
- [ ] Step 3 - Product page: Recommended-products.liquid section, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection
- [ ] Step 4 - Product grid of the Collection page, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection
- [ ] Step 5 - Collection grid of the Collection list page, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection
- [ ] Step 6 - Search results page, make sure to test different number of columns on mobile and desktop and see if the intrinsic size updates based on the different selection, searching for `a` should result in a blog page, products and a page in results

### Demo links

- [Store](https://tsx-demo.myshopify.com/?preview_theme_id=122870792256)
- [Editor](https://tsx-demo.myshopify.com/admin/themes/122870792256/editor)

### Checklist

### DIFF
--- a/snippets/card-collection.liquid
+++ b/snippets/card-collection.liquid
@@ -4,7 +4,8 @@
   Accepts:
   - card_collection: {Object} Collection Liquid object
   - media_aspect_ratio: {String} Size of the product image card. Values are "square" and "portrait". Default is "square" (optional)
-  - columns: {Number}
+  - columns_desktop: {Number} Number of columns on desktop. Default: 3 (optional)
+  - columns_mobile: {Number} Number of columns on mobile. Default: 2 (optional)
   - extend_height: {Boolean} Card height extends to available container space. Default: false (optional)
   - wrapper_class: {String} Wrapper class for card (optional)
 
@@ -28,6 +29,8 @@
     assign card_color_scheme = settings.collection_card_color_scheme
     assign card_style = settings.collection_card_style
   endif
+  assign desktop = columns_desktop | at_least: 1
+  assign mobile = columns_mobile | at_least: 1
 -%}
 
 <div class="card-wrapper animate-arrow {% if wrapper_class and wrapper_class != 'none' %}{{ wrapper_class }}{% else %}collection-card-wrapper{% endif %}">
@@ -62,8 +65,8 @@
               "
               src="{{ card_collection.featured_image | image_url: width: 1500 }}"
               sizes="
-                (min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: columns }}px,
-                (min-width: 750px) {% if columns > 1 %}calc((100vw - 10rem) / 2){% else %}calc(100vw - 10rem){% endif %},
+                (min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 100 | divided_by: desktop }}px,
+                (min-width: 750px) {% if mobile > 1 %}calc((100vw - 10rem) / 2){% else %}calc(100vw - 10rem){% endif %},
                 calc(100vw - 3rem)
               "
               alt=""
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -13,6 +13,8 @@
   - section_id: {String} The ID of the section that contains this card.
   - horizontal_class: {Boolean} Add a card--horizontal class if set to true. Default: false (optional)
   - horizontal_quick_add: {Boolean} Changes the quick add button styles when set to true. Default: false (optional)
+  - columns_desktop: {Number} Number of columns on desktop. Default: 4 (optional)
+  - columns_mobile: {Number} Number of columns on mobile. Default: 2 (optional)
 
   Usage:
   {% render 'card-product', show_vendor: section.settings.show_vendor %}
@@ -31,6 +33,8 @@
     if ratio == 0 or ratio == null
       assign ratio = 1
     endif
+    assign desktop = columns_desktop | at_least: 1
+    assign mobile = columns_mobile | at_least: 1
   -%}
   <div class="card-wrapper product-card-wrapper underline-links-hover">
     <div
@@ -64,7 +68,7 @@
                   {{ card_product.featured_media | image_url }} {{ card_product.featured_media.width }}w
                 "
                 src="{{ card_product.featured_media | image_url: width: 533 }}"
-                sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 130 | divided_by: 4 }}px, (min-width: 990px) calc((100vw - 130px) / 4), (min-width: 750px) calc((100vw - 120px) / 3), calc((100vw - 35px) / 2)"
+                sizes="(min-width: {{ settings.page_width }}px) {{ settings.page_width | minus: 130 | divided_by: desktop }}px, (min-width: 990px) calc((100vw - 130px) / {{ desktop }}), (min-width: 750px) calc((100vw - 120px) / {{ mobile }}), calc((100vw - 35px) / {{ mobile }})"
                 alt="{{ card_product.featured_media.alt | escape }}"
                 class="motion-reduce"
                 {% unless lazy_load == false %}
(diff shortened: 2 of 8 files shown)


snippets/card-product.liquid
```diff
@@ -31,6 +33,8 @@
     if ratio == 0 or ratio == null
       assign ratio = 1
     endif
+    assign desktop = columns_desktop | at_least: 1
+    assign mobile = columns_mobile | at_least: 1
```
ludoboludo, line 37: Here we could use better descriptive naming I think. Those names don't convey clearly what the value is tied to. Maybe `columns_amount_desktop` 🤷 or something else

snippets/card-collection.liquid
```diff
@@ -28,6 +29,8 @@
     assign card_color_scheme = settings.collection_card_color_scheme
     assign card_style = settings.collection_card_style
   endif
+  assign desktop = columns_desktop | at_least: 1
+  assign mobile = columns_mobile | at_least: 1
```
ludoboludo, line 33: Same here for the naming. Having something a bit more descriptive so it's easier to read would be great.

ludoboludo (commented):
A couple issues I noted: 

- With the collage section. We're loading the product and collection card without passing any `columns_desktop` value. Which means we're going to load the media like there was only 1 column which isn't great. We're going to need a bit of logic there I think to figure out the amount of column it should be using. Collage is a 3 columns grid where the card can take either 2/3, or 1/3, or 3/3 if there is only one block. Then on mobile depending on the amount of block mobile layout chosen it can be 1 or two columns. 
- With complementary products. We are currently not passing in any column amount. This one is tough cause there isn't any columns but it something coming up pretty small and will depend a bit of the size of the media column of the product page. If that column is small, the complementary product cards will use bigger images than when the media column is large.

Roi-Arthur:
@ludoboludo
> * With the collage section.

I _think_ this would cover the logic for how many columns to pass into the collage section blocks (both product & collection):
```liquid
{% liquid
  if section.blocks.size == 1
    assign columns_amount_desktop = 1
  elsif forloop.first and section.settings.desktop_layout == 'left' or forloop.last and section.settings.desktop_layout == 'right'
    assign columns_amount_desktop = 1.5
  else
    assign columns_amount_desktop = 3
  endif
  

  if section.settings.mobile_layout == 'column' or section.blocks.size == 1
    assign columns_amount_mobile = 1
  else
    if section.blocks.size == 3
      if forloop.first and section.settings.desktop_layout == 'left' or forloop.last and section.settings.desktop_layout == 'right'
        assign columns_amount_mobile = 1
      else 
        assign columns_amount_mobile = 2
      endif
    else
      assign columns_amount_mobile = 2
    endif
  endif
%}
```

For desktop:
- Using only 1 block we have 1 column
- Using more than 1 block, the `desktop_layout` setting ('left' or 'right') dictates which bock will be big or small:
  - The first block when 'left' or the last block when 'right' is big (3 columns times 2, so I think we want to divide by 1.5)
  - The other blocks are small (3 columns)

For mobile:
- If the layout is 'column' or if we have only one block we have only 1 column
- If the layout is 'collage', we're back to counting blocks and looking at `desktop_layout`:
  - The first block when `desktop_layout` is 'left' or the last block when `desktop_layout` is 'right' is big (1 column)
  - The other blocks are small (2 columns)

snippets/card-collection.liquid
```diff
@@ -4,7 +4,8 @@
   Accepts:
```
ludoboludo, line 4: To reply to your comment @Roi-Arthur 

I think we can change the logic a little. the `desktop` variable is assigned `at_least: 1` so in some cases we don't need to assign anything in the collage file. 
Here is the mobile logic you have but a bit reworked, I'd have to actually test it to make sure the `and` is read properly:
```liquid
{% liquid 
  unless section.settings.mobile_layout == 'column' or section.block.size == 1
    if section.settings.mobile_layout == 'collage'
      if section.blocks.size == 3 and forloop.first and section.settings.desktop_layout == left or forloop.last and section.settings.desktop_layout == right
        assign columns_amount_mobile = 1
      else
        assign columns_amount_mobile = 2
      endif
    endif
  endunless
%}
```

For the complementary products it's a bit trickier 🤔 I think from what I can see their size is going from `64px` wide up to potentially `140px` when it's on a product with no image. And `118px` on a product with images. 

Maybe we could do something a bit simpler that isn't quite perfect but good enough. Something like defaulting to `10` for desktop and `5` for mobile 👍 This way we don't go into logic that's a bit hard to understand. 

What do you think?

Roi-Arthur: > the desktop variable is assigned at_least: 1 so in some cases we don't need to assign anything in the collage file.

I did think of this after the fact and was wondering what's preferable: one the one hand we have more liquid but it's very readable and you know exactly what you're passing to the snippet so might be more user friendly, especially for people who are getting started. On the other we can definitely remove all cases where the assign is for 1 in terms of efficiency.

> I'd have to actually test it to make sure the and is read properly:

That was also my concern 😅
Haven't tried any of this yet; was just a draft starting from what sounded logical.

Roi-Arthur:
Here we go, I think we're covered now. The current changes adjust:
- The Collage section loads images with the logic we've mentioned before, taking into account the fact that we can't chain operators exactly like we want (hence a nested `unless` block)
- Make `column_amounts` more explicit in our card-product and card-collection snippets
- Adjust references to those snippets in main-product.liquid and cart-drawer.liquid snippets
  - In main-product.liquid, for the complementary product blocks I added the defaults we talked about (10 for desktop, 5 for mobile)
  - In cart-drawer.liquid, since we load the image for the line items, I've added defaults based on the size at which we load things, which appears to be a constant 150px (both desktop and mobile). As such I've made the defaults 5 for desktop and 2 for mobile.

snippets/card-collection.liquid
ludoboludo: Just a correction, you mentioned that `operators are evaluated from left to right` which isn't right. It's the opposite.

sections/collage.liquid
```diff
@@ -90,19 +90,63 @@
...
+
+                assign columns_amount_mobile = 1
+                unless section.settings.mobile_layout == 'column' or section.blocks.size == 1
+                  unless section.blocks.size == 3 and forloop.first and section.settings.desktop_layout == 'left'
+                    unless section.blocks.size == 3 and forloop.last and section.settings.desktop_layout == 'right'
+                      assign columns_amount_mobile = 2
+                    endunless
+                  endunless
+                endunless
+              %}
```
ludoboludo, line 141: I wonder if we need to repeat this twice 🤔 Since it just needs to be within the for loop. We could wrap it in an if statement `if block.type == 'collection' or block.type == 'product'` though that adds to the list of conditional already present.

Roi-Arthur: Yeah for this one I wasn't sure what to do; having the block of code only once does seem more appealing.

> it just needs to be within the for loop

My issue was "do we want to add an _if_ check when there's already a _case_ coming?
If we're fine with that I'm happy to move the logic before the case

nate-young:
Just wanted to see if there was any updates here? It looks like it's still happening in versions 15.0.2. 

Additionally, the same issue appears to effect article cards.

PR #2286, opened 2023-02-03
Fix mark element not getting the right color

ludoboludo:
Fix mark element getting a color from the browser which end up overwriting our default text color.


melissaperreault (dismissed):
🚢 🚢 🚢

kmeleta (approved):
Might as well commit the missing space, but fix looks good.

PR #1701, opened 2022-05-16
Connect logo and cover image to settings

sofiamatulis:
**PR Summary:** 

_Please include a short description (using non-technical terms, 1-2 sentences) about the changes you are introducing, what problem is being fixed and/or describe the benefit to merchants. This content will be used in our release notes for Dawn on [themes.shopify.com](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)._

**Why are these changes introduced?**

Fixes #0.

**What approach did you take?**

**Other considerations**

**Testing steps/scenarios**
- [ ] _List all the testing tasks that applies to your fix and help peers to review your work._

**Demo links**
_Please include a link to a demo store that includes preconfigured sections and settings to allow reviewers to easily test the features you are working on._

- [Store](url)
- [Editor](url)

**Checklist**
- [ ] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [ ] Notified the [help.shopify.com](https://help.shopify.com) team about updates to theme settings (current contact: Kimli)
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2307, opened 2023-02-15
Make header image preload and proper width

stufreen:
### PR Summary: 

Update header image tag to be crisper and preload.

We had a report from @melissaperreault of the logo image look fuzzy after this PR: https://github.com/Shopify/dawn/pull/1701

The previous logo image was rendered at various sizes which were not guaranteed to be a multiple of the logo's actual width. Also, the max size was 500px wide which is not wide enough for logos with a width of 260px or more. This changes it to always request 1) the logo width, and 2) 2X the logo width. This should result in a crisper image, including on high-DPI screens.

I also added the `preload: true` param to the image_tag. This should result in the logo being fetched more quickly by the browser. I think the header logo is a suitable place for this param since the logo will generally be small, above-the-fold and is critical branding for the merchant.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

Their logo will appear crisper and possibly load faster (esp. on slow networks).

### Testing steps/scenarios

- [ ] Open demo store
- [ ] Go to the Logo section of general theme settings. Upload a high resolution image (e.g 2000px wide)
- [ ] Visually examine the logo in the preview panel. It should look sharp on any screen.
- [ ] Inspect the img element in your browser dev tool. Check that it is requesting image size corresponding to the logo width setting.
- [ ] Also in the inspector, check the initial network request for the page. Look for a `link` header that includes the logo asset. The network waterfall should also show the logo image being fetched earlier in the waterfall.

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139395366934)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139395366934/editor)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -35,7 +35,7 @@
 
   {%- if section.settings.sticky_header_type == 'reduce-logo-size' -%}
     .scrolled-past-header .header__heading-logo {
-      width: 75%;
+      width: {{ settings.logo_width | times: 0.75 | round }}px;
     }
   {%- endif -%}
 
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -400,8 +400,8 @@
             {%- if settings.logo != blank -%}
               {%- assign logo_alt = settings.logo.alt | default: shop.name | escape -%}
               {%- assign logo_height = settings.logo_width | divided_by: settings.logo.aspect_ratio -%}
-              {%- assign sizes = settings.logo_width | append: "px" %}
-              {% capture widths %}{{ settings.logo_width }}, {{ settings.logo_width | times: 2 }}{% endcapture %}
+              {% capture sizes %}(max-width: {{ settings.logo_width | times: 2 }}px) 50vw, {{ settings.logo_width }}px{% endcapture %}
+              {% capture widths %}{{ settings.logo_width }}, {{ settings.logo_width | times: 1.5 | round }}, {{ settings.logo_width | times: 2 }}{% endcapture %}
               {{ settings.logo | image_url: width: 600 | image_tag:
                 class: 'header__heading-logo motion-reduce',
                 widths: widths,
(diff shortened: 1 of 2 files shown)


stufreen:
@eugenekasimov @ludoboludo Ready for another look

melissaperreault:
Thank you Stu!

When testing with a logo with smaller dimensions, we are preventing the logo to adopt the `Logo width` I set. [Video reference](https://screenshot.click/16-11-04b79-o7qgw.mp4)

eugenekasimov (dismissed):

stufreen:
@melissaperreault I see what you're saying. I just pushed a couple CSS changes. Now:
1. Desktop width setting will be respected for logos with any resolution/aspect ratio, except
2. At mobile sizes the logo may be resized to fit in the center 2 columns of the header

Have a try and let me know if it feels right to you now. If not let's pair on it!

eugenekasimov (approved):
🚀

sections/header.liquid
```diff
@@ -32,13 +32,10 @@
...
   }
 
-  .header__heading-logo {
-    max-width: {{ settings.logo_width }}px;
-  }
 
   {%- if section.settings.sticky_header_type == 'reduce-logo-size' -%}
     .scrolled-past-header .header__heading-logo {
-      width: 75%;
+      width: {{ settings.logo_width | times: 0.75 | round }}px;
```
ludoboludo, line 38: Since this doesn't reference the width from `.header__heading-logo` because we're only applying a max width on it, and instead referencing the setting. It doesn't really do anything on smaller screen when the space given to the logo itself is less than the setting's width. 

Say on mobile, the max amount of space the logo can be in is `150px`, and the `Desktop logo width` setting is at `190px` (at least `150px + (150 - 150 x 75%)`), then you end up having no `reduce logo size` effect.

stufreen:
@ludoboludo Ready for another look. I had to use a wrapper because, as mentioned before, putting `width: 100%` or `width: 75%` was causing the bug that @melissaperreault asked me to fix with the width not matching the `logo_width` setting.

ludoboludo (approved):
🚀

eugenekasimov (approved):

lezan:
Hello there,
do not want to be spammy with a thread issue, but I was checking the`image_tag`'s `preload` and I saw it is actually not working. I cannot find any mention to a `rel="preload" as="image"` of `header__heading-logo motion-reduce`. Can you confirm?
I find that because I was not able to preload an image in my theme so I moved to checked Dawn to be sure I was not missing something.
No sure I can pin you @melissaperreault @ludoboludo @stufreen, sorry if not.

krzksz:
Hey @lezan!

Using the `preload` option will make Shopify include the preload in the `Link` header in the main document response. It enables the resource to benefit from the Early Hints. If you're not seeing it there, it may be the case that you have a lot of preloads and we need we limit the number of files that end up in the header.

That said, I don't think the logo should be preloaded. Preloading should mainly be used for critical resources that can't be made early discoverable to the browser. A logo that uses a standard `` and is not lazyloaded will be discovered by the browser quite quickly.

I wrote about preloads and prefetches in the [article here](https://performance.shopify.com/blogs/blog/introduction-to-resource-hints#preloading-late-discovered-assets), you may find it useful.

krzksz:
As far as I know Dawn doesn't preload any files and the current limit is around 10, so you should at least be seeing some of them in the header.

PR #2337, opened 2023-02-28
Add animation script to theme layout

### LINKED ISSUE #2316: Add intersection observer for scroll-trigger animations
### Summary

Add an intersection observer to watch for when certain elements are scrolled into view.

### Details

We will be adding animations soon to fade in elements as they are scrolled into view. To power these animations we need to add an IntersectionObserver that observes certain elements.

@kjellr has created a prototype [here](https://github.com/Shopify/dawn/pull/2141/files#diff-ad413c098861cc8cc2776f05b60df136f3e0196456cf569b24a77e0ee3b32df5) that is a good starting point and also shows how the IntersectionObserver will fit in with the animations.

The general idea is that items we want to observer will be given a class such as `scroll-trigger`. The observer will watch those elements and add a new class such as `scrolled-into-view` to them.

This issue is to add the logic for the observer into a feature branch.

### Definition of done

- [ ] Add IntersectionObserver
- [ ] Consider how to handle elements that are above-the-fold, i.e. already scrolled into view when the page loads
- [ ] Investigate performance optimizations like removing the listener completely when there are no observed elements (i.e. all scrolled into view)
- [ ] Add a script for the editor to ensure it is triggered for new sections that are added

metamoni:

Fixes #2316 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/animations.js
+++ b/assets/animations.js
@@ -1,0 +1,25 @@
+function onIntersection(elements, observer) {
+  elements.forEach((element, index) => {
+    if (element.isIntersecting) {
+      const elementTarget = element.target;
+      elementTarget.classList.add('motion__trigger--animated');
+      elementTarget.setAttribute('style', `--animation-order: ${index};`);
+
+      observer.unobserve(elementTarget);
+    }
+  })
+}
+
+function initializeMotionTrigger() {
+  const motionTriggerElements = Array.from(
+    document.getElementsByClassName('motion__trigger')
+  );
+  if (motionTriggerElements.length === 0) return;
+
+  const observer = new IntersectionObserver(onIntersection, {
+    threshold: 0.5,
+  });
+  motionTriggerElements.forEach((element) => observer.observe(element));
+}
+
+initializeMotionTrigger();
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -32,6 +32,10 @@
     <script src="{{ 'constants.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
+    {%- if settings.motion_toggle -%}
+      <script src="{{ 'animations.js' | asset_url }}" id="AnimationScript" defer="defer"></script>
+    {%- endif -%}
+
     {{ content_for_header }}
 
     {%- liquid


assets/animations.js
```diff
@@ -0,0 +1,25 @@
...
+      const elementTarget = element.target;
+      elementTarget.classList.add('animation__trigger--active');
+      elementTarget.setAttribute('style', `--animation-order: ${index};`);
+
+      observer.unobserve(elementTarget);
+    }
+  })
+}
+
+function initializeAnimationTrigger() {
```
stufreen, line 13: I think we should avoid using the word "animation" in this module to keep a good separation of concerns. To me this observer is just about watching for when an element is scrolled into view, and has no opinion on what to do when that happens.

assets/animations.js
```diff
@@ -0,0 +1,25 @@
+function onIntersection(elements, observer) {
+  elements.forEach((element, index) => {
+    if (element.isIntersecting) {
+      const elementTarget = element.target;
+      elementTarget.classList.add('animation__trigger--active');
```
stufreen, line 5: Let's move `'animation__trigger--active'` and other such class names into a constant at the top of the file.

assets/animations.js
```diff
@@ -0,0 +1,25 @@
...
+      elementTarget.setAttribute('style', `--animation-order: ${index};`);
```
stufreen, line 6: What do you see this being used for? It might be helpful for cascading animations although those will depend on the screen size and layout.

assets/animations.js
```diff
@@ -0,0 +1,25 @@
...
+  );
+  if (animationTriggerElements.length === 0) return;
+
+  const observer = new IntersectionObserver(onIntersection, {
+    threshold: 0.5,
+  });
+  animationTriggerElements.forEach((element) => observer.observe(element));
+}
+
+initializeAnimationTrigger();
```
stufreen, line 25: Are we sure that the document has finished loading at this point?

assets/theme-editor.js
```diff
@@ -3,6 +3,18 @@ function hideProductModal() {
...
+  const scriptsToReload = document.querySelectorAll('[id^=AnimationsScript], [id^=EnableZoomOnHover]');
+  console.log('scripts', scriptsToReload);
+  if (!scriptsToReload.length) return;
+
+  scriptsToReload.forEach(script => {
+    const newScriptTag = document.createElement('script');
+    newScriptTag.setAttribute('src', script.src);
+    script.parentNode.replaceChild(newScriptTag, script);
+  });
+}
```
stufreen, line 16: Instead of this, could you just call `initializeAnimationTrigger` again?

If you wanted to more precise/targetted, you could update `initializeAnimationTrigger` to accept an optional query selector parameter that is a string. Then you could set up a second (or more) IntersectionObserver just for that section. E.g.

```
const containerEl = containerSelector ? containerClassName.querySelector(containerSelector) : document;
const animationTriggerElements = Array.from(
  containerEl.getElementsByClassName('animation__trigger')
);
```

sections/image-with-text.liquid
```diff
@@ -14,7 +14,11 @@
   }
 {%- endstyle -%}
 
-<div class="image-with-text image-with-text--{{ section.settings.content_layout }} page-width isolate{% if settings.text_boxes_border_thickness > 0 and settings.text_boxes_border_opacity > 0 and settings.media_border_thickness > 0 and settings.media_border_opacity > 0 %} collapse-borders{% endif %}{% unless section.settings.color_scheme == 'background-1' and settings.media_border_thickness > 0 and settings.text_boxes_shadow_opacity == 0 and settings.text_boxes_border_thickness == 0 or settings.text_boxes_border_opacity == 0 %} collapse-corners{% endunless %} section-{{ section.id }}-padding">
+{%- if settings.scroll_toggle -%}
+  <script src="{{ 'animations.js' | asset_url }}" id="AnimationsScript-{{ section.id }}"  defer="defer"></script>
```
stufreen, line 18: Question: I'm actually not sure this needs to be its own file. What do you think about adding it to `global.js`? We don't need to call it unless the setting is enabled and the file size is small.

assets/animations.js
metamoni: I know this is a pattern in some codebases, but I haven't seen it in Dawn. And since we're only using the class in one place, I don't see the benefit of it.

assets/animations.js
metamoni: This is for cascading animations, yes. And it's to replace [all this CSS](https://github.com/Shopify/dawn/blob/ebc1032459cabfeadbccc6fd3a9a43ae2a65bea5/assets/base.css#L3008-L3077) in the prototype.

assets/animations.js
metamoni: That's a fair comment, but it is the observer _for_ elements with the class name `animation__trigger`, so there's already a reference to that in the function. What would you suggest?

sections/image-with-text.liquid
metamoni: I didn't realise we could get access to the setting in our scripts. How do I do that?

assets/animations.js
stufreen: Unfortunately you'll need that CSS I think to accommodate different breakpoints. On product grids, for example, you will cascade 4 items at a time at desktop size, but 1 or 2 at mobile size.

Anyhow I suggest leaving this for a subsequent PR and focus on the intersection observer for now.

assets/animations.js
stufreen: There are two big benefits that I see here:
1) Let's you write self-documenting code. Instead of `elementTarget.classList.add('animation__trigger--active')` you'll have `elementTarget.classList.add(SCROLLED_INTO_VIEW_CLASSNAME)` which is more easily understandable.
2) It avoids nesting magic words and numbers deep in the code. Wherever we are passing configuration values, classnames, numbers, etc. we should really store them in variables at the top of classes/modules to avoid hunting around for them or guessing what the values may be.

assets/base.css
```diff
@@ -3116,3 +3116,64 @@ details-disclosure > details {
...
+}
+
+/* Respect the visitor's motion preferences. */
+@media (prefers-reduced-motion) {
+  .animation__trigger,
+  .animation__trigger--active {
+    opacity: 1;
+    animation: none;
+  }
+}
```
stufreen, line 3179: Let's leave the CSS for a subsequent PR - this one should be focussed on intersection observer and getting that pattern right

assets/theme-editor.js
```diff
@@ -26,6 +26,7 @@ document.addEventListener('shopify:block:deselect', function(event) {
 });
 
 document.addEventListener('shopify:section:load', () => {
+  initializeAnimationTrigger();
```
stufreen, line 29: Now that I think of it themes can use the [section rendering api](https://shopify.dev/docs/api/section-rendering) on storefronts so we have to watch for this event on storefronts as well, not just the editor. Let's add a `
document.addEventListener('shopify:section:load', () => {` to the scroll trigger code.

assets/animations.js
stufreen: Yeah I prefer `initializeScrollAnimationTrigger` - we could imagine using the intersection observer for other non-animation things in the future like analytics and lazy loading. Keeping it generic will make it more flexible.

layout/theme.liquid
```diff
@@ -32,6 +32,10 @@
     <script src="{{ 'constants.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
+    {%- if settings.animations_reveal_on_scroll -%}
+      <script src="{{ 'animations.js' | asset_url }}" id="AnimationsScript" defer="defer"></script>
```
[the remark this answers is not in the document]
kjellr: At the moment, the only other animation we're considering that uses any JS is the grow-on-scroll one. In a situation where the merchant has only selected one of these effects (scroll reveal or grow-on-scroll), I could see some benefit to not unnecessarily loading in the JS for the unused effect.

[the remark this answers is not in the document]
stufreen: I'm not 100% clear on why some of our JS is in separate `.js` files and most is in `global.js`. I suspect a mix of code organization and not wanting to load unnecessary JS.

If we want to keep all the animation JS in this file for organization reasons, we just need to be careful not to add scroll trigger class unnecessarily. In the next PR we can consider if we want to add the `scroll-animation__trigger` conditionally, i.e.
```liquid
class="{%- if settings.animations_reveal_on_scroll -%}scroll-animation__trigger{%- endif -%}"
```

LucasLacerdaUX (approved):
Looks good to me! :D

Issue #2341, opened 2023-02-28
[Slideshow] Add ambient movement to background


kjellr:
## Summary

Add ambient animation to the Slideshow section, as prototyped in #2229. 

## Details

This animation will be off by default. It should be implemented as a setting in a new dropdown, like so:
 
<img width="303" alt="Screenshot 2023-02-28 at 11 45 22 AM" src="https://user-images.githubusercontent.com/1202812/221921734-e4bc4249-c338-4fcf-b83b-baec73fe1f85.png">

Issue #2340, opened 2023-02-28
[Image with Text] Add ambient movement to background


kjellr:
## Summary

Add ambient animation to the Image with Text section, as prototyped in #2229. 

## Details

This animation will be off by default. It should be implemented as a section-specific setting in a new dropdown, like so:
 
<img width="303" alt="Screenshot 2023-02-28 at 11 45 22 AM" src="https://user-images.githubusercontent.com/1202812/221921339-08a536b5-a079-4fa8-b6a6-e800f0392ef8.png">

PR #2346, opened 2023-02-28
[Slideshow] Add fixed position background

### LINKED ISSUE #2345: [Slideshow] Add fixed position background
## Summary

Add a fixed position background option to the Slideshow section, as prototyped in #2229. 

## Details

This feature will be off by default. It should be implemented as a setting in a new dropdown, like so:
 
Screenshot 2023-02-28 at 2 05 49 PM

kjellr:
### PR Summary: 

Adds a fixed position background option to the Slideshow section. 

https://user-images.githubusercontent.com/1202812/221970091-6166b1e5-e574-4623-910a-49595553bcd2.mp4

Closes #2345.

This implements a new dropdown menu [as shown in the ticket](https://github.com/Shopify/dawn/issues/2345). When "Fixed background position" is selected, a CSS class is applied to the media elements within the Slideshow slides. This class sets the position of the media element to `fixed`, and uses `clip-path(0)` on the parent to ensure the image doesn't bleed outside of its intended container. 

The CSS for this is defined in `base.css` so that we can theoretically reuse it for the Image with Text (https://github.com/Shopify/dawn/issues/2340) and Image Banner (https://github.com/Shopify/dawn/issues/2341) sections.

### Other considerations

 ⚠ We may want to re-evaluate this approach, since it's causing some issues in Chrome. Sometimes, the pseudo element and/or slideshow content doesn't appear correctly when you scroll into it or advance onto the next slide. Scrolling a little or moving the mouse near the slideshow fixes this. From what I can infer, this is related to the use of `clip-path`. These issues are not present in Safari or Firefox. 😞 

If for some reason this approach falls through, we could also explore setting the `background-image` of the media container, and using `background-attachment: fixed`. This should avoid the issues, but is far more verbose and likely less performant, as it eliminates the ability for us to use `srcset`. 😕

<details>
<summary>📹 Example</summary>

https://user-images.githubusercontent.com/1202812/221972454-13d971eb-eb8c-4f38-b487-05ec369184f3.mp4

</details>

None. This is an optional feature. 

### Testing steps/scenarios

1. [Visit the demo store](https://os2-demo.myshopify.com/admin/themes/139452514326/editor).
2. Add and select a Slideshow section.
3. Look for the new "Image behavior" dropdown underneath the new "Animations" header in the sidebar.
4. Select "Fixed position background" from that dropdown.
5. Add various images, verify that they appear as expected.
6. Try various breakpoints, verify that they appear as expected.

### Checklist


assets/base.css
```diff
@@ -3048,3 +3048,14 @@ details-disclosure > details {
     outline: transparent solid 1px;
   }
 }
+
+/* Fixed background */
+
+.animate--fixed {
```
metamoni, line 3054: Regarding the bug you're seeing on Chrome, I tried [this solution](https://stackoverflow.com/questions/70296340/problems-with-css-clip-path-in-google-chrome) and it seemed to work for me. It's worth giving it a try and testing it more extensively.

kjellr: Ah yes, thanks for finding that. I thought my days of using the `translateZ(0)` hack were over, but it appears to still be relevant! 😅 I don't love kicking the GPU acceleration unnecessarily, but it may be our best option for now. 

I pushed this in a64ea39 for now, but used `-webkit-transform` instead of `transform` (so that it won't trigger unnecessarily for FF), and I also added a code comment explaining it.

stufreen (commented):
Not working for me unfortunately. I would be OK to use a `background-image` in this case - we can make it responsive using either several divs or the newish [image-set](https://developer.mozilla.org/en-US/docs/Web/CSS/image/image-set) CSS function. We will need to add some stuff for a11y if we go that route.

assets/base.css
kjellr: 😩 Oh whoops. You're correct. We'll need to re-evaluate this.

kimberlyoleiro:
Not working for me either. Both Chrome & Safari. No fixed background effect appearing

DavidSheerin:
Hey 👋  Just seen this issue and thought I'd toss in my two cents from a support perspective 🪙 🪙  

For the css property [background-attachment: fixed;](https://github.com/Shopify/dawn/pull/2346/commits/5fd57ead18fd46d48d883057b26a3c7406bb6c7f#diff-36ba6ae7db38743c020db0e42963e2d31e6e1b8bd758e3f95e23e79125050f31R88) that is being proposed, there are [limitations with it's usage in Safari on iOS](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment#browser_compatibility). When this property is used, it often creates frustration and/or confusion for merchants which we tend to see in tickets escalated to Theme Support. A [current example](https://github.com/Shopify/dawn-internal/issues/838) within Dawn is where the fixed gradient doesn't get applied to the header or footer on mobile and iPad. In a previous review, we found that this particular issue made up ~5% of issue tickets that had been escalated to our team. I suspect the fixed image effect would be more popular than the gradient and as a result, create more support debt - just something to consider if implementing. 

A similar effect was [previously used in the Narrative theme](https://github.com/Shopify/narrative/pull/294) that caused a lot of frustration for merchants at the time and to overcome the Safari issue on iOS, a function was added to the js to check the device and a class as a fallback. I don't believe it solved the problem as there were still [further issues](https://github.com/Shopify/narrative/issues/1071) reported and it's implementation goes against the zero js principle but thought I'd share for some added context.

metamoni (commented):
Tested this on Safari, Firefox, Edge, Chrome, with different breakpoints. It seems to work fine for me!

LucasLacerdaUX:
Closing this as it's inactive/old.

PR #2348, opened 2023-02-28
Add SKUs to dataLayer ecommerce objects

thomasjlee:
### PR Summary: 

- Adds a `skus` property to the ecommerce objects pushed to the dataLayer

(Technically, this is only used when the `purchase` event occurs, but it could be useful for other events in the checkout flow.)

This is needed to enable conversion tracking in Google Ads via Google Tag Manager.

By adding `skus`, we can access them in Google Tag Manager:

image

We can then create a GTM Trigger to check for the presence of a specific SKU. For example, this Trigger can be used to fire a Tag when a customer is checking out with an X-Carve Pro:

image

With this Trigger, we can create the appropriate Google Ads Conversion Tracking Tag so that we can track XCP conversions in Google Ads and measure the performance of our ad campaigns.

Apparently it's not possible to directly assign an array to a liquid variable, so I map over `checkout.line_items` to concatenate the SKUs into a comma-separated string. I pass that variable into each snippet, and from there we use `String.prototype.split` to cast it to an Array of strings.

image


thomasjlee:
Wrong base branch

PR #2349, opened 2023-03-01
Improve image sizes in the multicolumn section

### LINKED ISSUE #2298: Improper image sizes rendering in Multicolumn sections.
## Describe the enhancement you'd like
This issue is a part of a global problem described in #1944. To split a big issue into small parts I will be focusing here on image sizes we render particularly in **multicolumn** sections.

The general issue is we almost always render a bigger size of an image than we really need. That said, our theme performance is worse than it could be. Why does it happen? I can see three main reasons.

Our current logic for choosing a size as a reference
```
{%- capture sizes -%}
   (min-width: 990px) {% if section.blocks.size <= 2 %}710px{% else %}550px{% endif %}, (min-width: 750px) {% if 
   section.blocks.size == 1 %}710px{% else %}550px{% endif %}, calc(100vw - 30px)
{%- endcapture -%}
{{
    block.settings.image
    | image_url: width: 1420
    | image_tag:
        loading: 'lazy',
        sizes: sizes,
        widths: '275, 550, 710, 1420',
        class: 'multicolumn-card__image'
}}
```

1. When we decide what image size to render we don't account for DPR (Device pixel ratio). Since most of the modern devices have DPR >= 2 those sizes that provide as a condition in `sizes` are multiplied by 2 or more. As a result we always get the source image size at least 2 times bigger than we actually expect.
2. Choosing a source size for an image we only account the browser window size and a number of columns. However, we don't account for the **Image width** and/or  **Number of columns on mobile** settings in the Multicolumn sections. Those settings take a significant effect on the image rendered size and cannot be ignored.
3. Let's say we want to set the **Image ratio** setting in the Multicolumn sections as Portrait. If the original image has a landscape ratio that means we load a wide bigger picture and then crop it on a frontend side. The better solution would be to do it on a server.

eugenekasimov:
### PR Summary: 

This PR improves the way images are loaded in the multicolumn section. 

Fixes #2298.

I rewrote the logic of how we create our `srcset`. Now it accounts for the page width, desktop/tablet/mobile layouts, number of columns, the size of the image(full, half or third). The logic is based on the most common pixel density 2 and I also added options for pixel density 1. 

All calculations are based on the maximum possible image sizes for each media query and don't account for paddings/margin etc.

### Other considerations

- Maybe we could add more options for pixel density 3 since many latest mobile devices have it. 
- We could also add more breakpoints to cover more scenarios in terms of screen size.  

There is no visual impact. 

### Testing steps/scenarios

- [ ] Set the multicolumn section with images.
- [ ] Test different settings such a different image with and/or different number of columns and using the devtool check the intrinsic size and make sure it calculates right.  

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139456806934)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139456806934/editor)

### Checklist

### DIFF
--- a/sections/multicolumn.liquid
+++ b/sections/multicolumn.liquid
@@ -83,7 +83,8 @@
                   >
                     {%- liquid
                       assign numberOfColumns = section.settings.columns_desktop
-                      assign maxWidth = settings.page_width | times: 2
+                      assign numberOfColumnsMobile = section.settings.columns_mobile
+                      assign maxImageWidth = settings.page_width | times: 2
                       if section.settings.image_width == 'half'
                         assign imageWidth = 0.5
                       elsif section.settings.image_width == 'third'
@@ -93,17 +94,26 @@
                       endif
                     -%}
                     {%- capture widths -%}
-                      {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 | times: 2 }}, {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }}
+                      {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 | times: 2 }},
+                      {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }},
+                      {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }},
+                      {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
+
                     {%- endcapture -%}
-                    {% capture sizes %}(min-width: 990px) {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }}px {% endcapture %}
+                    {% capture sizes %}
+                      (min-width: 990px) {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }}px,
+                      (min-width: 750px) {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px,
+                      (min-width: 375px) {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px,
+                      {{ 375 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px
+                    {% endcapture %}
                     {{
                       block.settings.image
-                      | image_url: width: maxWidth
-                      | image_tag:
-                        loading: 'lazy',
-                        widths: widths,
-                        sizes: sizes,
-                        class: 'multicolumn-card__image'
+                      | image_url: width: maxImageWidth
+                      | image_tag: loading: 'lazy', widths: widths, sizes: sizes, class: 'multicolumn-card__image'
                     }}
                   </div>
                 </div>
--- a/sections/multicolumn.liquid
+++ b/sections/multicolumn.liquid
@@ -102,7 +102,7 @@
                       | image_url: width: max_image_width
                       | image_tag:
                         loading: 'lazy',
-                        widths: '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1500, 2000, 2500, 3000',
+                        widths: '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1250, 1500, 1750, 2000, 2250, 2500, 2750, 3000, 3250',
                         sizes: sizes,
                         class: 'multicolumn-card__image'
                     }}
--- a/sections/multicolumn.liquid
+++ b/sections/multicolumn.liquid
@@ -84,6 +84,9 @@
                     {%- liquid
                       assign number_of_columns = section.settings.columns_desktop
                       assign number_of_columns_mobile = section.settings.columns_mobile
+                      assign grid_space_desktop = number_of_columns | minus: 1 | times: settings.spacing_grid_horizontal | plus: 100 | append: 'px'
+                      assign grid_space_tablet = number_of_columns_mobile | minus: 1 | times: settings.spacing_grid_horizontal | plus: 100 | append: 'px'
+                      assign grid_space_mobile = number_of_columns_mobile | minus: 1 | times: settings.spacing_grid_horizontal | divided_by: 2 | plus: 30 | append: 'px'
                       if section.settings.image_width == 'half'
                         assign image_width = 0.5
                       elsif section.settings.image_width == 'third'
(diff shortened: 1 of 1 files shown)


sections/multicolumn.liquid
```diff
@@ -80,18 +80,39 @@
...
+                      {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
+
                     {%- endcapture -%}
+                    {% capture sizes %}
+                      (min-width: 990px) {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }}px,
+                      (min-width: 750px) {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px,
+                      (min-width: 375px) {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px,
+                      {{ 375 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}px
```
stufreen, line 110: The only issue I see with this is that it will always jump to the next size... E.g. if the screen width is 375 it will use an image that is 749w (assuming imageWidth of 1).

I think you can replace these three lines with:
```
calc(100vw * {{ imageWidth | divided_by: numberOfColumnsMobile }})
```

sections/multicolumn.liquid
```diff
@@ -80,18 +80,39 @@
...
+                    -%}
+                    {%- capture widths -%}
+                      {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 | times: 2 }},
+                      {{ settings.page_width | times: imageWidth | divided_by: numberOfColumns | round: 0 }},
+                      {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 989 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }},
+                      {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 749 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
+                      {{ 374 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
```
stufreen, line 103: I see the approach here (374 up to first breakpoint, 749 up to second breakpoint, etc.) but I don't think we need to be quite so targetted with the widths. The `sizes` are the important part. As long as there are enough sizes to cover most cases then we are good - the browser will select the best one based on `sizes`.

I would replace with the actual breakpoints to make it simple:
```
                      {{ 990 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
                      {{ 990 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }},
                      {{ 750 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
                      {{ 750 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
                      {{ 375 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 | times: 2 }},
                      {{ 375 | times: imageWidth | divided_by: numberOfColumnsMobile | round: 0 }}
```

Or you could even do something arbitrary like `50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1500, 2000` which will always have a close width.

stufreen (changes requested):
Way better than before! It looks like the previous version was based on the total number of blocks, which just seems wrong. There must have been an update but the `sizes` property wasn't updated.

sections/multicolumn.liquid
kmeleta: I think we can be more flexible/accurate here without much more complexity if we apply the same sort of math to css calculations replacing fixed values like `989` with `100vw`. This will essentially create a dynamic value that also factors in the viewport width, rather than just the max breakpoint width.

Also I think we often define 3 breakpoints and a non-specific mobile size that is essentially a max-width: 749px. I think if you make the change above you'll find you don't need the min-width 375 breakpoint
- `min-width: {{ settings.page_width}}` (image sizes won't vary above this breakpoint and can use a fixed sized like you have currently)
- `min-width: 990px` (image sizes vary with viewport between 990 and 1200)
- `min-width: 750px` (image sizes vary with viewport between 750 and 990)

sections/multicolumn.liquid
kmeleta: Missing a comma in one of these.

kmeleta: I don't like hardcoded widths that aim to be accurate (what we historically do). They take effort to decide on and are easily rendered obsolete with the tweak of one setting or css style. But I could be sold on the arbitrary hard coded list Stu suggested because it can more or less be repurposed to all sections without having to give the values a second thought.

With the current approach you have here the values are at least dynamic and not made obsolete as easily. It also takes the dozens of possible combinations of variables and providing fairly accurate widths without needing to actually add dozens of hard coded values and noise to the dom.

Though there a few things I don't really like now that I'm seeing this approach in practice..
- We can end up with kind of redundant values that are only a few pixels apart
- We can't really ensure these are in a sensible order so the dom output is actually pretty difficult to parse while debugging (maybe with more liquid shenanigans)
- More cognitive load in the liquid vs a simple comma separated string

sections/multicolumn.liquid
eugenekasimov: I decided to go with what Stu suggested 👆.

sections/multicolumn.liquid
```diff
@@ -81,17 +81,29 @@
...
+                      (min-width: 990px) {{ settings.page_width | times: image_width | divided_by: number_of_columns | round: 0 }}px,
+                      calc(100vw * {{ image_width }} / {{ number_of_columns_mobile }})
+                    {% endcapture %}
                     {{
                       block.settings.image
-                      | image_url: width: 1420
+                      | image_url: width: max_image_width
                       | image_tag:
                         loading: 'lazy',
+                        widths: '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1500, 2000, 2500, 3000',
```
kmeleta, line 105: Interested in what other devs think about going this route as it should be something we do moving forward. Assuming no one objects to it, I wonder if having a smaller increment than 500px for the larger images is worthwhile since those are the ones with the most cost. Maybe 250px or something?

stufreen (dismissed):

sections/multicolumn.liquid
eugenekasimov: Thanks for the feedback. I added a few more numbers. Also increased the largest one to 3250 to cover an edge case when page-width is 1600px and column number for desktop is 1.

sections/multicolumn.liquid
```diff
@@ -81,17 +81,29 @@
...
+                      endif
+                    -%}
+                    {% capture sizes %}
+                      (min-width: 990px) {{ settings.page_width | times: image_width | divided_by: number_of_columns | round: 0 }}px,
+                      calc(100vw * {{ image_width }} / {{ number_of_columns_mobile }})
+                    {% endcapture %}
                     {{
                       block.settings.image
-                      | image_url: width: 1420
+                      | image_url: width: max_image_width
```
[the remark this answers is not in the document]
eugenekasimov: I'm okay to hardcode 3200 as the last one instead of 3250 or replace the last one after 3000 with `max_image_width`. Thoughts?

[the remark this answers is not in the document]
kmeleta: I don't feel strongly. I could see breaking the arbitrary increment amount for the last value when we know the actual maximum we'll need. So we could update both to be 3200, rather than using 3250 just because we're incrementing the others by 250.

sections/multicolumn.liquid
```diff
@@ -81,17 +81,29 @@
...
+                        assign image_width = 0.5
+                      elsif section.settings.image_width == 'third'
+                        assign image_width = 0.33
+                      else
+                        assign image_width = 1
+                      endif
+                    -%}
+                    {% capture sizes %}
+                      (min-width: 990px) {{ settings.page_width | times: image_width | divided_by: number_of_columns | round: 0 }}px,
+                      calc(100vw * {{ image_width }} / {{ number_of_columns_mobile }})
```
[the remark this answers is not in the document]
eugenekasimov: I reviewed everything again and decided to add paddings. I think it's worth it.

sections/multicolumn.liquid
```diff
@@ -81,17 +81,30 @@
...
+                      if section.settings.image_width == 'half'
+                        assign image_width = 0.5
+                      elsif section.settings.image_width == 'third'
+                        assign image_width = 0.33
+                      else
+                        assign image_width = 1
+                      endif
+                    -%}
+                    {% capture sizes %}
+                      (min-width: {{ settings.page_width }}px) calc(({{ settings.page_width }} - 100 - ({{ settings.spacing_grid_horizontal }} * ({{ number_of_columns }} - 1))) * {{ image_width }} /  {{ number_of_columns }}),
```
stufreen, line 96: Can you put the spacing in variables? Having good variable names makes the code more self documenting:
```
                      assign grid_space_desktop = number_of_columns | minus: 1 | times: settings.spacing_grid_horizontal | plus: 100 | append: "px"
                      assign grid_space_tablet = number_of_columns_mobile | minus: 1 | times: settings.spacing_grid_horizontal | plus: 100 | append: "px"
                      assign grid_space_mobile = number_of_columns_mobile | minus: 1 | times: settings.spacing_grid_horizontal | plus: 30 | append: "px"
                    -%}
                    {% capture sizes %}
                      (min-width: {{ settings.page_width }}px) calc(({{ settings.page_width }}px - {{ grid_space_desktop }}) * {{ image_width }} / {{ number_of_columns }}),
                      (min-width: 990px) calc((100vw - {{ grid_space_desktop }}) * {{ image_width }} / {{ number_of_columns }}),
                      (min-width: 750px) calc((100vw - {{ grid_space_tablet }}) * {{ image_width }} / {{ number_of_columns_mobile }}),
                      calc(((100vw - {{ grid_space_mobile }}) * {{ image_width }} / {{ number_of_columns_mobile }})
                    {% endcapture %}
```

stufreen (approved):
Working really well from my testing - just a style comment

stufreen (approved):

melissaperreault (approved):
🎉 I tried to see if any of this change would be perceivable with a before and after comparison. ([Video reference](https://screenshot.click/10-46-gi8t2-i95vq.mp4)) Looks good!

PR #2342, opened 2023-02-28
[Image banner] Add ambient movement to background

kjellr:
### PR Summary: 

Adds an option for ambient animation to the Image Banner section's background image. 

https://user-images.githubusercontent.com/1202812/221939239-9930e1d4-f520-420b-93c2-4f5baf8001ca.mov

### Why are these changes introduced?

Closes #2339. 

### What approach did you take?

This implements a new dropdown menu [as shown in the ticket](#2339). When "Ambient movement" is selected, a CSS class is applied to the media elements within the Image Banner, triggering a long-duration CSS animation that slowly moves the image around within its frame. The CSS for this is defined in `base.css` so that we can theoretically reuse it for the Image with Text (#2340) and Slideshow (#2341) sections. 

As this feature is purely aesthetic, it supports the visitor's `prefers-reduced-motion` preference.

### Visual impact on existing themes

None, this is an optional setting that only adds new functionality. 

### Testing steps/scenarios

1. [Visit the demo store](https://os2-demo.myshopify.com/admin/themes/139450515478/editor). 
2. Add and select an Image Banner. 
3. Look for the new "Image behavior" dropdown underneath the new "Animations" header in the sidebar. 
4. Select "Ambient movement" from that dropdown. 
5. Add various images, verify that they appear as expected.
6. Try various breakpoints, verify that they appear as expected.

### Checklist
- [x] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [x] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [x] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [x] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [x] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2362, opened 2023-03-06
add scroll animation to image banner

metamoni:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -3120,3 +3120,71 @@ details-disclosure > details {
 .rte blockquote > * {
   margin: -0.5rem 0 -0.5rem 0;
 }
+
+/* Animations */
+
+.scroll-animation__trigger {
+  transition: opacity 0.55s ease-in-out;
+  opacity: 0;
+}
+
+.scroll-animation__trigger--active {
+  animation: 0.55s ease-in-out slideIn forwards;
+  animation-delay: calc(var(--animation-order) * 100ms);
+}
+
+@keyframes slideIn {
+  from {
+    transform: translateY(10vh);
+    opacity: 0;
+  }
+  to {
+    transform: translateY(0);
+    opacity: 1;
+  }
+}
+
+/* Section-specific motion behavior. */
+
+.banner.scroll-animation__trigger,
+.banner.scroll-animation__trigger--active {
+  opacity: 1;
+  animation: none;
+}
+
+.banner__media.scroll-animation__trigger,
+.banner.scroll-animation__trigger:after {
+  transition: opacity var(--duration-long) ease-in-out;
+}
+
+.banner__media.scroll-animation__trigger--active,
+.banner.scroll-animation__trigger--active:after {
+  animation: none;
+  opacity: 1;
+}
+
+.slideshow__slide:not(:first-of-type) .banner__media.scroll-animation__trigger {
+  opacity: 1;
+}
+
+/* Async loading prevents animation for Related Product cards. */
+.related-products .grid__item .scroll-animation__trigger {
+  opacity: 1;
+  animation: none;
+}
+
+/* Turn off animations if JS is turned off. */
+.no-js .scroll-animation__trigger,
+.no-js .scroll-animation__trigger--active {
+  opacity: 1;
+  animation: none;
+}
+
+/* Respect the visitor's motion preferences. */
+@media (prefers-reduced-motion) {
+  .scroll-animation__trigger,
+  .scroll-animation__trigger--active {
+    opacity: 1;
+    animation: none;
+  }
+}
--- a/sections/image-banner.liquid
+++ b/sections/image-banner.liquid
@@ -24,17 +24,22 @@
 {%- endif -%}
 
 {%- style -%}
-  #Banner-{{ section.id }}::after {
+  #Banner-{{ section.id }}.scroll-animation__trigger::after {
+    opacity: 0;
+  }
+
+  #Banner-{{ section.id }}::after,
+  #Banner-{{ section.id }}.scroll-animation__trigger--active::after {
     opacity: {{ section.settings.image_overlay_opacity | divided_by: 100.0 }};
   }
 {%- endstyle -%}
 
 <div
   id="Banner-{{ section.id }}"
-  class="banner banner--content-align-{{ section.settings.desktop_content_alignment }} banner--content-align-mobile-{{ section.settings.mobile_content_alignment }} banner--{{ section.settings.image_height }}{% if section.settings.stack_images_on_mobile and section.settings.image != blank and section.settings.image_2 != blank %} banner--stacked{% endif %}{% if section.settings.image_height == 'adapt' and section.settings.image != blank %} banner--adapt{% endif %}{% if section.settings.show_text_below %} banner--mobile-bottom{%- endif -%}{% if section.settings.show_text_box == false %} banner--desktop-transparent{% endif %}"
+  class="banner banner--content-align-{{ section.settings.desktop_content_alignment }} banner--content-align-mobile-{{ section.settings.mobile_content_alignment }} banner--{{ section.settings.image_height }}{% if section.settings.stack_images_on_mobile and section.settings.image != blank and section.settings.image_2 != blank %} banner--stacked{% endif %}{% if section.settings.image_height == 'adapt' and section.settings.image != blank %} banner--adapt{% endif %}{% if section.settings.show_text_below %} banner--mobile-bottom{%- endif -%}{% if section.settings.show_text_box == false %} banner--desktop-transparent{% endif %}{% if settings.animations_reveal_on_scroll %} scroll-animation__trigger{% endif %}"
 >
   {%- if section.settings.image != blank -%}
-    <div class="banner__media media{% if section.settings.image == blank and section.settings.image_2 == blank %} placeholder{% endif %}{% if section.settings.image_2 != blank %} banner__media-half{% endif %}">
+    <div class="banner__media media{% if section.settings.image == blank and section.settings.image_2 == blank %} placeholder{% endif %}{% if section.settings.image_2 != blank %} banner__media-half{% endif %}{% if settings.animations_reveal_on_scroll %} scroll-animation__trigger{% endif %}">
       {%- liquid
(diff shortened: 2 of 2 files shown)


kjellr:
Just leave a few UX notes for implementation here: 

For the Image Banner, the latest design I have working in #2141 fades in the `.banner__media` and its overlay shade, and simultaneously fades + slides-in the `banner__content`. The design reasoning behind that is that it's potentially overwhelming to have the entire full-screen background slide in (especially on initial page load). It also works well as a way to introduce some variety to the scroll experience when this section is lower on the page. 

For the Image with Text block, let's do the usual fade + slide-in for the entire section (This is how it works in #2141 today). I'd like to avoid animating each side individually, since these two areas sometimes appear as a unit. 

Here's a quick video of these two behaving this way: 

https://user-images.githubusercontent.com/1202812/223225374-408f2d34-3269-477c-a30c-b7079237faf6.mp4

sections/image-banner.liquid
```diff
@@ -24,17 +24,22 @@
 {%- endif -%}
 
 {%- style -%}
-  #Banner-{{ section.id }}::after {
+  #Banner-{{ section.id }}.scroll-animation__trigger::after {
+    opacity: 0;
+  }
```
kjellr, line 29: Not a blocking note, but I think this rule could technically be a more generalized one that isn't tied to this specific ID. On the other hand, for code organization's sake, it's kind of nice to keep it here next to the rule below.

assets/base.css
```diff
@@ -3120,3 +3120,60 @@ details-disclosure > details {
 .rte blockquote > * {
   margin: -0.5rem 0 -0.5rem 0;
 }
+
+/* Animations */
+
+.scroll-animation__trigger {
+  transition: opacity 0.55s ease-in-out;
```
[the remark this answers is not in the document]
stufreen: For easing, I think `ease-out` will look smoother. It's subtle, but we want the effect of the elements rushing in and settling into place from offscreen. With `ease-in-out` the startup also has some inertia, which is not what we want here IMO.

assets/base.css
```diff
@@ -3120,3 +3120,60 @@ details-disclosure > details {
...
+  opacity: 0;
+}
+
+.scroll-animation__trigger--active {
+  animation: 0.55s ease-in-out slideIn forwards;
+}
```
stufreen, line 3133: Do we need generic rules for `.scroll-animation__trigger` and `.scroll-animation__trigger--active` when there are specific rules below? I suggest _only_ using the specific rules on a section-by-section basis. I feel like `scroll-animation__trigger--active` shouldn't have any styles on its own.

assets/base.css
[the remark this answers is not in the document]
kjellr: Both those changes are ok on my end.

sections/image-banner.liquid
```diff
@@ -31,10 +31,10 @@
...
 >
   {%- if section.settings.image != blank -%}
-    <div class="banner__media media{% if section.settings.image == blank and section.settings.image_2 == blank %} placeholder{% endif %}{% if section.settings.image_2 != blank %} banner__media-half{% endif %}">
+    <div class="banner__media media{% if section.settings.image == blank and section.settings.image_2 == blank %} placeholder{% endif %}{% if section.settings.image_2 != blank %} banner__media-half{% endif %}{% if settings.animations_reveal_on_scroll %} scroll-animation__trigger{% endif %}">
```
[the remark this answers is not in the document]
kjellr: In this particular case, I prefer having two triggers. Since image banner can be quite large, you may not see the content transition at all if we tie them together.

metamoni:
> Recap of what I propose
> No overlap content layout: Reveal image and content container background together, than the content blocks
> Overlap content layout: Reveal each side individually

@melissaperreault, @kjellr this sounds like a great idea and I don't image it would be difficult to do. However, we don't have the "cascading" CSS in this PR, and adding that would change the scope a bit. The CSS that will handle delayed animation will be part of [this PR](https://github.com/Shopify/dawn/pull/2372) (draft) and will need to be reviewed and tested before we can introduce it into other sections.

I'd like to keep PRs small and perfect the cascading animation first, so let's leave the overlap content layout for a separate PR. When we come back to it, it should be a very short and easy task.

assets/base.css
metamoni: I think the Image banner is the only one that has specific styles, and these base classes cover everything else. Will we have more specific styles, @kjellr?

kjellr:
> I'd like to keep PRs small and perfect the cascading animation first, so let's leave the overlap content layout for a separate PR. When we come back to it, it should be a very short and easy task.

Sounds good, thanks @metamoni!

assets/base.css
kjellr: Yeah, most of our animations are shared across sections, so I think keeping this rule generic makes sense. The [Image Banner](https://github.com/Shopify/dawn/blob/8fcc99a8c31a7b5afb9189a607ad25643ff0fde2/assets/base.css#L3092-L3107) and [Slideshow](https://github.com/Shopify/dawn/blob/8fcc99a8c31a7b5afb9189a607ad25643ff0fde2/assets/base.css#L3109-L3111) are the only ones that I assigned specific styles to in my demo.

assets/section-image-banner.css
```diff
@@ -452,3 +452,20 @@
...
+
+.banner.scroll-animation__trigger,
+.banner.scroll-animation__trigger--active {
+  opacity: 1;
+  animation: none;
+}
+
+.banner__media.scroll-animation__trigger,
+.banner.scroll-animation__trigger:after {
+  transition: opacity var(--duration-long) ease-in-out;
```
stufreen, line 464: Should be `ease-out` easing as well

assets/base.css
```diff
@@ -3120,3 +3120,41 @@ details-disclosure > details {
 .rte blockquote > * {
   margin: -0.5rem 0 -0.5rem 0;
 }
+
+/* Animations */
+
+.scroll-animation__trigger {
```
stufreen, line 3126: Let's align on a BEM block name for animations if we're going to have global ones. I think in Kjell's [other ambient motion PR](https://github.com/Shopify/dawn/pull/2342/files#diff-00b03bd62b155daa89082ba52094c182e46422a9e28cd3ec601b3a2291b52718R3126) he used `animate` as the block and `ambient` as the modifier, e.g. `animate--ambient`. Let's be consistent between the two.

I had originally thought the scroll trigger JS and the animation CSS should be separate. In other words, the scroll trigger class does not have any styling on its own, and the styling is in another class. So the rule here would be something like:
```css
.scroll-trigger--active.animate--slide-in {
```

I still think this is the best approach since it will let us have different animations depending on the context. But if you want to keep it simple for now you could just rename this `animate__slide-in` and `animate__slide-in--active` and modify `animations.js` to match.

assets/base.css
```diff
@@ -3120,3 +3120,41 @@ details-disclosure > details {
...
+}
+
+/* Respect the visitor's motion preferences. */
+@media (prefers-reduced-motion) {
+  .scroll-animation__trigger,
+  .scroll-animation__trigger--active {
+    opacity: 1;
+    animation: none;
+  }
+}
```
stufreen, line 3160: Nit: Keep these rules next to `.scroll-animation__trigger` rules in the file since they are so closely related.

assets/base.css
LucasLacerdaUX: +1 to use separate classes for animation vs trigger. I was writing a similar suggestion.

The `.scroll-animation__trigger--active` name doesn't suggest that it's a slideIn animation and from the JS code, its purpose is just to be an indicator if the element has been in the viewport or not. I would opt for `.scroll-trigger--active.animate--slide-in` even if we don't really need other animation types right now as it keeps the `animations.js` as generic as it was designed to be.

assets/base.css
```diff
@@ -3120,3 +3120,41 @@ details-disclosure > details {
...
+
+/* Turn off animations if JS is turned off. */
+.no-js .scroll-animation__trigger,
+.no-js .scroll-animation__trigger--active {
+  opacity: 1;
+  animation: none;
+}
+
+/* Respect the visitor's motion preferences. */
+@media (prefers-reduced-motion) {
```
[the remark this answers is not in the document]
metamoni: I think this was based on our original plans to enable animations by default, but we don't have a final decision on this yet. Since we default to animations, unless the user has actively done something to turn them off, I find it more readable to follow the pattern where we use `@media (prefers-reduced-motion)`, but I don't have a strong opinion on it either way and I'm happy to change it

melissaperreault:
Thank you @metamoni, I opened the issue here https://github.com/Shopify/dawn/issues/2375

assets/section-image-banner.css
```diff
@@ -452,3 +452,20 @@
...
 }
+
+.banner.scroll-trigger--active,
+.banner.animate--slide-in {
+  opacity: 1;
+  animation: none;
+}
+
+.banner__media.scroll-trigger--active,
+.banner.scroll-trigger--active:after {
```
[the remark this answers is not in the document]
metamoni: I think it's working as expected after the latest changes:

https://user-images.githubusercontent.com/22390758/224655161-51b925c4-6e08-496a-81b9-7b1766822828.mp4

assets/section-image-banner.css
```diff
@@ -452,3 +452,15 @@
...
+  .banner__media.scroll-trigger,
+  .banner.scroll-trigger:after {
+    opacity: 0;
+    transition: opacity var(--duration-long) ease-out;
+  }
+
+  .banner__media.scroll-trigger--active {
+    opacity: 1;
+  }
+}
```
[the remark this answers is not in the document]
kjellr: @metamoni, if we add the `.banner.scroll-trigger` and `.banner-scroll-trigger--active` classes here, it removes the initial flash in my testing.

stufreen (approved):
LGTM! 🎩 on Chrome and Safari.

LucasLacerdaUX (approved):
LGTM :D

PR #2364, opened 2023-03-06
Fix the Page section's width. 

### LINKED ISSUE #2359: Page section has lost its width reference

Page section has lost its width reference, becoming too wide for readable content. [Video reference](https://screenshot.click/03-31-q8cbs-2dvut.mp4)

Thanks @melissaperreault  for catching this.

eugenekasimov:
### PR Summary: 

Fixes an issue when the page sections looses its width reference. 

Fixes #2359.

I lowered specificity for `.page-width:not(.drawer-menu)` class to its original one (0-1-0) by changing it to `.page-width:where(:not(.drawer-menu))`

The page section should be looking narrow now as it used to be. 

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139496783894/editor)

### Checklist


assets/base.css
```diff
@@ -248,7 +248,7 @@ html.no-js .no-js-hidden {
   padding: 0 1.5rem;
 }
 
-.page-width:not(.drawer-menu) {
+.page-width:not(:where(.drawer-menu)) {
```
ludoboludo, line 251: I feel like it's even easier to read like this: 
```suggestion
.page-width:where(:not(.drawer-menu)) {
```

eugenekasimov: Good suggestion, Ludo. Thank you.

melissaperreault (approved):
Thank you Eugene! Works well now! 🙌 🚢

assets/base.css
```diff
@@ -248,7 +248,7 @@ html.no-js .no-js-hidden {
   padding: 0 1.5rem;
 }
 
-.page-width:not(.drawer-menu) {
+.page-width:where(:not(.drawer-menu)) {
```
LucasLacerdaUX, line 251: I may be missing some context from the drawer menu PR, but I'm not sure if I understand the inclusion of `:not(.drawer-menu)` in this context. The `page-width` is a very generic class that is meant to be used across all elements on the page, not limited to the header, so increasing its specificity by adding a `:not()` to target a section-specific element sounds a bit off.

If our goal was to have `drawer-menu` take up to `100%` of the `max-width`, can we include a class like this:
```css
/* or maybe just .drawer-menu */
.page-width.drawer-menu {
  max-width: 100%;
}
```
instead, and keep `.page-width` as it was before? To me, that would make it a bit more clear as to why this element shouldn't have a `max-width`

As for the use of `where()`, this matches [the example of it in the specs](https://www.w3.org/TR/selectors-4/#zero-matches) so I'm totally aligned with the solution you proposed :DD 🎉 I'm just not sure if its need could be avoided in the first place.

LucasLacerdaUX: cc @ludoboludo do you have any thoughts on that?

Eugene tried the solution I suggested (override on the specific `drawer-menu`, rather than rewriteing the selector for the generic `page-width`) and it seems to produce the same results.

I like both solutions - it's just that the one using `where()` feels like the problem could be avoided in the first place if we didn't rewrite the generic selector.

eugenekasimov: ✅

LucasLacerdaUX (approved):

melissaperreault (approved):
🚢

PR #2368, opened 2023-03-07
[Blog post section] Fix slides size on mobile 

ludoboludo:
### PR Summary: 

This fixes the slider for the blog post section on mobile that doesn't show the right amount of total slides.

Fixes https://github.com/Shopify/dawn-private/issues/141

There are two approached here that could be taken: 

1. We overwrite the spacing in between each slide. Since right now it's a full width slide, we can force a gap value instead of using the one from the general settings (`horizontal space`).
2. We can reuse the same approach we have on the product page where the width isn't 100% - spacing but `calc(100% - 3rem - var(--grid-mobile-horizontal-spacing))`. This is great as well but has a visual impact. As it adds a `peek-a-boo` effect that wasn't there before: 
<details>
<summary>Screenshot comparison</summary>

</details>

~In this first iteration of the PR I went with approach `1`.~
Updated it so it uses approach 2 👍 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

None with option 1 and if we go with option 2, I shared a screenshot with the visual impact, adding the peek a boo effect which wasn't there before on mobile. 

### Testing steps/scenarios

- [ ] Add a blog post section on any template, 
- [ ] Go into the general `layout` settings and play with the value of `horizontal spacing`,
- [ ] Make sure the total pages is always calculated properly on mobile 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139489968150/editor)

### Checklist

### DIFF
--- a/assets/section-featured-blog.css
+++ b/assets/section-featured-blog.css
@@ -43,6 +43,12 @@
   margin-bottom: 1rem;
 }
 
+@media screen and (max-width: 749px) {
+  .blog__posts.articles-wrapper {
+    column-gap: 1.5rem;
+  }
+}
+
 @media screen and (min-width: 990px) {
   .blog__posts.articles-wrapper {
     margin-bottom: 0;
--- a/assets/section-featured-blog.css
+++ b/assets/section-featured-blog.css
@@ -43,12 +43,6 @@
   margin-bottom: 1rem;
 }
 
-@media screen and (max-width: 749px) {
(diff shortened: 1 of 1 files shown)


kmeleta (approved):
Approach 2 seems like the right approach here. The slider has the grid--peek class and I'm sure it was always intended to behave that way and it was just never updated with the variable padding.

eugenekasimov (approved):
🚀

PR #2372, opened 2023-03-08
animate Featured collection, Related products and Product grid sections

### LINKED ISSUE #2318: [Featured collection]: Add cascading fade in on scroll animation
## Summary

Add cascading fade in on scroll animation for Featured collection, Related products and Product grid sections based on explorations made in https://github.com/Shopify/dawn/pull/2141. Pair with UX to define the ideal strong outcome.

## Details

Reveal section content (Title and description) then each cards in cascading fade in on scroll for:
- [ ] Featured collection section
- [ ] Related products section
- [ ] Product grid section

### Setting
- [ ] The setting will be applied from the new `Animations` theme setting category, below `Layout` and above `Buttons`. The new setting label is offered via a checkbox and its label is; `Reveal sections on scroll`. 
- [ ] The setting will be off by default for the initial release, than turned on for the next one.

metamoni:
### PR Summary: 
Makes Featured collection, Related products and Product grid fade in on scroll.

Fixes #2318 

### Decision log

|---|---|---|---|---|
| 1 |  Use the `--animation-order` and `animation-delay` properties to create cascading animation |  Use [`nth-child` approach](https://github.com/Shopify/dawn/blob/3c11a40ddbb1cfd8059daf0ed171f4389213b317/assets/base.css#L3014-L3075) from prototype | Achieves the same thing in fewer lines of code, making it more readable. We're only setting a CSS attribute on each element in our script, in an already existing loop, so performance impact should be minimal | Relying on JS rather than CSS. However, we're not introducing a lot of new logic, we're only setting a new attribute |
| 2 |  Don't apply cascading animations to desktop and mobile carousels | Only cascade visible product cards in a carousel and stop cascading items that come into view when navigating left-right in the carousel | Achieving the alternative effect is very complicated, due to the many available permutations (variable number of columns on desktop and mobile) | No cascading effect in carousels |

Once this PR is merged, the  Featured collection, Related products and Product grid sections will scroll into view if the `Reveal on scroll` setting is enabled in Theme Settings > Animations.

_Featured collection_

https://user-images.githubusercontent.com/22390758/223750550-8b77b316-d659-4bfd-9860-efc73077dca4.mp4  

<br/>

_Product grid_

https://user-images.githubusercontent.com/22390758/223983227-13a02daa-fa28-47b4-a6bb-eb842b26c185.mp4  

<br/>

_Related products_

https://user-images.githubusercontent.com/22390758/224055462-0053d9e6-2323-474f-abbe-60c76214a59a.mp4

<br/>

#### Featured collection - Default product grid view (title and description slide in, products cascade in)
- [ ] Go to the theme editor
- [ ] Add an empty featured collection and verify that the title scrolls into view
- [ ] Verify that the empty collection items cascade into view
- [ ] Add a description and verify that it scrolls into view with the title
- [ ] Add a collection and verify that each item cascades into view
- [ ] Change the number of columns/products and verify that the section still animates correctly
- [ ] Enable the quick add button and verify that it slides in
- [ ] Check with different settings (see more detailed steps for carousel below)

#### Featured collection - Desktop and mobile carousel (title, description and container slide in, products do not cascade in)
- [ ] Enable carousel on desktop. The section should slide in but individual products should not cascade in
- [ ] Change different settings and verify that it still animates as expected
- [ ] While still in Desktop view, tick Enable swipe on mobile and verify that it still animates as expected
- [ ] Test with different mobile column settings
- [ ] Switch to Mobile view and test both the cascading grid and carousel versions
- [ ] While still in Mobile view, turn desktop carousel on/off and test that everything animates as expected on mobile

#### Product grid and related items
- [ ] Click "Shop all" to bring up the Product grid collection and verify that it animates as expected
- [ ] Change different settings and verify that the section still animates correctly
- [ ] Navigate to the product page and scroll to the bottom to reveal the Related products section. Verify that it animates as expected
- [ ] Change different settings and verify that the section still animates correctly
- [ ] [Turn on `Reduce motion` in your OS](https://mcmw.abilitynet.org.uk/how-to-reduce-motion-effects-in-macos-12-monterey) and verify that the animations stop working

### Demo links
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139494916118)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139494916118/editor)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -742,7 +742,7 @@ details > * {
   --duration-default: 200ms;
   --duration-long: 500ms;
   --duration-extra-long: 600ms;
-  --slide-in-animation: slideIn var(--duration-extra-long) ease-out forwards;
+  --animation-slide-in: slideIn var(--duration-extra-long) ease-out forwards;
 }
 
 .underlined-link,
@@ -1143,7 +1143,7 @@ summary::-webkit-details-marker {
   }
 
   .scroll-trigger--active .slider--tablet {
-    animation: var(--slide-in-animation);
+    animation: var(--animation-slide-in);
   }
 }
 
@@ -1179,7 +1179,7 @@ summary::-webkit-details-marker {
   }
 
   .scroll-trigger--active .slider--desktop {
-    animation: var(--slide-in-animation);
+    animation: var(--animation-slide-in);
   }
 }
 
@@ -3218,7 +3218,7 @@ details-disclosure > details {
   }
 
   .scroll-trigger--active.animate--slide-in {
-    animation: var(--slide-in-animation);
+    animation: var(--animation-slide-in);
     animation-delay: calc(var(--animation-order) * 100ms);
   }
 


kjellr:
This works really well, and the JS solution feels way more elegant than all the `nth-child` rules in previous iterations. I doubt we're really sacrificing much in terms of performance by it either. Nice work! 👏

Just noting that @metamoni and I chatted earlier and determined that it's ok for V1 to just fade/slide-in the entire Related Products section at once (rather than in a cascade, like other product grids). During initial prototypes, there were some issues around getting this section to animate correctly. 

I'm seeing some related bugginess here — if you scroll too quickly into the Related Products section, you get a little jump before the related products appear. [Here's a slow-motion video to demonstrate](https://screenshot.click/09-20-q43xv-bvb9f2.mp4). This isn't related to our work here, but it's a little more obvious because all the other animations are incredibly smooth. 😅 

If we think there's an easy fix for this, let's do it! Otherwise we can look into it later and I don't consider it a blocker.

assets/animations.js
```diff
@@ -2,10 +2,11 @@ const SCROLL_ANIMATION_TRIGGER_CLASSNAME = 'scroll-trigger';
 const SCROLL_ANIMATION_ACTIVE_CLASSNAME = 'scroll-trigger--active';
 
 function onIntersection(elements, observer) {
-  elements.forEach((element) => {
+  elements.forEach((element, index) => {
     if (element.isIntersecting) {
       const elementTarget = element.target;
       elementTarget.classList.add(SCROLL_ANIMATION_ACTIVE_CLASSNAME);
+      elementTarget.setAttribute('style', `--animation-order: ${index};`);
```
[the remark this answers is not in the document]
metamoni: @kjellr is this not the order that we're looking to animate the product cards in?

kjellr (dismissed):
This is working nicely for me — haven't stumbled upon any broken parts. 👍

assets/base.css
```diff
@@ -3148,6 +3148,7 @@ details-disclosure > details {
 
   .scroll-trigger--active.animate--slide-in {
     animation: slideIn var(--duration-extra-long) ease-out forwards;
```
melissaperreault, line 3150: For my personal knowledge, I always assumed anything coming into view is better with `ease-in` in general. Thoughts?

kjellr: Ease-out will slow it down a bit at the end, which feels more "real-life" to me personally. When we're driving our cars for instance, we generally aim to slow before a stop.

stufreen (dismissed):
Code LGTM. When testing I did find that
a) we might want to tweak the duration, easing, and slide-in amount
b) I do think we should add logic to NOT fade in the above-the-fold sections on initial pageload. This will also help the online store editor since it keeps re-running the animations each time a section is added.

These are both out of scope this PR though

assets/base.css
melissaperreault: Thank you!

melissaperreault (commented):
Thank you Monica! I tried to ignore the known bug and give a deep review testing different settings combinations.

**Observations**
- [ ] When `Enable carousel on desktop` or `Enable swipe on mobile` is checked, only the last slide is animated because the slider only brings up one more card at a time. I find the combination of the slider transition + the last card jarring. I am not convinced it adds the value for when we change slides as the reveal is partially shown, almost look like a bug. It's successful when the section comes into view though. Thoughts? 
- [ ] Notice on the _Mobile swipe_ video below, the `View all` comes in much later, probably once the full set of cards are animated? This happens when you set a high number of product to show. The delay was a surprise.

   - [Desktop carousel video](https://screenshot.click/15-32-73dt3-eu6k8.mp4)
   - [Mobile swipe video](https://screenshot.click/15-42-0ymte-1uppq.mp4)

kjellr:
> b) I do think we should add logic to NOT fade in the above-the-fold sections on initial pageload. This will also help the online store editor since it keeps re-running the animations each time a section is added.

I agree this is worth trying out. I've opened up #2383 so we can track this. 👍

melissaperreault (changes requested):
I noticed the slider would scroll vertically both in the Editor view and regular view while trying to replicate the _View all_ issue I raised.


https://user-images.githubusercontent.com/26230969/225677370-a53498b1-5a28-4fa9-9083-c603349f7887.mp4

metamoni:
@melissaperreault we agreed to disable the cascading animation when the carousel is enabled.

metamoni:
@melissaperreault we agreed to disable cascading on sliders for now

sections/featured-collection.liquid
```diff
@@ -78,7 +78,8 @@
         {%- for product in section.settings.collection.products limit: section.settings.products_to_show -%}
           <li
             id="Slide-{{ section.id }}-{{ forloop.index }}"
-            class="grid__item{% if show_mobile_slider or show_desktop_slider %} slider__slide{% endif %}"
+            class="grid__item{% if show_mobile_slider or show_desktop_slider %} slider__slide{% endif %}{% if settings.animations_reveal_on_scroll %}{% unless show_mobile_slider or show_desktop_slider %} scroll-trigger animate--slide-in{% endunless %}{% endif %}"
+            {% unless show_desktop_slider or show_mobile_slider %}data-cascade{% endunless %}
```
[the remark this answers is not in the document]
metamoni: I didn't see any dramatic doubling-up of the animation:

https://user-images.githubusercontent.com/22390758/228512701-1efdf956-dcb1-421c-a634-e8884d9057d3.mp4


Since we were already making use of `show_desktop_slider` and `show_mobile_slider`, I was trying to avoid duplicating these styles or having them in so many places. When we started doing animations, we always said we might have to tweak them at some point, and I think that relying on CSS too much will make it more difficult to do that, as we'll have to change things in lots of places, in a file that's already really large and difficult to navigate. Would it be a good idea to start moving these styles into their own file?

That being said, your suggestion made me realise there was an issue with my original code. The logic caused the cascading effect to stop working _on desktop_ when the mobile slider was enabled; and vice versa, cascading didn't work _on mobile_ when the desktop slider was enabled. Your suggestion fixes this 🎉

stufreen (dismissed):
LGTM 🚀

sections/featured-collection.liquid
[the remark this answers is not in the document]
stufreen: My main concern with using liquid is just about getting the logic right. That issue you mention about the slider not working on desktop if the mobile slider is enabled is a good example. We don't know what type of client (desktop/mobile, touchscreen/mouse, safari/chrome/firefox, etc) it is in the liquid files, so we can't add or remove any markup to try and target certain devices. We can only use media queries in CSS and JS for that.

I think it's fine to use `animate--slide-in` heavily but in some cases you want to only apply the animation at certain breakpoints. An alternative to the solution above would be to keep a `animate--slide-in` class on the slider parent element but disable it with a more specific rule when there's not a carousel.

Looking forward for how we can do the cascading, I think it's a matter for a future PR :) But some ideas:
* Use JS with element.animate and media queries to do the animations for carousels (not CSS)
* Add a `<style>` block to the `featured-collection` liquid file so that you can inject theme settings into the CSS, and use `nth-child` rules to remove animations for slides that are out of view.

stufreen (dismissed):

assets/base.css
```diff
@@ -742,6 +742,7 @@ details > * {
   --duration-default: 200ms;
   --duration-long: 500ms;
   --duration-extra-long: 600ms;
+  --slide-in-animation: slideIn var(--duration-extra-long) ease-out forwards;
```
kjellr, line 745: Not at all critical, but if we expect to have additional `animation` variables in the future, I wonder if we should name this `--animation-slide-in` to match the format of the `--duration-*` variables. 🤔

kjellr (dismissed):
One very minor comment, but this looks good to me! 🎉

PR #2385, opened 2023-03-13
[Image with Text] Add ambient movement to image

### LINKED ISSUE #2340: [Image with Text] Add ambient movement to background

kjellr:
### PR Summary: 

Adds an option for ambient animation to the Image with Text section's image.

https://user-images.githubusercontent.com/1202812/224763195-a7bb8532-a6a0-481a-9953-63930ec0f237.mp4

Closes #2340

Uses the same approach (and existing CSS) that was added in https://github.com/Shopify/dawn/pull/2342.

This implements a new dropdown menu https://github.com/Shopify/dawn/pull/2341. When "Ambient movement" is selected, a CSS class is applied to the Image with Text's image element, triggering a long-duration CSS animation that slowly moves the image around within its frame. As this feature is purely aesthetic, it supports the visitor's prefers-reduced-motion preference.

Like https://github.com/Shopify/dawn/pull/2342, it also conditionally loads in a slightly higher-res image when the ambient animation is active, in order to account for the scale effect that's applied to the image.

This feature is off by default. When it's turned on, the image with text's image will slowly animate ambiently.

### Decision log

|---|---|---|---|---|
| 1 |  Move section-specific animation-related strings to `sections.all.animation` to avoid duplication. | Repeat these string declarations every time they're used. | We're reusing the same strings across (at least) three sections, so declaring these in one place cuts down on duplicate code. | None, really. |

### Testing steps/scenarios

1. [Visit the demo store](https://os2-demo.myshopify.com/admin/themes/139518181398/editor).
2. Add and select an Image with Text.
3. Look for the new "Image behavior" dropdown underneath the new "Animations" header in the Image with Text's sidebar.
4. Select "Ambient movement" from that dropdown.
5. Add image backgrounds to individual slides, verify that they appear as expected.
6. Try various breakpoints, verify that they appear as expected.

### Checklist

### DIFF
--- a/assets/component-localization-form.css
+++ b/assets/component-localization-form.css
@@ -199,6 +199,7 @@ noscript .localization-selector.link {
   font-size: 1.4rem;
   letter-spacing: 0.06rem;
   height: auto;
+  background: transparent;
 }
 
 .header__localization .disclosure .localization-form__select:hover {
--- a/assets/component-menu-drawer.css
+++ b/assets/component-menu-drawer.css
@@ -217,6 +217,7 @@ details[open].menu-opening > .menu-drawer__submenu {
 .menu-drawer__utility-links {
   padding: 2rem;
   background-color: rgba(var(--color-foreground), 0.03);
+  position: relative;
 }
 
 .menu-drawer__account {
--- a/assets/section-featured-blog.css
+++ b/assets/section-featured-blog.css
@@ -55,7 +55,7 @@
 
 @media screen and (max-width: 749px) {
   .blog__post.article {
-    width: calc(100% - 3rem);
+    width: calc(100% - 3rem - var(--grid-mobile-horizontal-spacing));
   }
 }
 


locales/en.default.schema.json
```diff
@@ -1344,6 +1344,18 @@
...
+        },
+        "image_behavior": {
+          "options__1": {
+            "label": "None"
+          },
+          "options__2": {
+            "label": "Ambient movement"
+          },
+          "label": "Image behavior"
+        },
```
kjellr, line 1358: As noted in https://github.com/Shopify/dawn/pull/2383#discussion_r1134129110, it might make sense to have these strings be global since they're re-used across a few Sections.

melissaperreault (dismissed):
Thank you Kjell!

Works well! I tested
- [x] Focal point
- [x] Different screen sizes
- [x] Different file types
- [x] Different image sizes
- [x] Global theme settings

metamoni:
Not sure if this is something caused by your refactoring or just some code that snuck in, but the top padding of the Image with text section is larger than it used to be. Here's what it looks like on `main`:
main

and in this PR:
pr

metamoni (dismissed):
LGTM 👍

locales/en.default.schema.json
```diff
@@ -386,6 +386,18 @@
   },
   "sections": {
     "all": {
+      "animation": {
+        "label": "Animations",
```
LucasLacerdaUX, line 390: I wonder if we should use the key `label` here as this is used as a header `content` on the schema. We usually keep a similar structure to what's used in our schema to make mapping easier. However, I see that just below we're using a `section_padding_heading` which also doesn't map to anything, so, not a big deal right now 👀

LucasLacerdaUX (approved):
LGTM! :D

metamoni (approved):

PR #2394, opened 2023-03-14
Announcements slideshow

eugenekasimov:
### PR Summary: 

This PR changes the way announcement blocks are displayed. Instead of being stacked under each other now they are displayed as a slideshow.

#

Fixes [#398](https://github.com/Shopify/themes/issues/398).

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2226,9 +2226,20 @@ product-info .loading-overlay:not(.hidden) ~ *,
 }
 
 /* section-announcement-bar */
-.announcement-bar {
+.announcement-bar--bottom-border {
   border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.08);
+}
+
+.announcement-bar {
   color: rgb(var(--color-foreground));
+}
+
+.announcement-slider-container {
+  width: 100%;
+}
+
+.announcement-slider-container .slideshow__slide {
+  justify-content: center;
 }
 
 .announcement-bar__link {
@@ -2256,6 +2267,7 @@ product-info .loading-overlay:not(.hidden) ~ *,
 }
 
 .announcement-bar__message {
+  text-align: center;
   padding: 1rem 0;
   margin: 0;
   letter-spacing: 0.1rem;
--- a/assets/component-slider.css
+++ b/assets/component-slider.css
@@ -135,6 +135,10 @@ slider-component.slider-component-full-width {
   scroll-snap-align: center;
 }
 
+.announcement-bar .slider--everywhere {
+  margin-bottom: 0;
+}
+
 @media screen and (min-width: 990px) {
   .slider-component-desktop.page-width {
     max-width: none;
@@ -357,6 +361,37 @@ slider-component.slider-component-full-width {
   display: flex;
   align-items: center;
   justify-content: center;
+}
+
+.slider-button--next {
+  margin-right: -2rem;
+}
+
+.slider-button--prev {
+  margin-left: -2rem;
+}
+
+
+@media screen and (min-width: 750px) {
+  .slider-button--next {
+    margin-right: -3.8rem;
+  }
+
+  .slider-button--prev {
+    margin-left: -3.8rem;
+  }
+
+}
+
+@media screen and (min-width: 990px) {
+  .slider-button--next {
+    margin-right: -2rem;
+  }
+
+  .slider-button--prev {
+    margin-left: -2rem;
+  }
+
 }
 
 .slider-button:not([disabled]):hover {
--- a/assets/global.js
+++ b/assets/global.js
@@ -679,17 +679,16 @@ class SlideshowComponent extends SliderComponent {
   }
 
   setAutoPlay() {
-    this.sliderAutoplayButton = this.querySelector('.slideshow__autoplay');
+    if (this.querySelector('.slideshow__autoplay') !== null) {
+      this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
+      this.addEventListener('mouseover', this.focusInHandling.bind(this));
+      this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
+      this.addEventListener('focusin', this.focusInHandling.bind(this));
+      this.addEventListener('focusout', this.focusOutHandling.bind(this));
+      this.autoplayButtonIsSetToPlay = true;
+    }
     this.autoplaySpeed = this.slider.dataset.speed * 1000;
-
-    this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
-    this.addEventListener('mouseover', this.focusInHandling.bind(this));
-    this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
-    this.addEventListener('focusin', this.focusInHandling.bind(this));
-    this.addEventListener('focusout', this.focusOutHandling.bind(this));
-
     this.play();
-    this.autoplayButtonIsSetToPlay = true;
   }
 
   onButtonClick(event) {
--- a/sections/announcement-bar.liquid
+++ b/sections/announcement-bar.liquid
@@ -1,32 +1,215 @@
-{%- for block in section.blocks -%}
-  {%- case block.type -%}
-    {%- when 'announcement' -%}
-      <div class="announcement-bar color-{{ block.settings.color_scheme }} gradient" role="region" aria-label="{{ 'sections.header.announcement' | t }}" {{ block.shopify_attributes }}>
-        {%- if block.settings.text != blank -%}
-          {%- if block.settings.link != blank -%}
-            <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
-          {%- endif -%}
-              <div class="page-width">
-                <p class="announcement-bar__message {{ block.settings.text_alignment }} h5">
-                  <span>{{ block.settings.text | escape }}</span>
+{{ 'component-slideshow.css' | asset_url | stylesheet_tag }}
+
+<div
+  class="color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} announcement-bar--bottom-border{% endif %}"
+  {{ block.shopify_attributes }}
+>
+  <slideshow-component
+    class="announcement-bar page-width"
+    aria-roledescription="{{ 'sections.slideshow.carousel' | t }}"
+    aria-label="{{ section.settings.accessibility_info | escape }}"
+  >
+    {%- if section.blocks.size == 1 -%}
+      {%- for block in section.blocks -%}
+        {%- case block.type -%}
+          {%- when 'announcement' -%}
+            <div class="announcement-bar" role="region" aria-label="{{ 'sections.header.announcement' | t }}" {{ block.shopify_attributes }}>
+              {%- if block.settings.text != blank -%}
+                {%- if block.settings.link != blank -%}
+                  <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
+                {%- endif -%}
+                    <p class="announcement-bar__message h5">
+                      <span>{{ block.settings.text | escape }}</span>
+                      {%- if block.settings.link != blank -%}
+                        {% render 'icon-arrow' %}
+                      {%- endif -%}
+                    </p>
+                {%- if block.settings.link != blank -%}
+                  </a>
+                {%- endif -%}
+              {%- endif -%}
+            </div>
+        {%- endcase -%}
+      {%- endfor -%}
+    {%- endif -%}
+    {%- if section.blocks.size > 1 and section.settings.auto_rotate == false -%}
+      <div class="announcement-slider-container slider-buttons">
+        <button
+          type="button"
+          class="slider-button slider-button--prev"
+          name="previous"
+          aria-label="{{ 'sections.slideshow.previous_slideshow' | t }}"
+          aria-controls="Slider-{{ section.id }}"
+        >
+          {% render 'icon-caret' %}
+        </button>
+        <div
+          class="announcement-slider-container grid grid--1-col slider slider--everywhere"
+          id="Slider-{{ section.id }}"
+          aria-live="polite"
+          aria-atomic="true"
+          data-autoplay="{{ section.settings.auto_rotate }}"
+          data-speed="{{ section.settings.change_slides_speed }}"
+        >
+          {%- for block in section.blocks -%}
+            <div
+              class="slideshow__slide grid__item grid--1-col"
+              id="Slide-{{ section.id }}-{{ forloop.index }}"
+              {{ block.shopify_attributes }}
+              role="group"
+              aria-roledescription="{{ 'sections.slideshow.slide' | t }}"
+              aria-label="{{ forloop.index }} {{ 'general.slider.of' | t }} {{ forloop.length }}"
+              tabindex="-1"
+            >
+              <div class="announcement-bar" role="region" aria-label="{{ 'sections.header.announcement' | t }}" {{ block.shopify_attributes }}>
+                {%- if block.settings.text != blank -%}
                   {%- if block.settings.link != blank -%}
-                    {% render 'icon-arrow' %}
+                    <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
                   {%- endif -%}
-                </p>
+                      <p class="announcement-bar__message h5">
+                        <span>{{ block.settings.text | escape }}</span>
+                        {%- if block.settings.link != blank -%}
+                          {% render 'icon-arrow' %}
+                        {%- endif -%}
+                      </p>
+                  {%- if block.settings.link != blank -%}
+                    </a>
+                  {%- endif -%}
+                {%- endif -%}
               </div>
-          {%- if block.settings.link != blank -%}
-            </a>
-          {%- endif -%}
-        {%- endif -%}
+            </div>
+          {%- endfor -%}
+        </div>
+        <button
+          type="button"
+          class="slider-button slider-button--next"
+          name="next"
+          aria-label="{{ 'sections.slideshow.next_slideshow' | t }}"
+          aria-controls="Slider-{{ section.id }}"
+        >
+          {% render 'icon-caret' %}
+        </button>
       </div>
-  {%- endcase -%}
-{%- endfor -%}
+    {%- endif -%}
+    {%- if section.blocks.size > 1 and section.settings.auto_rotate -%}
+      <div class="announcement-slider-container slider-buttons">
+        <button
+          type="button"
+          class="slider-button slider-button--prev"
+          name="previous"
+          aria-label="{{ 'sections.slideshow.previous_slideshow' | t }}"
+          aria-controls="Slider-{{ section.id }}"
+        >
+          {% render 'icon-caret' %}
+        </button>
+        <div
+          class="announcement-slider-container grid grid--1-col slider slider--everywhere"
+          id="Slider-{{ section.id }}"
+          aria-live="polite"
+          aria-atomic="true"
+          data-autoplay="{{ section.settings.auto_rotate }}"
+          data-speed="{{ section.settings.change_slides_speed }}"
+        >
+          {%- for block in section.blocks -%}
+            <div
+              class="slideshow__slide grid__item grid--1-col"
+              id="Slide-{{ section.id }}-{{ forloop.index }}"
+              {{ block.shopify_attributes }}
+              role="group"
+              aria-roledescription="{{ 'sections.slideshow.slide' | t }}"
+              aria-label="{{ forloop.index }} {{ 'general.slider.of' | t }} {{ forloop.length }}"
+              tabindex="-1"
+            >
+              <div class="announcement-bar" role="region" aria-label="{{ 'sections.header.announcement' | t }}" {{ block.shopify_attributes }}>
+                {%- if block.settings.text != blank -%}
+                  {%- if block.settings.link != blank -%}
+                    <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
+                  {%- endif -%}
+                      <p class="announcement-bar__message h5">
+                        <span>{{ block.settings.text | escape }}</span>
+                        {%- if block.settings.link != blank -%}
+                          {% render 'icon-arrow' %}
+                        {%- endif -%}
+                      </p>
+                  {%- if block.settings.link != blank -%}
+                    </a>
+                  {%- endif -%}
+                {%- endif -%}
+              </div>
+            </div>
+          {%- endfor -%}
+        </div>
+        <button
+          type="button"
+          class="slider-button slider-button--next social-icons"
+          name="next"
+          aria-label="{{ 'sections.slideshow.next_slideshow' | t }}"
+          aria-controls="Slider-{{ section.id }}"
+        >
+          {% render 'icon-caret' %}
+        </button>
+      </div>
+    {%- endif -%}
+  </slideshow-component>
+</div>
 
 {% schema %}
 {
   "name": "t:sections.announcement-bar.name",
   "max_blocks": 12,
   "class": "announcement-bar-section",
+  "settings": [
+    {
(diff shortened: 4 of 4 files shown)


sections/announcement-bar.liquid
```diff
@@ -1,32 +1,213 @@
...
+            </div>
+      {%- endfor -%}
+    {%- endif -%}
+    {%- if section.blocks.size > 1 and section.settings.auto_rotate == false -%}
+      <div class="announcement-bar-slider slider-buttons">
+        <button
+          type="button"
+          class="slider-button slider-button--prev"
+          name="previous"
+          aria-label="{{ 'sections.slideshow.previous_slideshow' | t }}"
```
metamoni, line 39: This should be "Previous announcement".

assets/global.js
```diff
@@ -679,17 +679,16 @@ class SlideshowComponent extends SliderComponent {
...
   setAutoPlay() {
-    this.sliderAutoplayButton = this.querySelector('.slideshow__autoplay');
+    if (this.querySelector('.slideshow__autoplay') !== null) {
+      this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
+      this.addEventListener('mouseover', this.focusInHandling.bind(this));
+      this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
+      this.addEventListener('focusin', this.focusInHandling.bind(this));
+      this.addEventListener('focusout', this.focusOutHandling.bind(this));
+      this.autoplayButtonIsSetToPlay = true;
+    }
```
kmeleta, line 689: This might be a slightly cleaner option if you prefer
```suggestion
    if (this.querySelector('.slideshow__autoplay') === null) return;
 
    this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
    this.addEventListener('mouseover', this.focusInHandling.bind(this));
    this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
    this.addEventListener('focusin', this.focusInHandling.bind(this));
    this.addEventListener('focusout', this.focusOutHandling.bind(this));
    this.autoplayButtonIsSetToPlay = true;
```

kmeleta (commented):
Behavior is looking good so far, but mostly did a quick first pass on code. Let me know if any of my feedback is confusing or you disagree with it.

assets/component-slider.css
```diff
@@ -359,6 +363,39 @@ slider-component.slider-component-full-width {
   justify-content: center;
 }
 
+.slider-button--next {
+  margin-right: -2rem;
+  height: 100%;
+}
+
+.slider-button--prev {
```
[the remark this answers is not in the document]
eugenekasimov: I actually like this idea, this helps to keep the code a bit cleaner. Thanks.

assets/global.js
```diff
@@ -679,17 +679,16 @@ class SlideshowComponent extends SliderComponent {
   }
 
   setAutoPlay() {
-    this.sliderAutoplayButton = this.querySelector('.slideshow__autoplay');
+    if (this.querySelector('.slideshow__autoplay') !== null) {
```
ludoboludo, line 682: Do we need to check for this ? We only run this `setAutoPlay()` function if this is true: 
```javascript
if (this.slider.getAttribute('data-autoplay') === 'true') this.setAutoPlay();
```
Are we thinking of doing autoplay without a pause button ?

sections/announcement-bar.liquid
```diff
@@ -1,32 +1,213 @@
...
-            <a href="{{ block.settings.link }}" class="announcement-bar__link link link--text focus-inset animate-arrow">
-          {%- endif -%}
-              <div class="page-width">
-                <p class="announcement-bar__message {{ block.settings.text_alignment }} h5">
-                  <span>{{ block.settings.text | escape }}</span>
+{{ 'component-slideshow.css' | asset_url | stylesheet_tag }}
+{{ 'component-slider.css' | asset_url | stylesheet_tag }}
+
+<div
+  class="announcement-bar-container color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} announcement-bar--bottom-border{% endif %}"
```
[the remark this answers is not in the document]
eugenekasimov: Thanks for brining this up. I refactored all of these :)

assets/global.js
kmeleta: I could be mistaken,  didn't examine in-context. I just assumed if this.querySelector('.slideshow__autoplay') === null anyway, we wouldn't need to call play() or set a value for this.autoplaySpeed

eugenekasimov: The thing is that I don't use this `.slideshow__autoplay` class here. We apply this class in the slideshow component to pause/play buttons which I don't have here. That's why I added that logic above. So, in our case for announcement slideshow `this.querySelector('.slideshow__autoplay'` will always be `null`.

assets/component-slider.css
```diff
@@ -359,6 +363,39 @@ slider-component.slider-component-full-width {
...
+
+@media screen and (min-width: 990px) {
+  body:has( .section-header .header:not(.drawer-menu)) .announcement-bar-section .slider-button--next {
+    margin-right: -2rem;
+  }
+
+  body:has( .section-header .header:not(.drawer-menu)) .announcement-bar-section .slider-button--prev {
+    margin-left: -2rem;
+  }
+
```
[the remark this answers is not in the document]
eugenekasimov: Update from #help-accessibility channel:

"If the merchant sets it to auto-start and auto-advance, prefers-reduced-motion should override both of those IMO. So the user would click through the slides."

So, it's still an opinion, however, I don't have strong feelings against/for it. I'm okay to prevent the slider from rotation in case of a user having reduce-motion-on on their computer.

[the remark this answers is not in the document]
metamoni: > As far as I understand , `prefers-reduced-motion` should reduce motion but not remove it.

Not sure where you read this, but that's not quite right. `prefers-reduced-motion` is a media query that does whatever you tell it to do, allowing us to include animations/motion on the page while also making sure that they don't cause any issues or discomfort to people who suffer from vestibular disorders (or to those who simply don't like a lot of motion). We actually have a good example of using this query in recent [PRs](https://github.com/Shopify/dawn/pull/2342), where we're _only turning animations on_ if the user doesn't have `prefers-reduced-motion` enabled ([see code](https://github.com/Shopify/dawn/blob/e80839333aca4357e96eb81b2a1342a950159f8a/assets/base.css#L3139-L3148)).

What that Slack thread is saying is that slides should _not_ automatically change. There is [official guidance in WCAG](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html) for this, so it's not just an opinion. It's a level A requirement. The reason why we don't have it in the slideshow is that we didn't have much motion in Dawn before. Now that we're introducing more sections with different types of animations/motion, we want to make sure it's not all too overpowering. I think that's the reason why it was added as a requirement but @stufreen, who created the issue, should be able to clarify.

assets/component-slider.css
```diff
@@ -359,6 +363,39 @@ slider-component.slider-component-full-width {
   justify-content: center;
 }
 
+.announcement-bar .slider-button--next {
+  margin-right: -2rem;
+  height: 100%;
+}
+
+.announcement-bar .slider-button--prev {
+  margin-left: -2rem;
```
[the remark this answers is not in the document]
eugenekasimov: I found that -1.6rem fits the best

sections/announcement-bar.liquid
```diff
@@ -1,32 +1,216 @@
...
-                  <span>{{ block.settings.text | escape }}</span>
+{{ 'component-slideshow.css' | asset_url | stylesheet_tag }}
+{{ 'component-slider.css' | asset_url | stylesheet_tag }}
+
+<div
+  class="utility-bar color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} utility-bar--bottom-border{% endif %}"
+  {{ block.shopify_attributes }}
+>
+  {%- if section.blocks.size == 1 -%}
+    {%- for block in section.blocks -%}
```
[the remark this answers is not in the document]
eugenekasimov: fair enough, thanks

assets/component-slider.css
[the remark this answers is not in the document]
metamoni: Sounds good 🎉

assets/base.css
```diff
@@ -249,6 +249,10 @@ html.no-js .no-js-hidden {
   padding: 0 1.5rem;
 }
 
+body:has( .section-header .drawer-menu) .announcement-bar-section slideshow-component {
```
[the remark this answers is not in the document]
kmeleta: To summarize some thoughts from our pairing on this topic.. I'm actually curious if @YoannJailin thinks it's reasonable to remove the page-width limitation on the slider for _all_ menu types, not just drawer. Particularly when you consider the possibility of adding more content to the utility bar in the future. This would also come with the minor benefit of avoiding the use of :has and the possibility of different results in different browsers.

**Current** (with dropdown or mega menu):


**Proposed** (all):

assets/component-slider.css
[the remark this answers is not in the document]
eugenekasimov: It's [done](https://github.com/Shopify/dawn/pull/2394/commits/c0a1d44c1de46c226bbddb0fc4a760f8e979a187).

I also applied it to the slideshow and hide play/pause buttons when reduced-motion is on.

assets/base.css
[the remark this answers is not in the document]
YoannJailin: Hey @kmeleta, sorry i miss your comment. Usually i am always up for solutions that ease everyone's work! But I think we should stick on aligning the Bar content on the grid when Menu is set to dropdown or mega menu vs full width when it's a drawer on desktop. I am okay if a few users don't see the bar full width when menu is set to drawer. From the numbers i see on browsers that won't show the best behavior, the impact is minimal. I believe the result of having a full width top bar vs an on grid header would look misaligned.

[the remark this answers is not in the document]
eugenekasimov: I had a meeting with Yoann. From a UX perspective we still want to keep the current logic for the utility-bar and follow the header width. The reason is that for the wide screens it may look bad. Also, for the future when we consider adding more content there will be a bit different layout inside the utility-bar (it's presented on Figma).

We are agreed to keep` :has` for this case since the most popular browsers supports it and even those that don't won't have issues with functionality.

assets/component-slider.css
```diff
@@ -359,6 +363,39 @@ slider-component.slider-component-full-width {
   justify-content: center;
 }
 
+.slider-button--next {
```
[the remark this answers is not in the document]
metamoni: @eugenekasimov, @YoannJailin just to clarify: have we decided to leave this as is?

assets/base.css
```diff
@@ -2227,22 +2231,39 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
+  flex-wrap: wrap;
+  align-content: center;
+}
+
+.announcement-bar-slider,
+.announcement-bar-slider .slider {
+  width: 100%;
 }
 
 .announcement-bar__link {
```
kmeleta, line 2259: I'm aligned with the decision to remove the background hover effect, but now not sure if the arrow animation alone is enough feedback to visually indicate the hover state. I wonder if it might be worth reducing the initial opacity of the text and changing to 100% on hover, like many other links do ([example](https://screenshot.click/29-57-q7tlj-x8jos.mp4))

Not sure if this was considered or not, but wanted to run it by @YoannJailin just in case.

assets/base.css
```diff
@@ -2227,22 +2231,39 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
+.utility-bar {
+  height: 100%;
+}
+
+.utility-bar--bottom-border {
   border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.08);
+}
+
+.announcement-bar,
+.announcement-bar-slider__announcement {
```
kmeleta, line 2244: Not a big deal, but I'd probably have the __announcement use the announcement-bar namespace (rather than announcement-bar-slider) `.announcement-bar__announcement`

assets/component-slider.css
```diff
@@ -359,6 +363,45 @@ slider-component.slider-component-full-width {
...
+  outline-offset: -0.3rem;
+  box-shadow: 0 0 0 -0.2rem rgb(var(--color-foreground));
+}
+
+.announcement-bar .slider-button--prev {
+  margin-left: -1.6rem;
+  height: 100%;
+}
+
+
```
kmeleta, line 382:
```suggestion

```

assets/component-slider.css
```diff
@@ -359,6 +363,45 @@ slider-component.slider-component-full-width {
...
+@media screen and (min-width: 750px) {
+  .announcement-bar .slider-button--next {
+    margin-right: -3.8rem;
+  }
+
+  .announcement-bar .slider-button--prev {
+    margin-left: -3.8rem;
+  }
+
+}
```
kmeleta, line 392:
```suggestion
  }
}
```

assets/component-slider.css
```diff
@@ -359,6 +363,45 @@ slider-component.slider-component-full-width {
...
+.announcement-bar .slider-button--next:focus-visible,
+.announcement-bar .slider-button--prev:focus-visible {
+  outline-offset: -0.3rem;
+  box-shadow: 0 0 0 -0.2rem rgb(var(--color-foreground));
+}
+
+.announcement-bar .slider-button--prev {
+  margin-left: -1.6rem;
+  height: 100%;
+}
```
kmeleta, line 380:
```suggestion
.announcement-bar .slider-button--next {
  margin-right: -1.6rem;
  height: 100%;
}

.announcement-bar .slider-button--prev {
  margin-left: -1.6rem;
  height: 100%;
}

.announcement-bar .slider-button--next:focus-visible,
.announcement-bar .slider-button--prev:focus-visible {
  outline-offset: -0.3rem;
  box-shadow: 0 0 0 -0.2rem rgb(var(--color-foreground));
}
```

locales/en.default.schema.json
```diff
@@ -477,6 +477,14 @@
...
     "announcement-bar": {
       "name": "Announcement bar",
+      "settings": {
+        "auto_rotate": {
+          "label": "Auto-rotate slides"
+        },
+        "change_slides_speed": {
+          "label": "Change slides every"
+        }
+      },
```
kmeleta, line 487: Just calling out that I agree with duplicating these particular translations for the announcement bar vs reusing the entries from slideshow. 

These are not particularly verbose and even though slideshow code is being used as a layer of abstraction here, slideshow is a fully independent section on its own and I don't know if there's a strong case to rely on its settings here.

kmeleta (commented):
Basically gtg. Added a suggestion or two for consideration, but otherwise just some final formatting nitpicks before approval.

assets/base.css
YoannJailin: Hey! Agree an additional effect could help. Did some testing, the only matter i see with opacity is that some announcements may be clickable, some others not. Might look strange to switch between annoucements more or less transparent depending of they are clickable or not. 

Is it worth considering a line on hover when it's clickable? 
Capture d’écran, le 2023-03-30 à 14 22 58

eugenekasimov: @YoannJailin done ✅

assets/base.css
```diff
@@ -2246,22 +2250,43 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
+.announcement-bar,
+.announcement-bar__announcement {
   color: rgb(var(--color-foreground));
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+  align-content: center;
+}
```
kmeleta, line 2271: I'm not sure we need _all_ these styles applied to both of these classes. I know we've moved some classes around, so I feel like some of these are redundant now. I should have flagged this sooner though and I don't want to send us backward. If we feel we can confidently clean some of this up that'd be great but I won't block this PR for it.

assets/component-slider.css
```diff
@@ -359,6 +363,43 @@ slider-component.slider-component-full-width {
...
+  outline-offset: -0.3rem;
+  box-shadow: 0 0 0 -0.2rem rgb(var(--color-foreground));
+}
+
+@media screen and (min-width: 750px) {
+  .announcement-bar .slider-button--next {
+    margin-right: -3.8rem;
+  }
+
+  .announcement-bar .slider-button--prev {
```
kmeleta, line 387: Not something we need to address here or at all, just thinking aloud in case it appealed to anyone. With the arrows kind of serving as a frame, it might be a nice little improvement if the boundaries of the slider area had a little fade out/gradient overlay so the scrolling content doesn't just disappear at a harsh line. Something similar may have come up before though (maybe during slideshow development) and I know we didn't roll with that.

kmeleta (approved):
Seems like it's working beautifully to me 👍  Thanks for taking the time to consider all the heavily opinionated code nitpicks.

metamoni (approved):

YoannJailin (approved):

sections/announcement-bar.liquid
```diff
@@ -1,32 +1,154 @@
...
+        {
+          "value": "background-2",
+          "label": "t:sections.all.colors.background_2.label"
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "accent-1",
```
ludoboludo, line 126: Should the default be `accent-1` ? 
It used to be `background-1` when the setting was at block level 🤔

sections/announcement-bar.liquid
```diff
@@ -38,54 +160,6 @@
...
-            },
-            {
-              "value": "inverse",
-              "label": "t:sections.all.colors.inverse.label"
-            }
-          ],
-          "default": "accent-1",
-          "label": "t:sections.all.colors.label"
-        },
         {
```
ludoboludo, line 163: This probably a bit of problematic change for people updating. But wouldn't this benefit from being changed to a `inline_richtext` type. So we don't need the `url` setting that follows anymore 🤔

assets/global.js
```diff
@@ -675,21 +675,29 @@ class SlideshowComponent extends SliderComponent {
...
-
-    this.sliderAutoplayButton.addEventListener('click', this.autoPlayToggle.bind(this));
     this.addEventListener('mouseover', this.focusInHandling.bind(this));
     this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
     this.addEventListener('focusin', this.focusInHandling.bind(this));
     this.addEventListener('focusout', this.focusOutHandling.bind(this));
 
-    this.play();
-    this.autoplayButtonIsSetToPlay = true;
+    if(this.querySelector('.slideshow__autoplay') !== null) {
```
ludoboludo, line 693: Wouldn't this be the same ? cc: @kmeleta 
```suggestion
    if(this.querySelector('.slideshow__autoplay')) {
```

assets/global.js
```diff
@@ -731,16 +739,24 @@ class SlideshowComponent extends SliderComponent {
   }
 
   focusOutHandling(event) {
-    const focusedOnAutoplayButton = event.target === this.sliderAutoplayButton || this.sliderAutoplayButton.contains(event.target);
-    if (!this.autoplayButtonIsSetToPlay || focusedOnAutoplayButton) return;
-    this.play();
+    if (this.sliderAutoplayButton !== undefined ) {
```
ludoboludo, line 742: Same here, we could simply do this I think:
```suggestion
    if (this.sliderAutoplayButton ) {
```

assets/base.css
```diff
@@ -2246,22 +2250,43 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
+
+.utility-bar--bottom-border {
   border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.08);
+}
+
+.announcement-bar,
+.announcement-bar__announcement {
   color: rgb(var(--color-foreground));
+  width: 100%;
```
ludoboludo, line 2265: Not a huge fan of how far the arrows are from the content: [video](https://screenshot.click/31-12-sougk-1mo2o.mp4)

assets/component-slider.css
```diff
@@ -359,6 +363,43 @@ slider-component.slider-component-full-width {
...
 }
 
+.announcement-bar .slider-button--next {
+  margin-right: -1.6rem;
+  height: 100%;
+}
+
+.announcement-bar .slider-button--prev {
+  margin-left: -1.6rem;
+  height: 100%;
```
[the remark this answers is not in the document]
eugenekasimov: I had to proceed this way because Yoann wanted me to keep the default height of the announcement bar the same as it used to be. If we set a fix height for the buttons 44px then the minimal height of the announcement bar will be bigger.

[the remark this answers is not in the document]
eugenekasimov: @YoannJailin after discussing it with Ludo, I have to admit that his argument is quite valuable and important. I remember that we discussed it with you and you wanted to keep the default height of the announcement bar the same as it used to be. But because we added arrow buttons and there is a requirement for accessibility we should set those buttons at least 44px height and the announcement bar is going to have a minimum height of 44px as well.

assets/base.css
eugenekasimov: Hey @YoannJailin this needs your attention. Thanks.

sections/announcement-bar.liquid
```diff
@@ -1,32 +1,154 @@
...
+        },
+        {
+          "value": "inverse",
+          "label": "t:sections.all.colors.inverse.label"
+        }
+      ],
+      "default": "accent-1",
+      "label": "t:sections.all.colors.label"
+    },
+    {
```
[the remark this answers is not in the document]
eugenekasimov: Ludo and I have come to an agreement that we simply need to observe what happens after upgrading with announcements, and from that, we will determine if a warning about manual update settings for merchants is necessary.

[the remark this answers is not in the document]
ludoboludo: So I tested a manual update and the experience isn't too bad. 
Things you're losing: 
- the color scheme (moved to a section level setting instead of block)
- the text alignment, the setting was removed

But otherwise the blocks are keeping the links they had and coming up as a slider as expected. 
<details>
<summary>Screenshot</summary>


</details>

So we will still need to warn merchants when upgrading but it shouldn't be too bad of an experience.

assets/base.css
YoannJailin: Hello @eugenekasimov @ludoboludo i think it could be good to make the arrow icons more centered so the user can see them more easily / he won't have to go too far to hover it on large screens. 

Let's just add the additional options for now (selectors, social media icons) that will shrink the Announcements slider to the middle including the arrows. We can still make that narrow version of the slider default later, even if there's no other options activated.

assets/base.css
```diff
@@ -2246,22 +2250,78 @@ product-info .loading-overlay:not(.hidden) ~ *,
...
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+  align-content: center;
+}
+
+.announcement-bar-slider,
+.announcement-bar-slider .slider {
+  width: 100%;
+}
```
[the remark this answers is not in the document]
eugenekasimov: Good catch Ludo, thanks. 
Fixed [here](https://github.com/Shopify/dawn/pull/2394/commits/1cc56cc66dd1d0e534f411e6e7a2350b58fbc0b8)

ludoboludo (approved):

kmeleta (approved):
Still looks good to me with the latest changes :shipit:

melissaperreault:
Exciting team! I was just looking at the experience on our main branch and came here to share some observations. I'd like to raise a few key points for a follow up refinement for accessibility and usability before our next release.

1. **It is recommended to provide a "Play/Pause" control that stops the auto-rotating behaviour**, like the Slideshow section. This can then be used explicitly by users, in addition to the current behaviour where it is paused as soon as the user's mouse/pointer is over the slide/section.
2. **Once a user has explicitly activated one of these previous/next buttons, the autoplaying functionality should stop**. If users are currently reading/interacting with content in the carousel, and the slide that they are currently on is advanced and replaced with new content the experience will be disorienting, as they may not realize the reason why the content they were currently on suddenly disappeared.

PR #2401, opened 2023-03-14
Snippetize header menu components

kmeleta:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -144,285 +144,17 @@
 </svg>
 <{% if section.settings.sticky_header_type != 'none' %}sticky-header data-sticky-type="{{ section.settings.sticky_header_type }}"{% else %}div{% endif %} class="header-wrapper color-{{ section.settings.color_scheme }} gradient{% if section.settings.show_line_separator %} header-wrapper--border-bottom{% endif %}">
   <header class="header header--{{ section.settings.logo_position }} header--mobile-{{ section.settings.mobile_logo_position }} page-width{% if section.settings.menu_type_desktop == 'drawer' %} drawer-menu{% endif %}{% if section.settings.menu != blank %} header--has-menu{% endif %}">
+    {% comment %} drawer type menu {% endcomment %}
     {%- if section.settings.menu != blank -%}
-      <header-drawer data-breakpoint="{% if section.settings.menu_type_desktop == 'drawer' %}desktop{% else %}tablet{% endif %}">
-        <details id="Details-menu-drawer-container" class="menu-drawer-container">
-          <summary class="header__icon header__icon--menu header__icon--summary link focus-inset" aria-label="{{ 'sections.header.menu' | t }}">
-            <span>
-              {% render 'icon-hamburger' %}
-              {% render 'icon-close' %}
-            </span>
-          </summary>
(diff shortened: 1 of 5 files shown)


sections/header.liquid
```diff
@@ -145,282 +145,11 @@
...
-                    {%- endif -%}
-                  </ul>
-                </div>
-              </div>
-            </div>
-          </div>
-        </details>
-      </header-drawer>
+      {% render 'header-drawer' %}
     {%- endif -%}
```
ludoboludo, line 149: Could probably use a `{% liquid` tag here too like you did on like `187`

snippets/header-drawer.liquid
ludoboludo, line 1: I think you could run prettier on it. When I save locally it does reformat some of it 👍 
Probably same goes for the other files touched.

ludoboludo (commented):
Looking good 👍 Just a couple nitpicks

snippets/header-drawer.liquid
kmeleta: Done. I didn't notice much on the _new_ header.liquid but most of the snippets had changes.

sections/header.liquid
```diff
@@ -144,283 +144,14 @@
...
-            </div>
-          </div>
-        </details>
-      </header-drawer>
-    {%- endif -%}
+    {%- liquid
+      if section.settings.menu != blank
+        render 'header-drawer'
+      endif
+    -%}
```
ludoboludo, line 151: I was thinking of even including the header search logic that comes right after: 
```liquid
    {%- liquid
      if section.settings.menu != blank
        render 'header-drawer'
      endif

      if section.settings.logo_position == 'top-center' or section.settings.menu == blank
        render 'header-search', is_duplicate: true
      endif
    -%}
```

kmeleta: Hm I thought theme check was yelling at me about the  `, is_duplicate` format in the liquid block. Must have imagined it 🤷

ludoboludo (approved):
👍

metamoni (approved):
Looking good 🎉

ludoboludo (dismissed):

stufreen (approved):
Looks good with one question

snippets/header-search.liquid
```diff
@@ -0,0 +1,97 @@
+{% comment %}
+  Renders a header search modal. Should be used with 'header.liquid'
+
+  Accepts:
+  - is_duplicate: {Boolean} Prevents duplicate ids for subsequent instances
```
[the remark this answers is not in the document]
kmeleta: I'm not sure if my point was received, but this instance definitely doesn't make as strong of a case for that prop pattern as the other snippet I was matching so I decided to adjust.

> I wonder if we even need the id and for attributes if we could achieve the same thing by having the label as the parent of the input?

Well I definitely don't want to refactor every input in dawn to avoid a naming discussion regardless haha, but I imagine a major reason we chose the label for input pattern is so the label could access the :focus context of the input element in css (input:focus ~ label). Off hand I'm not aware of every detail considered in original decision, but I'm sure it wasn't made by accident or with ignorance to the label parent option.

PR #2412, opened 2023-03-15
Gift cards/add recipient

stufreen:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


assets/section-main-product.css
```diff
@@ -1473,3 +1473,129 @@ a.product__text {
...
+}
+
+.recipient-fields hr {
+  margin: 1.6rem auto;
+}
+
+.recipient-form > input[type='checkbox']:checked ~ .recipient-fields {
+  display: block;
+  animation: animateMenuOpen var(--duration-default) ease;
+}
```
kmeleta, line 1563: I wonder if we should provide more feedback to screenreader users that additional content gets revealed when checking the input 🤔 Maybe if we at least prevented the item from being successfully added to cart without a valid email this would be ok. But currently, it's possible you could check the checkbox and press enter to submit the form and add the item to the cart without noticing the additional fields. https://screenshot.click/17-02-h5lay-yhktj.mp4

kmeleta (commented):
This is largely approvable to me. Probably the biggest thing is if we can prevent invalid submission of the form if the fields are all empty. I left a couple comments that kind of touch on that, but I'm open to opinions about if this is a blocker or not.

snippets/gift-card-recipient-form.liquid
```diff
@@ -0,0 +1,164 @@
...
+  Accepts:
+  - product: {Object} product object.
+  - form: {Object} the product form object.
+  - section: {Object} section to which this snippet belongs.
+
+  Usage:
+  {% render 'gift-card-recipient-form', product: product, form: form, section: section %}
+{% endcomment %}
+
+<div class="customer">
```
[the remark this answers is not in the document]
ludoboludo: Oh I think that's something that needed tweaking. Because some styling applied on the customer pages matched what was applied here, I think this class was used but should be duplicated with a more accurate naming if it's still useful or remove if it's not necessary

assets/section-main-product.css
ludoboludo: Same feeling on my side. I brought the possibility to add something like toggling an attribute `aria-expanded="true"`.
[video](https://screenshot.click/15-39-qg6rs-igo6l.mp4)

kmeleta: Yeah unfortunately I'm pretty sure aria-expanded isn't applicable to inputs like this. At least it's not meant to be. So I was picturing it having to be more of a live region type solution.

assets/section-main-product.css
```diff
@@ -161,7 +161,7 @@
   flex: 0 0 100%;
   padding: 0;
   margin: 0 0 1.2rem 0;
-  max-width: 37rem;
+  max-width: 44rem;
```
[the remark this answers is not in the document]
fredma: This was changed to align form fields with buy buttons max-width.

https://user-images.githubusercontent.com/13537657/226006988-07567e97-037e-4964-a2f8-63fe15b8105c.mp4

snippets/gift-card-recipient-form.liquid
```diff
@@ -0,0 +1,164 @@
...
+            class="field__input"
+            id="Recipient-email-{{ section.id }}"
+            type="email"
+            placeholder="{{ 'recipient.form.email' | t }}"
+            name="properties[Recipient email]"
+            value="{{ form.email }}"
+            {% if form.errors contains 'email' %}
+              aria-invalid="true"
+              aria-describedby="RecipientForm-email-error-{{ section.id }}"
+            {% endif %}
```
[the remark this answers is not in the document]
fredma: The validation is happening on the back end if the beta flag `gift_card_add_recipient` is enabled for the time being.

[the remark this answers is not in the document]
kmeleta: Ah perfect. I just assumed we had that enabled on the test store already. I just turned it on for further testing.

assets/section-main-product.css
```diff
@@ -1473,3 +1473,129 @@ a.product__text {
...
   }
 }
+
+/* Recipient form */
+.recipient-form {
+  display: block;
+  position: relative;
+  --recipient-checkbox-line-height: 2.88rem;
+  /* (line-height - 1.6rem) / 2 */
+  --recipient-checkbox-margin-top: 0.64rem;
```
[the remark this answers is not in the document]
antiBaconMachine: I previously gave the opposite feedback to Fred when he wrote this 😂.

My thinking is that a calc using nothing but constants gives no benefit for dynamism as you have to go and manually change each constant each time you want to make a change _and_ it's worse for performance because you're spending clock cycles doing a calculation that will always be the same.

This answers your first question, no that variable is not used anywhere it's left over from when this was a calc, I have removed it and updated the comment.

I haven't put it back to a calc for reasons outlined above but if you prefer that, we can do it. Thanks.

locales/en.default.schema.json
```diff
@@ -827,6 +827,10 @@
             "show_dynamic_checkout": {
               "label": "Show dynamic checkout buttons",
               "info": "Using the payment methods available on your store, customers see their preferred option, like PayPal or Apple Pay. [Learn more](https://help.shopify.com/manual/using-themes/change-the-layout/dynamic-checkout)"
+            },
+            "show_gift_card_recipient": {
+              "label": "Show recipient form for gift card products",
+              "info": "When enabled, gift card products can optionally be sent to a recipient with a personal message."
```
ludoboludo, line 833: I think ideally here we would link to our help doc on how to add a gift card product: https://help.shopify.com/en/manual/products/gift-card-products/add-update-gift-card-products

Though doing it now means re requesting translations.

snippets/gift-card-recipient-form.liquid
```diff
@@ -0,0 +1,164 @@
...
+    >
+    <label class="recipient-checkbox" for="Recipient-Checkbox-{{ section.id }}">
+      <svg
+        width="1.6rem"
+        height="1.6rem"
+        viewBox="0 0 16 16"
+        aria-hidden="true"
+        focusable="false"
+      >
+        <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1.75"></rect>
```
ludoboludo, line 34: I think this was right when using a `stroke-width` of `1`. Facet filters are using that as well. 
```suggestion
        <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
```

snippets/gift-card-recipient-form.liquid
```diff
@@ -0,0 +1,164 @@
...
+      </svg>
+      {%- render 'icon-checkmark' -%}
```
ludoboludo, line 36: This is where the stroke seemed different. We could remove this `render` and use the same code as in the facets.liquid file: 
```html
<svg
  aria-hidden="true"
  class="icon icon-checkmark"
  width="1.1rem"
  height="0.7rem"
  viewBox="0 0 11 7"
  fill="none"
  xmlns="http://www.w3.org/2000/svg"
>
  <path d="M1.5 3.5L2.83333 4.75L4.16667 6L9.5 1" stroke="currentColor" stroke-width="1.75" stroke-linecap="round" stroke-linejoin="round" />
</svg>
```

But some styling will need to be tweaked as well as that svg won't be aligned properly when replacing the render. 
Seems like using those values could do the trick (removing the original height added to it):
```css 
    position: absolute;
    left: 0.3rem;
    z-index: 5;
    top: 0.4rem;
```

ludoboludo: Ill have a look at it, it should be a quick update 👍

assets/customer.css
```diff
@@ -239,6 +239,10 @@
     text-align: right;
   }
 
+  .customer td .properties {
+    text-align: left;
+  }
```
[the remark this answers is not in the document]
danielvan: I know this one is hard. You can ping @YoannJailin if needed.

I agree right aligned is best, but right now the stacking is weird. Could we force a line break? E.g.
Recipient email:
Ludo.segura
Recipient name:
Lu
Message:
Testing this out

We can't design a table or something custom here for gift cards, right?

assets/section-main-product.css
```diff
@@ -1473,3 +1473,129 @@ a.product__text {
...
+  flex-shrink: 0;
+}
+
+.recipient-form .icon-checkmark {
+  visibility: hidden;
+  position: absolute;
+  left: 0.28rem;
+  z-index: 5;
+  top: 0.4rem;
+}
```
ludoboludo, line 1546: I updated the checkmark 👍

assets/recipient-form.js
```diff
@@ -0,0 +1,139 @@
...
+        this.cartErrorUnsubscriber();
+      }
+    }
+
+    onChange() {
+      if (!this.checkboxInput.checked) {
+        this.clearInputFields();
+        this.clearErrorMessage();
+      }
+    }
```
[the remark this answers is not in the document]
fredma: See https://github.com/Shopify/dawn/pull/2672

PR #2425, opened 2023-03-17
Fix empty space in the drawer when there are no social media icons / account / localizations

### LINKED ISSUE #2414: Empty space in the drawer
There is an empty space in the drawer when there are no social media links, account or localizations. 

<details><summary>Screenshot</summary>
Screenshot 2023-03-15 at 5 29 20 PM
</details>

- Dawn Version: 7.0.1
- macOS Version 13.1

## Possible solution
It seems that after we added localization form it broke a css logic that previously hide this space if there were no icons.

We could probably change 
```
.menu-drawer__utility-links:has(ul:nth-child(2):empty) {
    display: none;
  }
```
to 
```
.menu-drawer__utility-links:has(ul:nth-child(n):empty) {
    display: none;
  }
```

eugenekasimov:
### PR Summary: 
This PR fixes empty space in the drawer when there are no social media icons. account or localizations. 

Fixes #2414 .

- Changed `padding: 2rem` to `padding: 0` as a default for`.menu-drawer__utility-links` container. 
- Created variables to store information about if there are social media links, account or localization forms.
- Added a few extra css classes for each scenario.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios
- [ ] Choose desktop menu - drawer.
- [ ] Add/Remove all social media links. Test it for desktop, tablet and mobile.
- [ ] Disable/enable `Show login link in the header of online store and at checkout ` in Admin. Test it for desktop, tablet and mobile.
- [ ] Add/remove localization forms. Test it for desktop, tablet and mobile.
- [ ] Try different combinations of three previous steps 👆. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139547312150/editor)

### Checklist

### DIFF
--- a/assets/component-menu-drawer.css
+++ b/assets/component-menu-drawer.css
@@ -220,18 +220,18 @@ details[open].menu-opening > .menu-drawer__submenu {
   position: relative;
 }
 
-.header.header--has-social .menu-drawer__utility-links {
+.header--has-social .menu-drawer__utility-links {
   padding: 2rem;
 }
 
 @media screen and (max-width: 749px) {
-  .header.header--has-account:where(:not(.header--has-social):not(.header--has-localizations)) .menu-drawer__utility-links {
+  .header--has-account:where(:not(.header--has-social):not(.header--has-localizations)) .menu-drawer__utility-links {
     padding: 2rem;
   }
 }
 
 @media screen and (max-width: 989px) {
-  .header.header--has-localizations:where(:not(.header--has-social)) .menu-drawer__utility-links {
+  .header--has-localizations:where(:not(.header--has-social)) .menu-drawer__utility-links {
     padding: 2rem;
   }
 }


assets/component-menu-drawer.css
```diff
@@ -235,7 +235,7 @@ details[open].menu-opening > .menu-drawer__submenu {
 }
 
 @media screen and (min-width: 750px) {
-  .menu-drawer__utility-links:has(ul:nth-child(2):empty) {
+  .menu-drawer__utility-links:has(ul:empty) {
```
ludoboludo, line 238: That doesn't work properly because locale selectors should be coming up in the drawer on tablet sizes (`750px - 990px`). And with this it doesn't account for it. 

Because we're hiding the wrapper just because of social media when we're not checking on the locales being enabled or not. 

[Video](https://screenshot.click/17-32-qlayx-1zd15.mp4)

So maybe it can be a different selector: 
```css
.menu-drawer__utility-links:has(.menu-drawer__account + ul:empty) {
```
Something to test 🤔

ludoboludo (changes requested):

assets/component-menu-drawer.css
ludoboludo: Damn, so yeah then maybe we need to rely on adding a class on the wrapper to know when locales are enable in the header.

kmeleta: You're correct (I just edited my message to account for that actually haha), but it is _fairly_ bad support atm. If we think this is minor enough, that's fine, but probably still only is it really helps us avoid a much more complex fix.

eugenekasimov:
@kmeleta  @ludoboludo

I changed my approach and now it seems I covered all possible scenarios without using JS or `:has`.

assets/component-menu-drawer.css
```diff
@@ -215,11 +215,27 @@ details[open].menu-opening > .menu-drawer__submenu {
...
   background-color: rgba(var(--color-foreground), 0.03);
   position: relative;
 }
 
+.menu-drawer__utility-links--social-media-links {
+  padding: 2rem;
+}
+
+@media screen and (max-width: 749px) {
+  .menu-drawer__utility-links--account {
```
ludoboludo, line 228: Probably too much just to prevent the styling from being applied multiple times. 😅 
```suggestion
  .menu-drawer__utility-links--account:where(:not(.menu-drawer__utility-links--social-media-links):not(.menu-drawer__utility-links--localizations)) {
```

assets/component-menu-drawer.css
```diff
@@ -215,11 +215,27 @@ details[open].menu-opening > .menu-drawer__submenu {
...
+    padding: 2rem;
+  }
+}
+
+@media screen and (max-width: 989px) {
+  .menu-drawer__utility-links--localizations {
```
ludoboludo, line 234: Maybe too much, but this prevents repeating the same styling 3 times 🤷 
```suggestion
  .menu-drawer__utility-links--localizations:where(:not(.menu-drawer__utility-links--social-media-links)) {
```

snippets/header-drawer.liquid
```diff
@@ -110,7 +110,19 @@
...
             </ul>
           </nav>
-          <div class="menu-drawer__utility-links">
+          {%- liquid
+            assign social_links = false
+            assign localization_forms = false
+
+            if settings.social_twitter_link != blank or settings.social_facebook_link != blank or settings.social_pinterest_link != blank or settings.social_instagram_link != blank or settings.social_youtube_link != blank or settings.social_vimeo_link != blank or settings.social_tiktok_link != blank or settings.social_tumblr_link != blank or settings.social_snapchat_link != blank
+              assign social_links = true
+            endif
```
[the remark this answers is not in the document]
eugenekasimov: That's a good idea Ludo, I totally agree. Thanks!

snippets/header-drawer.liquid
```diff
@@ -110,7 +110,19 @@
...
+
+            if localization.available_countries.size > 1 and section.settings.enable_country_selector or section.settings.enable_language_selector and localization.available_languages.size > 1
+              assign localization_forms = true
+            endif
```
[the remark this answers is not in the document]
eugenekasimov: This one I believe should work if I add variables that will store
`localization.available_countries.size > 1 and section.settings.enable_country_selector section.settings.enable_language_selector and localization.available_languages.size > 1`
to theme.liquid then we can reuse it for both the header and footer.

assets/component-menu-drawer.css
eugenekasimov: Honestly, since there is no harm I'd keep the classes simple and more readable 🙂. At the same time, I don't strong feeling about it. I'm okay to add what you suggest if @Ken votes for this too 🙂.

snippets/header-drawer.liquid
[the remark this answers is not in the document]
ludoboludo: I think it's similar no ? In the footer it seems like we're checking if all of them are empty. While here you check if any of them has a value.

snippets/header-drawer.liquid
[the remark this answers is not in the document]
ludoboludo: Oh actually here it would be just for the header. Cause we're checking on the `section.settings`. 

So you could add it to `header.liquid` and pass in that value to the `header-drawer` snippet if you need to reference it there as well.

assets/component-menu-drawer.css
kmeleta: I guess I kind of like this type of thing being handled via classes a lot of the time. Not necessarily ridiculous BEM monstrosities like `menu-drawer__utility-links--social-media-links` 😅  but generally. 

To poke at this in another way though (sorry), the bug we're fixing right now is specific to the drawer utility links element, but the conditions we're checking are potentially applicable to anything in the header. Would it make sense to add classes like `header--has-account`, `header--has-localizations`, and `header--has-social` at a higher level so we can potentially make use of them in the future without having to rewrite all this code? I've previously had to (or wanted to) refactor a lot modifier classes and liquid conditionals done too deep in the tree.

eugenekasimov: It's wise. Thanks Ken for a good idea. 

It's done ✅

snippets/header-drawer.liquid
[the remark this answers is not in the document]
ludoboludo: We're not adding this ? I don't mind the way it currently is. We can always re work it later. It's not hard to read in its current state.

ludoboludo (approved):
👍

assets/component-menu-drawer.css
```diff
@@ -215,11 +215,27 @@ details[open].menu-opening > .menu-drawer__submenu {
...
   background-color: rgba(var(--color-foreground), 0.03);
   position: relative;
 }
 
+.header.header--has-social .menu-drawer__utility-links {
+  padding: 2rem;
+}
+
+@media screen and (max-width: 749px) {
+  .header.header--has-account:where(:not(.header--has-social):not(.header--has-localizations)) .menu-drawer__utility-links {
```
kmeleta, line 228: Are we just using :where() to reduce specificity here? If so, I suppose that's reasonable. Otherwise it shouldn't be necessary.

assets/component-menu-drawer.css
```diff
@@ -215,11 +215,27 @@ details[open].menu-opening > .menu-drawer__submenu {
 }
 
 .menu-drawer__utility-links {
-  padding: 2rem;
+  padding: 0;
   background-color: rgba(var(--color-foreground), 0.03);
   position: relative;
 }
 
+.header.header--has-social .menu-drawer__utility-links {
```
kmeleta, line 223: I don't think any of these rules you have need the additional .header class in the selector here do you?. Those classes shouldn't belong to any other type of element and because we use BEM, it should be clear enough that `header--has-social` is effectively also  targeting `.header`

kmeleta (approved):
Looks good. Left a little nitpick, but not a big deal.

assets/component-menu-drawer.css
eugenekasimov: Yeah, I forgot to mention it, without :where it stays with the same specificity.

PR #2458, opened 2023-03-22
Add aria-expanded and aria-control attribbutes to gift card checkbox

antiBaconMachine:
### PR Summary: 

Add aria-expanded and aria-control attribbutes to gift card checkbox

The intent here is to give context to screen readers that the gift card recipient form is being hidden and shown. 

I'm using the example here https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/ as a reference but I'm not getting great results from mac voiceover. Android talkback is better, when I highlight the checkbox it correctly announces the collapsed/expanded state. It does seem that voiceover is lacking in support for aria-expanded.

### Other considerations

I should note that it is only the checkbox that gets the state not the label which would be more commonly selected as it's bigger. I think this is a good first step to solving this issue, though we may want to do more.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

None

### Testing steps/scenarios

Activate your preferred screenreader and listen to the results as you toggle the checkbox on and off. 

### Demo links

- [Store]( https://9f55df.myshopify.com/?preview_theme_id=132571070654)
- [Editor](https://9f55df.myshopify.com/admin/themes/132571070654/editor)

### Checklist

### DIFF
--- a/snippets/gift-card-recipient-form.liquid
+++ b/snippets/gift-card-recipient-form.liquid
@@ -23,6 +23,8 @@
       name="{{ gift_card_recipient_control_flag }}"
       disabled
       unchecked
+      aria-expanded="false"
+      aria-controls="Recipient-fields-{{ section.id }}"
     >
     <label class="recipient-checkbox" for="Recipient-Checkbox-{{ section.id }}">
       <svg
@@ -76,7 +78,7 @@
         {%- endif -%}
       </ul>
     </div>
-    <div class="recipient-fields">
+    <div class="recipient-fields" id="Recipient-fields-{{ section.id }}">
       <hr>
       <div class="recipient-fields__field">
         <div class="field">
--- a/assets/recipient-form.js
+++ b/assets/recipient-form.js
@@ -59,6 +59,9 @@ if (!customElements.get('recipient-form')) {
       if (!this.checkboxInput.checked) {
         this.clearInputFields();
         this.clearErrorMessage();
+        this.checkboxInput.setAttribute('aria-expanded', false)
+      } else {
+        this.checkboxInput.setAttribute('aria-expanded', true)
       }
     }
 


snippets/gift-card-recipient-form.liquid
```diff
@@ -23,6 +23,8 @@
       name="{{ gift_card_recipient_control_flag }}"
       disabled
       unchecked
+      aria-expanded="false"
```
metamoni, line 26: I think the reason why this isn't working as you expected it is because it's missing a role. aria-expanded works with the following [associated roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles). A button, for example, has an implicit role, so you wouldn't need to add one. But for an element without an implicit role, such as a div or this input, the screen reader will require more context to understand what the widget does.

I'm not super familiar with this area, but I'd be happy to take a look at it myself if you add some detailed testing steps (some simple instructions on how to get to this screen/checkbox would be helpful). Feel free to reach out if you'd like to pair too.

metamoni: Thanks, @ludoboludo. @antiBaconMachine and I have discussed this, as well as potential solutions. Here's a summary:
- we both agreed that a checkbox might not be the best element to use here
- I suggested changing the element to a `button` and using one of the following patterns: an [accordion](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) or a [disclosure](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/) (which we use in the drawer menu)
- I suggested checking with UX to find a good pattern and adjust the design if needed

ludoboludo: Sweet, thanks for the update 🙂

bakura10:
I am facing some issues here as well. What about creating an aria-owns relationship between the checkbox input and the div that owns the content? I think the checkbox is the right element here for holding the decision :).

Issue #2262, opened 2023-01-27
[Header] Localization selectors follow up for no-js


andrewetchen:
### Follow up for no-JS header language + currency selectors

- [ ] Add and style no-JS header language + currency selectors on desktop
- [ ] Add and style no-JS header language + currency selectors in menu drawer

### Related

- https://github.com/Shopify/dawn/pull/2258
- https://github.com/Shopify/dawn/issues/2240

### Notes

- Follow existing patterns for no-JS button styling
- Have PR approved before next release

### Screenshot

**no-JS header language + currency selectors on desktop**

![image](https://user-images.githubusercontent.com/28404165/215125433-ce3ef924-7be8-4504-9cf4-063104fa8b02.png)

**no-JS header language + currency selectors in menu drawer**

![image](https://user-images.githubusercontent.com/28404165/218801410-82507152-3466-4fa9-b2b0-712e58f65922.png)

PR #2485, opened 2023-04-03
[Header] No-js behaviour for locale selectors

ludoboludo:
### PR Summary: 

Adding the locale selector for the no-js behaviour in the header

Fixes https://github.com/Shopify/dawn/issues/2262

In the original PR, the no script behaviour was there at first so I re added it with some extra styling tweaks.

Visual impact only on the no JS environment which is a very niche use case. 

### Testing steps/scenarios

- [ ] Make sure the locale selectors are enabled in the header
- [ ] then disable JS in for your browser tab/window and make sure the local selectors are still coming up and working as expected. (To disable JS in chrome, you can right click on the page then `Inspect`, click on the settings wheel in the inspect tool, scroll all the way to the bottom and find the `Debugger > Disable JavaScript`

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139634049046)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139634049046/editor)

### Checklist

### DIFF
--- a/sections/header.liquid
+++ b/sections/header.liquid
@@ -248,9 +248,25 @@
     <div class="header__icons{% if section.settings.enable_country_selector or section.settings.enable_language_selector %} header__icons--localization header__localization{% endif %}">
       <div class="desktop-localization-wrapper">
         {%- if section.settings.enable_country_selector and localization.available_countries.size > 1 -%}
-          <localization-form class="small-hide medium-hide">
+          <noscript class="small-hide medium-hide">
+            {%- form 'localization', id: 'HeaderCountryMobileFormNoScript', class: 'localization-form' -%}
+              <div class="localization-form__select">
+                <h2 class="visually-hidden" id="HeaderCountryMobileLabelNoScript">{{ 'localization.country_label' | t }}</h2>
+                <select class="localization-selector link" name="country_code" aria-labelledby="HeaderCountryMobileLabelNoScript">
+                  {%- for country in localization.available_countries -%}
+                    <option value="{{ country.iso_code }}" {%- if country.iso_code == localization.country.iso_code %} selected{% endif %}>
+                      {{ country.name }} ({{ country.currency.iso_code }} {{ country.currency.symbol }})
+                    </option>
+                  {%- endfor -%}
+                </select>
+                {% render 'icon-caret' %}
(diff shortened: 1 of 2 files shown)


sections/header.liquid
```diff
@@ -248,9 +248,26 @@
     <div class="header__icons{% if section.settings.enable_country_selector or section.settings.enable_language_selector %} header__icons--localization header__localization{% endif %}">
       <div class="desktop-localization-wrapper">
         {%- if section.settings.enable_country_selector and localization.available_countries.size > 1 -%}
-          <localization-form class="small-hide medium-hide">
+          <noscript class="small-hide medium-hide">
+            {%- form 'localization', id: 'HeaderCountryMobileFormNoScript', class: 'localization-form' -%}
+              <div class="localization-form__select">
+                <h2 class="visually-hidden" id="HeaderCountryMobileLabelNoScript">{{ 'localization.country_label' | t }}</h2>
```
stufreen, line 254: Why `h2` and not a `label`?

sections/header.liquid
```diff
@@ -248,9 +248,26 @@
...
+                </select>
+                {% render 'icon-caret' %}
+              </div>
+              <button class="button button--tertiary">{{ 'localization.update_country' | t }}</button>
+            {%- endform -%}
+          </noscript>
+
+          <localization-form class="small-hide medium-hide no-js-hidden">
             {%- form 'localization', id: 'HeaderCountryForm', class: 'localization-form' -%}
-              <div class="no-js-hidden">
```
stufreen, line 253: Non blocking but just curious: do you know why `no-js-hidden` was on the inner div before?

stufreen (approved):
LGTM. It really makes the header big, but it works!

sections/header.liquid
ludoboludo: Reusing the approach from the footer and also what's in the JS approach. I believe the reason why is because it helps navigate to it more easily when using a screen reader and using the VoiceOver rotor which shows you all the page's headings.

ludoboludo:
Yeah I've noticed this too @eugenekasimov. I think it's ok as is. It's the same behaviour we've had for the locale selectors in the footer. 🤷

PR #2410, opened 2023-03-15
animate Collage on scroll

metamoni:
### Why are these changes introduced?

Fixes #2319 

### Decision log

| # | Decision | Alternatives | Rationale | Downsides |
|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist
- [x] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [x] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [x] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [x] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [x] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2405, opened 2023-03-15
Animate featured product and product info

metamoni:
### PR Summary: 
Adds fade in on scroll to the following areas
- Main product (product info)
- Featured product

### Why are these changes introduced?

Fixes #2390 

### What approach did you take?
Replicated animations from the [original prototype](https://github.com/Shopify/dawn/pull/2141)

### Visual impact on existing themes
Will animate product info and featured product on scroll, when `Reveal sections on scroll` is enabled

https://user-images.githubusercontent.com/22390758/225381696-d6932ba5-770d-464f-8a7a-6829648944e9.mp4

### Testing steps/scenarios
- [ ] Enable `Reveal sections on scroll` under `Animations` in [Theme settings](https://os2-demo.myshopify.com/admin/themes/139535613974/editor?context=theme)
- [ ] On the home page, add a Featured Product section and verify that it fades in on scroll
- [ ] Navigate to the product page and verify that the product information section fades in on scroll

### Demo links
- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139535613974)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139535613974/editor)

### Checklist
- [x] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [x] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [x] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [x] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [x] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [x] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2499, opened 2023-04-06
Fine-tune the slideIn animation

kjellr:
### PR Summary: 

Minor adjustments to the slideIn scroll animation. Intent is to strike a natural-feeling balance between subtle and apparent. 

<details><summary>Video Comparison</summary>

#### Before:

https://user-images.githubusercontent.com/1202812/230602988-77319ac0-2668-4382-9343-3e66341a0b94.mp4

#### After:

https://user-images.githubusercontent.com/1202812/230602977-93eb1ed3-8be2-4eb5-99ba-37652c863e3c.mp4

</details>

- Decreasing the `translateY` distance, in an effort to tone down the overall motion.  
- Decreasing the speed of the animation from `--duration-extra-long` to a new `--duration-medium` value. (The standalone `animate--fade-in` speed remains the same, since that animation tends to cover larger areas).

### Testing steps/scenarios

- Visit the demo store, scroll around. 
- Compare to [the existing animation here](https://os2-demo.myshopify.com/?preview_theme_id=139658493974). 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139665866774)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139665866774/editor)

### Checklist


assets/base.css
```diff
@@ -3302,7 +3302,7 @@ details-disclosure > details {
 
   @keyframes slideIn {
     from {
-      transform: translateY(10vh);
+      transform: translateY(3rem);
```
ludoboludo, line 3305: It's much better with this 👍 

I wonder if we should also reduce the speed at which the animation run 🤔 I think first visits are fine but then when you keep navigating, you don't want to have the user to start wondering weather it's the time it takes for things to load or if we're slowing down their experience for the sake of animation. 

What do you think ? I find `--duration-default` to be pretty decent.

kjellr: `--duration-default` felt a little too fast to me. I pushed a compromise in 128ad25 — let me know how that feels to you.

ludoboludo: Yeah it's looking nice 👌 Should be a good starting point and we can tune it more in the future if needed 👍

kjellr: Great! I've taken this out of draft and opened it up for reviews. 🌟

ludoboludo (approved):
Looking good, just have one comment

assets/base.css
```diff
@@ -744,9 +744,10 @@ details > * {
 :root {
   --duration-short: 100ms;
   --duration-default: 200ms;
+  --duration-medium: 400ms;
   --duration-long: 500ms;
   --duration-extra-long: 600ms;
```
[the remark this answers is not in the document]
kjellr: The long duration is still used for product card hovers (we can probably re-evaluate that later... I suspect that this new 400ms one will also work there). The extra-long duration is also still used when we fade-in large items on scroll, so they don't appear to flash.

metamoni (approved):

assets/base.css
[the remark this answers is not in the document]
stufreen: I think `extra-long` was only used for the new animations, right? Can it be removed?

stufreen (approved):
LGTM. I think it's a bit snappier.

One thing you could experiment with is different easings, but longer durations... I find cubic bezier better for things like this because it takes a bit more time settling into place. E.g. https://cubic-bezier.com/#.17,.67,.55,.98

assets/base.css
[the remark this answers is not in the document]
kjellr: We're still using `extra-long` for the standalone [opacity fade-in](https://github.com/Shopify/dawn/blob/f9bb7bcf390989ec2f04de6ee1c8430c06d0c0f3/assets/base.css#L3284). We use this for large images, and I found it starts to seem like flashing when we move to anything quicker.

kjellr:
While the improvement here felt pretty good, I thought I'd try another pass using @stufreen's `cubic-bezier()` suggestion. In [aba46c1](https://github.com/Shopify/dawn/pull/2499/commits/aba46c11d67d0a26a2a558e689b76875827a4dab), I used a slightly longer transition, with a tweaked version of `ease-out`. I also slightly shortened the timing for items that cascade in. 

This version feels pretty good to me too. If you're scrolling slow, the animation is there but not overwhelming. If you're scrolling quickly, the animation's pretty subtle, and I kinda like that. 

Give it a try and let me know what you think! It might also be worth testing this once https://github.com/Shopify/dawn/pull/2405 and https://github.com/Shopify/dawn/pull/2410 are merged in so we have a better sense of it all.

stufreen (approved):
Looks great! Subtly better now

ludoboludo (approved):

PR #2502, opened 2023-04-07
[Motion] No animation on added/edited section

ludoboludo:
### PR Summary: 

Prevent having animation while being in the editor. 

Fixes https://github.com/Shopify/themes/issues/400

I added some logic that only runs in the `DOMContentLoaded` listener, `shopify:section:reorder`, and `shopify:section:load`. So it should run when the page loads, when you're re ordering a section or when you're editing a section in the editor. 
It adds a new class that overwrite the other ones. 

### Other considerations

At first I was thinking of removing the existing classes we had but I though it was cleaner/clearer to have a specific class to prevent `fade in` and `slide in` animations. 

### Decision log

|---|---|---|---|---|
| 1 | ~Do not apply animation on elements in the viewport on page load~  | Animate on page load all elements  | It can be a bit jarring to always have animated things show up.  | 🤷   |
| 2 | Do not apply animation on elements being edited in the theme editor  | Animate elements everytime they reload  | it is not necessary to show the animation every time a merchant edit their section  | Merchant could wonder why the animation isn't applied when a section reloads   |

In the editor, prevent a section from being animated anytime it's re rendered due to setting changes. 

### Testing steps/scenarios

- [ ] Test in the editor while tweaking a section and/or its blocks

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139674681366/editor)

### Checklist

### DIFF
--- a/assets/animations.js
+++ b/assets/animations.js
@@ -13,11 +13,11 @@ function onIntersection(elements, observer) {
   });
 }
 
-function initializeScrollAnimationTrigger(rootEl = document, isDesignMode = false) {
+function initializeScrollAnimationTrigger(rootEl = document, isDesignModeEvent = false) {
   const animationTriggerElements = Array.from(rootEl.getElementsByClassName(SCROLL_ANIMATION_TRIGGER_CLASSNAME));
   if (animationTriggerElements.length === 0) return;
 
-  if (isDesignMode) {
+  if (isDesignModeEvent) {
     animationTriggerElements.forEach((element) => {
       element.classList.add('scroll-trigger--design-mode');
     });


kjellr:
This feels pretty good! @ludoboludo, how do you feel about the experience with these changes? 

I do sort of miss the initial slide-in of some above-the-fold elements, but maybe that's because I've been seeing them slide in for so long now. Let's all test this and let it sit for a few days before we merge in.

ludoboludo:
After tweaking our animations which are feeling much nicer now, we agreed not to prevent them from happening for elements in the viewport on load. So this PR (description updated) is more about preventing the animation to happen when not necessary in the theme editor.

kjellr:
Thanks for revising the PR, @ludoboludo! This feels pretty good. Is it possible to turn off all the animations during this zoomed-out reorder state? 

https://user-images.githubusercontent.com/1202812/231459646-7ca14c11-5cee-4eb5-a407-91acf9487c1a.mp4

assets/animations.js
```diff
@@ -25,4 +42,9 @@ function initializeScrollAnimationTrigger(rootEl = document) {
 
 window.addEventListener('DOMContentLoaded', () => initializeScrollAnimationTrigger());
 
-document.addEventListener('shopify:section:load', (event) => initializeScrollAnimationTrigger(event.target));
+if (Shopify.designMode) {
+  document.addEventListener('shopify:section:load', (event) => initializeScrollAnimationTrigger(event, event.target));
+  document.addEventListener('shopify:section:reorder', (event) =>
```
ludoboludo, line 47: @kjellr I added something for the section being re ordered. One thing to note, when I checked, I can only prevent the animation on the section I'm currently trying to re order. Any other section has the animation re running. 

When I checked, `section:load` event wasn't triggered on other sections, so it must be a full page reload. I think it's ok though as long as the section we're currently moving around isn't animating every time we're moving its order. 

[Here is a little video of how it looks](https://screenshot.click/13-18-rxdnv-x24jn.mp4)

kjellr: Hmm ok. I'd definitely prefer to have the other animations stop on reorder too. Can we add a body class with some `!important` rules or something to disable all animations when you're in that mode or something? 😅 

Either way, what we have here is an improvement, so happy to get this merged in and tackle improvements separately. I'll add an approval for now, with the caveat that if we come up with a quick fix to stop all animation on reorder, we should take it.

kjellr (approved):
This generally looks good. Add a few additional notes in https://github.com/Shopify/dawn/pull/2502#discussion_r1165775252.

assets/animations.js
ludoboludo: Ok so I can actually do it by making it remove the animation from all present sections that usually do animate. One of the potential issue with this, is that if you're at the top of a page and you haven't fully scrolled down yet and triggered the animations and you're reordering one of your first section on the page, nothing will be animated on further navigation on this page until page reload. 

[Here is a video ](https://screenshot.click/13-24-76sba-d2e80.mp4)cause it's seem hard to explain without visual aid.

kjellr: Ah, nice work! This seems like a pretty good solution. From watching the video, I'm not concerned about the subsequent sections not animating. That seems like a reasonable tradeoff in this case. If you're happy with the dev solution, let's push it up here and test it out.

assets/animations.js
```diff
@@ -13,10 +13,18 @@ function onIntersection(elements, observer) {
   });
 }
 
-function initializeScrollAnimationTrigger(rootEl = document) {
+function initializeScrollAnimationTrigger(event, rootEl = document) {
```
stufreen, line 16: It seems like the only point of passing in `event` here is the check on line 22.

Would it be more self-explanatory to have a parameter like `isEditorEvent` instead which is a boolean?

assets/animations.js
```diff
@@ -25,4 +33,7 @@ function initializeScrollAnimationTrigger(rootEl = document) {
 
 window.addEventListener('DOMContentLoaded', () => initializeScrollAnimationTrigger());
 
-document.addEventListener('shopify:section:load', (event) => initializeScrollAnimationTrigger(event.target));
+if (Shopify.designMode) {
+  document.addEventListener('shopify:section:load', (event) => initializeScrollAnimationTrigger(event, event.target));
+  document.addEventListener('shopify:section:reorder', (event) => initializeScrollAnimationTrigger(event));
```
stufreen, line 38: If you don't pass in an element to `initializeScrollAnimationTrigger` it will create a new intersection observer for the entire document, which already has one from line 34. Can we just create the new observer for the section(s) that was reordered?

assets/animations.js
```diff
@@ -13,10 +13,18 @@ function onIntersection(elements, observer) {
...
 
-function initializeScrollAnimationTrigger(rootEl = document) {
+function initializeScrollAnimationTrigger(event, rootEl = document) {
   const animationTriggerElements = Array.from(rootEl.getElementsByClassName(SCROLL_ANIMATION_TRIGGER_CLASSNAME));
   if (animationTriggerElements.length === 0) return;
 
+  if (Shopify.designMode) {
+    animationTriggerElements.forEach((element) => {
+      if (event) {
+        element.classList.add('scroll-trigger--in-viewport');
```
metamoni, line 23: We can improve this class name, as it's applied to all elements that have the `scroll-trigger` class on them and not just the ones currently in your viewport. I think this name would be better suited for when we solve the [above-the-fold issue](https://github.com/Shopify/themes/issues/400). How about something like `scroll-trigger--design-mode` or `scroll-trigger--editor-mode`?

assets/animations.js
```diff
@@ -13,10 +13,16 @@ function onIntersection(elements, observer) {
   });
 }
 
-function initializeScrollAnimationTrigger(rootEl = document) {
+function initializeScrollAnimationTrigger(rootEl = document, isDesignMode = false) {
   const animationTriggerElements = Array.from(rootEl.getElementsByClassName(SCROLL_ANIMATION_TRIGGER_CLASSNAME));
   if (animationTriggerElements.length === 0) return;
 
+  if (isDesignMode) {
```
LucasLacerdaUX, line 20: Is `isDesignMode` simply `Shopify.designMode` or is there another use case where these won't be equal?

If they're the same thing, do we need this another variable/parameter instead of simply checking for `Shopify.designMode`?

assets/animations.js
ludoboludo: I had it for the specific section in the first place but it worked in a way that only the re ordered section was not getting animated while every other section was ([video](https://screenshot.click/17-02-bkbxc-6bz2f.mp4)). 

What I did just now was adding an early return within my if statement. So that if the event triggering the initialization is coming from the editor (`(re)load` or `reorder`) so we're not observing more element as it's unnecessary since the classes applied prevent the animation from happening anyway. 

https://github.com/Shopify/dawn/pull/2502/commits/e89ed0294309e4ece9879885f7dddd0c19ff0f55

PR #2506, opened 2023-04-10
Add unique ids to all header menu items

### LINKED ISSUE #2336: Duplicate IDs in navigation menu
If two different top-level navigation items both have second-level items with third-level items at the same position, the `details` elements in the second-level items will have duplicate IDs in the tablet/mobile menu.

HTML elements should not have duplicate IDs

- Dawn Version: 8.0.0
- Firefox Version 110
- macOS Version 13.2.1

## Possible solution
Add top-level position to the ID generation strategy

Issue can be seen here by running [the axe browser extension test ](https://www.deque.com/axe/) on [this site](https://sterner-stuff-test.myshopify.com/)
password - rtohre

### LINKED ISSUE #2419: Add ids and class names to menu items
### Summary

Update menu items in header to have class names and ids that will unlock targeting with CSS.

### Details

We added a Custom CSS feature at the bottom of each section that allows merchants to configure custom CSS without going off the update path.

However, none of our menu items have CSS classes or ids, making them difficult to style. This issue is to add `class` and/or `id` attributes to menu items that will allow merchants to target specific menu items.

An example use case would be a merchant wanting to apply a very custom styling to only one menu item without editing code.

Implementer should decide what the ideal `id` and/or `class` naming pattern should be to ensure uniqueness and flexibility.

kmeleta:
### PR Summary: 

Added unique ids for header navigation links for easier styling via Custom CSS.

Fixes #2419 
Fixes #2336 

Gives all header menu items, in all levels, unique ids to allow easier targeting with CSS selectors when using the Header section Custom CSS setting.

I considered two approaches, one primarily using classes and one using explicit ids. The idea with classnames would be to give each ul/list-menu element a class representing it's level and likely cleaning up/consolidating the classes used for the menu items themselves (li, a, and summary). But even with clearer classes I did not find it exceptionally easy to target _only_ the specific item you want in any moderately complex menu structure. 

I could see there being other benefits to adding additional classes for us (which we can add as needed), but for the purpose of improving custom css targeting of menu items, using ids seems like the better approach. 

### Decision log

|---|---|---|---|---|
| 1 | Use ids on the link elements | - Use classes<br>- Use classes and ids<br>- Use ids on ul or li item | - Results in the most minimal selector | - No real advantage for us or developer ergonomics |
| 2 | Include ancestor link handles in the id string (e.g `#HeaderMenu-Apparel-Shoes-Louise`) | - Use only current link's handle and add additional liquid logic to ensure uniqueness (e.g `#HeaderMenu-Louise`)<br>- Use only current link's handle and the menu level to minimize risk of duplicates | - Duplicates are not possible unless the same menu has 2 links that are actually duplicates for some reason | - Level 2 and 3 links will likely have very long ids |

No visual changes should be noticeable.

### Testing steps/scenarios
Ensure ids display for every menu item and there are no duplicates, checking the following..
- All menu types Dropdown, megamenu, drawer
- All 3 levels of menus
- Menu items with and with child links

You should be able to target and style any menu item on any level via the header's custom css

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139685298198/editor)

### Checklist

### DIFF
--- a/snippets/header-drawer.liquid
+++ b/snippets/header-drawer.liquid
@@ -25,7 +25,10 @@
                 <li>
                   {%- if link.links != blank -%}
                     <details id="Details-menu-drawer-menu-item-{{ forloop.index }}">
-                      <summary class="menu-drawer__menu-item list-menu__item link link--text focus-inset{% if link.child_active %} menu-drawer__menu-item--active{% endif %}">
+                      <summary
+                        id="HeaderMenu-{{ link.handle }}"
+                        class="menu-drawer__menu-item list-menu__item link link--text focus-inset{% if link.child_active %} menu-drawer__menu-item--active{% endif %}"
+                      >
                         {{ link.title | escape }}
                         {% render 'icon-arrow' %}
                         {% render 'icon-caret' %}
@@ -45,6 +48,7 @@
                               <li>
                                 {%- if childlink.links == blank -%}
                                   <a
+                                    id="HeaderMenu-{{ link.handle }}-{{ childlink.handle }}"
                                     href="{{ childlink.url }}"
                                     class="menu-drawer__menu-item link link--text list-menu__item focus-inset{% if childlink.current %} menu-drawer__menu-item--active{% endif %}"
                                     {% if childlink.current %}
@@ -55,7 +59,10 @@
                                   </a>
                                 {%- else -%}
                                   <details id="Details-menu-drawer-submenu-{{ forloop.index }}">
-                                    <summary class="menu-drawer__menu-item link link--text list-menu__item focus-inset">
+                                    <summary
+                                      id="HeaderMenu-{{ link.handle }}-{{ childlink.handle }}"
+                                      class="menu-drawer__menu-item link link--text list-menu__item focus-inset"
+                                    >
                                       {{ childlink.title | escape }}
                                       {% render 'icon-arrow' %}
                                       {% render 'icon-caret' %}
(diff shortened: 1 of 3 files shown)


snippets/header-drawer.liquid
```diff
@@ -55,7 +59,10 @@
                                   </a>
                                 {%- else -%}
                                   <details id="Details-menu-drawer-submenu-{{ forloop.index }}">
-                                    <summary class="menu-drawer__menu-item link link--text list-menu__item focus-inset">
+                                    <summary
+                                      id="HeaderMenu-{{ link.handle }}-{{ childlink.handle }}"
```
kmeleta, line 63: I'm also not married to the dash separated format here. I considered doing `HeaderMenu_{{ link.handle }}_{{ childlink.handle }}` because the handles will already being using dashes. But I wasn't convinced there was a significant need to clearly indicate the separation.

snippets/header-dropdown-menu.liquid
```diff
@@ -12,7 +12,10 @@
         {%- if link.links != blank -%}
           <header-menu>
             <details id="Details-HeaderMenu-{{ forloop.index }}">
-              <summary class="header__menu-item list-menu__item link focus-inset">
+              <summary
+                id="HeaderMenu-{{ link.handle }}"
```
stufreen, line 16: Is there potential/likelihood to have multiple elements with the same id, since we render both the mobile drawer and dropdown menu? I believe that's invalid HTML and will show a console error.

If so you could add it as a class or data attribute instead - I think it achieves the same goal of being able to target the elements with CSS. Alternatively you could have different prefix for each menu (e.g. `HeaderDrawer`) which would make it even more targetable but a little more CSS.

The naming convention looks good to me 👍

snippets/header-drawer.liquid
```diff
@@ -54,8 +58,11 @@
                                     {{ childlink.title | escape }}
                                   </a>
                                 {%- else -%}
-                                  <details id="Details-menu-drawer-submenu-{{ forloop.index }}">
```
kmeleta, line 57: I noticed this existing ID was also resulting in duplicate ids, so I've updated the id to use a similar format as the other link elements ids I'm adding this PR. This was also an issue for the grandchild child below, and in the dropdown menu.

PR #2523, opened 2023-04-13
[Image behavior] Fixed background position

LucasLacerdaUX:
Base PR to add `Fixed background position` option for `Image behavior` on sections.

- We're using `clip-path` alongside `position: fixed` to recreate a `background-attachment: fixed` for `img` elements. 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### DIFF
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -158,7 +158,25 @@
   left: auto;
 }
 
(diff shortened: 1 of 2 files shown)


ludoboludo (approved):
Seems good but we will see as we test the approach for different sections.

stufreen (approved):

PR #2532, opened 2023-04-14
[Animation] Add conditional around data-cascade attribute

ludoboludo:
### PR Summary: 

We currently have two approaches to adding the `data-cascade` attribute. One is to apply it at all time even when animations aren't enabled and one is to only add it when necessary. This PR is to use the same approach for all and only add the attribute when necessary. 

To polish what we have and use the same approach everywhere. 

Wrapping the existing code in an if statement. 

None

### Testing steps/scenarios

- [ ] Make sure that all files touched are still animating as expected

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139710103574/editor)

### Checklist


kjellr (approved):
Looks good to me. No noticeable difference while testing. 👍

PR #2540, opened 2023-04-17
Add datepicker for scheduling gift cards

antiBaconMachine:
### PR Summary: 

Adds date picker for scheduling gift card sending for a future date.

Closes a great number of merchant frustrations

We're using a native date picker for now as this was deemed to be the best starting point int terms of utility and accessibility. As the feature matures we may choose to implement something custom which is more themable but this is a good starting position.

Other theme developers may use a different date picker as long as the line itmem property is in ISO format YYYY-MM-DD

We also now use error messages which contain the field as part of the error string which is much better for localisation. Unfortunately it would be a breaking change to introduce this, so we've made it opt in my keying on the new _shopify_offset parameter. It's a bit strange to use this property as the switch but it works.

We want the TZ offset so that we can schedule the gift card send for a sensible time in the user's local timezone rather than in the shop timezone which might cause confusion.

Whilst I was here I also cleaned up a minor bug with empty line item properties being sent even when the checkbox was not checked. I've done this simply by disabling the fields whenever the checkbox is unchecked.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store]( https://shop1.shopify.scheduling-theme-with-validation.ollie-edwards.eu.spin.dev/products/gift-card)
- [Editor](https://shop1.shopify.scheduling-theme-with-validation.ollie-edwards.eu.spin.dev/admin/themes/1/editor)

### Checklist

### DIFF
--- a/locales/es.json
+++ b/locales/es.json
@@ -482,7 +482,9 @@
       "message_label": "Mensaje (opcional)",
       "message": "Mensaje",
       "max_characters": "{{ max_chars }} caracteres máx.",
-      "email_label_optional_for_no_js_behavior": "Correo electrónico de la persona destinataria (opcional)"
+      "email_label_optional_for_no_js_behavior": "Correo electrónico de la persona destinataria (opcional)",
+      "send_on": "AAAA-MM-DD",
+      "send_on_label": "Enviar el (opcional)"
     }
   }
 }
--- a/locales/pt-BR.json
+++ b/locales/pt-BR.json
@@ -482,7 +482,9 @@
       "message_label": "Mensagem (opcional)",
       "message": "Mensagem",
       "max_characters": "Máximo de {{ max_chars }} caracteres",
-      "email_label_optional_for_no_js_behavior": "E-mail do destinatário (opcional)"
+      "email_label_optional_for_no_js_behavior": "E-mail do destinatário (opcional)",
+      "send_on": "AAAA-MM-DD",
+      "send_on_label": "Enviar em (opcional)"
     }
   }
 }
--- a/locales/pt-PT.json
+++ b/locales/pt-PT.json
@@ -482,7 +482,9 @@
       "message_label": "Mensagem (opcional)",
       "message": "Mensagem",
       "max_characters": "Máximo de {{ max_chars }} caracteres",
-      "email_label_optional_for_no_js_behavior": "E-mail do destinatário (opcional)"
+      "email_label_optional_for_no_js_behavior": "E-mail do destinatário (opcional)",
+      "send_on": "AAAA-MM-DD",
+      "send_on_label": "Enviar a (opcional)"
     }
   }
 }
--- a/locales/ro-RO.json
+++ b/locales/ro-RO.json
@@ -482,7 +482,9 @@
       "message_label": "Mesaj (opțional)",
       "message": "Mesaj",
       "max_characters": "Maximum {{ max_chars }} (de) caractere",
-      "email_label_optional_for_no_js_behavior": "Adresă de e-mail destinatar (opțional)"
+      "email_label_optional_for_no_js_behavior": "Adresă de e-mail destinatar (opțional)",
+      "send_on": "DD.MM.YYYY",
+      "send_on_label": "Trimite pe (opțional)"
     }
   }
 }
--- a/locales/sk-SK.json
+++ b/locales/sk-SK.json
@@ -492,7 +492,9 @@
       "message_label": "Správa (voliteľné)",
       "message": "Správa",
       "max_characters": "Maximálny počet znakov: {{ max_chars }}",
-      "email_label_optional_for_no_js_behavior": "E-mail príjemcu (voliteľné)"
+      "email_label_optional_for_no_js_behavior": "E-mail príjemcu (voliteľné)",
+      "send_on": "RRRR-MM-DD",
+      "send_on_label": "Dátum odoslania (voliteľné)"
     }
   }
 }
--- a/locales/sl-SI.json
+++ b/locales/sl-SI.json
@@ -492,7 +492,9 @@
       "message_label": "Sporočilo (izbirno)",
       "message": "Sporočilo",
       "max_characters": "Največje št. znakov: {{ max_chars }}",
-      "email_label_optional_for_no_js_behavior": "E-poštni naslov prejemnika (izbirno)"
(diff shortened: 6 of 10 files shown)


assets/recipient-form.js
```diff
@@ -1,142 +1,164 @@
...
-            inputElement.setAttribute('aria-invalid', true);
-            inputElement.setAttribute('aria-describedby', errorMessageId);
-          });
-        }
-
-        this.errorMessage.innerText = body;
+      } else {
+        this.enableInputFields();
       }
+    }
```
ludoboludo, line 71: More nitpick for readability: 

```javascript
onChange() {
  if (this.checkboxInput.checked) {
    this.enableInputFields();
  } else {
    this.clearInputFields();
    this.disableInputFields();
    this.clearErrorMessage();
  }
}

```

assets/recipient-form.js
```diff
@@ -1,142 +1,164 @@
...
+      this.checkboxInput = this.querySelector(`#Recipient-Checkbox-${ this.dataset.sectionId }`);
+      this.checkboxInput.disabled = false;
+      this.hiddenControlField = this.querySelector(`#Recipient-Control-${ this.dataset.sectionId }`);
+      this.hiddenControlField.disabled = true;
+      this.emailInput = this.querySelector(`#Recipient-email-${ this.dataset.sectionId }`);
+      this.nameInput = this.querySelector(`#Recipient-name-${ this.dataset.sectionId }`);
+      this.messageInput = this.querySelector(`#Recipient-message-${ this.dataset.sectionId }`);
+      this.sendonInput = this.querySelector(`#Recipient-send_on-${ this.dataset.sectionId }`);
+      this.offsetProperty = this.querySelector(`#Recipient-Offset-${ this.dataset.sectionId }`);
+      if (this.offsetProperty) this.offsetProperty.value = new Date().getTimezoneOffset();
```
ludoboludo, line 14: Out of curiosity, is there any back end validation/erroring if the date entered isn't in the future?

Say I make a mistake and enter `2022` as the year.

assets/recipient-form.js
```diff
@@ -1,142 +1,164 @@
...
-          const textField = field.querySelector('.error-message');
-          if (textField) textField.innerText = '';
-        });
+    enableInputFields() {
+      this.disableableFields().forEach((field) => field.disabled = false);
+    }
+
+    disableInputFields() {
+      this.disableableFields().forEach((field) => field.disabled = true);
+    }
```
ludoboludo, line 96: Could this be something to add to `this.clearInputFields()` instead of its own function?

I see that `this.clearInputFields()` is run twice and in one instance, `this.disableInputFields()` isn't run, in the `resetRecipientForm()`. Would it make sense to also run it there ?

snippets/gift-card-recipient-form.liquid
```diff
@@ -106,7 +105,6 @@
           {% render 'icon-error' %}
           <span class="error-message">
             {%- if form.errors contains 'email' -%}
-              {{ form.errors.translated_fields.email | capitalize }}
```
ludoboludo, line 109: When there is an error, the messaging is pretty vague.That seem to be because I'm testing locally. 
Shouldn't every store behave the same?
Here is what I'm getting:

ludoboludo (commented):
A few comments, mostly nipticks. A few questions as well to make sure I understand the context 🙂 

Nothing blocking. 👍

snippets/gift-card-recipient-form.liquid
```diff
@@ -181,18 +178,60 @@
...
       type="hidden"
       name="{{ gift_card_recipient_control_flag }}"
       value="if_present"
       id="Recipient-Control-{{ section.id }}"
     >
+    <input
+      type="hidden"
+      name="properties[__shopify_offset]"
+      value=""
+      id="Recipient-Offset-{{ section.id }}"
```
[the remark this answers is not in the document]
antiBaconMachine: Yep it's as you say, the difference between UTC and local time. It exists because UX wanted the user's date input to be interpreted from their perspective rather than from the shops's perspective. It means the gift card is sent on the morning of the day selected in the user's timezone. If we didn't do it then we could only guarantee +- 1 day. In terms of naming, yes it's still changeable. It's an internal detail that users of flagship themes should basically never be aware of, but it could make a difference to other theme developers.

assets/recipient-form.js
antiBaconMachine: Yeah, we only allow from today until 90 days currently, though that is an arbitrary choice.

assets/recipient-form.js
antiBaconMachine: It's different because we never want to clear the offset field but we do want to disable it. It's because offset is not user input but it does sit along side user input and use the same line item property mechanism, so there are two similar use cases but not the same.

snippets/gift-card-recipient-form.liquid
antiBaconMachine: So this is an temporary edge case that you should only see if javascript is disabled. That's clearly not the case in this screenshot so very interested in a replication for how you got this.

snippets/gift-card-recipient-form.liquid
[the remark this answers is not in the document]
ludoboludo: Great sounds good. I think we could match the name of the JS method: `getTimezoneOffset`. So maybe `Recipient-timezone-offset-{{ section.id }}`

assets/recipient-form.js
ludoboludo: Sounds good 👍

snippets/gift-card-recipient-form.liquid
antiBaconMachine: Oh that's fine, the backend that supports the new error messages isn't merged yet. It should look better in the spin links above.

For some more context we originally just copied what the customer form does which is concatenate error messages with their field name in liquid like `{{ name }}{{ message }}` which we knew was bad but we were pressed for time so we didn't mess with the precedent. Now we have some time to revaluate we think it's pretty unacceptable to do this so we're fixing the messages so they are localised properly. That would be a breaking change if not handled properly so we're using the presence of the offset field as a hint the theme can cope with the new better error strings.

ludoboludo (approved):
I requested translations 👍 so we will have to wait on them before being able to merge

Issue #2539, opened 2023-04-17
[Animation] performance issue


ludoboludo:
We're noticing some performance issues due to the animation running. 
![](https://screenshot.click/17-47-3zags-4t7k4.png)

Let's explore what could be done to mitigate it

PR #2543, opened 2023-04-18
Performance tweaks to our animation.js approach to trigger animations

ludoboludo:
### PR Summary: 

Explore more performant ways to trigger our theme animations

Fixes https://github.com/Shopify/dawn/issues/2539 & https://github.com/Shopify/dawn-private/issues/179

- I added `scroll-trigger--active` on all elements that have the `scroll-trigger` class. This way on page load it runs the animation by default. Then in the JS we remove the active class only for elements that aren't in the viewport on page load. This way hidden sections/elements will get the animation using the intersection observer once coming into the viewport
- I added the `--animation-order: {{ forloop.index }};` to a few elements that are using the `data-cascade` attribute to prevent a flashing animation. 
- I moved the `animation.js` code into a `<script>` tag instead of calling the file. 

### Decision log

|---|---|---|---|---|
| 1 | Apply animation/transition on page load and use JS to remove where unnecessary  | Apply only the animation when needed  | The existing code adds the animation when needed but that cause an issue in the time it takes to render everything. So it impacts our performance score.  |   |

Shouldn't have any visual impact. 

### Testing steps/scenarios

- [ ] Test all sections that currently have the animation added to them
- [ ] Test performance and make sure it passes the CI check

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139797397526/editor)

### Checklist


layout/theme.liquid
```diff
@@ -33,7 +33,7 @@
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
     {%- if settings.animations_reveal_on_scroll -%}
-      <script src="{{ 'animations.js' | asset_url }}" id="AnimationsScript" defer="defer"></script>
+      <script src="{{ 'animations.js' | asset_url }}" defer="defer"></script>
```
kmeleta, line 36: I was going to ask if you found significant improvements adding the script via snippet over this approach. Not sure I could tell.

layout/theme.liquid
```diff
@@ -33,7 +33,7 @@
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
     {%- if settings.animations_reveal_on_scroll -%}
-      <script src="{{ 'animations.js' | asset_url }}" id="AnimationsScript" defer="defer"></script>
```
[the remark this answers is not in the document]
kmeleta: I was wondering too. Didn't find any references to it or anything.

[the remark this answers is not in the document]
metamoni: My bad, I forgot to remove that from an old PR

layout/theme.liquid
ludoboludo: Actually not too sure, here is what the CI has given me: 
- `animation.js` approach: 77, 82, 81
- inline in `<script>`: 90, 78, 82, 89

But when testing in webpagetest.org the results are pretty similar. 🤷

assets/animations.js
```diff
@@ -1,14 +1,18 @@
 const SCROLL_ANIMATION_TRIGGER_CLASSNAME = 'scroll-trigger';
-const SCROLL_ANIMATION_ACTIVE_CLASSNAME = 'scroll-trigger--active';
+const SCROLL_ANIMATION_OFFSCREEN = 'scroll-trigger--offscreen';
```
stufreen, line 2: Nit: Change to SCROLL_ANIMATION_OFFSCREEN_CLASSNAME to match the other constant

sections/collage.liquid
```diff
@@ -30,6 +30,7 @@
           {{ block.shopify_attributes }}
           {% if settings.animations_reveal_on_scroll %}
             data-cascade
+            style="--animation-order: {{ forloop.index }};"
```
stufreen, line 33: Adding these seems to add a lot of inline styles for a very limited effect: if a cascading section is rendered above the fold the cascading will still work before the JS takes over and handles it with the IntersectionObserver. This might happen on collection template for instance.

Personally I am OK to remove these inline styles and have those elements just fade in at the same time if they're above the fold. From a UI perspective the cascading is mostly nice when you are scrolling the page, not on initial page load. From a code perspective I feel like they add more complexity relative to the benefit.

sections/video.liquid
```diff
@@ -64,21 +64,23 @@
...
         {%- else -%}
-          {{ section.settings.video
-            | video_tag:
-              image_size: "1100x",
-              loop: section.settings.enable_video_looping,
-              controls: true,
-              muted: false
+          {{
+            section.settings.video
+            | video_tag: image_size: '1100x', loop: section.settings.enable_video_looping, controls: true, muted: false
```
stufreen, line 83: Non blocking: I guess these are just prettier formatting? I suppose we may see a few of these now that formatOnSave is enabled by default.

stufreen (changes requested):

sections/video.liquid
ludoboludo: Exactly 👍

sections/collage.liquid
ludoboludo: I don't have a strong feeling about this. On this one we should confirm with a designer if that works for them or if we keep the `animation order` 👍  

Here is the difference shown [in video](https://screenshot.click/21-15-zcxdp-avt3l.mp4). @melissaperreault what do you think ? 🤔

melissaperreault: Hey! I have a strong preference to keep it all the same as it felt like 2 different animations when we removed the cascade on page load.

stufreen: OK - I do have concerns about the maintainability of doing cascading two ways (one for no-JS/above-the-fold, one with JS) but if we feel strongly the design is important then this seems like the way to do it.

stufreen (approved):
LGTM 🎩  🏎️ 

Can you update the PR to have a more descriptive name?

ludoboludo:
> Also, why is it merging into `animate-featured-product`? I think it should merge into `main` and we'll rebase the other PR.

It was done this way to compare the CI scores 🤷 Ill change the base branch back to main 👍

sections/collage.liquid
melissaperreault: I'd love to have other UX point of view though and align to find the perfect trade off! The grid in its form did gave that impression of difference that I found less aligned or consistent with the flow as you discover the next row items.

LucasLacerdaUX (approved):
Looks good to me :D

metamoni (approved):

sections/collage.liquid
kjellr: I agree with Meli here — I think it's important to keep the cascade everywhere. 👍

PR #2547, opened 2023-04-19
[Collage] Add ambient effect on image block

ludoboludo:
### PR Summary: 

Add ambient movement to the image block in the Collage section

Fixes https://github.com/Shopify/dawn/issues/2442

Add the new setting option for the block reusing the same approach as other sections that already have it. 
Added the logic for the image sizes. Though I used a different approach just for the sake of exploration to see if we could have something a bit easier to update in the future if we decided to use a different multiplier for the size of the images (`1.2` which isn't documented anywhere I believe) or if we wanted to change the default sizes we use. 

Adds a new animation option for image block

### Testing steps/scenarios

- [ ] Go to the collage section and add an image block
- [ ] Make sure the image size is the right one
- [ ] Make sure that the movement applied on the image never uses too small of an image, revealing some whitespace. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139809652758/editor)

### Checklist

### DIFF
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -36,7 +36,7 @@
             {%- when 'image' -%}
               <div class="collage-card {% if section.settings.card_styles == 'none' %}global-media-settings{% else %}card-wrapper {{ section.settings.card_styles }} color-{{ settings.card_color_scheme }}{% endif %}">
                 <div
-                  class="media media--transparent ratio"
+                  class="media media--transparent ratio{% if block.settings.image_behavior != 'none' %} animate--{{ block.settings.image_behavior }}{% endif %}"
                   {% if block.settings.image != blank %}
                     style="--ratio-percent: {{ 1 | divided_by: block.settings.image.aspect_ratio | times: 100 }}%"
                   {% else %}
@@ -45,6 +45,23 @@
                 >
                   {%- if block.settings.image != blank -%}
                     {%- liquid
+                      assign default_widths_array = '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1250, 1500, 1750, 2000, 2250, 2500, 2750, 3000, 3200' | split: ','
+                      assign widths_array_final = '' | split: ''
+
+                      if block.settings.image_behavior == 'ambient'
+                        assign max_width = '3840'
+                        assign size_multiplier = 1.2
+                        for width in default_widths_array
+                          assign width_final = width | times: size_multiplier | split: ' '
+                          assign widths_array_final = widths_array_final | concat: width_final
+                        endfor
+                        assign widths = widths_array_final | join: ', '
+                      else
+                        assign max_width = '3200'
+                        assign size_multiplier = 1
+                        assign widths = default_widths_array | join: ', '
+                      endif
+
                       if section.settings.desktop_layout == 'left'
                         assign large_block = forloop.first
                       else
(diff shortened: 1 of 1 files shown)


sections/collage.liquid
```diff
@@ -45,6 +45,23 @@
...
+                      if block.settings.image_behavior == 'ambient'
+                        assign max_width = '3840'
+                        assign size_multiplier = 1.2
+                        for width in default_widths_array
+                          assign width_final = width | times: size_multiplier | split: ' '
+                          assign widths_array_final = widths_array_final | concat: width_final
+                        endfor
+                        assign widths = widths_array_final | join: ', '
+                      else
+                        assign max_width = '3200'
```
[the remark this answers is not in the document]
ludoboludo: Oh yeah that'd make sense 👍

sections/image-banner.liquid
```diff
@@ -33,11 +33,13 @@
   if section.settings.image_behavior == 'ambient'
     assign half_width = '60vw'
     assign full_width = '120vw'
+    assign max_width = 4608
     assign stacked_sizes = '(min-width: 750px) 60vw, 120vw'
-    assign widths = '450, 660, 900, 1320, 1800, 2136, 2400, 3600, 7680'
+    assign widths = '450, 660, 900, 1320, 1800, 2136, 2400, 3600, 4608'
```
ludoboludo, line 38: Let me know here if there was a specific reason for this 🤔 Because the section is full width I could see how very wide screen could be an issue but the gap between `3600` and `7680` seemed odd. 
cc: @kjellr

sections/slideshow.liquid
```diff
@@ -133,20 +133,18 @@
               assign height = block.settings.image.width | divided_by: block.settings.image.aspect_ratio | round
               if section.settings.image_behavior == 'ambient'
                 assign sizes = '120vw'
-                assign widths = '450, 660, 900, 1320, 1800, 2136, 2400, 3600, 7680'
+                assign widths = '450, 660, 900, 1320, 1800, 2136, 2400, 3600, 4608'
```
ludoboludo, line 136: similar thing here. Let me know if the `7680` was meant for a specific use case.

sections/image-banner.liquid
kmeleta: fwiw even if 7680 (or anything remotely near that) was called for by the viewport, I'm not sure we would want to allow the requesting of such a large asset.

kjellr: Yeah, I think we're probably okay losing that value. 👍

sections/collage.liquid
```diff
@@ -45,6 +45,23 @@
                 >
                   {%- if block.settings.image != blank -%}
                     {%- liquid
+                      assign default_widths_array = '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1250, 1500, 1750, 2000, 2250, 2500, 2750, 3000, 3200' | split: ','
```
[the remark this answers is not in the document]
ludoboludo: I went back to using the original approach used in other places

sections/collage.liquid
```diff
@@ -56,31 +64,25 @@
...
-                          widths: '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1250, 1500, 1750, 2000, 2250, 2500, 2750, 3000, 3200'
+                        | image_url: width: 3840
+                        | image_tag: loading: 'lazy', sizes: sizes, widths: widths
                       }}
                     {%- else -%}
                       {%- capture sizes -%}
-                           (min-width: {{ settings.page_width }}px) calc(({{ settings.page_width }}px - 100px) * 1 / 3 - {{ grid_space_desktop }}),
-                           (min-width: 750px) calc((100vw - 100px) * 1 / 3 - {{ grid_space_desktop }}),
-                           {% if section.settings.mobile_layout == 'collage' %}calc((100vw - 30px) / 2 - {{ grid_space_mobile }}){% else %}calc(100vw - 30px){% endif %}
+                           (min-width: {{ settings.page_width }}px) calc(({{ settings.page_width | times: size_multiplier }}px - 100px) * 1 / 3 - {{ grid_space_desktop }}),
```
kmeleta, line 78: Getting some wonky sizes loading with these changes on collage. Haven't seen the issue in the other sections yet.

ludoboludo: Tho if you look at the network tab in the inspect tool, it's loading the expected image size right ? 

We're basing our `sizes` attribute values on what we already had by multiplied by `1.2`. So the images we're getting should be a bit bigger than their containers. 

There is also a PR from folks in theme support to fix an issue with the image logic in that section: https://github.com/Shopify/dawn/pull/2277

sections/collage.liquid
```diff
@@ -46,6 +46,14 @@
                 >
                   {%- if block.settings.image != blank -%}
                     {%- liquid
+                      assign widths = '50, 75, 100, 150, 200, 300, 400, 500, 750, 1000, 1250, 1500, 1750, 2000, 2250, 2500, 2750, 3000, 3200'
+                      assign size_multiplier = 1
+
+                      if block.settings.image_behavior == 'ambient'
+                        assign widths = '60, 90, 120, 180, 240, 360, 480, 600, 900, 1200, 1500, 1800, 2100, 2400, 2700, 3000, 3300, 3600, 3840'
+                        assign size_multiplier = 1.2
```
kmeleta, line 54: Another general change idea, lmk what you think. Might be nice to also somehow link these multipliers to the value in the css. Or at least document that the 1.2 corresponds to both values. Maybe a comment in the css would suffice. I could certainly see changing that css value to tweak the motion and not remembering to update the multiplier in the sections.

kmeleta (approved):
Didn't notice any issues with the motion itself in collage. Generally happy with where we landed on the approach in the logic. Just left a couple small things to consider.

sections/collage.liquid
ludoboludo: Very good point 👍

sections/collage.liquid
```diff
@@ -163,21 +165,33 @@
...
+                              calc({{ page_width }}px - 100px)
                             {%- else -%}
-                              {{- settings.page_width | minus: 100 | times: 0.67 | round }}px
+                              {{- page_width | minus: 100 | times: 0.67 | round }}px
                             {%- endif -%}
                             , (min-width: 750px)
-                            {%- if section.blocks.size == 1 %} calc(100vw - 100px){% else %} 500px{% endif -%}
-                            , calc(100vw - 30px)
+                            {%- if section.blocks.size == 1 %} calc({{ 100 | times: size_multiplier }}vw - 100px){% else %} {{ 500 | times: size_multiplier }}px{% endif -%}
+                            , calc({{ 100 | times: size_multiplier }}vw - 30px)
```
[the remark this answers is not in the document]
ludoboludo: What I've done seem to give us numbers a bit higher than if I went with the approach you mention. So I've made changes so it's more accurate and close to the size of the image we're using. 
👍

kmeleta:
Changes look good to me. Thanks Ludo.

kjellr:
Just raising a discussion point here from slack: 

We need to make a UX decision around whether or not this setting should be available on a section- or block-level for collage. In all other occurrences, this is a section-level setting, even when there are interior blocks (like Slideshow). 

cc @melissaperreault

kjellr:
I'm removing myself from this for now to stop GitHub from reminding me everyday about it. 😅 But when we pick it back up, feel free to add me back in!

PR #2559, opened 2023-04-24
[Image with text] Add slide-in animation to the text content when set to overlap

### LINKED ISSUE #2375: [Image with text] Reveal in cascade for overlap content layout
### Context
Recap of what I propose
- No overlap content layout: Reveal image and content container background together, than the content blocks
- Overlap content layout: Reveal each side individually

>@melissaperreault, @kjellr this sounds like a great idea and I don't image it would be difficult to do. However, we don't have the "cascading" CSS in this PR, and adding that would change the scope a bit. The CSS that will handle delayed animation will be part of [this PR](https://github.com/Shopify/dawn/pull/2372) (draft) and will need to be reviewed and tested before we can introduce it into other sections.

_Originally posted by @metamoni in https://github.com/Shopify/dawn/issues/2362#issuecomment-1461494881_

ludoboludo:
### PR Summary: 

Add the slide in animation to the text box when using the `overlap` content layout. 

Fixes #2375 

I didn't use the `data-cascade` approach. I just added the class for the slide in animation. Doing it this way makes sure we're sliding in that container when it comes into view. 
It can be vertically aligned at the top, middle or bottom of the image.

Depending on the image format and the text content's length, we can have different situation that should be tested while reviewing the PR: 
- landscape image
- portrait image
- long text content that is bigger than the image
- text content smaller than the image

A delay in the animation for the text box when set to overlap. 

### Testing steps/scenarios

- landscape image
- portrait image
- long text content that is bigger than the image
- text content smaller than the image
- test mobile as well

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139832066070/editor)

### Checklist


melissaperreault (approved):
Thank you! Works as expected! 🚢

sections/image-with-text.liquid
```diff
@@ -50,7 +50,7 @@
     <div class="image-with-text__text-item grid__item">
       <div
         id="ImageWithText--{{ section.id }}"
-        class="image-with-text__content image-with-text__content--{{ section.settings.desktop_content_position }} image-with-text__content--desktop-{{ section.settings.desktop_content_alignment }} image-with-text__content--mobile-{{ section.settings.mobile_content_alignment }} image-with-text__content--{{ section.settings.height }} gradient color-{{ section.settings.color_scheme }} content-container"
+        class="image-with-text__content image-with-text__content--{{ section.settings.desktop_content_position }} image-with-text__content--desktop-{{ section.settings.desktop_content_alignment }} image-with-text__content--mobile-{{ section.settings.mobile_content_alignment }} image-with-text__content--{{ section.settings.height }} gradient color-{{ section.settings.color_scheme }} content-container{% if settings.animations_reveal_on_scroll and section.settings.content_layout == 'overlap' %} scroll-trigger animate--slide-in{% endif %}"
```
stufreen, line 53: This looks ok but it's not exactly the same effect. What it's doing here:
1. The parent `image-with-text` div is sliding in when it comes into viewport
2. The `image-with-text__content` div is sliding in when it comes into the viewport. This is usually after the parent because it's further down the page.

If you want to be more precise about the animations:
1. Move the `scroll-trigger` and `animate--slide-in` class to `image-with-text__media-item` `and `image-with-text__text-item`
2. Add `data-cascade` to each div if `section.settings.content_layout == 'overlap'`

stufreen (changes requested):

sections/image-with-text.liquid
metamoni, line 53: Is there an element below this wrapper that we can place the animation on? Right now, when you use a background colour, the colour fades in as well and it looks a bit off. In previous PRs, @kjellr asked to make sure that the background for different elements doesn't fade in, but the individual elements do.

**Original**

https://user-images.githubusercontent.com/22390758/234898759-3c75e22f-965f-4d0e-877b-ba66cf7a08cf.mp4

**Slow motion**


https://user-images.githubusercontent.com/22390758/234898873-cd9d5bba-de8e-4de6-9aba-ff7163962203.mp4

kjellr: It doesn't bother me too much in this case — the rule about the background color seems less important to me when it's only taking up half of the screen like this.

melissaperreault: Thank you for putting words on what originally bothered me, but I assume it was expected because of the transitions. I am curious to see how this could experience without fade!

ludoboludo: Here is how it looks, ill push the change so you can test. 

There was something else I needed to fix in the logic. Make sure to test when there isn't any overlap too :) 

[Video example](https://screenshot.click/27-41-92inc-twnjv.mp4)

kjellr: Yeah, I think the fade is preferable. Let's use that!

ludoboludo:
Alrighty, I will close this PR for now then and we can reopen in the future if needed 👍

PR #2568, opened 2023-04-25
Fix fade in animation and CSS specificity for sliders

ludoboludo:
### PR Summary: 

Fixing the fade in animation as it was messed with when we merged the performance PR

The fade in animation wasn't applied as it was getting the `opacity: 1` applied right away so no animation seemed to happen. 

I used a CSS keyframe animation like for the `slide-in` effect. This way we're animating the opacity change and using similar values as for slide-in

I think we need to align on whether or not it should be using the same values here: 
```css
--animation-fade-in: fadeIn var(--duration-extra-long) cubic-bezier(0, 0, 0.3, 1) forwards;
```

### Other considerations

I think we need to align on whether or not it should be using the same values here: 
```css
--animation-fade-in: fadeIn var(--duration-extra-long) cubic-bezier(0, 0, 0.3, 1) forwards;
```

### Testing steps/scenarios

- [ ] Test the animation on elements using the fade in one. Image banner, templates heading. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139840880662/editor)

### Checklist


assets/base.css
```diff
@@ -747,6 +747,7 @@ details > * {
   --duration-long: 500ms;
   --duration-extra-long: 600ms;
   --animation-slide-in: slideIn var(--duration-extra-long) cubic-bezier(0, 0, 0.3, 1) forwards;
+  --animation-fade-in: fadeIn var(--duration-default) cubic-bezier(0, 0, 0.3, 1) forwards;
```
kjellr, line 750:
```suggestion
  --animation-fade-in: fadeIn var(--duration-extra-long) cubic-bezier(0, 0, 0.3, 1) forwards;
```
I'd like to keep the longer duration here, so the speed matches the fade in speed when they're happening together. With `--duration-default`, I find it's harder to notice the fade-in at all.

assets/base.css
```diff
@@ -3279,12 +3280,11 @@ details-disclosure > details {
 
   .scroll-trigger.animate--fade-in,
   .scroll-trigger.animate--slide-in {
-    transition: opacity var(--duration-extra-long) ease-out;
-    opacity: 0;
+    opacity: 0.1;
```
kjellr, line 3283: It doesn't make much of a difference in practice, but I'm curious why we're using `0.1` instead of `0` here?

kjellr (approved):
Approved with just a couple small comments. 👍

assets/base.css
ludoboludo: It was something to test that Stu found out. It has an impact on performance scores. 

There is some details mentioned in here: https://www.debugbear.com/blog/opacity-animation-poor-lcp

kjellr: Oh, interesting! Sounds good to me. 👍

ludoboludo: Thanks for saving the day @kjellr 

Looks like having the opacity set to `1` on the element and applying the animation without the `forwards` fill-mode solved the issue 🤞 🙌

kjellr (approved):
LGTM with the latest commit. I'd have really expected a flicker since we're setting the opacity to `1` and then animating it from `0.1` to `1`, but I can't find one, even when I slow things down: 


https://user-images.githubusercontent.com/1202812/234668487-da8d547e-b44d-4586-8ab9-1e04ee630fa6.mp4

https://user-images.githubusercontent.com/1202812/234668638-622088dd-2375-4b23-8f8d-dd09cdf63d83.mp4

assets/base.css
```diff
@@ -3279,12 +3280,12 @@ details-disclosure > details {
...
```
LucasLacerdaUX, line 3283: I'm seeing some stuttering on the demo link, specifically on `slide-in` animations.

Do you think this is related to this PR?

https://user-images.githubusercontent.com/37168033/234676244-3f64f652-df77-4645-9c29-1c634d387b23.mp4

LucasLacerdaUX (commented):
One small issue

assets/base.css
ludoboludo: Good catch 👍 pushed the fix.

assets/base.css
kjellr: I'm curious: Does the `0.1` opacity performance fix still work if the opacity is set to `0.01` instead? 

Something about the animations with 10% opacity just feels less smooth — I think it's because you can still see dark items really clearly _before_ they animate in. Here's an example in slow-motion. Note that the first frame you see is of the dark box on the left: 

https://user-images.githubusercontent.com/1202812/234870650-cecb2c59-962d-47c5-8ccc-93e5d5fa18e9.mp4

That kinda interrupts the impression of sequential left-to-right loading. If we're able to set the opacity to 1% instead of 10%, then the effect is more as intended: 

https://user-images.githubusercontent.com/1202812/234870915-3b37c904-61f9-4ab7-a3b1-5c9a5e725a8d.mp4

stufreen:
Changes LGTM too

LucasLacerdaUX (approved):
Looks and works as expected now!

PR #2569, opened 2023-04-25
Refactor color schemes implementation

kmeleta:
### PR Summary: 

Adds improved color scheme support.

Fixes https://github.com/Shopify/dawn-private/issues/146

Refactors the color schemes implementation in Dawn to utilize new schemas and editor UI, keeping theme behaviors mostly the same as before.

- Defined a scheme definition in `settings_schema.json` and roles via [`color_scheme_group`](https://shopify.dev/docs/themes/architecture/settings/input-settings#color_scheme_group) which is identical across all schemes. 
- Defined color scheme defaults in `settings_data.json` to match original Dawn colors.
- Replaced color scheme `select` type settings with new [`color_scheme`](https://shopify.dev/docs/themes/architecture/settings/input-settings#color_scheme) settings in all sections, blocks, and theme settings.
- Refactored CSS color variable mapping in `theme.liquid` using color schemes liquid drop.
- Refactored scheme specific classname and legacy variable references (i.e `.background-1` and `--color-base-text`) which no longer have the same meaning or purpose.
- Added a color scheme setting for shadows instead of inheriting from `--color-base-text`
- Removed `Icons` theme settings, translation strings, and `color-icon-` classes. Decorative icons now inherit the in-scope scheme's text color. (Technically visually breaking)

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


locales/en.default.schema.json
```diff
@@ -114,49 +114,29 @@
...
         },
-        "colors_background_2": {
-          "label": "Background 2"
+        "shadow": {
+          "label": "Shadow"
         },
-        "gradient_background_2": {
-          "label": "Background 2 gradient"
-        }
+        "info": "Scheme 1 is used as a fallback for sections and templates without a color scheme setting."
```
kmeleta, line 139: @melissaperreault could you verify all the language changes in here are good to go?

melissaperreault, line 139:
```suggestion
        "info": "Scheme 1 is used for areas without a color scheme setting."
```

@silviacela @katycobb Can you please help remind me if this was the last version we aligned on? 🙏

stufreen:
@kmeleta @melissaperreault I am curious what the process was for picking new colour schemes for each setting. Did you have to go through each section?

assets/component-rating.css
```diff
@@ -14,6 +14,7 @@
 }
 
 .rating-star {
+  --color-rating-star: rgb(var(--color-foreground));
```
kmeleta, line 17: I could see the loss of accent color for rating stars being one of the more noticeable cases. So I've at least captured the color in a variable up at the .rating-star scope so would be easy to change via custom css. Even though the color gets used in a gradient style within a pseudo element.

`.rating-star { --color-rating-star: #000; }`

kmeleta, line 17: I could see the loss of accent color for rating stars being one of the more noticeable cases. So I've at least captured the color in a variable up at the .rating-star scope so would be easy to change via custom css. Even though the color gets used in a gradient style within a pseudo element.

`.rating-star { --color-rating-star: #000; }`

sections/collage.liquid
```diff
@@ -186,7 +186,10 @@
                     </button>
                   </div>
                 </modal-opener>
-                <modal-dialog id="PopupModal-{{ block.id }}" class="modal-video media-modal color-background-1">
+                <modal-dialog
+                  id="PopupModal-{{ block.id }}"
+                  class="modal-video media-modal color-{{ settings.color_schemes | first }}"
```
kmeleta, line 191: This is an example of where I'm dynamically referencing the "first" (and fallback) scheme rather than "background-1" specifically.

layout/theme.liquid
```diff
@@ -51,6 +51,46 @@
...
+            --color-scheme-background-gradient: {{ scheme.settings.background }};
+          {% endif %}
+          --color-scheme-text: {{ scheme.settings.text.red }},{{ scheme.settings.text.green }},{{ scheme.settings.text.blue }};
+          --color-scheme-button: {{ scheme.settings.button.red }},{{ scheme.settings.button.green }},{{ scheme.settings.button.blue }};
+          --color-scheme-button-label: {{ scheme.settings.button_label.red }},{{ scheme.settings.button_label.green }},{{ scheme.settings.button_label.blue }};
+          --color-scheme-secondary-button-label: {{ scheme.settings.secondary_button_label.red }},{{ scheme.settings.secondary_button_label.green }},{{ scheme.settings.secondary_button_label.blue }};
+          --color-scheme-shadow: {{ scheme.settings.shadow.red }},{{ scheme.settings.shadow.green }},{{ scheme.settings.shadow.blue }};
+        }
+      {% endfor %}
+
```
kmeleta, line 72: The color variables/styles below are ones that were previously (and ideally) in base.css. But with the possibility of added themes with generated scheme ids, we need the color scheme drop.

assets/template-giftcard.css
```diff
@@ -107,25 +107,25 @@ h2,
...
 .button:hover {
-  box-shadow: 0 0 0 0.2rem rgb(var(--color-base-accent-1));
+  box-shadow: 0 0 0 0.2rem rgb(var(--color-button));
 }
 
 .button--secondary {
-  color: rgb(var(--color-base-outline-button-labels));
-  box-shadow: 0 0 0 0.1rem rgb(var(--color-base-outline-button-labels));
+  color: rgb(var(--color-secondary-button-label));
+  box-shadow: 0 0 0 0.1rem rgb(var(--color-secondary-button-label));
```
stufreen, line 119: Looks good - box-shadow is being used as an outline here, not a shadow, so it's ok to use the label color.

templates/gift_card.liquid
```diff
@@ -52,16 +52,23 @@
...
+        --color-button: {{ scheme.settings.button.red }},{{ scheme.settings.button.green }},{{ scheme.settings.button.blue }};
+        --color-button-label: {{ scheme.settings.button_label.red }},{{ scheme.settings.button_label.green }},{{ scheme.settings.button_label.blue }};
+        --color-secondary-button-label: {{ scheme.settings.secondary_button_label.red }},{{ scheme.settings.secondary_button_label.green }},{{ scheme.settings.secondary_button_label.blue }};
+        --color-shadow: {{ scheme.settings.shadow.red }},{{ scheme.settings.shadow.green }},{{ scheme.settings.shadow.blue }};
+
+        {% assign badge_scheme_id = settings.sold_out_badge_color_scheme | append: '' %}
+        {% assign badge_scheme = settings.color_schemes[badge_scheme_id] %}
+        --color-soldout-badge: {{ badge_scheme.settings.background.red }},{{ badge_scheme.settings.background.green }},{{ badge_scheme.settings.background.blue }};
+        --color-soldout-badge-label: {{ badge_scheme.settings.text.red }},{{ badge_scheme.settings.text.green }},{{ badge_scheme.settings.text.blue }};
```
stufreen, line 71: Can you explain what's going on with the sold out badge colors? Is it a secondary color_scheme setting?

sections/collage.liquid
stufreen: Do you need this `color` class at all now? Won't the variables for the first scheme be defined in `:root`?

sections/announcement-bar.liquid
```diff
@@ -99,32 +99,10 @@
...
-        },
-        {
-          "value": "inverse",
-          "label": "t:sections.all.colors.inverse.label"
-        }
-      ],
-      "default": "accent-1",
-      "label": "t:sections.all.colors.label"
+      "label": "t:sections.all.colors.label",
+      "default": "accent-1"
```
stufreen, line 105: NB: It's interesting that we set a `default` value for this setting, since we have no way of knowing if that scheme still exists in `settings_data.json`. I guess it makes sense for fresh installs (because the theme dev controls the default color schemes) but not always for adding new sections.

sections/main-password-footer.liquid
```diff
@@ -1,4 +1,5 @@
-{%- if section.settings.color_scheme == 'background-1' -%}<hr>{%- endif -%}
+{% assign scheme1 = settings.color_schemes | first %}
+{%- if section.settings.color_scheme == scheme1 -%}<hr>{%- endif -%}
```
stufreen, line 2: Why do we only show `hr` if it's `background-1`? Does that still work with the new schemes?

snippets/card-product.liquid
```diff
@@ -319,7 +319,7 @@
           {%- if card_product.available == false -%}
             <span
               id="Badge-{{ section_id }}-{{ card_product.id }}"
-              class="badge badge--bottom-left color-{{ settings.sold_out_badge_color_scheme }}"
+              class="badge badge--bottom-left color-{{ settings.sold_out_badge_color_scheme }}t"
```
stufreen, line 322: Accident?

snippets/email-signup-banner-background.liquid
```diff
@@ -12,5 +12,5 @@
...
...52.3l-.1 279.3v.7h483.1v-.2c.1-8.5 0-280.1 0-280.1z"/><path class="st0" d="M1548.1 623.4c12.4-133.5 115.6-239.8 243.5-252.6h-1.3c-127.9 12.7-229.7 119-242.2 252.6zM2068.7-.1V371h107.2V-.1z"/><path class="st4" d="M135.9-.1C59.6-.1-2.3 61.8-2.3 138.1V649.9l138.2-.4V-.1z"/><g><path class="st0" d="M721.2 134.3c-17.9-14.5-40.6-23.4-65.1-24.4-1.4 0-2.9-.1-4.3-.1-61.1 0-110.5 49.5-110.5 110.5s49.5 110.5 110.5 110.5h2.6c25.2-.6 48.5-9.7 66.8-24.5 25.1-20.2 41.2-51.3 41.2-86s-16.1-65.8-41.2-86z"/><path class="st4" d="M870.4-.1H432.7v217.5c0 24.9 4.2 48.9 12 71.3 13.3 38.8 37.2 72.9 68.1 98.7 33.3 27.8 74.5 45.5 119.9 49.4h21.1s3.7.6 5.5.5c11.1-.2 22.1-1.4 33-3.3 9.8-1.8 19.5-4.3 29-7.4 14-4.7 27.6-10.8 40.4-18.3 2.9-1.7 5.7-3.5 8.6-5.3 17.5-11.4 33.1-25.1 46.7-40.9l.1-.1c.1-.2.4-.4.5-.6 8-9.5 15.2-19.5 21.5-30.3l1.7-2.8c9-15.8 15.9-33.1 20.7-50.6v-62l69-66.6c-37.4-38.5-60.1-91.4-60.1-149.2zM721.2 306.3c-18.3 14.8-41.5 23.9-66.8 24.5h-2.6c-61.1 0-110.5-49.5-110.5-110.5s49.5-110.5 110.5-110.5c1.4 0 2.9 0 4.3.1 24.6 1 47.1 9.9 65.1 24.4 25.1 20.2 41.2 51.3 41.2 86-.1 34.8-16.1 65.8-41.2 86z"/><path class="st0" d="M870.4-.1s0-.1 0 0c0 57.8 22.7 110.8 60 149.2L1084.9-.1s-214.5-.2-214.5 0z"/></g>
```
stufreen, line 15: In the old one the colors are the same (--color-base-accent-1) but in the changed one they are different. Is this intentional?

templates/gift_card.liquid
```diff
@@ -105,9 +112,7 @@
             {{ gift_card.balance | money }}
           {% endif %}
         </h1>
-        {%- if gift_card.enabled == false or gift_card.expired -%}
-          <p class="badge badge--{{ settings.sold_out_badge_color_scheme }}">{{ 'gift_cards.issued.expired' | t }}</p>
-        {%- endif -%}
```
stufreen, line 110: Did you intend to remove the `if` statement here?

stufreen (changes requested):
Looking great! I tophatted and couldn't find anything that looked broken, but I don't have as much context on the different setting combos that might break it.

I have left mostly a bunch of questions for my own understanding, and a few cleanup suggestions.

One big change from the old colour scheme is that Scheme 1 really does a lot of heavy lifting. It's good that we added the `info` text saying this. I guess it will be obvious once the user start's editing.

layout/theme.liquid
```diff
@@ -51,6 +51,46 @@
       {{ body_font_bold_italic | font_face: font_display: 'swap' }}
       {{ settings.type_header_font | font_face: font_display: 'swap' }}
 
+      {% for scheme in settings.color_schemes %}
+        {% if forloop.index == 1 %}
+          :root,
+        {% endif %}
+        .color-{{ scheme.id }} {
+          --color-scheme-background: {{ scheme.settings.background.red }},{{ scheme.settings.background.green }},{{ scheme.settings.background.blue }};
```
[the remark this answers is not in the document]
kmeleta: I added a decision entry for this in the PR description, but this gist is that is the same syntax we're doing in Dawn now so I don't think it's too controversial to keep using it for the time being. I want to use `color.rgb` but it returns its value in a space-sperated format that the legacy css rgba isn't compatible with. So I'd need to update ~300 references of `rgba(0,0,0,0)` to use `rgb(0 0 0 / 0)`. I ultimately decided to not include in this PR because there's enough to look at here already.

assets/component-rating.css
kmeleta: We don't. So quite I'm happy to simplify if we don't think it's valuable. But I'm basically just making it easier to revert the visual change without having to dissect the line below and determine for yourself that only the first instance of color-foreground is the one you need to change to get your color back.

`background: linear-gradient(90deg, var(--color-foreground) var(--percent), rgba(var(--color-foreground), 0.15) var(--percent));`

sections/main-password-footer.liquid
kmeleta: Realized this code was actually for the password footer after I took these screenshots, but the answer is the same

layout/theme.liquid
```diff
@@ -51,6 +51,46 @@
       {{ body_font_bold_italic | font_face: font_display: 'swap' }}
       {{ settings.type_header_font | font_face: font_display: 'swap' }}
 
+      {% for scheme in settings.color_schemes %}
+        {% if forloop.index == 1 %}
+          :root,
```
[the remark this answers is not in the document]
kmeleta: Effectively. This makes "scheme 1" the fallback scheme for areas that don't have a scheme setting or a scheme parent to inherit from, like the site background. Though it’s not as much a “default” as it is a “fallback”. Our sections mostly all default to background-1, which works nicely when the default and the fallback schemes are the same. If our fallback isn't background-1, the site background and the sections (in their initial state) will be different colors.

So.. an alternative is to specifically check for `background-1` rather than index == 1 to match the default scheme we use in most of our sections, but my current thinking is we shouldn't commit to specific scheme ids existing. Figured I'd call that option out though in case we want to challenge that.

layout/theme.liquid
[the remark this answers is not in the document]
kmeleta: > One big change from the old colour scheme is that Scheme 1 really does a lot of heavy lifting

This isn't really changed, we're just being transparent about it now. None of our colors are inherently global anymore (only existing within the context of a particular scheme). So it seemed important to clarify that some areas implicitly use a scheme.

config/settings_schema.json
```diff
@@ -37,78 +37,62 @@
...
+          },
+          "links": "secondary_button_label",
+          "icons": "text",
+          "primary_button": "button",
+          "on_primary_button": "button_label",
+          "primary_button_border": "button",
+          "secondary_button": "background",
+          "on_secondary_button": "secondary_button_label",
+          "secondary_button_border": "secondary_button_label"
+        }
```
ludoboludo, line 95: This is specifically to do the mapping of the color scheme icon in the editor right ?

config/settings_schema.json
```diff
@@ -37,78 +37,62 @@
...
+          {
+            "type": "color",
+            "id": "secondary_button_label",
+            "label": "t:settings_schema.colors.settings.secondary_button_label.label"
+          },
+          {
+            "type": "color",
+            "id": "shadow",
+            "label": "t:settings_schema.colors.settings.shadow.label"
+          }
```
ludoboludo, line 79: Something interesting is that if you add a color scheme, you can't delete/remove it ? 
I added one just to see how it would come up but now I'm stuck with and don't see a [way to remove it](https://screenshot.click/26-05-dnelm-artr4.png).

layout/theme.liquid
```diff
@@ -51,6 +51,46 @@
...
+        --gradient-background: var(--color-scheme-background-gradient);
+        --color-foreground: var(--color-scheme-text);
+        --color-shadow: var(--color-scheme-shadow);
+        --color-button: var(--color-scheme-button);
+        --color-button-text: var(--color-scheme-button-label);
+        --color-link: var(--color-scheme-secondary-button-label);
+        --color-badge-foreground: var(--color-foreground);
+        --color-badge-background: var(--color-background);
+        --color-badge-border: var(--color-foreground);
+        --payment-terms-background-color: rgb(var(--color-scheme-background));
```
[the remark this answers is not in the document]
ludoboludo: Based on your example it seem to make sense and adds a safety layer to tweaking variable values in specific scenarios.

layout/theme.liquid
```diff
@@ -51,6 +51,46 @@
...
+
+        color: rgba(var(--color-foreground), 0.75);
+        background-color: rgb(var(--color-background));
```
ludoboludo, line 91: do we need this in `:root` if it's also applied on `body` below on line `230-231`?

locales/en.default.schema.json
```diff
@@ -114,49 +114,29 @@
     "colors": {
       "name": "Colors",
       "settings": {
-        "colors_solid_button_labels": {
-          "label": "Solid button label",
-          "info": "Used as foreground color on accent colors."
```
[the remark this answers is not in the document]
ludoboludo: yah you'd see an error outputting the string you're using to point to the translation file. So visually we should be able to see it pretty easily if we check each settings in the theme 😅

sections/image-with-text.liquid
```diff
@@ -14,7 +14,9 @@
   }
 {%- endstyle -%}
 
-<div class="image-with-text image-with-text--{{ section.settings.content_layout }} page-width isolate{% if settings.text_boxes_border_thickness > 0 and settings.text_boxes_border_opacity > 0 and settings.media_border_thickness > 0 and settings.media_border_opacity > 0 %} collapse-borders{% endif %}{% unless section.settings.color_scheme == 'background-1' and settings.media_border_thickness > 0 and settings.text_boxes_shadow_opacity == 0 and settings.text_boxes_border_thickness == 0 or settings.text_boxes_border_opacity == 0 %} collapse-corners{% endunless %} section-{{ section.id }}-padding{% if settings.animations_reveal_on_scroll %} scroll-trigger animate--slide-in{% endif %}">
+{% assign scheme1 = settings.color_schemes | first %}
```
ludoboludo, line 17: Here this is an added safety if the schemes were to be re ordered ? 

I don't think we really need this 🤔 if that's the only reason. I don't think we have really ever re ordered our color scheme 🤷 

Though the extra safety isn't a big addition.

sections/main-password-footer.liquid
ludoboludo: It was just because the color difference between schemes didn't require the line/border separation I believe. 
Seem to work as I'd expect but something a designer could confirm.

sections/main-search.liquid
```diff
@@ -256,7 +256,7 @@
                                 </h3>
                               </div>
                               <div class="card__badge {{ settings.badge_position }}">
-                                <span class="badge color-background-1">{{ 'templates.search.page' | t }}</span>
+                                <span class="badge color-{{ settings.color_schemes | first }}">{{ 'templates.search.page' | t }}</span>
```
ludoboludo, line 259: New pattern introduced with this. In a few files already, this would be the type of thing, if we're keeping it that we will need to share with all devs/designers. 

I'm not sure we need the extra caution as I don't feel like it's very likely that we re order our color schemes. 

So I'm torn between the extra safety it adds vs the new pattern.

ludoboludo (commented):
We should test how this is going to work with manual theme upgrades 🙂 
Maybe that was tested by another team already ?

Reviewed the files for now and gonna spend some time testing the themes and different sections/templates/etc.

assets/section-main-product.css
```diff
@@ -282,8 +282,8 @@ fieldset.product-form__input .form__label {
 }
 
 .no-js .product-form__submit.button--secondary {
-  --color-button: var(--color-base-accent-1);
-  --color-button-text: var(--color-base-solid-button-labels);
+  --color-button: var(--color-button);
+  --color-button-text: var(--color-button-text);
```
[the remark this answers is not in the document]
kmeleta: Oh good eye. I obviously didn't notice that, but I double checked and that wasn't doing what it should have been anyway. It's trying to make a override things and make a secondary button display as primary for nojs. So I actually need to reset the button colors back to their initial values from the color scheme.

config/settings_data.json
```diff
@@ -3,17 +3,63 @@
...
-      "colors_accent_2": "#334fb4",
-      "gradient_accent_2": "",
-      "colors_text": "#121212",
-      "colors_outline_button_labels": "#121212",
-      "colors_background_1": "#ffffff",
-      "gradient_background_1": "",
-      "colors_background_2": "#f3f3f3",
-      "gradient_background_2": "",
+      "color_schemes": {
+        "background-1": {
```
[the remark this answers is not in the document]
kmeleta: We were going to update them to scheme1, scheme2, etc, but for now they're staying as is to support a theme update migration. But I guess it's also come in handy when discussing stuff like the difference between "scheme 1" (the first scheme) and "background-1" (a scheme that we _intend_ to be the first scheme) 😅

config/settings_data.json
```diff
@@ -3,17 +3,63 @@
...
+          "settings": {
+            "background": "#FFFFFF",
+            "background_gradient": "",
+            "text": "#121212",
+            "button": "#121212",
+            "button_label": "#FFFFFF",
+            "secondary_button_label": "#121212",
+            "shadow": "#121212"
```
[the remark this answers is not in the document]
kmeleta: Shadows used to use the _base_ text color so it would change it you were to change that color, but if not, they would be #121212 at all times in all schemes.

config/settings_schema.json
kmeleta: Yep. It will also eventually help apps and things safely use the same colors for similar elements in a theme. For example, you can confidently assume.. 
a) the `primary_button` color is the color the theme is using for buttons
b) `on_primary_button` _should_ have adequate contrast against `primary_button`

config/settings_schema.json
kmeleta: Yeah, that's not a thing right now. But you could remove or reorder them from settings_data.json

sections/image-with-text.liquid
kmeleta: We couldn't reorder schemes before. At least not in a way that made any difference, because we default to background-1, regardless of anything else. That was never _not_ a background-1 scheme. Now reordering, changing the name, or removing the `background-1` scheme entirely changes the meaning of "scheme 1 is used as a fallback". It's still true in that whatever the first scheme defined in settings_data is used as the fallback, but if that's not background-1 then any logic like this that's counting on that will break.

assets/template-giftcard.css
```diff
@@ -287,18 +287,18 @@ h2,
...
+  box-shadow: 0 0 0 0.1rem rgb(var(--color-button)), 0 0 0 0.3rem rgb(var(--color-background)),
+    0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
 }
 
 .button--secondary:focus-visible {
-  box-shadow: 0 0 0 0.1rem rgb(var(--color-base-outline-button-labels)),
-    0 0 0 0.3rem rgb(var(--color-base-background-1)), 0 0 0.5rem 0.4rem rgba(var(--color-base-text), 0.3);
+  box-shadow: 0 0 0 0.1rem rgb(var(--color-secondary-button-label)), 0 0 0 0.3rem rgb(var(--color-background)),
+    0 0 0.5rem 0.4rem rgba(var(--color-foreground), 0.3);
 }
```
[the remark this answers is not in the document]
kmeleta: Gross, thanks for catching that. I was accidentally using `--color-text` instead of `color-foreground` in the gift card template so that color was blank here.

melissaperreault (commented):
Thank you Ken!!

First pass results, I know I played with very questionable color choice 😅 :

- [ ] [Cart quantity input](https://screenshot.click/26-10-3epot-h1pfo.png) 
   - [ ] **Desktop**: Background should fallback to the solid color when there is a gradient, like on the product template. 
   - [ ] **Mobile**: The fallback color [is not there strangely](https://screenshot.click/26-18-b9si5-jotzs.mp4)
- [ ] Already exists, but had to note it; the [product image border](https://screenshot.click/26-24-84pj3-a33rf.png) of the Quick add modal doesn't fallback to the solid color when there is a gradient like other places. e.g. Product template, [popup notification](https://screenshot.click/26-26-7wnbm-9h8cg.png), cart thumbnail, and zoom lightbox.

sections/main-search.liquid
melissaperreault: I like the safety it gives if a merchants ever rename the color ids to match their reality or if they remove it, swap it in the code. For the rest I'll leave it to your expertise! 

It's less likely a pattern we encourage with the new system, to reorder the schemes, mostly that the first `Scheme 1` became the new baseline starting point for the whole site and this was our only way of doing it for MVP until we learn more how merchants will use, etc. 

In our system, `Scheme 1` is the parent of all the schemes and the subsequent one are meant to create diversity and rhythm where needed.

ludoboludo:
Other than what was already mentioned in the existing reviews, it looks good to me 👍 Didn't find any other issues.

layout/password.liquid
```diff
@@ -37,6 +37,46 @@
...
+          --color-scheme-text: {{ scheme.settings.text.red }},{{ scheme.settings.text.green }},{{ scheme.settings.text.blue }};
+          --color-scheme-button: {{ scheme.settings.button.red }},{{ scheme.settings.button.green }},{{ scheme.settings.button.blue }};
+          --color-scheme-button-label: {{ scheme.settings.button_label.red }},{{ scheme.settings.button_label.green }},{{ scheme.settings.button_label.blue }};
+          --color-scheme-secondary-button-label: {{ scheme.settings.secondary_button_label.red }},{{ scheme.settings.secondary_button_label.green }},{{ scheme.settings.secondary_button_label.blue }};
+          --color-scheme-shadow: {{ scheme.settings.shadow.red }},{{ scheme.settings.shadow.green }},{{ scheme.settings.shadow.blue }};
+        }
+      {% endfor %}
+
+      {% for scheme in settings.color_schemes %}
+        .color-{{ scheme.id }},
```
irenenam, line 60: Is there a way to generate this style in the first loop?  (line 40)
Maybe we can set this in liquid variable?

layout/theme.liquid
kmeleta: Adding to this.. I ended up separating these out like ^ that original version so `:root` isn't also receiving the background-color styles. That's what was causing disprencies between main and my branch.

layout/theme.liquid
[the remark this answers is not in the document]
stufreen: You don't actually use the `--color-scheme-whatever` ones anywhere else though, right? I would delete to keep it as simple as possible until we really need it.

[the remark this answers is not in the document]
kmeleta: I think there are 4? instances of them are used in the PR currently. Not to say we couldn't work around that in a way that doesn't use this pattern, but I do still like the pattern.

stufreen (changes requested):
Just took another pass and looked at all your comments. There are a few changes not addressed, so just to summarize my suggestions:
1. Remove `--color-scheme-` variables in layouts since they are not used elsewhere
2. Consider removing the `color-{{ settings.color_schemes | first }}` pattern if the variables for the first/fallback scheme are already set.
3. Remove unnecessary CSS variable in component-rating.css

melissaperreault (approved):
🎉 Alright I give up, I couldn't break anything. ❤️‍🔥 🚢

stufreen (approved):
LGTM 🚀 

@melissaperreault I wonder if we should have a `color_scheme` setting called "Default scheme" or something that lets the user select a scheme other than the first one. We could then use that for all of the places where we don't have a color scheme setting.

melissaperreault:
> I wonder if we should have a color_scheme setting called "Default scheme" or something that lets the user select a scheme other than the first one. We could then use that for all of the places where we don't have a color scheme setting.

Yeah! 😅 Big long conversations went over this potential solution of calling one of the theme "Default scheme", we considered other alternative as well. I'll share further conversation of the rationale we got from the team. I think it's worth resurfacing those doubts as we learn more how merchants and partners will use colors moving forward.

PR #2572, opened 2023-04-26
Correct pagination classes

cfxd:
### PR Summary: 

Corrects the prev/next pagination arrow classes.

For some reason, these class names were reversed.


ludoboludo (approved):
Looking good, thanks for taking the time to do this @cfxd  👍

PR #2585, opened 2023-04-27
Separate collection filters into its own section

cfxd:
### PR Summary: 

Move the facets and storefront filtering into its own separate section.

These components add significant size and page load time to the product grid section and they should be separate because they provide distinct features. This becomes especially noticeable when using the Section Rendering API to load collection products into the product grid. When the filters are no longer part of the product grid section, there's a considerable size savings (I've seen >= 55%) and time savings (I've seen >= 40%).

### Testing steps/scenarios

The best test is using infinite scrolling to load the product grid section first with, and then without, the filters included as part of that section.

1. Open Demo link [#&#x2060;1](https://infinite-scroll-demo.myshopify.com/collections/infinite-scroll) and [#&#x2060;2](https://infinite-scroll-demo.myshopify.com/collections/infinite-scroll-separate-filter-section) below, each in its own separate tab
2. Open your dev tools of choice and switch to its Network tab
3. Add "page" in the Network tab's search/filter to see only the results of the sections being loaded as you scroll
4. Scroll down several pages and note the sizes and load times of each request
5. Compare sizes and load times between tabs
6. Perform filtering as you would on any other collection

### Comparison with/without this PR

comparison

### Demo links

1. [Product grid with filters in the **_same_** section and infinite scrolling](https://infinite-scroll-demo.myshopify.com/collections/infinite-scroll) (password **infinite**)
1. [Product grid with filters in a **_separate_** section and infinite scrolling](https://infinite-scroll-demo.myshopify.com/collections/infinite-scroll-separate-filter-section) (password **infinite**)
1. [Product grid with filters in the **_same_** section](https://infinite-scroll-demo.myshopify.com/collections/all) (password **infinite**)
1. [Product grid with filters in a **_separate_** section](https://infinite-scroll-demo.myshopify.com/collections/all-products-separate-filter-section) (password **infinite**)

### Checklist

### DIFF
--- a/assets/facets.js
+++ b/assets/facets.js
@@ -76,19 +76,20 @@ class FacetFiltersForm extends HTMLElement {
   }
 
   static renderProductGridContainer(html) {
-    document.getElementById('ProductGridContainer').innerHTML = new DOMParser()
-      .parseFromString(html, 'text/html')
-      .getElementById('ProductGridContainer').innerHTML;
+    const productGridContainer = new DOMParser().parseFromString(html, 'text/html').getElementById('ProductGridContainer');
+    if (!productGridContainer) return;
+    document.getElementById('ProductGridContainer').innerHTML = productGridContainer.innerHTML;
   }
 
   static renderProductCount(html) {
-    const count = new DOMParser().parseFromString(html, 'text/html').getElementById('ProductCount').innerHTML;
+    const count = new DOMParser().parseFromString(html, 'text/html').getElementById('ProductCount');
+    if (!count) return;
     const container = document.getElementById('ProductCount');
     const containerDesktop = document.getElementById('ProductCountDesktop');
-    container.innerHTML = count;
+    container.innerHTML = count.innerHTML;
     container.classList.remove('loading');
     if (containerDesktop) {
-      containerDesktop.innerHTML = count;
+      containerDesktop.innerHTML = count.innerHTML;
       containerDesktop.classList.remove('loading');
     }
   }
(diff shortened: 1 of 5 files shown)


stufreen:
Thanks for contributing @cfxd ! One of us from the Dawn team will have a look as soon as we can.

assets/facets.js
```diff
@@ -164,6 +165,9 @@ class FacetFiltersForm extends HTMLElement {
       {
         section: document.getElementById('product-grid').dataset.id,
       },
+      {
+        section: document.getElementById('main-collection-filters').dataset.id,
+      },
```
ludoboludo, line 170: One thing I noticed while testing is that with those changes we're getting two network requests while if I look at Dawn untouched I'm getting one request when filtering 🤔 One for the filter and one for the product grid. 

They're both lighter individually but since there are 2, I end up with 7kb x 2. While on untouched Dawn I get 11kb and only one request. 

Unless I'm missing something while testing.

cfxd: That seems expected with the way the script currently grabs one section per request @ https://github.com/Shopify/dawn/blob/02aaed23526b99fd6beb851cd3ff1b751960acd3/assets/facets.js#L46

ludoboludo: Then I'm not sure it's worth decoupling the filters from the product grid 🤔 
Is there a benefit to having 2 smaller fetch vs 1 that's bigger but equal or lower than the two individual ones ?

Issue #2451, opened 2023-03-20
[Cards and Buttons] 3D "Lift" effect on hover


kjellr:
## Describe the enhancement you'd like

Add a 3D Lift effect on hover for card and button elements. 

Example: 

https://user-images.githubusercontent.com/1202812/226442119-287f15aa-de86-4e00-aead-ed24f36c768e.mp4

## Additional context/screenshots
 
- Inspired by the explorations in #2136 and #2229.
- The settings for this feature will be part of the new "Animations" section in Theme Settings, also utilized by https://github.com/Shopify/dawn/pull/2367. 
- Settings Mockup (Content is TBD):
 
<img width="337" alt="Screenshot 2023-03-20 at 3 09 23 PM" src="https://user-images.githubusercontent.com/1202812/226441726-cee39de8-b182-4d4e-aa98-5f8fa1d80ea0.png">

PR #2589, opened 2023-04-28
[3D lift animation] Raise hovered card above adjacent cards

kjellr:
### PR Summary: 

Fix shadows cast by cards with 3D Lift enabled

As noted over in https://github.com/Shopify/dawn/issues/2451#issuecomment-1526151249, there's a minor bug in the way shadows are cast by cards when the 3D Lift effect is active. The card appears to rise, but its shadow appears _behind_ the adjacent cards. This breaks the illusion, making the effect less effective visually. 

I added a slightly increased `z-index` to the parent container of cards and collage items. 

We _usually_ apply hover effects to the `card-wrapper` class, but increasing the z-index there doesn't solve the issue. I had to reach one level higher in the DOM to get the card to actually appear lifted. Not totally ideal, but it's what worked. 🤷

Before|After
---|---
Screenshot 2023-04-28 at 10 32 56 AM|Screenshot 2023-04-28 at 10 24 41 AM

Before|After
---|---
Screenshot 2023-04-28 at 11 08 30 AM|Screenshot 2023-04-28 at 10 24 50 AM

### Testing steps/scenarios

- Visit the demo store
- Visit Theme Settings > Animations, and select the "3d lift" hover effect. 
- Hover over product cards, ensure they appear to lift above their neighbors. 
- Hover over collection items, ensure they appear to lift above their neighbors. 

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139861590038)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139861590038/editor)

### Checklist


assets/base.css
```diff
@@ -3353,6 +3353,11 @@ details-disclosure > details {
     transform: rotate(0.5deg); /* Less intense rotation for collage items. */
   }
 
+  .animate--hover-3d-lift .grid__item:hover,
+  .animate--hover-3d-lift .collage__item:hover {
+    z-index: 1; /* Make sure the hovered card is the topmost card. */
+  }
```
kjellr, line 3359: As noted in the PR description, I would've preferred to target `card-wrapper:hover` as usual, but changing the z-index there won't actually lift the card + shadow above it's neighbors. I had to target the level above like this to make that happen.

assets/base.css
```diff
@@ -3353,6 +3353,11 @@ details-disclosure > details {
     transform: rotate(0.5deg); /* Less intense rotation for collage items. */
   }
 
+  .animate--hover-3d-lift .grid__item:hover,
```
ludoboludo, line 3356: applying it to `grid__item` scares me a little. It feels to generic/general. Lots of elements that aren't getting that effect I believe would still be using this class. 
So maybe we could be more specific at targeting containers of elements that do have the 3d lift applied to them 🤔 

A couple ways we could go about it: 
- use a selector that is yet to be supported on all browsers, so the experience isn't the same everywhere but it's a "cute" fix not specifically functional. So using grid__item:hover:has(.card, .collection__card, etc)`
- target something a bit more specific. `collection-list__item`, `.product-grid .grid__item`, `.complementary-slide li`

assets/base.css
ludoboludo, line 3359: For general shadows we're applying some padding to account for the shadow spread. Which we're not with this and it can cut off the shadow, specifically on sliders: [video](https://screenshot.click/28-21-amct2-zh7o9.mp4)

Something similar could be useful to look into. Maybe as a follow up for the next release.

ludoboludo (commented):
I noticed while playing with shadows that when you have some horizontal shadow using a negative value, it's now [showing that shadow on top](https://screenshot.click/28-25-e15vs-3vrd0.png) of the previous card when it wasn't before. 

Though not introduced in that PR but with 3D lift. Was that noted before and we're ok with it? I don't think it's a blocker but something for the merchants to figure out. 

Due to this: [screenshot](https://screenshot.click/28-27-8qu8i-3d0tb.png)

assets/base.css
kjellr: 👍 I targeted more specific classes in https://github.com/Shopify/dawn/pull/2589/commits/777fa46e2077d48c9d981c6c0ee5802cfdd22d8e. I _think_ I got them all: product grids, collection lists, collage items, blog articles, and complementary products. Let me know if you can think of any other contexts we should hit.

assets/base.css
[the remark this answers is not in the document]
kjellr: Good catch. This should be fixed with the combo of https://github.com/Shopify/dawn/pull/2589/commits/777fa46e2077d48c9d981c6c0ee5802cfdd22d8e and https://github.com/Shopify/dawn/pull/2589/commits/fc40e5e0a6b02a25ef09f6510d3b8b211092dd34. For the latter, I had to increase the `z-index` from `1` to `2` to get it to work here. I'm not entirely sure why — `1` worked great in every other context. 🤷

kjellr: Yeah, this would be a good followup. There's a bug with complementary products where the shadow doesn't extend out of its container: 

Screenshot 2023-04-28 at 4 13 05 PM

assets/base.css
```diff
@@ -3367,6 +3367,14 @@ details-disclosure > details {
     transform: rotate(0.5deg); /* Less intense rotation for collage items. */
   }
 
+  .animate--hover-3d-lift .product-grid .grid__item:hover,
+  .animate--hover-3d-lift .collection-list .grid__item:hover,
+  .animate--hover-3d-lift .collage__item:hover,
+  .animate--hover-3d-lift .blog-articles .article,
```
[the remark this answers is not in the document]
kjellr: Definitely. Updated in https://github.com/Shopify/dawn/pull/2589/commits/0cdd592503c809be6b7c1cbe3be66886fc4d6016.

bredowmax:
Thanks @kjellr @ludoboludo for opening this ticket!

Currently, it looks like we use **box shadow** on each of the cards - this creates issues whenever a shadow is larger than the gap between elements/cards because we can't separate the 'z-index' of the card and its shadow. 

We already use the drop shadow for the card hover animation. I wonder - is it an option to have a **drop shadow** on the parent element (i.e. grid)? ([Drop shadow vs. box shadow](https://css-tricks.com/breaking-css-box-shadow-vs-drop-shadow/))

I got this working in a Dawn test theme by adding a simple drop shadow filter to the collection grid like this:
```
ul.grid {
  filter: drop-shadow(2px 4px 6px #00000020);
}
```


By the way - I noticed that  once we sorted out the issue with adjacent cards, having the card hover shadow _below_ adjacent cards actually looks pretty good. This may actually be a viable option because it may prevent cards becoming unreadable when overlapped by shadows?
| Shadow below | Shadow above |
| ------------- | ------------- |
| Screenshot 2023-04-29 at 3 39 35 PM  | Screenshot 2023-04-29 at 3 39 29 PM  |


<details><summary>Shadow exploration, created in Figma</summary>

Screenshot 2023-04-29 at 3 46 47 PM

</details>


PS:
- Browser support: [Caniuse states 97.5% support](https://caniuse.com/css-filters)
- Performance: I can imagine that performance may suffer - but testing it on my own devices, I haven't felt any difference in performance

assets/base.css
```diff
@@ -3200,7 +3200,6 @@ details-disclosure > details {
   .animate--hover-3d-lift .deferred-media .deferred-media__poster-button {
     transition: transform var(--duration-long) ease, box-shadow var(--duration-long) ease;
     transform-origin: center;
-    z-index: 1;
```
kjellr, line 3203: @ludoboludo Just a heads up: I think we're ok removing this rule entirely for now. We're overriding it for cards with the new changes below, and I don't think it was actually necessary for buttons anyway — they appear to work ok in my testing as-is.

kjellr: This will fix the 3D Lift portion of #2596.

assets/base.css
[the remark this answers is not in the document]
ludoboludo: `z-index: 2;` seem to create its own issue with the search template: 


I don't think it's a huge deal. But ideally it doesn't happen. I think one way to go about it, would be to apply on the `product-grid`, `isolation: isolate;` so we basically scope the stacking elements and their z-index to this wrapper. 

I'n not seeing a negative impact in adding the property so far.

melissaperreault (approved):
LGTM! 🚢 🎉

ludoboludo (approved):
I think it's looking good 👍

assets/base.css
```diff
@@ -3225,6 +3224,18 @@ details-disclosure > details {
     transform: rotate(0.5deg); /* Less intense rotation for collage items. */
   }
 
+  .animate--hover-3d-lift .product-grid .grid__item:hover,
+  .animate--hover-3d-lift .collection-list .grid__item:hover,
+  .animate--hover-3d-lift .collage__item:hover,
+  .animate--hover-3d-lift .blog-articles .article:hover,
+  .animate--hover-3d-lift .complementary-slide li:hover {
+    z-index: 2; /* Make sure the hovered card is the topmost card. */
```
stufreen, line 3232: Giving the hovered item a bigger z-index does tend to create a flickering effect in some cases since z-index is either 0 or 2. It would be pretty tricky to animate the shadow, though. I can imagine fancy and complicated ways of doing it using two shadows: a top shadow and a bottom shadow that fade in and out depending on hover.

https://user-images.githubusercontent.com/14841718/235761801-b8018379-cb1b-4868-a8cb-d5ddad301769.mov

stufreen (approved):
Just one comment to consider. Personally I'm not sure the flickering effect is better than the shadows being below, but I will leave it to you if you would prefer it. We could add a follow-up issue to revisit if we think it's worthwhile.

PR #2595, opened 2023-05-01
[Announcement bar] Slides animation.

### LINKED ISSUE #2550: Announcement bar: Tweak slideshow animation
### Summary

Update slide animation in the announcement bar to be vertical or fade-in fade-out, instead of sliding right-to-left.

### Designs

https://www.figma.com/proto/ZywFj2r8xci8rMHriRYfUX/Banner-Animation-Tests?page-id=0%3A1&node-id=1-542&scaling=min-zoom&starting-point-node-id=1%3A542&show-proto-sidebar=1

Confirm with UX which option to use (leaning towards options 2 or 3)

### Details

In https://github.com/Shopify/themes/issues/398 we updated the announcement bar to be a slideshow, instead of stacking announcements.

However the animation is a little too prominent and we want to make it more subtle.

This issue is to update the animation in one of two ways (to be determined):
1. Cross-fade between slides (fade in/fade-out)
2. Slide in from bottom to top with a fade-in fade-out of opacity

eugenekasimov:
### PR Summary: 

This PR adds animation for slides in the announcement bar. 

Fixes #2550.

1. Changed `scroll-behavior:` from `smooth` to `auto` to prevent the current way of visual animation. 
2. Added classes describing animation behaviour and keyframed them. 
3. In JS added a few methods to apply animation and update the current slide in the announcement bar. Added a delay for changing slides to give some time for animation.

[Before](https://screenshot.click/15-09-hf2hk-61w3h.mp4)

[After](https://screenshot.click/15-12-eqp59-73rs0.mp4)

### Testing steps/scenarios

- [ ] Add a few announcements to the announcement bar.
- [ ] Go through slides back and forth and make sure that it responds right. When you click for the next slide it should be shifting to the left and vice versa.  
- [ ] Make sure that when you reach the last slide and keep clicking next it should return you to the first slide with opposite animation as though you clicked to the previous one. Same rule applies when clicking backward.
- [ ] In the settings set the slider for auto-rotating. Make sure that animation works properly. 
- [ ] Add the slideshow section and make sure that the current changes don't affect it.  

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139872010262/editor)

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -892,8 +892,6 @@ class SlideshowComponent extends SliderComponent {
   }
 
   setSlideVisibility(event) {
-    if (!this.wasClicked) {
-    }
     this.sliderItemsToShow.forEach((item, index) => {
       const linkElements = item.querySelectorAll('a');
       if (index === this.currentPage - 1) {
@@ -913,14 +911,6 @@ class SlideshowComponent extends SliderComponent {
       }
     });
     this.wasClicked = false;
-  }
-
-  removeAnimation(element) {
-    Array.from(element.classList).forEach(className => {
-      if (className.startsWith('announcement-bar-slider--fade')) {
-        element.classList.remove(className);
-      }
-    });
   }
 
   applyAnimationToAnnouncementBar(button = 'next') {
@@ -949,8 +939,8 @@ class SlideshowComponent extends SliderComponent {
     nextSlide.classList.add(`${animationClassIn}-${direction}`);
 
     setTimeout(() => {
-      this.removeAnimation(currentSlide);
-      this.removeAnimation(nextSlide);
+      currentSlide.classList.remove(`${animationClassOut}-${direction}`);
+      nextSlide.classList.remove(`${animationClassIn}-${direction}`);
     }, this.announcerBarAnimationDelay * 2);
   }
 


assets/global.js
```diff
@@ -728,6 +738,16 @@ class SlideshowComponent extends SliderComponent {
...
         }, {once: true});
+
+        button.addEventListener('click', () => {
+          if (button.name === 'next') {
+            this.applyAnimation(button.name);
+            this.updateAnnouncementBarSlide(button.name);
+          } else {
+            this.applyAnimation(button.name);
+            this.updateAnnouncementBarSlide(button.name);
+          }
```
KaichenWang, line 749:
```suggestion         
          this.applyAnimation(button.name);
          this.updateAnnouncementBarSlide(button.name);
```

assets/global.js
```diff
@@ -764,9 +784,17 @@ class SlideshowComponent extends SliderComponent {
     } else if (isLastSlide && event.currentTarget.name === 'next') {
       this.slideScrollPosition = 0;
     }
-    this.slider.scrollTo({
-      left: this.slideScrollPosition,
-    });
+    if (this.querySelector('.announcement-bar-slider')) {
```
KaichenWang, line 787: Can we save `this.querySelector('.announcement-bar-slider')` value to a class field, instead of querying multiple times?

assets/global.js
```diff
@@ -867,6 +906,55 @@ class SlideshowComponent extends SliderComponent {
...
+    requestAnimationFrame(() => {
+      this.querySelectorAll('.announcement-bar__message').forEach((slide) => {
+        if (this.currentSlide === this.allAnnouncements.length && button === 'previous') {
+          slide.classList.add(`announcement-bar-slider--fade-out-next`);
+          setTimeout(() => {
+            slide.classList.remove(`announcement-bar-slider--fade-out-next`);
+          }, 200);
+          setTimeout(() => {
+            slide.classList.add(`announcement-bar-slider--fade-in-next`);
+          }, 200);
```
KaichenWang, line 919:
```suggestion
          setTimeout(() => {
            slide.classList.remove(`announcement-bar-slider--fade-out-next`);
            slide.classList.add(`announcement-bar-slider--fade-in-next`);
          }, 200);
```

assets/global.js
```diff
@@ -764,9 +784,17 @@ class SlideshowComponent extends SliderComponent {
...
     }
-    this.slider.scrollTo({
-      left: this.slideScrollPosition,
-    });
+    if (this.querySelector('.announcement-bar-slider')) {
+      setTimeout (() => {
+        this.slider.scrollTo({
+          left: this.slideScrollPosition,
+        });
+      }, 200);
```
KaichenWang, line 792: I'd suggest saving the `200` value as a constant, because I see it's being used in a few places

assets/global.js
```diff
@@ -867,6 +906,55 @@ class SlideshowComponent extends SliderComponent {
...
+      this.querySelectorAll('.announcement-bar__message').forEach((slide) => {
+        if (this.currentSlide === this.allAnnouncements.length && button === 'previous') {
+          slide.classList.add(`announcement-bar-slider--fade-out-next`);
+          setTimeout(() => {
+            slide.classList.remove(`announcement-bar-slider--fade-out-next`);
+          }, 200);
+          setTimeout(() => {
+            slide.classList.add(`announcement-bar-slider--fade-in-next`);
+          }, 200);
+          setTimeout(() => {
```
KaichenWang, line 920: Can some of these `setTimout` be avoided if we leverage [animation delay](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay)?

assets/global.js
```diff
@@ -688,9 +688,17 @@ class SliderComponent extends HTMLElement {
...
+      setTimeout (() => {
+        this.slider.scrollTo({
+          left: this.slideScrollPosition,
+        });
+      }, 200);
+    } else {
+      this.slider.scrollTo({
+        left: this.slideScrollPosition,
+      });
+    }
```
KaichenWang, line 701:
```suggestion
    const delay = this.querySelector('.announcement-bar-slider') ? 200 : 0;
    setTimeout (() => {
      this.slider.scrollTo({
        left: this.slideScrollPosition,
      });
    }, delay);
```

KaichenWang (commented):
The animation looks good! Left some code suggestions.

Also, I noticed if I click next a few times quickly, the animation for returning to the 1st announcement is not playing at the correct moment anymore (between slide 4 -> 5, instead of 5 -> 1)

https://github.com/Shopify/dawn/assets/3710548/ac5d613b-41d3-4c29-bd2c-b51fffb7fc86

assets/global.js
eugenekasimov: I tried but it didn't work for me. Maybe I missed something but that what I tried:

changed this
```
slide.classList.add(`announcement-bar-slider--fade-out-previous`);
setTimeout(() => {
  slide.classList.remove(`announcement-bar-slider--fade-out-previous`);
  slide.classList.add(`announcement-bar-slider--fade-in-previous`);
}, this.delay);
setTimeout(() => {
  slide.classList.remove(`announcement-bar-slider--fade-in-previous`);
}, this.delay * 2);
```
to this
```
slide.classList.add(`announcement-bar-slider--fade-out-previous`);
slide.classList.add(`announcement-bar-slider--fade-in-previous`);
setTimeout(() => {
  slide.classList.remove(`announcement-bar-slider--fade-out-previous`);
  slide.classList.remove(`announcement-bar-slider--fade-in-previous`);
}, this.delay * 2);
```
and added `animation-delay: 200ms` to `announcement-bar-slider--fade-in-previous`

eugenekasimov:
@ludoboludo @KaichenWang 

I pushed updates. I tried to account for your suggestions regarding CSS and combined them.

@KaichenWang , regarding this quick clicking issue I think it's a very edge case and you really need to click faster than 200ms to reproduce it. I don't think it's reasonable to cover that case 🤔.

assets/global.js
KaichenWang: Ok I think we can keep the `setTimeout`. One other question is why we need to apply classes on every slide on every button click? Could we only need to apply animation on the outgoing and incoming slide?

assets/global.js
```diff
@@ -867,6 +893,49 @@ class SlideshowComponent extends SliderComponent {
     });
   }
 
+  applyAnimation(button = 'next') {
+    requestAnimationFrame(() => {
+      this.querySelectorAll('.announcement-bar__message').forEach((slide) => {
+        if (this.currentSlide === this.allAnnouncements.length && button === 'previous') {
```
KaichenWang, line 899: We can move the logic into a separate function:
``` JS
function updateSlideClass(slide, action) {
  slide.classList.add(`announcement-bar-slider--fade-out-${action}`);
  setTimeout(() => {
    slide.classList.remove(`announcement-bar-slider--fade-out-${action}`);
    slide.classList.add(`announcement-bar-slider--fade-in-${action}`);
  }, this.delay);
  setTimeout(() => {
    slide.classList.remove(`announcement-bar-slider--fade-in-${action}`);
  }, this.delay * 2);
}
```
``` JS
if (this.currentSlide === this.allAnnouncements.length && button === 'previous') {
    updateSlideClass.call(this, slide, 'next');
  } else if (this.currentSlide === 1 && button === 'next') {
    updateSlideClass.call(this, slide, 'previous');
  } else {
    updateSlideClass.call(this, slide, button);
  }
```

KaichenWang:
@eugenekasimov
 
> regarding this quick clicking issue I think it's a very edge case and you really need to click faster than 200ms to reproduce it. I don't think it's reasonable to cover that case

We can open a separate issue for this. Agreed it's an edge case, but we should try to understand why it's happening.

assets/global.js
```diff
@@ -867,6 +893,49 @@ class SlideshowComponent extends SliderComponent {
     });
   }
 
+  applyAnimation(button = 'next') {
```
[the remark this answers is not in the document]
eugenekasimov: okay, I mark this as resolved since we aligned on it.

assets/global.js
```diff
@@ -752,7 +761,19 @@ class SlideshowComponent extends SliderComponent {
   }
 
   onButtonClick(event) {
-    super.onButtonClick(event);
+    // This is copied from the parent component because we need to add a delay.
```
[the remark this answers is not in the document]
eugenekasimov: works good. thanks Kai!

sections/announcement-bar.liquid
```diff
@@ -1,95 +1,121 @@
 {{ 'component-slideshow.css' | asset_url | stylesheet_tag }}
 {{ 'component-slider.css' | asset_url | stylesheet_tag }}
 
+{%- liquid
+  assign social_icons = true
+  if settings.social_facebook_link == blank and settings.social_instagram_link == blank and settings.social_youtube_link == blank and settings.social_tiktok_link == blank and settings.social_twitter_link == blank and settings.social_pinterest_link == blank and settings.social_snapchat_link == blank and settings.social_tumblr_link == blank and settings.social_vimeo_link == blank
+    assign social_icons = false
+  endif
+-%}
```
ludoboludo, line 9: I'm thinking this could actually live within the social-icons snippet 🤔 But maybe something to change another time

assets/base.css
```diff
@@ -584,6 +584,7 @@ details > * {
 :root {
   --duration-short: 100ms;
   --duration-default: 200ms;
+  --duration-announcement-bar: 250ms;
```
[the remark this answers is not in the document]
eugenekasimov: It doesn't to me but it was Dan and Kim call :)

sections/announcement-bar.liquid
eugenekasimov: Let's change it in the future.

assets/global.js
```diff
@@ -753,20 +762,45 @@ class SlideshowComponent extends SliderComponent {
...
+        // Remove transition class to slider after animation ends
+        this.slider.classList.remove(this.sliderTransitionSelector);
+      }, this.delay * 2);
+    }
+
+    if (this.setPositionTimeout) clearTimeout(this.setPositionTimeout);
+    this.setPositionTimeout = setTimeout (() => {
+      this.slider.scrollTo({
+        left: position,
+      });
```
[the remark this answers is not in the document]
eugenekasimov: inside itself?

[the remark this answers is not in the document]
ludoboludo: Oh no, I didn't realize. It got me confused cause the function is so similar for the SliderComponent and SlideshowComponent. And named the same way as well though does it differently, one with a timeout and one without.

assets/base.css
```diff
@@ -2209,6 +2211,56 @@ product-info .loading-overlay:not(.hidden) ~ *,
   letter-spacing: 0.1rem;
 }
 
+.announcement-bar-slider--fade-in-next,
+.announcement-bar-slider--fade-in-previous,
+.announcement-bar-slider--fade-out-next,
+.announcement-bar-slider--fade-out-previous {
```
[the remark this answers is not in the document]
ludoboludo: Ah yeah it's because it should be `[class*="announcement-bar-slider--fade"]`

assets/global.js
```diff
@@ -771,20 +780,34 @@ class SlideshowComponent extends SliderComponent {
...
       this.slideScrollPosition = 0;
     }
-    this.slider.scrollTo({
-      left: this.slideScrollPosition,
-    });
+
+    this.setSlidePosition(this.slideScrollPosition);
+
+    this.appleAnimationToAnnouncementBar(event.currentTarget.name);
+  }
```
ludoboludo, line 802: One thing I noticed is that all the logic is added to be tackled onButtonClick when it should also be running on scroll/swipe 😬 

If you use it on mobile and swipe instead of you will see that it doesn't apply all the changes necessary.

So I think you could either look into: 
- applying some of the `announcement-bar-slider--fade-***` class in liquid by default. Since you know which slide will be showing on page load and which ones won't. 
- look at applying the same transitions in the `update()` function where it's checking what is the current page and what's the previous page. If you apply the logic you added to `onButtonClick` there it should work in all cases. Tho it might be triggered too late compared to when the button is clicked 🤔  


From what I'm seeing it's working ok on page load. No animation is applied but at least all slides are being shown. If you click on the previous or next button though, and then swipe later, the slides are hidden. Ideally they would work whatever the method you're using to go to the next slide is. 

https://github.com/Shopify/dawn/assets/28613840/9913deb6-a047-47a5-90a6-616f1291effd

assets/global.js
```diff
@@ -889,6 +912,43 @@ class SlideshowComponent extends SliderComponent {
...
 
+  removeAnimation(element) {
+    Array.from(element.classList).forEach(className => {
+      if (className.startsWith('announcement-bar-slider--fade')) {
+        element.classList.remove(className);
+      }
+    });
+  }
+
+  appleAnimationToAnnouncementBar(button = 'next') {
```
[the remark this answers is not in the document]
eugenekasimov: because I use Apple macbook 🤣

nice catch Ludo :)

assets/global.js
eugenekasimov: Thanks for flagging this Ludo. What if we just restrict animation if you swipe/scroll it?

Honestly, in my opinion, when you swipe it looks way more better with the original effect rather than the current one? Because when you swipe you want it to be swiped and not fade out. Thoughts?

Issue #2578, opened 2023-04-26
[Animations] Disable slide-in animation when filters are changed


kjellr:
## Describe the current behavior

Currently, our slide-in animations are re-triggered every time a filter is updated: 

https://user-images.githubusercontent.com/1202812/234629251-2b76cce6-7b21-4e91-9c41-1974d1373c7d.mp4

## Describe the expected behavior

This behavior is a little overbearing — there's no need for such a substantial animation every time a box is checked. Let's consider preventing the sequential animation from re-firing when the products list is re-rendered, or replacing the sequential animation with a simple all-at-once fade in on re-render instead.

PR #2597, opened 2023-05-01
[Facet filters] Remove animation from product grid re render

ludoboludo:
### PR Summary: 

This PR explores what we could do to limit the amount of animation when using facet filters and re rendering the product grid. 
We're not applying the animation on product cards that are coming up above the fold. But are on the rest of them when scrolled into view. 

Fixes https://github.com/Shopify/dawn/issues/2578

- I added a class `scroll-trigger--facet-cancel` on each `grid__item` from the `facets.js` file, when we're re rendering the elements. 
- Added some CSS styling to cancel the animation and set the opacity to `1` when the class mentioned above is applied. 
- In the `animation.js` Im removing the class in question if it's not within the viewport. 

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139873255446/editor)

### Checklist

### DIFF
--- a/assets/facets.js
+++ b/assets/facets.js
@@ -79,6 +79,13 @@ class FacetFiltersForm extends HTMLElement {
     document.getElementById('ProductGridContainer').innerHTML = new DOMParser()
       .parseFromString(html, 'text/html')
       .getElementById('ProductGridContainer').innerHTML;
+
+    document
+      .getElementById('ProductGridContainer')
+      .querySelectorAll('.grid__item')
+      .forEach((element) => {
+        element.classList.add('scroll-trigger--facet-cancel');
+      });
   }
 
   static renderProductCount(html) {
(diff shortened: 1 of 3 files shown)


kjellr (approved):
I like it! This eliminates the unnecessarily repetitive behavior that existed before when the animation would re-render, and still retains the ability to have items animate in on scroll. Nice work!

melissaperreault (approved):
LGTM! 🎉 🚢

assets/facets.js
```diff
@@ -79,6 +79,13 @@ class FacetFiltersForm extends HTMLElement {
     document.getElementById('ProductGridContainer').innerHTML = new DOMParser()
       .parseFromString(html, 'text/html')
       .getElementById('ProductGridContainer').innerHTML;
+
+    document
+      .getElementById('ProductGridContainer')
+      .querySelectorAll('.grid__item')
```
stufreen, line 85: Nit: It's the same elements, but would it be more self-documenting to select elements with the `scroll-trigger` class?

stufreen (approved):
Good point, it's better to keep the animation for offscreen elements 👍

PR #2608, opened 2023-05-03
Fix color for image in password

sofiamatulis:
Editor: https://os2-demo.myshopify.com/admin/themes/139893768214/editor

The background image was updated to use `background` instead of `accent` with the introduction of color schemes: https://github.com/Shopify/dawn/commit/d9d02dd0e28b385957c6543b45594dd4669ad29e#diff-6e5f0d3612802cfb04d17d3549be6264862ef203ec25ebb9dd765ffabf2791a1L11

We want to update it to use the `Solid button background` color (play with that in the settings to see it change) which is: 

`{{ scheme.settings.button.red }},{{ scheme.settings.button.green }},{{ scheme.settings.button.blue }}`
 `--color-button`


melissaperreault (approved):
LGTM! 🎉

Issue #2609, opened 2023-05-03
Fixed position elements jitter on mobile due to viewport size and address bars


ludoboludo:
## Describe the current behavior
Just realized something about some of the work we did and are about to do. When we do fixed position images, Image behavior: fixed background position or zoom in in the future, it does this wonky thing on mobile where as you land on the page, you have the address bar showing at the bottom (iPhone), then you start scrolling and it disappears. And whenever you scroll back up a little it comes back.
Which is annoying! Cause when using the fixed position, the image is taking 100% width and height of the viewport so there is a bit of a jump happening when the address bar disappears and go away. :shakes_fist:
It's a pain point in the industry in general, not just us FYI.

## Describe the expected behavior
Ideally no jitter at all time. 

## Version information (Dawn, browsers and operating systems)

- Dawn Version: 9.0.0
- Chrome Version 108.0.5359.124
- macOS Version 13.1

## Possible solution
Here are some ways people tackle it:
use a resize listener to adjust styling
new viewport units dvh and dvw are dynamic viewport height/width but you still get a little choppy change to the image :shrug:

## Additional context/screenshots

https://user-images.githubusercontent.com/28613840/236041708-fe8c6d76-cddc-45d2-93c9-a41d94590804.MP4

ludoboludo:
I think that would fit under the motion design umbrella yeah since it was introduced due to some work in that project. 
Oh but my PR linked should have closed this issue. 

It might be a thing that comes back when doing the zoom on scroll effect but we can be deal with it in the PR itself. 

The fix was shipped in the release branch as we have a fixed option for image banner 👍

PR #2611, opened 2023-05-03
[Image behavior] fixed background jitter fix on mobile

ludoboludo:
### PR Summary: 

On mobile there is a jitter happening on the image due to the address bar showing/hiding depending on your scrolling. 

Fixes https://github.com/Shopify/dawn/issues/2609

Applied a `height: 100vh;` as it doesn't take into account the address bar space being taken. This way there are no jitter when it's appearing and disappearing 👍 

### Testing steps/scenarios

- [ ] Add an image to the image banner section and set `image behavior` to `fixed background position`. 
- [ ] Test it on mobile, ideally both android and iPhone. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139894456342/editor)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -3127,6 +3127,7 @@ details-disclosure > details {
 .animate--fixed > img:not(.zoom):not(.deferred-media__poster-button),
 .animate--fixed > svg:not(.zoom):not(.deferred-media__poster-button) {
   position: fixed;
+  height: 100vh;
 }
 
 /* Animations */


eugenekasimov (approved):

melissaperreault (approved):
LGTM! 🎉 Tested on both phones works well! Tested with different image ratio too! Good find!

Dumb question: The only thing I wonder is if there will be some incompatibility if we ever offer a new banner height size like a `full height` with the help of `vh` values.

assets/base.css
```diff
@@ -3127,6 +3127,7 @@ details-disclosure > details {
 .animate--fixed > img:not(.zoom):not(.deferred-media__poster-button),
 .animate--fixed > svg:not(.zoom):not(.deferred-media__poster-button) {
   position: fixed;
+  height: 100vh;
```
kjellr, line 3130: Should we pair this with a `width: auto` to safeguard against stretching if future styles interfere here? I'm always overly cautious about images when it comes to that. 😅

ludoboludo: yeah I agree, I was thinking of that but also not seeing an impact on the images I was testing. Probably safer to add it 👍

ludoboludo: Actually messes with things 😅 and we already have declarations that set the width and max-width to `100%` while the height is set to `100% too. 

https://user-images.githubusercontent.com/28613840/236268356-8d5f4922-ee8d-4b9e-93ad-6c058cd3a51f.mp4

I could also set the height on the image wrapper instead 🤔

kjellr: Oh well. We've got another `width: 100%` rule on the wrapper too, so I'm not sure that's much better. I think what we have here seems to work ok.

ludoboludo:
> LGTM! 🎉 Tested on both phones works well! Tested with different image ratio too! Good find!
> 
> Dumb question: The only thing I wonder is if there will be some incompatibility if we ever offer a new banner height size like a `full height` with the help of `vh` values.

I don't think so as this is only applied when we have a fixed background. And when we do, the image will always be the full size of the viewport (100%). Though that 100% means something different on mobile depending on the browser. So by using 100vh we're making sure the look is consistent on mobile despite the different behaviours of dynamic address bars and what not.

sofiamatulis (approved):

PR #2617, opened 2023-05-09
[bug] Horizontal scrolling on mobile. 

eugenekasimov:
### PR Summary: 

Should be a quick fix for horizontal scrolling issue on mobile while the announcement bar slider has two or more slides. 

Fixes #0.

Add overflow-x: hidden to classes that cause the issue. 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


assets/base.css
```diff
@@ -2110,6 +2110,7 @@ product-info .loading-overlay:not(.hidden) ~ *,
 /* utility-bar */
 .utility-bar {
   height: 100%;
+  overflow-x: hidden;
```
ludoboludo, line 2113: I wonder if this is the best way to go about it. I think there is a also another way for mobile to deal with it 🤔 Here we might create this fix when maybe we should revisit our spacing approach ?

We're using negative margin, `-1.6rem`, for the slider buttons that isn't matching the mobile padding `1.5rem`. 

I think this could be a more logical fix 🤔 If in fact it does also solve our issue.

eugenekasimov: I assume we have the same issue for tablets, unfortunately I don't have a real tablet to test it, but it's the same thing when we have a negative margin and it's not matching with padding.

I think unless this solution ruins something it's simple and works.

ludoboludo: I think you can just change the `1.6` value to `1.5` and leave the rest untouched 🤔 

We don't have the same issue for tablet devices 👍 
On tablet sizes, over 750px and under 990px, the padding on the sides is `5rem` while the negative margin applied to the buttons is `-3.8rem` so we should be safe here 🎉

ludoboludo: Yeah I don't think it's much of an issue for me to keep the tablet look how it currently is. I think that alignment is ideally what we would have on mobile actually. But since the svg icon is a different size compared to the hamburger menu icon it makes it a bit trickier to align. 

There is also the possibility to add some extra negative margin to the icon. So the `0.1rem` we removed by going from `1.6` -> `1.5`  can be added to `.slider-button--prev .icon` through `margin-left: -0.1rem;`

Though I don't really think we need to go to that level of nitpicking. Curious to see what designers think 👀

melissaperreault: I am okay with the alignment proposed! 👍

melissaperreault (approved):
LGTM!

PR #2627, opened 2023-05-11
Variant List

sofiamatulis:
### PR Summary: 

https://www.figma.com/file/3y5RP5ET0JvqORkRFpyWmH/Variants-List-Design-Review?type=design&node-id=0-1&t=ursrFVPa2NlGVTkC-0

- Introduces a variant list section for the PDP

I am creating a new Bulk Buying Section. Some of the design is very similar to cart so we can re-visit this in the future to figure out ways to refactor/optimize for both sections

### Other considerations
This PR doesnt include
- ~~Sticky bar (will be in a follow up PR)~~ https://github.com/Shopify/dawn/pull/2690
- ~~No JS experience~~ https://github.com/Shopify/dawn/pull/2714
~~- Qty rules (info icon doesn't work for now) - https://github.com/Shopify/dawn/pull/2712~~
- ~~There are placeholders for the subtotals (in progress)~~
- ~~Filter (descoped)~~
~~- Errors~~
- ~~Mobile~~ https://github.com/Shopify/dawn/pull/2690

### Decision log

- Showing discounts and buying plans only in the cart page (as we currently do)

### Testing steps/scenarios

- [ ] Test on a product with no variant, one variant or multiple variants
- [ ] Add to cart, remove from cart and update cart
- [ ] Test the different settings (show image, show sku, color schemes etc)

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139987845142/editor)


assets/variant-list.css
```diff
@@ -0,0 +1,368 @@
...
+
+.variant__contents {
+  position: relative;
+}
+
+.variant__items {
+  max-height: 90rem;
+  overflow-y: scroll;
+  padding-bottom: 4rem;
+}
```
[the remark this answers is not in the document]
sofiamatulis: Interesting - can't seem to reproduce it but added it. Can you confirm it's working for you?

sofiamatulis:
> 

@eugenekasimov Thanks for flagging this! I think we are following something similar to cart where it follows the qty rules (min 1 in this case). What do you think @edmund-teh ? https://screenshot.click/31-44-jzafx-tj6li.mp4

sofiamatulis:
> 
> Does this complicate things? @sofiamatulis

@edmund-teh @eugenekasimov It gets a bit more complex with qty rules. When you get to qty rules if the min is 4 and youre at 4, the - should delete it from cart. Ill look a bit more into it. But perhaps we can add this as an enhancement and assess if we should add it.

We also need to see how this affects cart and the existing behaviour

edmund-teh:
> > Does this complicate things? @sofiamatulis
> 
> @edmund-teh @eugenekasimov It gets a bit more complex with qty rules. When you get to qty rules if the min is 4 and youre at 4, the - should delete it from cart. Ill look a bit more into it. But perhaps we can add this as an enhancement and assess if we should add it.
> 
> We also need to see how this affects cart and the existing behaviour

Good points @sofiamatulis, I definitely see where you're coming from with factoring in the qty rules. I'm okay to leave this for now. 

I wonder if it's only really weird in the case that @eugenekasimov pointed out where you have 1 in cart and click ing the - should delete it from cart.

In any case, this may be more of a product decision that @nicklepine could weigh in on.

assets/variant-list.css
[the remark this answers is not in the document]
eugenekasimov: I tested it in Chrome. Here is a short video 
https://screenshot.click/01-05-7bms3-0s45b.mp4

eugenekasimov:
Here is another issue I found during testing the feature on my real iPhone.
[Pls watch the video](https://screenshot.click/RPReplay_Final1685644316.MP4.mp4)

@edmund-teh I don't have much expertise in UX but it seems like not a great experience on mobile having a scrolling element inside a page which is also scrollable. Thoughts? 🤔

edmund-teh:
> Here is another issue I found during testing the feature on my real iPhone. [Pls watch the video](https://screenshot.click/RPReplay_Final1685644316.MP4.mp4)
> 
> @edmund-teh I don't have much expertise in UX but it seems like not a great experience on mobile having a scrolling element inside a page which is also scrollable. Thoughts? 🤔

Thanks @eugenekasimov— you're right this isn't very good from a UX perspective. I wonder if there are any alternatives or things we can do differently on mobile? Maybe there's a way to disable the inner element scroll when you're within that container on mobile because we aren't adapting any of the sticky footer features. 

Let's see if @sofiamatulis has any ideas, as well.

sofiamatulis:
> > Here is another issue I found during testing the feature on my real iPhone. [Pls watch the video](https://screenshot.click/RPReplay_Final1685644316.MP4.mp4)
> > @edmund-teh I don't have much expertise in UX but it seems like not a great experience on mobile having a scrolling element inside a page which is also scrollable. Thoughts? 🤔
> 
> Thanks @eugenekasimov— you're right this isn't very good from a UX perspective. I wonder if there are any alternatives or things we can do differently on mobile? Maybe there's a way to disable the inner element scroll when you're within that container on mobile because we aren't adapting any of the sticky footer features.
> 
> Let's see if @sofiamatulis has any ideas, as well.

I apologize. It isnt supposed to be doing that :) ill have it fixed

assets/variant-list.css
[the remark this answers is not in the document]
eugenekasimov: TIL:
Partially this depends on OS settings but it also depends on a device we use. For example, we may all have set it to `Automatically based ...` but the result will be different depending on whether you use the mouse or trackpad. 
 
Screenshot 2023-06-06 at 11 58 08 AM

snippets/variant-list-row.liquid
```diff
@@ -0,0 +1,233 @@
...
+  Accepts:
+  - item: {Object} Variant or Product object
+  - sku: {String} Sku of Product or Variant (optional)
+  - image: {Object} Product or Variant image (optional)
+  Usage:
+  {% render 'variant-list-row' variant: variant %}
+{% endcomment %}
+
+<tr class="variant-item" id="Variant-{{ variant.id }}">
+  <td class="variant-item__contain">
```
kmeleta, line 13: `__contain` is a new pattern and not very clear. I'd actually think this class would wrap all the cells or is at least used as a generic container in each column. Could we choose a more informative name? Maybe using "media", "label", or "title"

snippets/variant-list-row.liquid
```diff
@@ -0,0 +1,233 @@
...
+      <div class="variant-item__quantity-info">
+        {% render 'icon-info' %}
+      </div>
+      {%- if variant.available == false or quantity_rule_soldout -%}
+        <span class="variant-item__sold-out h4"> {{ 'products.product.sold_out' | t }} </span>
+      {%- else -%}
+        {% # theme-check-disable %}
+        {% assign cart_qty = cart | item_count_for_variant: variant.id %}
+        {% # theme-check-enable %}
+        {% render 'quantity-input', variant: variant %}
```
kmeleta, line 116: When tabbing through these the focus outlines can appear above the totals
https://screenshot.click/09-04-g3d0b-xh638.mp4

assets/variant-list.css
```diff
@@ -0,0 +1,370 @@
...
+  margin: 0 1.5rem 0 0;
+  min-width: 1.5rem;
+  min-height: 1.5rem;
+}
+
+.variant-item__sold-out {
+  opacity: 0.7;
+}
+
+.variant-item variant-list-remove-button {
```
kmeleta, line 170: These remove buttons don't appear to be focusable via tabbing. I'm not not sure how effective (or accessible) the hide/show pattern these use is. Might be nicer if these were persistent on rows with a quantity or at least the button showed up when the whole row was hovered or contained focus.

snippets/variant-list-row.liquid
kmeleta: It's also a little annoying that the window won't automatically scroll when the focus gets to the border of the totals element, since that's effectively serving at the viewport boundary now. Maybe some scroll-padding could assist with that? Not sure how the sticky changes do or will affect these behaviors though.

sections/variant-list.liquid
```diff
@@ -0,0 +1,182 @@
...
+      id="shopping-cart-variant-item-status"
+      aria-live="polite"
+      aria-hidden="true"
+      role="status"
+    >
+      {{ 'accessibility.loading' | t }}
+    </p>
+  </form>
+
+  {%- unless product.has_only_default_variant -%}
```
kmeleta, line 99: Totals element doesn't receive gradient and also doesn't appear full width as the rest of the variant list is which is noticeable when you use a gradient color scheme on the section.

sections/variant-list.liquid
```diff
@@ -0,0 +1,182 @@
...
+  <form class="variant__contents critical-hidden" method="post">
+    <div class="variant__items" id="main-variant-items">
+      <div class="js-contents">
+        <table class="variant-items">
+          <caption class="visually-hidden">
+            {{ 'sections.cart.title' | t }}
+          </caption>
+          <thead>
+            <tr>
+              <th class="caption-with-letter-spacing" colspan="1" scope="col">
```
kmeleta, line 34: Do we need the colspans on the cell headers if we don't have any cases of colspans greater than 1? I didn't notice a difference in VO announcements, but maybe there's some impact on mobile.

assets/variant-list.css
```diff
@@ -0,0 +1,370 @@
...
+
+.product-option+.product-option {
+  margin-top: 0.4rem;
+}
+
+.product-option * {
+  display: inline;
+  margin: 0;
+}
+
```
kmeleta, line 256: Ugh.. is all this type of stuff duplicated from the cart styles? I see what you mean a refactor could be in order, but can we find a middle ground for now? Like maybe consider utilizing that stylesheet for this section too, by conditionally loading it on pdp and using the existing `.cart-item ...` and `.product-option` classnames in the variant list markup? 

We can still keep the variant specific classes if we need specific styles for that, but including the existing classes would probably prevent a lot dupes. `<tr class="variant-item cart-item">`

kmeleta (commented):
Did a first pass, mostly on the code stuff before looking at the PR that builds on this. I'd like to clean up some naming conventions given the complexity we're going to be adding with these features, but I'd also like to see if we can borrow some of the existing styles from cart so we don't have to duplicate quite as much.

assets/variant-list.css
kmeleta: > The styles are slightly different and to avoid increasing the scope of testing, it's something that will be a fast follow up and not something in this PR

I'll leave it to you, but this seems like the kind of thing that would be get out of the way now. It would highlight the differences between the two cases that we should test, make every PR built on this quicker and easier to review, and perhaps most importantly, would be safer than having to go back later and risk introducing regressions.

sections/variant-list.liquid
sofiamatulis: Since this is related to the sticky bar, ill address it in https://github.com/Shopify/dawn/pull/2690

snippets/variant-list-row.liquid
sofiamatulis: This is fixed in the sticky bar PR https://github.com/Shopify/dawn/pull/2690 

👀 you can see it works as expected:https://screenshot.click/09-03-itupl-a9o4e.mp4

assets/variant-list.css
edmund-teh: @sofiamatulis I don't mind this feedback with the remove buttons, especially if it's an accessibility concern. 
Let's make them persistent on rows with a quantity.

snippets/variant-list-row.liquid
sofiamatulis: `media` is already the one that wraps the image below it and `title` didn't make sense to me because i show the name again below in a different screensize. What about `variant-item__content`?

assets/variant-list.css
sofiamatulis: @edmund-teh to confirm the behaviour, if there are any items in cart of that row, youll always see the trash can?

sections/variant-list.liquid
sofiamatulis: Good point!

KaichenWang (commented):
Great work! I left some code comments! 

As @kmeleta already suggested, it may be beneficial to look into reducing duplication with cart logic, but can be in a follow up.

assets/variant-list.js
```diff
@@ -0,0 +1,213 @@
+class VariantListRemoveButton extends HTMLElement {
+  constructor() {
+    super();
+
+    this.addEventListener('click', (event) => {
+      event.preventDefault();
+      const variantList = this.closest('variant-list');
```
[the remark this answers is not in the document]
sofiamatulis: Fair! I was following something similar to `cart.js` but I can definitely switch it to id

sections/variant-list.liquid
```diff
@@ -0,0 +1,182 @@
...
+
+<variant-list
+  class="page-width section-{{ section.id }}-padding color-{{ section.settings.color_scheme }} gradient"
+  id="variant-list"
+  data-id="{{ section.id }}"
+>
+  <form class="variant__contents critical-hidden" method="post">
+    <div class="variant__items" id="main-variant-items">
+      <div class="js-contents">
+        <table class="variant-items">
```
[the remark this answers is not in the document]
sofiamatulis: Good point! Updated it - let me know your thoughts

sofiamatulis:
Messaging update: https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4107-155292&t=sgGFEPCmn0G3IzpN-0

snippets/variant-list-row.liquid
```diff
@@ -0,0 +1,190 @@
...
+        {% render 'icon-info' %}
+      </div>
+      {% render 'quantity-input', variant: variant %}
+      {%- if cart_qty > 0 -%}
+        <variant-list-remove-button
+          id="Remove-{{ variant.id }}"
+          data-index="{{ variant.id }}"
+        >
+          <a
+            href="{{ item.url_to_remove }}"
```
eugenekasimov, line 121: I think this url is doing nothing in our case. I see that this was copied from the cart where we iterate through items. 

1. `Items` in the cart is an array of `line_items`.  [source](https://shopify.dev/docs/api/liquid/objects/cart#cart-items)
2. `line_item` has a parameter `url_to_remove`. [source](https://shopify.dev/docs/api/liquid/objects/line_item#line_item-url_to_remove)
3. So, it works perfectly in the cart. 

In our case we iterate through variants and `variant` doesn't have a parameter `url_to_remove`.

edmund-teh (changes requested):
@sofiamatulis, I wanted to add some quick notes here first:

- [x] Can we re-order the section settings and follow what's being shown on Figma? 

Original PR: https://screenshot.click/PR-Section-Settings.png
Figma: https://screenshot.click/Figma-Section-Settings.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3469-113365&t=bGAADqZ4uJbnVoME-4)

1. Show images
2. Show SKUs
3. Enable filtering
4. Color scheme
5. Section padding (w/ heading)
6. Theme Settings
7. Custom CSS

- [x]  Minor grammar fixes: 
- `Show image` should be `Show images`
- `Show SKU` should be `Show SKUs`

This was all approved by Content Design (@katycobb) when we did a pass together a week or two ago.

- [ ] Do we need to show and account for global theme settings for this section? 
Figma: https://screenshot.click/Global-Theme-Settings.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4254-108019&t=bGAADqZ4uJbnVoME-4)

Which control the following:
- border, corner radius, shadow, etc for the hoverover / popover
- border, corner radius, shadow, etc for product images
- border, corner radius, shadow, etc for qty selectors

- [ ] Can we re-order sections and put the **Quick order list** section in the section picker underneath **Collection list**?  It's currently at the bottom of the list underneath the **Page** section.

Original PR: https://screenshot.click/PR-Section-picker-order.png
Figma: https://screenshot.click/Figma-Section-picker-order.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3434-92523&t=bGAADqZ4uJbnVoME-4)

I remember working with @nicklepine on the ordering of this.

- [ ] Are we able to have a fixed number of variants shown in the section picker preview for the **Quick order list** section?

Currently, if we are adding the section to a product with many, many variants, the preview is shown super tall and the content is indistinguishable:

Example 1: https://screenshot.click/Section-picker-preview-height-1.png
Example 2: https://screenshot.click/Section-picker-preview-height-2.png

cc: @KaichenWang, Kai if you have any ideas here, please let us know!

edmund-teh:
Sorry @sofiamatulis, I know we had decided on this previously, but after conversations with other UX stakeholders, they really pushed for this being the best experience for our merchants. 

We need to have the View cart button with Remove all items under it. The additional changes with the "✔️ 10 items added" messaging also factored in here.

- [ ] 3. Re-arrange order of the footer elements for Mobile + Tablet viewports
Change: https://screenshot.click/Re-arrange-order-Change.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4291-114581&mode=design&t=TE46n0Hf3fS4TL6R-4)

sofiamatulis:
> * [ ]  Can we tweak the hover interaction details + spacing, sizing with the qty picker input/component?
> * Figma: https://screenshot.click/Figma-Desktop-Hover-Qty-picker.png
> * Video reference: https://screenshot.click/Desktop-Hover-qty-picker.mp4
> * [Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4333-115791&mode=design&t=xCpLNXUf3i7MVGzX-4)

This is part of this PR here: https://github.com/Shopify/dawn/pull/2712 It doesnt include all interactions. Could you take a look at that PR to ensure it's what you mean? 👀

edmund-teh (commented):
@sofiamatulis, there's something weird going on here when resizing down the page at tablet viewports, etc.: https://screenshot.click/Responsive-sizing-bug.png

- [x] Can we follow this Figma reference for different viewports? [Figma here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3702-111988&mode=design).

- [x] Can we make the text sizing for `✓ 1 item added` to `.caption-large`? 
https://screenshot.click/Item-added-text-size.png, [Figma for reference](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3469-115338&mode=design&t=aRohVp13ytzqDKBS-4)

edmund-teh (commented):
@sofiamatulis A few type classes changes here:

- [x] Can we change `.variant-item__name` to `font-size: 1.6rem, 
letter-spacing: 0.06rem` to better match `.price` and the variant total?
https://screenshot.click/variant-name-sku.png
https://screenshot.click/price-unit-price-reference.png


- [x] Can we change the SKU type class to `.unit-price` with `font-size: 1.1rem,
letter-spacing: 0.04rem` to better match the unit price per measure under the price per item?
https://screenshot.click/variant-name-sku.png
https://screenshot.click/price-unit-price-reference.png

[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3469-115338&mode=design&t=7C3Bofh9UnTBGCOG-4)

edmund-teh:
@sofiamatulis I had some previous UX feedback around there being too many loading spinner animations (especially when you see two) on both the `Price` and `Variant total` columns. 

It can be a bit distracting and actually misrepresentative if the price isn’t actually changing sometimes (ex: when they haven’t hit the second price break for volume pricing).

- [ ] Can we limit this to only 1 loading spinner animation under the Variant total column?

https://screenshot.click/Single-loader-animation.png
https://screenshot.click/Single-loader-animation-example.mp4

Or if this isn't possible and is a technical constraint (which I recall we have talked about in the past)? 
- [ ] Can we at least suppress the loader spinning animation under the Price column to only show when the actual price is changing (reaching a new price break tier, etc):

https://screenshot.click/Double-loader-only-when-price-change.mp4

sofiamatulis:
> * [ ]  The `Cancel` and `Remove all [x] items from your cart?` should be the same font-sizing.
>   `font-size: 1.5rem, letter-spacing: 0.06rem`

@edmund-teh I have updated the messaging. For `cancel`, it uses button-tertiary which is 1.2rem. I want to ensure you want me to modify this before doing so.

locales/en.default.schema.json
```diff
@@ -2707,6 +2707,20 @@
...
       }
     },
+    "variant-list": {
+      "name": "Quick order list",
+      "settings": {
+        "show_image": {
+          "label": "Show images"
+        },
+        "show_sku": {
+          "label": "Show SKUs"
```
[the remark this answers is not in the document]
sofiamatulis: Visual was provided

locales/en.default.json
```diff
@@ -330,6 +329,29 @@
     },
     "page": {
       "title": "Page title"
+    },
+    "variant_list": {
+      "product_total": "Product subtotal",
+      "view_cart": "View cart",
```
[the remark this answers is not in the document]
sofiamatulis: Visual was provided

locales/en.default.schema.json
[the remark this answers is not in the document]
sofiamatulis: Visual was provided

locales/en.default.json
```diff
@@ -330,6 +329,29 @@
...
+      "each": "{{ money }}/ea",
+      "product": "Product",
```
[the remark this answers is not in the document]
sofiamatulis: Visual was provided

locales/en.default.json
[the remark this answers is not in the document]
sofiamatulis: Visual was provided

edmund-teh (changes requested):
A few mobile notes, re: spacing and gaps— @sofiamatulis

- [x] For variants that are displaying a standard price per item, not even one with a strikethrough price before it— there is a weird space or gap in front.
https://screenshot.click/Mobile-Price-weird-space.png

- [ ] Can we reduce the gap between the product image and the product info? I understand it's doing this because we've separated the content into 3 columns, but is there any way to tie the product image and product info more closely together?
https://screenshot.click/Mobile-Product-image-gap.png
https://screenshot.click/Mobile-Figma-Gap-reference.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4291-108288&mode=design&t=gyrVsRxcd8ry2fb5-4)

- [ ] When the `Show images` setting is unchecked, we should have the Product info all the way to the left under the `Variant` column. It's still showing a gap even when the setting is unchecked.
https://screenshot.click/Mobile-No-images-still-gap.png

edmund-teh:
@sofiamatulis feel free to ignore this one, haha... just a question re:sticky footer. 

Would it be possible to "start" the sticky footer from either the first variant row or even from the column titles? 
See what I mean in this video here: https://screenshot.click/27-09-yrqvu-jdgfj.mp4

https://screenshot.click/Desktop-Current-PR-Stickyfooter.png
https://screenshot.click/Desktop-Stickyfooter-1st-row.png

This isn't blocking the release or part of the review. It could be something that's explored with a fast follow, etc.

edmund-teh:
@sofiamatulis 

Some of this may not be possible, but can we have some of these alignments be centered? I get that it's tricky if SKUs are turned on/off and if they have price per unit of measure, etc.

https://screenshot.click/Desktop-Alignments-centering.png

edmund-teh:
@sofiamatulis Feedback from UX stakeholders:

- [ ] Are we able to make the column titles `1.1rem` or `11px`? Would we have to change `.caption-with-letter-spacing`? 
From `font-size: 1rem` and changing things to `font-size: 1.1rem`
Current PR: https://screenshot.click/Current-PR-Column-titles.png
https://screenshot.click/caption-with-letter-spacing-1.1.png

edmund-teh:
> > Some of this may not be possible, but can we have some of these alignments be centered? I get that it's tricky if SKUs are turned on/off and if they have price per unit of measure, etc.
> > https://screenshot.click/Desktop-Alignments-centering.png
> 
> @edmund-teh Do you want me to decrease the spacing between the title and the sku to match the spacing in the price and unit price?

@sofiamatulis Yes, please! This exactly. With the price and variant total, I was just hoping they would perfectly align if you were to scan your eye across the row, side-by-side (or as close as possible).

sofiamatulis:
> 

@edmund-teh the total is aligned 👀 All 1.6rem padding top: https://screenshot.click/04-21-gnzxv-4hegc.mp4

edmund-teh:
> > 
> 
> @edmund-teh the total is aligned 👀 All 1.6rem padding top: https://screenshot.click/04-21-gnzxv-4hegc.mp4

Oh weird, @sofiamatulis. I still see it being slightly... just a hair off. Can you see what I mean in this screenshot: https://screenshot.click/Slight-misalignment.png

sofiamatulis:
> > > 
> > 
> > 
> > @edmund-teh the total is aligned 👀 All 1.6rem padding top: https://screenshot.click/04-21-gnzxv-4hegc.mp4
> 
> Oh weird, @sofiamatulis. I still see it being slightly... just a hair off. Can you see what I mean in this screenshot: https://screenshot.click/Slight-misalignment.png

I can't seem to reproduce it 🤔  https://screenshot.click/04-29-zc74t-yagqt.png

edmund-teh:
@KaichenWang Sorry! Three bits of feedback re: error messaging. 

- [ ] I just noticed in testing with this PR. And I'm not even quite sure how this can be reproduced, but here's a quick vid: https://screenshot.click/Product-inventory-Error-bug.mp4

- [ ] Can we align the icon and the error messaging when a row has no quantity rules and/or volume pricing?
https://screenshot.click/Current-PR-Error-alignment-off.png

- [ ] We had some feedback that the spacing between the error message and its parent UI could be improved. Can we make things a bit tighter here? The current implementation makes it hard to distinguish if it’s related to the input above or below.
https://screenshot.click/Figma-Error-tighter-spacing.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4592-102380&mode=design&t=pwynUUCQnntJmpr7-4)

edmund-teh:
@sofiamatulis A few bits of Mobile and one smaller viewport feedback:

- [ ] Please tell me if this isn't possible, but can we have the `Remove all items` button default to being centered when there is no messaging about products being added or removed? The current footer elements stack on Mobile and have wasted space. 
https://screenshot.click/Current-PR-Mobile-elements.png
https://screenshot.click/Figma-Mobile-elements.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4592%3A100688&mode=design&t=pwynUUCQnntJmpr7-1)

- [ ] The `Remove all items` confirmation messaging (`Remove all 6 items from cart?`)  isn't centered on Mobile. Can we also have the `Remove all` button the same size as the `View cart` button (minimum width)?
https://screenshot.click/Current-PR-Mobile-Remove-all-confirmation.png
https://screenshot.click/Figma-Mobile-Remove-all-confirmation.png
[Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4592%3A101372&mode=design&t=pwynUUCQnntJmpr7-1)

- [ ] Can we fix this misalignment on smaller viewports with no QR or VP?
https://screenshot.click/Current-PR-Smaller-Viewport-Total-items-misalignment.png

sofiamatulis:
> Sorry, do you mind checking this out again @sofiamatulis?

I think I found the issue! Should be fixed now!

sofiamatulis:
> * Please tell me if this isn't possible, but can we have the `Remove all items` button default to being centered when there is no messaging about products being added or removed? The current footer elements stack on Mobile and have wasted space.
>   https://screenshot.click/Current-PR-Mobile-elements.png
>   https://screenshot.click/Figma-Mobile-elements.png
>   [Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4592%3A100688&mode=design&t=pwynUUCQnntJmpr7-1)

@edmund-teh 

> * Please tell me if this isn't possible, but can we have the `Remove all items` button default to being centered when there is no messaging about products being added or removed? The current footer elements stack on Mobile and have wasted space.
>   https://screenshot.click/Current-PR-Mobile-elements.png
>   https://screenshot.click/Figma-Mobile-elements.png
>   [Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=4592%3A100688&mode=design&t=pwynUUCQnntJmpr7-1)

For visibility: we have decided to center this 👌

edmund-teh:
@sofiamatulis 

- [ ] Can we reduce `padding-left` to `0` here, so things vertically align properly on Mobile?
https://screenshot.click/Current-PR-Mobile-padding-misalignment-v2.png
https://screenshot.click/Mobile-padding-left-v2.png

- [ ] Can we fix the vertical misalignment here with variants that have no images, but images are still turned on on smaller viewports?
https://screenshot.click/Current-PR-Smaller-viewport-misalignment-no-image-v2.png

- [ ] Can the `Soldout` type class match our recent changes with everything being the same as price at `1.6 rem` or `16px`?
https://screenshot.click/Soldout-old-font-sizing.png

edmund-teh:
@sofiamatulis 

Sorry, another alignment issue here:

- [ ] `Price` and `Variant total` elements align nicely, but the `Variant name` with the `SKU` under are misaligned. Is there anyway to fix this?
https://screenshot.click/Variant-name-SKU-misalignment.png

edmund-teh:
@KaichenWang @sofiamatulis 

I found another issue with error messaging where the `Variant total` column wraps? It's causing some peculiar jumping on the page: https://screenshot.click/Errors-Variant-total-wrapping-bug.mp4

edmund-teh:
@sofiamatulis 

Global settings for inputs are not respecting `Corner radius` for the hoverover highlight behind the input. 
- [ ] Can we apply this to the hoverover, as well, to have both elements match?

Current PR: https://screenshot.click/Hoverover-input-corner-radius.png 
(hover over is still square, while the corner radius is set to `8px`)

Inputs global settings: https://screenshot.click/Global-settings-input-corner-radius.png

sofiamatulis:
> 

@edmund-teh This seems to be an old version. Do you mind going in incognito to ensure? They have the same padding as you can see here? Or do you mean the title and price aligning? I believe the line height is different. Do you wanna change it?

https://screenshot.click/06-33-vhh4o-nbids.mp4

edmund-teh:
> > 
> 
> @edmund-teh This seems to be an old version. Do you mind going in incognito to ensure? They have the same padding as you can see here? Or do you mean the title and price aligning? I believe the line height is different. Do you wanna change it?
> 
> https://screenshot.click/06-33-vhh4o-nbids.mp4

Sorry! @sofiamatulis, yeah I meant the title and price aligning. Can we please have these matching so the title, the price, and the variant total align properly? That would probably mean the same line-height, yeah.

sofiamatulis:
> > > 
> > 
> > 
> > @edmund-teh This seems to be an old version. Do you mind going in incognito to ensure? They have the same padding as you can see here? Or do you mean the title and price aligning? I believe the line height is different. Do you wanna change it?
> > https://screenshot.click/06-33-vhh4o-nbids.mp4
> 
> Sorry! @sofiamatulis, yeah I meant the title and price aligning. Can we please have these matching so the title, the price, and the variant total align properly? That would probably mean the same line-height, yeah.
> 
> This is from your video screenshot too: https://screenshot.click/From-video.png

@edmund-teh to confirm, do you want me to take the price line height and apply it to the title?

PR #2630, opened 2023-05-11
[Image with text] zoom in scroll

ludoboludo:
### PR Summary: 

Adding a new `image behaviour` option to the image with text section. 

Fixes https://github.com/Shopify/dawn/issues/2344

Reusing the same approach used for the Image banner section. 

### Other considerations

I'm reusing the fixed positioning from the `fixed image` option but I think it's worth trying to use `position: absolute` instead so we don't load very big images. And it would work better with the different sizes image banner and image with text media can have. 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139933646870/editor)

### Checklist

### DIFF
--- a/assets/animations.js
+++ b/assets/animations.js
@@ -41,8 +41,10 @@ function initializeScrollAnimationTrigger(rootEl = document, isDesignModeEvent =
 // Zoom in animation logic
 function initializeScrollZoomAnimationTrigger() {
   const animationTriggerElements = Array.from(document.getElementsByClassName(SCROLL_ZOOM_IN_TRIGGER_CLASSNAME));
-  const scaleAmount = 0.2 / 100;
+
   if (animationTriggerElements.length === 0) return;
+
+  const scaleAmount = 0.2 / 100;
 
   animationTriggerElements.forEach((element) => {
     let elementIsVisible = false;
@@ -73,11 +75,14 @@ function percentageSeen(element) {
   const elementHeight = element.offsetHeight + borderHeight;
 
   if (elementPositionY > scrollY + viewportHeight) {
+    // If we haven't reached the image yet
     return 0;
   } else if (elementPositionY + elementHeight < scrollY) {
+    // If we've completely scrolled past the image
     return 100;
   }
 
+  // When the image is in the viewport
   const distance = scrollY + viewportHeight - elementPositionY;
   let percentage = distance / ((viewportHeight + elementHeight) / 100);
   return Math.round(percentage);
--- a/assets/animations.js
+++ b/assets/animations.js
@@ -57,11 +57,15 @@ function initializeScrollZoomAnimationTrigger() {
 
     element.style.setProperty('--zoom-in-ratio', 1 + scaleAmount * percentageSeen(element));
 
-    window.addEventListener('scroll', () => {
-      if (!elementIsVisible) return;
+    window.addEventListener(
+      'scroll',
+      throttle(() => {
+        if (!elementIsVisible) return;
 
-      element.style.setProperty('--zoom-in-ratio', 1 + scaleAmount * percentageSeen(element));
-    });
+        element.style.setProperty('--zoom-in-ratio', 1 + scaleAmount * percentageSeen(element));
+      }),
+      { passive: true }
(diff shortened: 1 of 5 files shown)


sections/image-banner.liquid
```diff
@@ -33,7 +33,7 @@
   assign full_width = '100vw'
   assign widths = '375, 550, 750, 1100, 1500, 1780, 2000, 3000, 3840'
 
-  if section.settings.image_behavior == 'ambient'
+  if section.settings.image_behavior == 'ambient' or section.settings.image_behavior == 'zoom-in'
```
kjellr, line 36: This PR is looking good! I noticed it has these Image Banner changes from https://github.com/Shopify/dawn/pull/2623/ in here too. 

I tested the Image with Text-specific changes, and those all seem to work well in my testing too. 👍

assets/animations.js
```diff
@@ -36,7 +38,59 @@ function initializeScrollAnimationTrigger(rootEl = document, isDesignModeEvent =
...
+    const observer = new IntersectionObserver((elements) => {
+      elements.forEach((entry) => {
+        elementIsVisible = entry.isIntersecting;
+      });
+    });
+    observer.observe(element);
+
+    element.style.setProperty('--zoom-in-ratio', 1 + scaleAmount * percentageSeen(element));
+
+    window.addEventListener('scroll', () => {
```
KaichenWang, line 60: Ideally, we'd want to throttle all scroll event listeners.

assets/animations.js
```diff
@@ -36,7 +38,59 @@ function initializeScrollAnimationTrigger(rootEl = document, isDesignModeEvent =
...
 }
 
-window.addEventListener('DOMContentLoaded', () => initializeScrollAnimationTrigger());
+// Zoom in animation logic
+function initializeScrollZoomAnimationTrigger() {
+  const animationTriggerElements = Array.from(document.getElementsByClassName(SCROLL_ZOOM_IN_TRIGGER_CLASSNAME));
+  const scaleAmount = 0.2 / 100;
+  if (animationTriggerElements.length === 0) return;
+
+  let scrollPositionY = 0;
```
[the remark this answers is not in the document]
ludoboludo: Nothing, good catch 👍

assets/animations.js
ludoboludo: Let me know if you think `100` is enough. It doesn't seem to impact too poorly the animation this way 👍

assets/base.css
```diff
@@ -3182,8 +3182,8 @@ details-disclosure > details {
 
 .animate--zoom-in > img,
 .animate--zoom-in > svg {
-  transition: transform var(--duration-short) linear;
-  transform: scale(var(--zoom-in-ratio));
+  transition: scale var(--duration-short) linear;
+  scale: var(--zoom-in-ratio);
```
kmeleta, line 3186: Sorry.. I'm seeing a shadow issue that I don't see with ambient motion selected or on the image banner PR. It affects iwt but I expect it's from something added on that PR. Can we see if there's anything we can do to avoid that?

https://screenshot.click/16-26-o8uak-0l6yh.mp4

ludoboludo: Should be good to re reviewed 👍 I was reusing a `clip-path` from the fixed image styling in the image banner and that was unnecessary

kmeleta: That does fix the image with text issue, but without the clipping I seem to get a strange effect where the fixed images (from image banner) are visible behind certain sections like image with text, multicolumn, etc. This only happens with the zoom in effect selected on an image banner https://screenshot.click/20-05-6yruf-k9vdi.mp4 Are you seeing this?

ludoboludo: I had removed it here but it's necessary for the image banner section. On the other PR I had added it into the image banner CSS file and I just did this here too now, so hopefully that should be it 👌

kmeleta: Gotcha, yeah all looks good now 👍

PR #2631, opened 2023-05-12
[Footer] Remove Global Media settings inheritance from images

LucasLacerdaUX:
### PR Summary: 
Reverts a decision made in #2112 which applied Global Media Settings to images in the Footer.

They were causing layout issues in some themes, essentially preventing merchants from using a logo / transparent background image.

- Kept the wrapper class, removed the `global-media-settings` and the `box-shadow` styles.
- 
This should revert the look on image blocks back to what it was before `8.0.0`.
Any images using media settings in the footer will no longer inherit these styles.

### Testing steps/scenarios
- [ ] Update global media settings to use shadows, borders, etc.
- [ ] Add Image block to the footer. It should NOT inherit global media styles. 
  - [ ] Change the image size and confirm that it's applied to the image.
- [ ] Add a Brand Information image to the footer. It should NOT inherit global media styles. 
  - [ ] Change the image size and confirm that it's applied to the image.

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=139936661526)
- [Editor](https://os2-demo.myshopify.com/admin/themes/139936661526/editor)

### DIFF
--- a/assets/section-footer.css
+++ b/assets/section-footer.css
@@ -402,9 +402,7 @@
 }
 
 .footer-block__image-wrapper {
-  box-shadow: var(--media-shadow-horizontal-offset) var(--media-shadow-vertical-offset) var(--media-shadow-blur-radius)
-    rgba(var(--color-shadow), var(--media-shadow-opacity));
-  margin-bottom: max(calc(2rem + var(--media-shadow-vertical-offset) * var(--media-shadow-visible)), 2rem);
+  margin-bottom: 2rem;
   overflow: hidden !important;
 }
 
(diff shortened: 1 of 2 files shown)


sofiamatulis (commented):
The image in the footer is working correctly and not inheriting the global settings 👍 

I added a brand image and text and it's not showing at all 🤔 Am I missing a step

https://screenshot.click/12-38-2swb7-kd9u3.png

eugenekasimov (approved):

sofiamatulis (approved):
I think it was cached! Works as expected 🎉

PR #2633, opened 2023-05-12
Send the timezone offset as string instead of integer

fredma:
### PR Summary: 

This fixes an issue when sending timezone offset as negative integer. This change ensures the offset is sent as string which avoid the cart add issue for gift cards with recipient.

### Why are these changes introduced?

The cart serialization of line item properties with negative integer does not work. Using a string instead works. 

### What approach did you take?

Casting.

### Visual impact on existing themes

No visual impact.

### Checklist
- [ ] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [ ] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [ ] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #2636, opened 2023-05-12
Send timezone offset as string instead of integer

ludoboludo:
### PR Summary: 

This fixes an issue when sending timezone offset as negative integer. This change ensures the offset is sent as string which avoid the cart add issue for gift cards with recipient.

The cart serialization of line item properties with negative integer does not work. Using a string instead works. 

The latest version of recipient form sends a new line item property __shopify_offset which can be a negative number (based on the customer timezone). When its negative, request to add gift card to the cart will fail (cart serialization code doesnt accept negative number)

Casting.

Hide whitespace when checking the files changed

No visual impact.

### Checklist

### DIFF
--- a/assets/recipient-form.js
+++ b/assets/recipient-form.js
@@ -1,164 +1,162 @@
 if (!customElements.get('recipient-form')) {
-  customElements.define('recipient-form', class RecipientForm extends HTMLElement {
-    constructor() {
-      super();
-      this.checkboxInput = this.querySelector(`#Recipient-checkbox-${ this.dataset.sectionId }`);
-      this.checkboxInput.disabled = false;
-      this.hiddenControlField = this.querySelector(`#Recipient-control-${ this.dataset.sectionId }`);
-      this.hiddenControlField.disabled = true;
-      this.emailInput = this.querySelector(`#Recipient-email-${ this.dataset.sectionId }`);
-      this.nameInput = this.querySelector(`#Recipient-name-${ this.dataset.sectionId }`);
-      this.messageInput = this.querySelector(`#Recipient-message-${ this.dataset.sectionId }`);
-      this.sendonInput = this.querySelector(`#Recipient-send-on-${ this.dataset.sectionId }`);
-      this.offsetProperty = this.querySelector(`#Recipient-timezone-offset-${ this.dataset.sectionId }`);
-      if (this.offsetProperty) this.offsetProperty.value = new Date().getTimezoneOffset();
+  customElements.define(
+    'recipient-form',
+    class RecipientForm extends HTMLElement {
+      constructor() {
+        super();
+        this.checkboxInput = this.querySelector(`#Recipient-checkbox-${this.dataset.sectionId}`);
+        this.checkboxInput.disabled = false;
+        this.hiddenControlField = this.querySelector(`#Recipient-control-${this.dataset.sectionId}`);
+        this.hiddenControlField.disabled = true;
+        this.emailInput = this.querySelector(`#Recipient-email-${this.dataset.sectionId}`);
+        this.nameInput = this.querySelector(`#Recipient-name-${this.dataset.sectionId}`);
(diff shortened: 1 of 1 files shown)


assets/recipient-form.js
```diff
@@ -1,164 +1,162 @@
...
+        this.checkboxInput = this.querySelector(`#Recipient-checkbox-${this.dataset.sectionId}`);
+        this.checkboxInput.disabled = false;
+        this.hiddenControlField = this.querySelector(`#Recipient-control-${this.dataset.sectionId}`);
+        this.hiddenControlField.disabled = true;
+        this.emailInput = this.querySelector(`#Recipient-email-${this.dataset.sectionId}`);
+        this.nameInput = this.querySelector(`#Recipient-name-${this.dataset.sectionId}`);
+        this.messageInput = this.querySelector(`#Recipient-message-${this.dataset.sectionId}`);
+        this.sendonInput = this.querySelector(`#Recipient-send-on-${this.dataset.sectionId}`);
+        this.offsetProperty = this.querySelector(`#Recipient-timezone-offset-${this.dataset.sectionId}`);
+        if (this.offsetProperty) this.offsetProperty.value = new Date().getTimezoneOffset().toString();
```
ludoboludo, line 16: Original comment from @fredma : 
> The backend already casts the property to int if its a string, so this is safe.

sofiamatulis: @ludoboludo I think the original change was only this line, but the whole file seems to be changed 🤔 

https://github.com/Shopify/dawn/pull/2633/files#diff-b155c49124449ce256c03e05ec95dc95b232976c969766f023e30278c1833210L14

ludoboludo: Yeah formatting happened. I don't think Fred was using the prettier settings we do. So if you hide the whitespace while reviewing there should not be any change affecting the rest of the code:

sofiamatulis: I still see a lot of other lines being affected. Is that expected?

ludoboludo: Can you share a screenshot of what you're seeing ? 
From my side it's just different formatting, spacing, semi colons, parentheses, etc. And that change with `.toString()`

sofiamatulis: Sorry, what I mean is that since the file is already formatted in `main`, shouldnt the changes only be `toString`?

https://screenshot.click/12-09-6bevf-5f0j2.png

andy-polhill: I think those other changes are still just whitespace, it's just the GitHub whitespace filter not quite isolating them.

(interesting that prettier didn't do it's thing out of the box)

LucasLacerdaUX:
@ludoboludo Code looks good. How do we test this?

assets/recipient-form.js
kmeleta: > I dont see all the changes when I save it 🤔 but maybe i have different settings? (how customElements .define line went to a different line

When I prettify this file (outside this PR) I get the same type of [formatting](https://screenshot.click/12-13-2szle-uk8ew.mp4) seen here. I'm not 100% but I think prettier in its default state tends to break complex arguments like an anon function, or in this case a class definition, into their own lines even if the line was within our defined print width of 120. So this is expected as far as I can tell. 

Sofia, we can pair to see if/why your editor isn't using the same formatting as ours. Being a new addition, we've definitely had some cases of it not taking effect and instead using your existing editor/workspace preferences. Quitting vscode usually fixes it though. Either way, I would treat the formatting as a blocker.

ludoboludo: The formatting in `settings.json` should be like this: 


[Source](https://shopify.dev/docs/themes/tools/liquid-prettier-plugin#visual-studio-code)

sofiamatulis: Ah! Mystery solved 😅 thanks Ludo! Definitely not a blocker - agreed

PR #2651, opened 2023-05-17
Gradient fix for transparent background medias and cards

### LINKED ISSUE #2640: Reveal on Scroll Animation breaking gradient on Image with Text
Problem we're trying to solve: The Reveal on Scroll Animation is impacting the gradient application for image with text section

**Expected behaviour:** 1 gradient application should span the content container and Image placeholder

**Current behaviour:** 2 gradients are applied: one for content container and one for image placeholder 

This has implications for transparent png images uploaded

https://screenshot.click/12-03-vejkm-x70n8.mp4

ludoboludo:
### PR Summary: 

The issue that we're dealing with here is how gradients behave when animation is turned ON vs when it's OFF. 

The animation is creating a new stacking context on elements receiving the `animate--slide-in` due to a `transform` being applied. 
It means the gradient, which was matching the gradient applied on the general background is now based on the element size itself in which it is. 

Here we're matching some of those new styles due to animation when no animation is applied so we have some consistency. It's a visual change!

[Here is more context](https://screenshot.click/18-25-8toxq-fasqn.gif).
[And a little gif comparison](https://screenshot.click/18-25-8toxq-fasqn.gif)

Fixes #2640

Because the issue introduced via the animations isn't something we have a fix for at the moment. We're wanting to still keep some consistency between the looks and therefor change how the gradient was applied in certain scenarios. 

### Decision log

|---|---|---|---|---|
| 1 | A decision taken was to, for the time being, apply the same behaviour when animations are off. Instead of trying to fix some of the behaviours/changes when animations are on.   |   |   |   |

Here are some comparison with before and after. What it looked like without animation before and how it looks with the fixes in this PR:

#### Product cards: 

<details>
<summary>Same color schemes, No gradient</summary>

**Current:**
19-28-4r5qd-oxfiz

---
**With fix:**
19-29-kzd51-x0gs7
</details>

<details>
<summary>Same color schemes, scheme 1 gradient</summary>

**Current:**
19-26-utuha-pkoac

---
**With fix:**
19-27-mwkcv-z8lky

</details>

<details>
<summary>Different color schemes, scheme 1 & 2 gradients</summary>

**Current:**
19-25-kbo6e-brogn

---
**With fix:**
19-24-qabwu-usyhv
</details>

#### Image banner (with container):

<details>
<summary>Same color schemes, scheme 1 gradient</summary>

19-14-1zt5a-a4g0l
</details>

<details>
<summary>Different color schemes, scheme 1 & 2 gradients</summary>

19-15-8ugtx-2xgy9
</details>

#### Image with text:

<details>
<summary>Different schemes, scheme 1 & 2 gradients</summary>

**Current:**
iwt-current-diff-schemes-png

---
**With fix:**
iwt-fix-diff-schemes-png
</details>

<details>
<summary>Same scheme, scheme 1 gradient, with shadows</summary>

**Current:**
iwt-current-same-scheme-both-styles-png

---
**With fix:**
iwt-fix-same-scheme-both-styles-png

</details>

<details>
<summary>Different schemes, with overlap</summary>

**Current:**
iwt-current-diff-schemes-overlap-png

---
**With fix:**
iwt-fix-diff-schemes-overlap-png
</details>

<details>
<summary>Same schemes, with overlap</summary>

**Current:**
iwt-current-same-scheme-overlap-png

---
**With fix:**
iwt-fix-same-scheme-overlap
</details>

#### Multirow:

<details>
<summary>Scheme 1 with transparent image</summary>

19-17-0vqtw-4a83a
</details>

<details>
<summary>Scheme 2 with transparent image</summary>

19-18-28zup-q4vee
</details>

Are impacted: 
- cards (products, blog, collections)
- content containers in the image banner and slideshow section
- image banner 
- multirow
- collage (added support for gradient for the image and video block)
- collapsible content (image background so PNGs work as expected)
- video section (image background so PNGs work as expected)

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139964416022/editor)

### Checklist

### DIFF
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -118,8 +118,8 @@
   padding: var(--image-padding);
 }
 
+/* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
 .collage-card.gradient {
-  /* Needed for gradient continuity with or without animation */
   background-attachment: local;
 }
 
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -111,12 +111,15 @@
 }
 
 .collage-card {
-  background: rgb(var(--color-background));
   height: 100%;
   position: relative;
   border-radius: var(--border-radius);
   border: var(--border-width) solid rgba(var(--color-foreground), var(--border-opacity));
   padding: var(--image-padding);
+}
+
+.collage-card.gradient {
+  background-attachment: local;
 }
 
 .collage-card:after {
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -120,7 +120,7 @@
 
 /* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
 .collage-card.gradient {
-  background-attachment: local;
+  transform: perspective(0);
 }
 
 .collage-card:after {
--- a/assets/collage.css
+++ b/assets/collage.css
@@ -119,6 +119,7 @@
 }
 
 .collage-card.gradient {
+  /* Needed for gradient continuity with or without animation */
   background-attachment: local;
 }
 
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -16,6 +16,7 @@
 }
 
 .collapsible-content__media {
+  /* Needed for gradient continuity with or without animation */
   background: transparent;
 }
 
@@ -86,6 +87,7 @@
 .collapsible-row-layout .accordion {
   border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
   margin-bottom: 1.5rem;
+  /* Needed for gradient continuity with or without animation */
   background-attachment: local;
 }
 
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -15,7 +15,7 @@
   }
 }
 
-.scroll-trigger .collapsible-content__media.gradient {
+.collapsible-content__media {
   background: transparent;
 }
 
@@ -86,6 +86,7 @@
 .collapsible-row-layout .accordion {
   border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
   margin-bottom: 1.5rem;
+  background-attachment: local;
 }
 
 .collapsible-row-layout .accordion summary,
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -87,8 +87,8 @@
 .collapsible-row-layout .accordion {
   border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
   margin-bottom: 1.5rem;
-  /* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
-  background-attachment: local;
+  /* Needed for gradient continuity with or without animation, the transform scopes the gradient to its container which happens already when animation are turned on */
+  transform: perspective(0);
 }
 
 .collapsible-row-layout .accordion summary,
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -15,8 +15,8 @@
   }
 }
 
+/* Needed for gradient continuity with or without animation so that transparent PNG images come up as we would expect */
 .collapsible-content__media {
-  /* Needed for gradient continuity with or without animation */
   background: transparent;
 }
 
@@ -87,7 +87,7 @@
 .collapsible-row-layout .accordion {
   border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
   margin-bottom: 1.5rem;
-  /* Needed for gradient continuity with or without animation */
+  /* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
   background-attachment: local;
 }
 
--- a/assets/collapsible-content.css
+++ b/assets/collapsible-content.css
@@ -15,6 +15,10 @@
   }
 }
 
+.scroll-trigger .collapsible-content__media.gradient {
+  background: transparent;
+}
+
 .collapsible-content__media--small {
   height: 19.4rem;
 }
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -52,7 +52,6 @@
 .card--standard .card__inner:after {
   content: '';
   position: absolute;
-  z-index: -1;
   width: calc(var(--border-width) * 2 + 100%);
   height: calc(var(--border-width) * 2 + 100%);
   top: calc(var(--border-width) * -1);
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -64,6 +64,7 @@
 
 .card--card.gradient,
 .card__inner.gradient {
+  /* Needed for gradient continuity with or without animation */
   background-attachment: local;
 }
 
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -62,7 +62,8 @@
     rgba(var(--color-shadow), var(--shadow-opacity));
 }
 
-.card--card.gradient {
+.card--card.gradient,
+.card__inner.gradient {
   background-attachment: local;
 }
 
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -62,14 +62,14 @@
     rgba(var(--color-shadow), var(--shadow-opacity));
 }
 
+/* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
 .card--card.gradient,
 .card__inner.gradient {
-  /* Needed for gradient continuity with or without animation */
   background-attachment: local;
 }
 
+/* Needed for gradient continuity with or without animation so that transparent PNG images come up as we would expect */
 .card__inner.color-background-1 {
-  /* Needed for gradient continuity with or without animation */
   background: transparent;
 }
 
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -68,6 +68,11 @@
   background-attachment: local;
 }
 
+.card__inner.color-background-1 {
+  /* Needed for gradient continuity with or without animation */
+  background: transparent;
+}
+
 .card .card__inner .card__media {
   overflow: hidden;
   /* Fix for Safari border bug on hover */
--- a/assets/component-card.css
+++ b/assets/component-card.css
@@ -62,10 +62,10 @@
     rgba(var(--color-shadow), var(--shadow-opacity));
 }
 
-/* Needed for gradient continuity with or without animation, background-attachment: local scopes the gradient to its container which happens automatically when a transform is applied (animation on scroll) */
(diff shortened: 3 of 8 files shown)


assets/collapsible-content.css
```diff
@@ -15,6 +15,10 @@
   }
 }
 
+.scroll-trigger .collapsible-content__media.gradient {
+  background: transparent;
+}
```
ludoboludo, line 20: I wonder if the background transparency should also be applied when the animation isn't on: 
<details>
<summary>Screenshot</summary>


</details>

assets/video-section.css
```diff
@@ -4,6 +4,10 @@
   padding-bottom: calc(var(--ratio-percent) - var(--media-border-width));
 }
 
+.scroll-trigger:where(.gradient.video-section__media) {
+  background: transparent;
+}
```
ludoboludo, line 9: Video section with animation and without. 

<details>
<summary>Comparison screenshot</summary>


</details>

assets/component-card.css
```diff
@@ -52,7 +52,6 @@
 .card--standard .card__inner:after {
   content: '';
   position: absolute;
-  z-index: -1;
```
ludoboludo, line 55: This has the biggest impact as it means, shadows aren't hidden in the back anymore but showing on top of the preceding element: 
<details> 
<summary>Screenshot</summary>


</details>

assets/component-image-with-text.css
```diff
@@ -299,6 +297,14 @@
   z-index: 2;
 }
 
+.image-with-text:not(.image-with-text--overlap) .color-background-1 {
+  background: transparent;
+}
+
+.image-with-text .gradient:not(.color-background-1) {
+  background-attachment: local;
+}
```
ludoboludo, line 306: I think this should also be applied when it's `.color-background-1` and set to Overlap. Also In that same situation, setting the image background to transparent maybe ? Checkout the border and background applied to the image with a transparent background:

assets/component-image-with-text.css
```diff
@@ -51,8 +51,6 @@
 .image-with-text__media--placeholder:after {
   content: '';
   position: absolute;
-  width: 100%;
-  height: 100%;
```
ludoboludo, line 55: This was not doing anything, that's why we're removing it ?

assets/component-image-with-text.css
kmeleta, line 306: This first style is to preserve some of the background-1 cases that are possible in image with text, like these..


The second rule basically removes the fixed global gradient look for all non-background-1 cases so image with text displays the same with and without animations enabled https://screenshot.click/18-32-mrcd0-5ujji.mp4

ludoboludo: cc: @kmeleta

assets/section-image-banner.css
```diff
@@ -8,6 +8,10 @@
   text-align: center;
 }
 
+.banner__box.gradient {
+  background-attachment: local;
+}
```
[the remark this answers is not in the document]
kmeleta: That makes sense to me, but I guess I fall slightly more in favor of keeping the background-1 cases preserved where possible. Maybe leave it as you have for now and see what others think.

assets/component-image-with-text.css
kmeleta: Removing it fixes a small bug with the placeholder shadow that was preexisting and unrelated to this PR, but it's much uglier when removing persistent gradient on the media like we're doing in this PR https://screenshot.click/19-59-lvmkd-swxu2.mp4

assets/component-image-with-text.css
kmeleta: Right, so in light of our decision to minimize (or completely avoid) background-1 type exceptions, I'm still opting to treat at least the overlap with local gradient in all cases for now. Additionally I feel making the content transparent can end up wonky pretty easily and making the media transparent without the content feels odd to me. But entirely possible still if we want to push back on that more.

kjellr: Can you remind me of the downside to making the background transparent elsewhere (like on Product cards) too? It seems like it would minimize the overall changes for merchants today. If I recall, we had some concerns about shadows overlapping text, but if merchants keep the shadows reasonable I think we'll be okay. 🤔 

<details>
<summary>Example</summary>

Before|After
---|---
Screenshot 2023-05-23 at 2 35 14 PM|Screenshot 2023-05-23 at 2 36 03 PM

</details>

assets/collapsible-content.css
kjellr: Yeah, probably. 👍

assets/component-image-with-text.css
kjellr: I think so, yes. Especially for this overlap version. 

In general though, I'm starting to get the feeling that we should make a few more exceptions for `background-1`. [Use cases like this](https://github.com/Shopify/dawn/pull/2651#discussion_r1202866791), as well as [maybe product cards](https://github.com/Shopify/dawn/pull/2651#discussion_r1202857830) feel like they could benefit from that. 

I know that breaks our consistency decision a little bit, but seeing this in action, I think we'll have slightly better visual outcomes if we make a few more exceptions. I'll do some more testing in the morning and form a more cohesive opinion.

assets/component-card.css
ludoboludo: This is outdated. Without animation the shadow isn't on top of the previous card anymore. It happens only when animation is on.

assets/component-image-with-text.css
kmeleta: For cards I didn't think there were too many concerns other than potential shadow stuff. Though I thought we were interested in "correcting" the way gradients were applied to card style cards anyway so that the gradient was applied in a more expected way. I can be convinced of either. @ludoboludo any thoughts?

melissaperreault:
## Observations [WIP]

**Image with text**
- It gives back power to other sections that had similar outcome. The impact on that section feels comfortable for a few reasons; the most important one is that it gives better purpose for other sections where floating content on an on grid or full width image may be more relevant. e.g. Image banner
- For that reason above, for `Multirow`, I wonder if it's worth keeping the `Scheme 1` behave differently, maybe it's still okay too. Just a thought!

assets/video-section.css
melissaperreault: Is this meant to document that there are no changes between before and after?

assets/section-image-banner.css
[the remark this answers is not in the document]
kjellr: 👍 Cool, that works for me. Going to mark this one as resolved.

assets/collapsible-content.css
kjellr: Yeah, we should definitely apply transparency here. The full gradients are bound to look weird when they’re crammed into those short, long shapes: 

Screenshot 2023-05-24 at 1 29 06 PM

...but also, [we display those accordions as transparent](https://screenshot.click/24-33-08jtv-fbosz.png) within the product details, so making them transparent here would allow merchants to have the Collapsible Content section match that.

ludoboludo: Oh sorry in my screenshots I meant for the image specifically and how the shadow either show or not behind the transparent image.

I think if the layout is set to row containers they shouldn't be set to transparent. Cause in this case you might as well set the layout to `No container`. I don't think it looks pretty, specifically with the radial gradient but that's up to the merchant to pick something nice. 
With a different gradient it doesn't look bad:

kjellr (approved):
Aside from [this minor change](https://github.com/Shopify/dawn/pull/2651/files#r1204564399) (which I consider a nice-to-have), I think this is good to go. The changes are relatively minimal for our existing themes by default, and it's relatively easy to explain the intent of the new gradient behavior. Also, if the updated gradients don't work for a merchant, it's easier to change than before due to the new Color Schemes interface. 👍

assets/collapsible-content.css
```diff
@@ -82,6 +86,7 @@
 .collapsible-row-layout .accordion {
   border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
   margin-bottom: 1.5rem;
+  background-attachment: local;
```
ludoboludo, line 89: Added this to avoid this difference:

assets/collapsible-content.css
ludoboludo: Change should be done for the image transparency 👍

assets/component-image-with-text.css
kjellr: Just closing the loop here: I think we're fine with the cards approach as-is. If merchants want to have the backgrounds be transparent, they can change from "Card" to "Standard" in most cases.

kjellr:
👋 I found one more small use case I think we should make transparent: 

When the card style is set to "Standard" and the color scheme is the `background-1`, we should make the background behind the images transparent. This is in line with what we do for the image banner — it takes something that's supposed to appear transparent against the background and maintains that appearance. 

Current|Expected
---|---
Screenshot 2023-05-26 at 7 47 40 AM|Screenshot 2023-05-26 at 7 49 23 AM

melissaperreault:
🔍  I found a bug on `Image banner` and `Slideshow` where we don't preserve the color scheme on the container for Scheme 2 and beyond. [Video reference for Slideshow](https://screenshot.click/26-59-2tmtx-df32z.mp4)

assets/component-card.css
```diff
@@ -62,6 +62,12 @@
     rgba(var(--color-shadow), var(--shadow-opacity));
 }
 
+.card--card.gradient,
+.card__inner.gradient {
+  /* Needed for gradient continuity with or without animation */
+  background-attachment: local;
+}
+
 .card .card__inner .card__media {
```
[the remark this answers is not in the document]
kjellr: Cool, thanks for the clarification. Yeah, I think what you pushed here makes sense. Merchants can always add a border if they want one. 👍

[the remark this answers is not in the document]
ludoboludo: Yeah or create a specific scheme they can use in that scenario 👌

assets/collage.css
```diff
@@ -111,14 +111,18 @@
...
-  background: rgb(var(--color-background));
   height: 100%;
   position: relative;
   border-radius: var(--border-radius);
   border: var(--border-width) solid rgba(var(--color-foreground), var(--border-opacity));
   padding: var(--image-padding);
 }
 
+.collage-card.gradient {
+  /* Needed for gradient continuity with or without animation */
```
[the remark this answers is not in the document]
ludoboludo: For `2.` the reason I didn't do that (even though I think ideally it's better), is because sometimes it's a whole declaration that's the fix but sometimes it's just one declaration and not the whole thing. For example: 
```css
.collapsible-row-layout .accordion {
  border: var(--text-boxes-border-width) solid rgba(var(--color-foreground), var(--text-boxes-border-opacity));
  margin-bottom: 1.5rem;
  /* Needed for gradient continuity with or without animation */
  background-attachment: local;
```

So I went with a similar approach everywhere but could mix it up where we can if you think it's better 👍

eugenekasimov (approved):

melissaperreault (commented):
**_To be discussed_**
- [ ] Product image lightbox is [not considering the corner radius](https://screenshot.click/30-09-tzbac-qqdjp.png) anymore

**_Not a blocker_**
- [ ] Little corner visible when we have 3D Lift hover effect turned on. Not a blocker to me! ([Video reference](https://screenshot.click/30-57-fcu0o-urtw9.mp4))

kjellr (approved):
The remaining issues are all pretty minor (I honestly had trouble figuring out how to reproduce them), so I think they're fine to follow up on later. In the meantime, we're good to go.

PR #2652, opened 2023-05-17
Split the slideshow and announcement bar slider

### LINKED ISSUE #2604: Split the slideshow and announcement bar into two components in JS.
## Describe the issue

We built the announcement bar slider on top of the slideshow-component. For the first time that was ok. Now we have added lots of things to the announcement bar slider that have nothing common with the slideshow-component. This makes adding new features and maintaining the announcement bar slider difficult.

## Solution

We need to split the slideshow and the announcement bar slider into two independent components. Because these two components still have a lot of common code there is a chance that we will need to make a new component that will have all common code while the slideshow-component and the announcement bar slider will have their own code.

eugenekasimov:
### PR Summary: 

In this PR I separated the slideshow and announcement bar slider into two different components.  

Fixes #2604 .

Originally the slider for the announcement bar was built on top of the slideshow component by adding extra code to the existing component. Since the two components have common code I ended up having tree components:
Screenshot 2023-05-17 at 9 31 47 AM

N/A

### Testing steps/scenarios

- [ ] Add a few announcements in the announcement bar. Make sure that everything works as before. 
- [ ] Add the slideshow section and ensure that everything works as before.

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/139963400214/editor)

### Checklist

### DIFF
--- a/locales/cs.schema.json
+++ b/locales/cs.schema.json
@@ -1625,8 +1625,8 @@
               "info": "V rámci platebních metod dostupných v obchodě uvidí zákazníci tu, kterou nejvíce preferují, jako například PayPal nebo Apple Pay. [Zjistit více](https://help.shopify.com/manual/using-themes/change-the-layout/dynamic-checkout)"
             },
             "show_gift_card_recipient": {
-              "label": "Zobrazit formulář s informacemi o příjemci u produktů dárkových karet",
-              "info": "Dárkové karty je možné volitelně v naplánované datum posílat přímo příslušnému příjemci společně s osobní zprávou. [Zjistit více](https://help.shopify.com/manual/online-store/themes/customizing-themes/add-gift-card-recipient-fields)"
+              "label": "Zobrazit formulář s informacemi o příjemci u dárkových karet",
+              "info": "Umožňuje kupujícím odeslat dárkové karty v naplánované datum společně s osobní zprávou. [Zjistit více](https://help.shopify.com/manual/online-store/themes/customizing-themes/add-gift-card-recipient-fields)"
             }
           }
         },
--- a/locales/da.schema.json
+++ b/locales/da.schema.json
@@ -1625,8 +1625,8 @@
               "info": "Via de tilgængelige betalingsmetoder i din butik ser kunderne deres foretrukne mulighed, som f.eks. PayPal eller Apple Pay. [Få mere at vide](https://help.shopify.com/manual/using-themes/change-the-layout/dynamic-checkout)"
             },
             "show_gift_card_recipient": {
-              "label": "Vis formular til modtageroplysninger for gavekortprodukter",
-              "info": "Gavekortprodukter kan sendes direkte til modtageren på en planlagt dato sammen med en personlig besked. [Få mere at vide](https://help.shopify.com/manual/online-store/themes/customizing-themes/add-gift-card-recipient-fields)"
+              "label": "Vis formularer til modtageroplysninger for gavekort",
+              "info": "Giver købere mulighed for at sende gavekort på en planlagt dato sammen med en personlig besked. [Få mere at vide](https://help.shopify.com/manual/online-store/themes/customizing-themes/add-gift-card-recipient-fields)"
             }
           }
         },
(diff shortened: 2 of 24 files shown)


assets/global.js
```diff
@@ -879,7 +892,42 @@ class SlideshowComponent extends SliderComponent {
...
+
+    this.announcementBarArrowButtonWasClicked = false;
+
+    [this.reducedMotion, this.desktopLayout].forEach((mediaQuery) => {
+      mediaQuery.addEventListener('change', () => {
+        if (this.slider.getAttribute('data-autoplay') === 'true') this.setAutoPlay();
+      });
+    });
+
+    [this.prevButton, this.nextButton].forEach((button) => {
```
LucasLacerdaUX, line 909: This can extend the `onButtonClick` function that is already attached to these elements:

```js
  onButtonClick(event) {
    super.onButtonClick(event);
    this.announcementBarArrowButtonWasClicked = true;
  }
```

assets/global.js
```diff
@@ -740,15 +722,6 @@ class SlideshowComponent extends SliderComponent {
     this.addEventListener('mouseleave', this.focusOutHandling.bind(this));
     this.addEventListener('focusin', this.focusInHandling.bind(this));
     this.addEventListener('focusout', this.focusOutHandling.bind(this));
-
-    if (this.querySelector('.slideshow__autoplay')) {
-      this.sliderAutoplayButton = this.querySelector('.slideshow__autoplay');
```
LucasLacerdaUX, line 745: Can you explain in more details why the `SlideshowSharedComponent` differs from the `SlideComponent`, for instance? 

As I look at these classes, I'm questioning whether or not we actually need three components instead of two: the `SlideshowSharedComponent` isn't used anywhere - it's just there to be extended. But if it were to be used on its own, it wouldn't work: both the `focusInHandling` and `focusOutHandling` functions are making reference to this `this.sliderAutoplayButton` that no longer exits in this component, as well as `announcementBar`-related elements that are no longer there.

assets/global.js
```diff
@@ -879,7 +892,42 @@ class SlideshowComponent extends SliderComponent {
...
+  }
+
+  focusOutHandling() {
+    if (!this.reducedMotion.matches && !this.announcementBarArrowButtonWasClicked && this.desktopLayout.matches) this.play();
+  }
+
+  focusInHandling(event) {
+    if (this.querySelector('.announcement-bar-slider').contains(event.target)) this.pause();
+  }
+}
```
LucasLacerdaUX, line 928: If these are not using the inherited `focusInHandling` and `focusOutHandling`, do we need them in the `SlideshowSharedComponent` class at all?

LucasLacerdaUX (commented):
I don't have the full context behind these components as I haven't reviewed the announcement bar PRs.

I know that the `AnnouncementBar` is pretty different so separating it into it's own component makes sense, but the difference between `SlideshowSharedComponent`, `SlideComponent` and `SlideshowSectionComponent` isn't so clear to me. Could you add more details as to which functions you want each component to have?

If we want to keep both the `SlideshowSectionComponent` and the `SlideshowSharedComponent`, we need to make sure the latter works as a generic component - right now, it's using references to invalid elements that are specific to the ones inheriting from it.

assets/global.js
```diff
@@ -771,48 +739,15 @@ class SlideshowComponent extends SliderComponent {
...
-      if (focusedOnAutoplayButton && this.autoplayButtonIsSetToPlay) {
-        this.play();
-      } else if (this.autoplayButtonIsSetToPlay) {
-        this.pause();
-      }
-    } else if (this.querySelector('.announcement-bar-slider').contains(event.target)) {
-      this.pause();
-    }
+  focusInHandling() {
+    this.pause();
```
eugenekasimov, line 750: same case here 👆

assets/global.js
```diff
@@ -879,7 +873,45 @@ class SlideshowComponent extends SliderComponent {
...
+    this.announcementBarArrowButtonWasClicked = false;
+    this.desktopLayout = window.matchMedia('(min-width: 750px)');
+    this.reducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)');
+
+    [this.reducedMotion, this.desktopLayout].forEach((mediaQuery) => {
+      mediaQuery.addEventListener('change', () => {
+        if (this.slider.getAttribute('data-autoplay') === 'true') this.setAutoPlay();
+      });
+
+      if (this.slider.getAttribute('data-autoplay') === 'true') this.setAutoPlay()
```
eugenekasimov, line 891: This is the only code that I repeat in both components although it's the same. But I could't find a way to keep it in the parent element. I think it's okay.

assets/global.js
eugenekasimov: I changed it as you suggested, thank you.

assets/global.js
eugenekasimov: I think I can resolve this one as we discussed it during our pairing session.

Issue #1391, opened 2022-02-22
Slideshow + Image Banner when no container are limited to white text on image


nicklepine:
**Describe the current behavior**
When opting to remove the container on slideshow or image banner, text is always white. This is not ideal when the image a merchant wishes to use is light. 

We should let merchants pick the text color, or have them pick between white or their globally set text color. 

**Update: May 17th, 2023**
This issue was flagged by leadership twice. We will prioritize a fix

melissaperreault:
### A few discussion points to consider

- Remove the override of white applied to Text and Button and keep relying on the color scheme to change the elements
- Keep the overlay opacity setting but also offer an content container opacity so we can give more flexibility and not rely on creating always dramatic banner to offer good contrast.

kjellr:
This is a tricky one because it depends so much on the overall darkness of the image. Also, in schemes with dark text colors: if we _don't_ make the text white by default, the text will be unreadable whenever the image overlay opacity past a certain percentage. The overlay color is hardcoded to black. 

As noted above, I think the simplest UX solution might be to give folks a checkbox that allows them to choose between white or the text color set in the scheme.

We might also consider changing Dawn's default homepage layout to use the content container on desktop. That won't fix the problem, but it will at least provide users with something that works less awkwardly with Color Schemes by default.

kjellr:
We found some confusing bits to #2663, so here's a more modest proposal for consideration as well. This one aims to make it more clear that the color scheme is only used when the container is displayed, by reordering settings and placing the info messaging before the color scheme setting instead of after it.

The changes are: 

1. Move the "Show container on desktop" checkbox down below the "Desktop content alignment" dropdown. 
2. Move the "Visible when container displayed." info text up, in between the "Show container on desktop" field and the color palette field, and rewrite to be more clear. 

Before|After
---|---
<img width="300" alt="Screenshot 2023-05-23 at 11 25 10 AM" src="https://github.com/Shopify/dawn/assets/1202812/974d9b9b-54af-4c27-8e00-b919142bab01">|<img width="300" alt="Screenshot 2023-05-23 at 11 24 52 AM" src="https://github.com/Shopify/dawn/assets/1202812/aeec18da-356b-4c3d-9d5e-28278f6ac6c5">

cc @katycobb: do you have any thoughts on that rewrite of the "Visible when container displayed" copy there? We want to be super clear that it's in reference to the color palette control.

melissaperreault:
> We could add it before the existing help text

[...]

PR #2663, opened 2023-05-19
Remove forced white text color in Image Banner & Slideshow

kjellr:
### PR Summary: 

Always allow text and button color customization within the Image banner and Slideshow sections. 
Important: Color scheme may need to be manually changed to  prevent visual change and ensure adequate contrast.

Fixes https://github.com/Shopify/dawn/issues/1391

This PR does a few things: 

1. On Image Banner and Slideshow sections, it removes the forced white color for text and buttons when the content container is hidden. 
2. On Dawn's default homepage template, it changes the Image Banner's default color palette from `background-1` (which uses black text) to `inverse` (which uses white text). 
3. It removes the "Visible when container displayed." that showed underneath the color palette control for Image Banner & Slideshow, since it's no longer relevant. 

For new merchants, this approach maintains the acceptable contrast ratio that existed by default on the top of the homepage template for Dawn (the text and buttons are still white, so there's no perceptable change.)

In other areas, we cannot guarantee an acceptable ratio today anyway: If the background image is light, and/or there's no overlay applied, white text is unreadable. In the past, the merchant was stuck with that being broken. With this new approach, the merchant can fix it by applying a new color scheme. 

<details>
<summary>🎥 Video</summary>

https://github.com/Shopify/dawn/assets/1202812/79643d16-d8f5-483b-8e52-b3673402be12

</details>

### Other considerations

- [ ] This should be fine for new merchants, but if a merchant upgrades to this change automatically, they may lose their previously-white text. Is there anything we can do about that? 
- [ ] We need to double check and ensure that this works right with the starter content in our other themes that use Image Banners too. 

### Visual changes
If selected color scheme doesn't use white text, image banners and slideshows without content containers may have a visual change. Shops will need to manually update their color scheme to account for this.

### Testing steps/scenarios

1. [Visit the demo store](https://os2-demo.myshopify.com/admin/themes/139977031702/editor). 
2. Select the Image Banner at the top of the page. Change the color scheme, and note that the text + button color updates.
3. Edit the color scheme, ensure that the new colors also appear.
4. Repeat with a freshly-inserted Image Banner section and a Slideshow section. 

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/140005474326/editor)

### Checklist

### DIFF
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -209,9 +209,6 @@
 
   .banner:not(.banner--mobile-bottom):not(.email-signup-banner) .banner__box {
     background: transparent;
-    --color-foreground: 255, 255, 255;
-    --color-button: 255, 255, 255;
-    --color-button-text: 0, 0, 0;
   }
 
   .banner:not(.banner--mobile-bottom) .banner__box {
@@ -220,9 +217,7 @@
     box-shadow: none;
   }
 
-  .banner:not(.banner--mobile-bottom) .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
+  .banner--desktop-transparent .button--secondary {
     --alpha-button-background: 0;
   }
 
@@ -335,9 +330,6 @@
 @media screen and (min-width: 750px) {
   .banner--desktop-transparent .banner__box {
     background: transparent;
-    --color-foreground: 255, 255, 255;
-    --color-button: 255, 255, 255;
-    --color-button-text: 0, 0, 0;
     max-width: 89rem;
     border: none;
     border-radius: 0;
@@ -345,8 +337,6 @@
   }
 
   .banner--desktop-transparent .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
     --alpha-button-background: 0;
   }
 
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -217,7 +217,7 @@
     box-shadow: none;
   }
 
-  .banner--desktop-transparent .button--secondary {
+  .banner .button--secondary {
     --alpha-button-background: 0;
   }
 
--- a/assets/section-image-banner.css
+++ b/assets/section-image-banner.css
@@ -217,7 +217,7 @@
     box-shadow: none;
   }
 
-  .banner .button--secondary {
+  .banner:not(.banner--mobile-bottom) .button--secondary {
     --alpha-button-background: 0;
   }
 
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1186,9 +1186,6 @@
           },
           "label": "Desktop content alignment"
         },
-        "color_scheme": {
-          "info": "Visible when container displayed."
-        },
         "mobile": {
           "content": "Mobile Layout"
         },
@@ -3043,9 +3040,6 @@
             },
             "image_overlay_opacity": {
               "label": "Image overlay opacity"
-            },
-            "color_scheme": {
-              "info": "Visible when container displayed."
             },
             "text_alignment_mobile": {
               "label": "Mobile content alignment",
--- a/locales/cs.schema.json
+++ b/locales/cs.schema.json
@@ -2572,9 +2572,6 @@
         "image_overlay_opacity": {
           "label": "Neprůhlednost překryvu obrázku"
         },
-        "color_scheme": {
-          "info": "Zobrazuje se v případě, že se zobrazuje kontejner."
-        },
         "show_text_below": {
           "label": "Na mobilním zařízení zobrazovat obsah pod obrázkem",
           "info": "Nejlepších výsledků dosáhnete použitím obrázku s poměrem stran 16:9. [Zjistit více](https://help.shopify.com/manual/shopify-admin/productivity-tools/image-editor#understanding-image-aspect-ratio)"
@@ -2815,9 +2812,6 @@
             },
             "image_overlay_opacity": {
               "label": "Neprůhlednost překryvu obrázku"
-            },
-            "color_scheme": {
-              "info": "Zobrazuje se v případě, že se zobrazuje kontejner."
             },
             "text_alignment_mobile": {
               "label": "Zarovnání obsahu v mobilním prostředí",
--- a/locales/da.schema.json
+++ b/locales/da.schema.json
@@ -2572,9 +2572,6 @@
         "image_overlay_opacity": {
           "label": "Billedoverlejringens uigennemsigtighed"
         },
-        "color_scheme": {
-          "info": "Synlig, når objektbeholderen vises."
-        },
         "show_text_below": {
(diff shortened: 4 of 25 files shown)


assets/section-image-banner.css
```diff
@@ -220,9 +217,7 @@
     box-shadow: none;
   }
 
-  .banner:not(.banner--mobile-bottom) .button--secondary {
-    --color-button: 255, 255, 255;
```
kmeleta, line 224: > This should be fine for new merchants, but if a merchant upgrades to this change automatically, they may lose their previously-white text. Is there anything we can do about that?

Unless I'm thinking about this wrong, even if we added a migration to automatically update a theme to, say, inverse scheme, we can't be sure that scheme actually is using white (or even a relatively accessible color). Though this will be a manual update I'm sure so it shouldn't be "automatic" in that way at least, but still breaking.

assets/section-image-banner.css
```diff
@@ -335,18 +330,13 @@
...
     max-width: 89rem;
     border: none;
     border-radius: 0;
     box-shadow: none;
   }
 
   .banner--desktop-transparent .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
     --alpha-button-background: 0;
```
kmeleta, line 340: Although I obviously know why this is happening, this is a weird result. Or at least it's different than the desktop equivalent. I'd expect the secondary button to stay transparent unless the mobile container was enabled https://screenshot.click/19-41-z63lp-yb7ol.mp4

locales/en.default.schema.json
```diff
@@ -1195,9 +1195,6 @@
           },
           "label": "Desktop content alignment"
         },
-        "color_scheme": {
-          "info": "Visible when container displayed."
-        },
```
kmeleta, line 1200: Ideally we go thru and remove these from all the translated files, though it technically won't hurt too much.

kmeleta (commented):
Just jumping on some quick feedback. Approach seems reasonable enough to me.

assets/section-image-banner.css
eugenekasimov: that's a good call 👆

kjellr: Yeah, this is the tricky bit. I'm not sure how, but we might consider keeping the CSS in place for existing merchants somehow.

kjellr: Could we do the reverse? Add a setting like "Use white text and buttons when container is not displayed." We'd want it unchecked by default, but checked automatically when folks upgrade?

It's not all that different, but I think it reinforces that the default (expected) behavior would be that their color schemes always work, and that checking a box would override that.

eugenekasimov: Sounds like a potential plan. But we still need to keep in mind that if we are going to force i.e `inverse` color scheme, it may not have white color for text/button if it was changed by a merchant previously.

kjellr: If we do this plan, any existing merchants who upgrade don't need their color scheme changed at all, right? Since they'll have the box checked, and will keep the forced white color scheme. And any new merchants will just have Dawn's default schemes, so inverse should work fine. Or am I missing a detail?

kmeleta: Part of me actually likes that adding a setting would keep the shortcut to white text around. It wasn't perfect logic or UX but the original decision to use white still has _some_ merit in a lot of cases. Not to say I at all love the idea of having to add a setting for essentially deprecated functionally, of course 😅 

If we go this route we should seek opinions about the wording. I wish there was a concise way to kind of hint at why you might ever use such a setting. Not a real recommendation but, you know something like.. `[ ] Override color scheme with white for maximum contrast`

kmeleta: > We'd want it unchecked by default, but checked automatically when folks upgrade?

Hmm.. this will all just be thinking aloud more than a recommendation, but in the case of it being a white text setting that defaults to `false`, _I think_ we'd need a migration to add the new setting entry with the desired value (`true`) to upgrading themes.

If the schema default for the new setting was `true` however..
- Upgrading shops would automatically get the correct result because of the absence of the setting in their templates
- New theme installs could receive preset sections with the setting `false` if we add them, but..
- Any newly added (non-preset) sections would still fallback the schema default (`true`)
- In a subsequent release maybe, finally change the schema default to `false`

Does that train of thought make sense?

kjellr: It does. The key objective here is that we want new merchants to have their color scheme text used by default. It sounds like with that plan, we'd get there, but it would take a couple releases. If it's possible to get there in a single release, that would be preferable. But I'd totally defer to someone with more knowledge of our upgrade/migration capabilities — I'm not sure what we can and can't do.

kjellr:
Closing in favor of #2668 for now.

kjellr:
After some offline discussions, we're reopening this one. Should be ready for more dev reviews, and it needs [a fix for this issue](https://github.com/Shopify/dawn/pull/2663#discussion_r1199269637).

eugenekasimov (commented):
I can see that two things left to fix:

1. Remove translations from all locales files. 
2. Address the issue when buttons on mobile inherit background colors from active container on desktop. 

The rest looks good to me.

assets/section-image-banner.css
kjellr: I pushed 1c9af8e, which makes this work again. I _think_ it's fine to have the background be transparent in all cases like that, but let me know if you notice anything unexpected.

eugenekasimov: hm, i worked for me locally, but it's not working here 😅🤔

eugenekasimov: I pulled your latest changes from this PR and works on my local computer. I guess some magic is happening with connection between github and our demo store. If I go to the editor and check the code there I can see that CSS file hasn't been updated 
Screenshot 2023-05-25 at 11 05 03 AM

eugenekasimov: I reconnected the editor and updated the link in the description to the PR. Now it looks everything working good.

kmeleta (commented):
Looking good. Just found one remaining discrepancy.

assets/section-image-banner.css
```diff
@@ -220,9 +217,7 @@
     box-shadow: none;
   }
 
-  .banner:not(.banner--mobile-bottom) .button--secondary {
-    --color-button: 255, 255, 255;
-    --color-button-text: 255, 255, 255;
+  .banner .button--secondary {
     --alpha-button-background: 0;
```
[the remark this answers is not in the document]
kjellr: Thanks! Fixed in 78dacd0.

assets/section-image-banner.css
kjellr: Just to close the loop here officially, we're ok with this being a visual change on merchant storefronts since they'll have a chance to review. Going to mark this one as resolved as a result.

eugenekasimov (approved):

melissaperreault (approved):

kmeleta (approved):
:shipit:

PR #2672, opened 2023-05-25
Use live region to announce recipient form

fredma:
### PR Summary: 

This PR implements support for accessibility by announcing the recipient form when the checkbox is ticked/unticked on the gift card product page

Fixes https://github.com/Shopify/core-issues/issues/55523

The [original plan](https://github.com/Shopify/dawn/pull/2458) was to use `aria-expanded` and `aria-controls` but those are not supported for a `checkbox` role. To make it work, we have to use a supported role e.g. `button` but to align with UX design, we are keeping the checkbox as planned.

So the alternative is now using a live region to announce the form when the checkbox is ticked/unticked

### Decision log

|---|---|---|---|---|
| 1 | use live region  |  `aria-expanded` | `aria-expanded` is not supported for `checkbox` role  |   |

No impact on themes unless a11y feature is turned on.
See [video](https://screenshot.click/26-15-0kt82-8gqj8.mp4)

### Testing steps/scenarios

- [ ] Load a gift card
- [ ] Turn on a11y feature e.g. VoiceOver on Mac
- [ ] Focus on the checkbox, tick and check the a11y prompt output

### Demo links

- [Store](https://fred-test-gift-card.myshopify.com/products/gcp)

### Checklist


snippets/gift-card-recipient-form.liquid
```diff
@@ -74,6 +74,15 @@
         {%- endif -%}
       </ul>
     </div>
+    <p
+      id="Recipient-fields-live-region-{{ section.id }}"
+      class="visually-hidden"
+      aria-live="polite"
```
ludoboludo, line 80: I don't believe this is needed. From the mdn docs about `role="status"`:

snippets/gift-card-recipient-form.liquid
```diff
@@ -74,6 +74,15 @@
...
+      aria-hidden="true"
```
ludoboludo, line 81: Here I think we could potentially approach it differently. What you have now is a toggling between hidden and not hidden. 
We could instead, have this `p` always there, visually hidden the way you have it and use the JS, not to toggle `aria-hidden`, but to inject the text. 

Because it's a `role="status"`, anytime something is added in that container, it will announce it if I'm not mistaking. 

For the strings, you can add them in `theme.liquid` like we do for: 
```javascript
      window.accessibilityStrings = {
        imageAvailable: `{{ 'products.product.media.image_available' | t: index: '[index]' }}`,
        shareSuccess: `{{ 'general.share.success_message' | t }}`,
        pauseSlideshow: `{{ 'sections.slideshow.pause_slideshow' | t }}`,
        playSlideshow: `{{ 'sections.slideshow.play_slideshow' | t }}`,
      };
```

This way it follows the pattern we've used in other part of the theme (eg: SKU on product page)

ludoboludo (commented):
Looking goo 👍 I'd just suggest a potential different approach but in the end it seem to work either way 🙂

mzarud (approved):
Looked this over and tested, all works as expected. 👍

Walla-dev:
Is it possible to add this functionality to a highly customized, older theme? I put the recipient form on a separate digital gift card product template and attached custom product properties to it, but that didn't seem to affect sending the gift card to the intended recipient. Would I have to make a custom app to accomplish this using the storefront API?

It seems native to the Shopify platform now since the "Gift Card Confirmation" notification now includes this liquid tag in the notification template: {% if gift_card.recipient %} - but what exactly do I need in place for it to occur?

I have both of these included in my theme now:
gift-card-recipient-form.liquid
recipient-form.js

Screen Shot 2023-09-26 at 2 45 49 PM

PR #2693, opened 2023-06-05
Add release/v10.0.0 branch fixes to main

ludoboludo:
We've added a few fixes in our `/release/10.0.0` branch that are yet to be merged back into `/main`. This aims to do so. 

We've had quite a few bug fixes but also some new stuff merged into `/main`. So let's be extra cautious here with what is being changed and make sure there is only what we need.


kmeleta (approved):
lgtm!

PR #2732, opened 2023-06-23
Volume pricing - price per item

eugenekasimov:
### PR Summary: 

This PR adds a price per unit next to quantity input field on PDP for products that have volume pricing. 

Fixes #0.

### Things I still need to work on here
- ~~Change hardcoded text to locales~~
- ~~Isolate some scope of JS that causes an errors in console now.~~ 
- ~~For some reason, when you switch to another variant and add to cart 2+ items, the qty input field is not reseting to the min qty.~~
- ~~Probably we need to add loading spinner~~

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

If a product has volume pricing you will see a price per item next to quantity input field on PDP. 

### Testing steps/scenarios

- [ ] Login as B2B
- [ ] Go to a product that has volume pricing and make sure there is a price per item next to  quantity input field
- [ ] Start changing qty in quantity input field back and forward and make sure the ppi changes as well
- [ ] Switch to another variant and make sure it still works
- [ ] Start adding an item to cart buy pressing add to cart button, make sure it updates ppi after you added something to cart
- [ ] Try different cart type. Switch to drawer cart. Add an item to cart and start changing qty in cart. Make sure ppi changes accordingly. 

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/140204081174/editor)

### Checklist


edmund-teh:
Sorry, I just realized you're still working on this @eugenekasimov!
I'm sure you're already aware of this bug: https://screenshot.click/23-49-qg18n-pk7v9.mp4

Please feel free to ignore me and continue working on this... 😆

eugenekasimov:
>

eugenekasimov:
> Sorry, I just realized you're still working on this @eugenekasimov! I'm sure you're already aware of this bug: https://screenshot.click/23-49-qg18n-pk7v9.mp4
> 
> Please feel free to ignore me and continue working on this... 😆

Nice catch, I fixed it.

scottgmeadows (approved):
UX is looking nice :)

sections/main-product.liquid
```diff
@@ -235,36 +235,60 @@
...
+                    assign curentQtyForVolumePricing = cart_qty | plus: product.selected_or_first_available_variant.quantity_rule.min
+
+                  -%}
+                  {%- if product.quantity_price_breaks_configured? -%}
+                    <div class='price-per-item'>
+                      {%- if curentQtyForVolumePricing  < volumePricingArray.last.minimum_quantity -%}
+                        <span>at</span>
+                        <span>{{ product.selected_or_first_available_variant.price | money_with_currency }}/ea</span>
+                      {%- else -%}
+                        {%- for price_break in volumePricingArray -%}
```
sofiamatulis, line 282:
```suggestion
                        {%- for price_break in volume_pricing_array -%}
```

sections/main-product.liquid
```diff
@@ -235,36 +235,60 @@
...
+                    </button>
+                  </quantity-input>
+                  {%- liquid
+                    assign volumePricingArray = product.selected_or_first_available_variant.quantity_price_breaks | sort: 'price'
+                    assign curentQtyForVolumePricing = cart_qty | plus: product.selected_or_first_available_variant.quantity_rule.min
+
+                  -%}
+                  {%- if product.quantity_price_breaks_configured? -%}
+                    <div class='price-per-item'>
+                      {%- if curentQtyForVolumePricing  < volumePricingArray.last.minimum_quantity -%}
```
sofiamatulis, line 278:
```suggestion
                      {%- if current_qty_volume_pricing  < volume_pricing_array.last.minimum_quantity -%}
```

sections/main-product.liquid
```diff
@@ -235,36 +235,60 @@
...
+                    />
+                    <button class="quantity__button no-js-hidden" name="plus" type="button">
+                      <span class="visually-hidden">
+                        {{- 'products.product.quantity.increase' | t: product: product.title | escape -}}
+                      </span>
+                      {% render 'icon-plus' %}
+                    </button>
+                  </quantity-input>
+                  {%- liquid
+                    assign volumePricingArray = product.selected_or_first_available_variant.quantity_price_breaks | sort: 'price'
```
sofiamatulis, line 272:
```suggestion
                    assign volume_pricing_array = product.selected_or_first_available_variant.quantity_price_breaks | sort: 'price'
```

sections/main-product.liquid
```diff
@@ -235,36 +235,60 @@
...
+                    assign curentQtyForVolumePricing = cart_qty | plus: product.selected_or_first_available_variant.quantity_rule.min
```
sofiamatulis, line 273:
```suggestion
                    assign current_qty_volume_pricing = cart_qty | plus: product.selected_or_first_available_variant.quantity_rule.min
```

assets/global.js
```diff
@@ -157,28 +157,88 @@ class QuantityInput extends HTMLElement {
     super();
     this.input = this.querySelector('input');
     this.changeEvent = new Event('change', { bubbles: true });
-
+    if (this.input.hasAttribute('data-volume-pricing')) {
+      this.isVolumePricing = this.querySelector('input[name="quantity"]').dataset.volumePricing;
```
sofiamatulis, line 161:
```suggestion
      this.isVolumePricing = this.querySelector('input[name="quantity"]').dataset.volumePricing === 'true';
```

assets/global.js
```diff
@@ -157,28 +157,88 @@ class QuantityInput extends HTMLElement {
...
     );
   }
 
   quantityUpdateUnsubscriber = undefined;
+  updatePricePerItemUnsubscriber = undefined;
 
   connectedCallback() {
     this.validateQtyRules();
     this.quantityUpdateUnsubscriber = subscribe(PUB_SUB_EVENTS.quantityUpdate, this.validateQtyRules.bind(this));
+    if (this.isVolumePricing === 'true') {
```
sofiamatulis, line 175:
```suggestion
    if (this.isVolumePricing) {
```

assets/global.js
```diff
@@ -157,28 +157,88 @@ class QuantityInput extends HTMLElement {
...
       this.quantityUpdateUnsubscriber();
     }
+    if (this.updatePricePerItemUnsubscriber) {
+      this.updatePricePerItemUnsubscriber();
+    }
   }
 
   onInputChange(event) {
     this.validateQtyRules();
+    if (this.isVolumePricing === 'true') this.updatePricePerItem();
```
sofiamatulis, line 194:
```suggestion
    if (this.isVolumePricing) this.updatePricePerItem();
```

sofiamatulis (commented):
Added a few comments 👌 Looking good

Ill re-test when we have the spinner/translations and look a bit more at the code

assets/constants.js
```diff
@@ -5,4 +5,5 @@ const PUB_SUB_EVENTS = {
   quantityUpdate: 'quantity-update',
```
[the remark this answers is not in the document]
eugenekasimov: Okay, I expected this question :). I thought that it makes sense to add the spinner only for cases when we request updated data through API because that is the case when we potentially may have a delay. If you try to add to cart or using cart-drawer change qty in cart you will see the spinner.

Does it make sense?

sections/main-product.liquid
```diff
@@ -235,36 +235,79 @@
...
+                            >
+                              <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+                            </svg>
+                          </span>
+                        </span>
+                        <div class='price-per-item'>
+                          <span>{{ 'products.product.volume_pricing.at_price' | t }} </span>
+                          {%- if current_qty_for_volume_pricing  < volume_pricing_array.last.minimum_quantity -%}
+                            {%- assign varian_price = product.selected_or_first_available_variant.price | money_with_currency -%}
+                            <span>{{ 'products.product.volume_pricing.price_each' | t: price: varian_price }}</span>
```
sofiamatulis, line 296:
```suggestion
                            <span>{{variant_price }}/ each</span>
```

sections/main-product.liquid
```diff
@@ -235,36 +235,79 @@
...
+                              xmlns="http://www.w3.org/2000/svg"
+                            >
+                              <circle class="path" fill="none" stroke-width="6" cx="33" cy="33" r="30"></circle>
+                            </svg>
+                          </span>
+                        </span>
+                        <div class='price-per-item'>
+                          <span>{{ 'products.product.volume_pricing.at_price' | t }} </span>
+                          {%- if current_qty_for_volume_pricing  < volume_pricing_array.last.minimum_quantity -%}
+                            {%- assign varian_price = product.selected_or_first_available_variant.price | money_with_currency -%}
```
sofiamatulis, line 295:
```suggestion
                            {%- assign variant_price = product.selected_or_first_available_variant.price | money_with_currency -%}
```

eugenekasimov:
Okay team, let me summarize my thoughts. I also noticed that there is a bit of a misunderstanding of terminology, so I attached a quick explanation video. 

1. I'm not going to add the spinner when we change price by interacting with the input field because there is no network request.
2. We make the spinner every time you press the add to cart button or change the quantity in the cart because these make network requests. 
3. There is a known behaviour that we see the spinner each time we interact with the cart even when the price doesn't change. Although it may look not ideal and I personally agree with that I don't think that should be a blocker for this PR. Also, due to complexity, it's going to be quite time consuming to try to change this behaviour. I think we should address it in the future. 
https://screenshot.click/28-46-faf8q-a160g.mp4

sofiamatulis:
> I think I'm okay with this, given our tight timelines and that there is no network request when changing the price by interacting with the input field.
> 
> But, is this something that could be a fast follow? Or should it be a blocker? @sofiamatulis

@edmund-teh If UX is ok with this, I agree it can be a follow up. So the follow up behaviour would be:

- Only show spinner when price changes (by adding to cart **or** playing with the picker

MVP version
- Showing the spinner when buyer adds to cart (price changing or not(
- Not showing the spinner at all when the price changes.

cc @nicklepine 

Would that match variant list? 🤔 Because variant list each time you click on the picker you add to cart or remove to cart (which is a network request). That means that the spinner shows every time

assets/product-info.js
```diff
@@ -59,7 +59,7 @@ if (!customElements.get('product-info')) {
 
       fetchQuantityRules() {
         if (!this.currentVariant || !this.currentVariant.value) return;
-        this.querySelector('.quantity__rules-cart .loading-overlay').classList.remove('hidden');
+        this.querySelectorAll('.quantity__rules-cart .loading-overlay').forEach((overlay) => overlay.classList.remove('hidden'));
```
KaichenWang, line 62: Showing loading spinner could also be handled by the new `price-per-item` component instead

KaichenWang (commented):
I think the overall logic is good! I would suggest moving your logic to a new custom HTML component. The `quantity-input` component is meant to be more generic and used throughout the theme and may not always need to display a price per item; so it would be good to separate the logic.

sections/main-product.liquid
```diff
@@ -235,36 +235,79 @@
...
+                      {% render 'icon-plus' %}
+                    </button>
+                  </quantity-input>
+                  {%- liquid
+                    assign volume_pricing_array = product.selected_or_first_available_variant.quantity_price_breaks | sort: 'price'
+                    assign current_qty_for_volume_pricing = cart_qty | plus: product.selected_or_first_available_variant.quantity_rule.min
+
+                  -%}
+                  {%- if product.quantity_price_breaks_configured? -%}
+                    <label class="price-per-item__label form__label" for="Price-per-item-{{ section.id }}">
```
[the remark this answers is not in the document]
eugenekasimov: We don't need it here 👍

locales/en.default.json
```diff
@@ -142,7 +142,9 @@
       "volume_pricing": {
         "title": "Volume Pricing",
         "note": "Volume pricing available",
-        "minimum": "{{ quantity }}+"
+        "minimum": "{{ quantity }}+",
+        "at_price": "at",
```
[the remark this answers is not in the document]
eugenekasimov: Honestly, I couldn't figure a better way. We don't want to add "at" to "price_each": "{{ price }}/ea" because as far as I understand we use "price_each" for volume-pricing and we don't want to see "at" there 🤔
Screenshot 2023-06-29 at 10 17 13 AM

assets/global.js
```diff
@@ -157,28 +157,107 @@ class QuantityInput extends HTMLElement {
...
+    const volumePricing = document.querySelector('volume-pricing[id^="Volume-"]');
+    this.qtyPricePairs = [];
+
+    volumePricing.querySelectorAll('li').forEach(li => {
+      const qty = parseInt(li.querySelector('span:first-child').textContent);
+      const price = (li.querySelector('span:not(:first-child):last-child').textContent);
+      this.qtyPricePairs.push([qty, price]);
+    });
+
+    this.qtyPricePairs.reverse();
```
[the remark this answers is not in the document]
eugenekasimov: That's a good idea. Thanks.

assets/global.js
```diff
@@ -206,6 +205,103 @@ class QuantityInput extends HTMLElement {
 
 customElements.define('quantity-input', QuantityInput);
 
+class PricePerItem extends HTMLElement {
```
sofiamatulis, line 208: Can you move this to its own file? We wont need it for every page. This way we dont load it every time

assets/product-form.js
```diff
@@ -40,6 +40,8 @@ if (!customElements.get('product-form')) {
         }
         config.body = formData;
 
+        publish(PUB_SUB_EVENTS.cartUpdateStarted);
```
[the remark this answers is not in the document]
eugenekasimov: That how it was previously. Kai suggested to move it to price-per-item component just to keep all about price per item in one place. I agree with that approach, it makes more sense to me. I don't see a big difference and have a strong preference on it. If everyone is okay I can move this add/remove to the sources of the events.

locales/en.default.json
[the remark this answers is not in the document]
KaichenWang: They can be 2 different strings
```
"price_at_each": "at {{ price }}/ea",
"price_each": "{{ price }}/ea"
```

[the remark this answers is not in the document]
eugenekasimov: That's not actually working for me. The way I update the price per item is I use an array in JS where I add data from liquid which already looks like `$100.00 CAD/ea`.

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+        let pricePerItem = pair[1];
+        const pricePerItemCurrent = document.querySelector('.price-per-item span:last-child');
+        pricePerItemCurrent.innerHTML = pricePerItem;
+        break;
+      }
+    }
+  }
+
+  updateInputOnProductPage() {
+    this.input.value = Number(this.input.min);
```
KaichenWang, line 70: Why is this needed?

Is `this.input` the quantity input? Wouldn't this mess with the `min` set by the `setQuantityBoundries()` function?

sections/main-product.liquid
```diff
@@ -235,36 +236,75 @@
...
+                      <span class="visually-hidden">
+                        {{- 'products.product.quantity.decrease' | t: product: product.title | escape -}}
+                      </span>
+                      {% render 'icon-minus' %}
+                    </button>
+                    <input
+                      class="quantity__input"
+                      type="number"
+                      name="quantity"
+                      data-variant-id="{{ product.selected_or_first_available_variant.id }}"
```
KaichenWang, line 251: `data-variant-id` and  `data-volume-pricing` could be moved to the `<price-per-item>` element?

assets/global.js
```diff
@@ -1034,6 +1033,10 @@ class VariantSelects extends HTMLElement {
         const volumePricingSource = html.getElementById(
           `Volume-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`
         );
+
+        const pricePerItemDestination = document.querySelector(`price-per-item .price-per-item`);
```
KaichenWang, line 1037: We need to query by a unique id. Otherwise we'll run into issues with quick buy, because it's possible that multiple `price-per-item` elements are on the page.

See the examples in the same file that leverage `this.dataset.originalSection` and `this.dataset.section`

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+  updatePricePerItem(updatedCartQuantity) {
+    this.getVolumePricingArray();
+
+    const enteredQty = parseInt(this.input.value);
+    this.currentQtyForVolumePricing = this.getCartQuantity(updatedCartQuantity) + enteredQty;
+
+    for (let pair of this.qtyPricePairs) {
+      if (this.currentQtyForVolumePricing >= pair[0]) {
+        let pricePerItem = pair[1];
+        const pricePerItemCurrent = document.querySelector('.price-per-item span:last-child');
```
KaichenWang, line 62:
```suggestion
        const pricePerItemCurrent = this.querySelector('.price-per-item span:last-child');
```
Also, could we give the element a class instead of querying for generic `span:last-child`?

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
+class PricePerItem extends HTMLElement {
+  constructor() {
+    super();
+    this.input = this.querySelector('input');
```
KaichenWang, line 4: Which input are we trying to query for here? I would try to query for something less generic. I think you could add a unique id to the input in Liquid and use that to query using `document.querySelector`

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+    this.variantId = this.querySelector('input[name="quantity"]').dataset.variantId;
```
KaichenWang, line 5: Is this `this.querySelector('input[name="quantity"]')` the same element as `this.input`?

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+    this.input.addEventListener('change', this.onInputChange.bind(this));
```
KaichenWang, line 6: We don't have to do this in this PR, but it would be better to add your event listener to an pub/sub. The quantity input publishes an event when its input value changes and the `price-per-item` can subscribe to that event.

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+
+  updateInputOnProductPage() {
+    this.input.value = Number(this.input.min);
+  }
+
+  getCartQuantity(updatedCartQuantity) {
+    if (updatedCartQuantity || updatedCartQuantity === 0) {
+      return updatedCartQuantity;
+    } else {
+      const cartQuantity = parseInt(document.querySelector('input[name="quantity"]').dataset.cartQuantity);
```
KaichenWang, line 77: Which input are we querying for here? There could be multiple inputs that match this on the page.

assets/price-per-item.js
```diff
@@ -0,0 +1,96 @@
...
+      return cartQuantity;
+    }
+  }
+
+  getVolumePricingArray() {
+    const volumePricing = document.querySelector('volume-pricing[id^="Volume-"]');
```
KaichenWang, line 83: 1 - There could be multiple volume-pricing tables on the page when quick buy is enabled.

2 - We're re-running this every time the quantity input value changes, which is expensive. We should cache this data somehow or obtain if some other way.

sections/main-product.liquid
```diff
@@ -235,36 +236,75 @@
...
-                      max="{{ product.selected_or_first_available_variant.quantity_rule.max }}"
-                    {% endif %}
-                    step="{{ product.selected_or_first_available_variant.quantity_rule.increment }}"
-                    value="{{ product.selected_or_first_available_variant.quantity_rule.min }}"
-                    form="{{ product_form_id }}"
-                  />
-                  <button class="quantity__button no-js-hidden" name="plus" type="button">
-                    <span class="visually-hidden">
-                      {{- 'products.product.quantity.increase' | t: product: product.title | escape -}}
+                <price-per-item>
```
KaichenWang, line 239: I don't think you need to nest the `quantity-input` inside the `price-per-item`. You could keep them separate and query the quantity input from the DOM using a unique id.

assets/product-form.js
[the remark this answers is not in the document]
sofiamatulis: cc @KaichenWang thoughts?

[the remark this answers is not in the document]
KaichenWang: I think this is fine. Without the `cartUpdateStarted` event, the `price-per-item` component is unaware of when an item is being added to the cart and won't know when to start showing its loader.

assets/price-per-item.js
eugenekasimov: The input already has `id="Quantity-{{ section.id }}"` I added `data-section-id="{{ section.id }}"` to the component and now I can query it `this.input = document.getElementById(`Quantity-${this.dataset.sectionId}`);`

assets/product-form.js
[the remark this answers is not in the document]
sofiamatulis: Can't we just add the loader to the `product-form` when the request is sent? Im just trying to be mindful of not creating new events if we dont need to

assets/price-per-item.js
eugenekasimov: Actually I found a bug that was happening when you switched variant because I never actually updated variant ID in my logic. Fixed.

assets/price-per-item.js
eugenekasimov: 1 - I cannot imagine how this is possible.

assets/product-form.js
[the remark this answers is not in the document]
eugenekasimov: We can but in this case we will also need to add it to cart.js. In my opinion a new event is not that harmful and we can use less code.

locales/en.default.json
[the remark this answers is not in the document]
sofiamatulis: @eugenekasimov Wouldnt creating two strings like what Kai suggested work?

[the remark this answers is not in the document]
eugenekasimov: not really, I explained why in my previous post 👆

assets/price-per-item.js
```diff
@@ -0,0 +1,100 @@
...
+        break;
+      }
+    }
+  }
+
+  getCartQuantity(updatedCartQuantity) {
+    if (updatedCartQuantity || updatedCartQuantity === 0) {
+      return updatedCartQuantity;
+    } else {
+      const cartQuantity = parseInt(this.input.dataset.cartQuantity);
```
sofiamatulis, line 81:
```suggestion
      return parseInt(this.input.dataset.cartQuantity);
```

assets/price-per-item.js
```diff
@@ -0,0 +1,100 @@
...
+      return cartQuantity;
```
sofiamatulis, line 82:
```suggestion
```

locales/en.default.json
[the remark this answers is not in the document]
eugenekasimov: Okay, Kai suggested me another approach to keep my current logic and it seems working.

locales/en.default.json
```diff
@@ -144,6 +144,7 @@
         "title": "Volume Pricing",
         "note": "Volume pricing available",
         "minimum": "{{ quantity }}+",
+        "price_at_each": "at {{ price }}/ea",
```
sofiamatulis, line 147: Can we request translations now? 👀

assets/price-per-item.js
```diff
@@ -0,0 +1,109 @@
...
+
+      updatePricePerItem(updatedCartQuantity) {
+        // updatedCartQuantity is undefined when qty is updated on product page. We need to sum entered qty and current qty in cart.
+        if (updatedCartQuantity === undefined) {
+          const enteredQty = parseInt(this.input.value);
+          this.currentQtyForVolumePricing = this.getCartQuantity(updatedCartQuantity) + enteredQty;
+        // updatedCartQuantity is not undefined when qty is updated in cart. We need to sum qty in cart and min qty for product.
+        } else {
+          this.currentQtyForVolumePricing = this.getCartQuantity(updatedCartQuantity) + parseInt(this.input.min);
+        }
```
sofiamatulis, line 75:
```suggestion
 const enteredQty = parseInt(this.input.value);

this.currentQtyForVolumePricing = updatedCartQuantity ? this.getCartQuantity(updatedCartQuantity) + parseInt(this.input.min) : this.getCartQuantity(updatedCartQuantity) + enteredQty;
```

assets/price-per-item.js
```diff
@@ -0,0 +1,109 @@
...
+          }
+        }
+      }
+
+      getCartQuantity(updatedCartQuantity) {
+        if (updatedCartQuantity || updatedCartQuantity === 0) {
+          return updatedCartQuantity;
+        } else {
+          return parseInt(this.input.dataset.cartQuantity);
+        }
```
sofiamatulis, line 92:
```suggestion
const newCartQty = if (updatedCartQuantity || updatedCartQuantity === 0) ? updatedCartQuantity : parseInt(this.input.dataset.cartQuantity);

return newCartQty
```

assets/price-per-item.js
sofiamatulis: I think you could try

`this.currentQtyForVolumePricing = updatedCartQuantity === undefined ? (this.getCartQuantity(updatedCartQuantity) + parseInt(this.input.min)) : (this.getCartQuantity(updatedCartQuantity) + enteredQty);`

eugenekasimov: this should work :)

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+
+        // updatedCartQuantity is undefined when qty is updated on product page. We need to sum entered qty and current qty in cart.
+        // updatedCartQuantity is not undefined when qty is updated in cart. We need to sum qty in cart and min qty for product.
+        this.currentQtyForVolumePricing = updatedCartQuantity === undefined ? this.getCartQuantity(updatedCartQuantity) + enteredQty : this.getCartQuantity(updatedCartQuantity) + parseInt(this.input.min);
+
+        for (let pair of this.qtyPricePairs) {
+          if (this.currentQtyForVolumePricing >= pair[0]) {
+            let pricePerItem = pair[1];
+            const pricePerItemCurrent = document.querySelector(`price-per-item[id^="Price-Per-Item-${this.dataset.sectionId}"] .price-per-item`);
+            pricePerItemCurrent.innerHTML = pricePerItem;
```
sofiamatulis, line 78:
```suggestion
            pricePerItemCurrent.innerHTML = pair[1];
```

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+      updatePricePerItem(updatedCartQuantity) {
+        const enteredQty = parseInt(this.input.value);
+
+        // updatedCartQuantity is undefined when qty is updated on product page. We need to sum entered qty and current qty in cart.
+        // updatedCartQuantity is not undefined when qty is updated in cart. We need to sum qty in cart and min qty for product.
+        this.currentQtyForVolumePricing = updatedCartQuantity === undefined ? this.getCartQuantity(updatedCartQuantity) + enteredQty : this.getCartQuantity(updatedCartQuantity) + parseInt(this.input.min);
+
+        for (let pair of this.qtyPricePairs) {
+          if (this.currentQtyForVolumePricing >= pair[0]) {
+            let pricePerItem = pair[1];
```
sofiamatulis, line 76:
```suggestion
```

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+            const pricePerItemCurrent = document.querySelector(`price-per-item[id^="Price-Per-Item-${this.dataset.sectionId}"] .price-per-item`);
+            pricePerItemCurrent.innerHTML = pricePerItem;
+            break;
+          }
+        }
+      }
+
+      getCartQuantity(updatedCartQuantity) {
+        const newCartQty = (updatedCartQuantity || updatedCartQuantity === 0) ? updatedCartQuantity : parseInt(this.input.dataset.cartQuantity);
+        return newCartQty
```
sofiamatulis, line 86:
```suggestion
return (updatedCartQuantity || updatedCartQuantity === 0) ? updatedCartQuantity : parseInt(this.input.dataset.cartQuantity);
```

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+          this.getVolumePricingArray();
+        });
+        this.cartUpdateStartedUnsubscriber = subscribe(PUB_SUB_EVENTS.cartUpdateStarted, (event) => {
+          if (event.variantId !== this.variantId) return;
+          this.querySelector('.loading-overlay').classList.remove('hidden');
+        });
+
+        this.updatePricePerItemUnsubscriber = subscribe(PUB_SUB_EVENTS.cartUpdate, (response) => {
+          this.querySelector('.loading-overlay').classList.add('hidden');
+          // Item was added to cart via product page
```
sofiamatulis, line 31: Do we need these comments? 👀

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+      getCartQuantity(updatedCartQuantity) {
+        const newCartQty = (updatedCartQuantity || updatedCartQuantity === 0) ? updatedCartQuantity : parseInt(this.input.dataset.cartQuantity);
+        return newCartQty
+      }
+
+      getVolumePricingArray() {
+        const volumePricing = document.getElementById(`Volume-${this.dataset.sectionId}`);
+        this.qtyPricePairs = [];
+
+        volumePricing.querySelectorAll('li').forEach(li => {
```
sofiamatulis, line 93: I wonder if we should target for the class instead of `li`? 🤔

sofiamatulis (approved):
Works well! Added small nitpicks

assets/cart.js
```diff
@@ -6,6 +6,8 @@ class CartRemoveButton extends HTMLElement {
       event.preventDefault();
       const cartItems = this.closest('cart-items') || this.closest('cart-drawer-items');
       cartItems.updateQuantity(this.dataset.index, 0);
+
```
sofiamatulis, line 9: This will be the wrong price when JS is off (variant change and qty amount):

https://screenshot.click/06-42-ojxav-eyfja.png

I think we hide it for now since we show it in the dropdown the price (the same way we hide the table for now and well rethink it in no JS)

cc @KaichenWang @edmund-teh

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
+if (!customElements.get('price-per-item')) {
+  customElements.define(
+    'price-per-item',
+    class PricePerItem extends HTMLElement {
+      constructor() {
+        super();
+        this.input = document.getElementById(`Quantity-${this.dataset.sectionId}`);
+        this.variantId = document.getElementById(`Price-Per-Item-${this.dataset.sectionId}`).dataset.variantId;
```
KaichenWang, line 8: I think this is equivalent: 

```suggestion
        this.variantId = this.dataset.variantId;
```

assets/price-per-item.js
```diff
@@ -0,0 +1,103 @@
...
+
+        this.getVolumePricingArray();
+      }
+
+      updatePricePerItemUnsubscriber = undefined;
+      cartUpdateStartedUnsubscriber = undefined;
+      variantIdChangedUnsubscriber = undefined;
+
+      connectedCallback() {
+        // Update varinatId if variant is switched on product page
```
KaichenWang, line 19:
```suggestion
        // Update variantId if variant is switched on product page
```

KaichenWang (commented):
I'm not sure it's caused by this PR, but the Quick Add is causing a JS error. I'm still trying to figure out why this is happening.

Screenshot 2023-07-06 at 5 49 15 PM 1

assets/price-per-item.js
eugenekasimov: I find them helpful at least now when we are going to work with this code for featured product for example.

eugenekasimov:
> I'm not sure it's caused by this PR, but the Quick Add is causing a JS error. I'm still trying to figure out why this is happening.
> 
> Screenshot 2023-07-06 at 5 49 15 PM 1

I found it and fixed. I missed `"` in one place.

Issue #2742, opened 2023-06-26
Image with Text - Ambient Movement Animation Broken


kimberlyoleiro:
**Expected behaviour:**
Image with ambient movement fills full container width & height, and reflects ambient movement animation

**Actually behaviour:** 
Image is stuck, without movement, in bottom right quadrant of the container

**Version:**
Dawn v10

https://admin.shopify.com/store/kimberly-oleiro/themes/129112375352/editor

![Image](https://user-images.githubusercontent.com/107572834/248935297-27669d67-caf8-44e2-8660-9d6ab482d1c9.png)

**Relates to:**
https://github.com/Shopify/dawn/pull/2385

PR #2744, opened 2023-06-27
[Image with text] Fix ambient movement and zoom in on placeholder

ludoboludo:
### PR Summary: 

Fixes the ambient background animation on the placeholder image

Fixes https://github.com/Shopify/dawn/issues/2742

Changed the left, top and transform values when the ambient and zoom in animation is enabled. 

### Testing steps/scenarios

- [ ] Add an image with text section
- [ ] Set the image behaviour to be ambient or zoomed in and confirm it's working as expected (the position of the SVG is looking good)

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/140174688278/editor)

### Checklist

### DIFF
--- a/assets/component-image-with-text.css
+++ b/assets/component-image-with-text.css
@@ -74,7 +74,7 @@
   fill: currentColor;
 }
 
-.image-with-text__media--placeholder.animate--ambient > svg {
+.image-with-text__media--placeholder:is(.animate--ambient, .animate--zoom-in) > svg {
   top: 0;
   left: 0;
   transform: translate(0);
--- a/assets/component-image-with-text.css
+++ b/assets/component-image-with-text.css
@@ -74,6 +74,12 @@
   fill: currentColor;
 }
 
+.image-with-text__media--placeholder.animate--ambient > svg {
+  top: 0;
+  left: 0;
+  transform: translate(0);
+}
+
 .image-with-text__content {
   align-items: flex-start;
   display: flex;


kimberlyoleiro:
Looks good to me https://hllvkum3xlhy2x07-45840990230.shopifypreview.com

assets/component-image-with-text.css
```diff
@@ -74,7 +74,7 @@
   fill: currentColor;
 }
 
-.image-with-text__media--placeholder.animate--ambient > svg {
+.image-with-text__media--placeholder:is(.animate--ambient, .animate--zoom-in) > svg {
   top: 0;
```
lougoncharenko, line 78: Asking question so I can enhance my learning: why did you use :is() vs. :where() ?

ludoboludo: The main difference is in specificity.

`:is()`  will use the selectors within it to count towards the specificity of the overall selector.
`:where()` will default to a specificity of `0`. 

So in our case, because the selector used to be `.class.anotherClass` then it would have a specificity count of `2`. By using `:is()` I make sure I keep that same specificity. 

If I had used `:where()` then the specificity would be lower. 

<details>
<summary>[Here](https://specificity.keegan.st/) is a specificity calculator and its result: </summary>


</details>

[Here is a useful breakdown](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) 🙂

lougoncharenko: Oh wow! This is a very helpful diagram! Thank you so much for taking the time to explain the difference to me !

eugenekasimov (approved):
👍

assets/component-image-with-text.css
```diff
@@ -74,6 +74,12 @@
   fill: currentColor;
 }
 
+.image-with-text__media--placeholder:is(.animate--ambient, .animate--zoom-in) > svg {
+  top: 0;
+  left: 0;
+  transform: translate(0);
```
kmeleta, line 80: Did you find you needed to set the initial translate to 0?

PR #2746, opened 2023-06-27
Update cart unit price to reflect of line item

sofiamatulis:
### PR Summary: 

Original ticket: https://github.com/orgs/Shopify/projects/6397/views/5?pane=issue&itemId=30821952

it makes sense to use the discounted unit price in cart (currently we don't). So we can use `line_item.unit_price` instead of `line_item.variant.unit_price`

UX: The unit price needs to show cost clarity of how much it will cost per unit values - the strike-through price is not important at this line level because we already display it on the total price with discount in a larger font size.

### Demo links

Test cart page and cart drawer

- [Editor](https://os2-demo.myshopify.com/admin/themes/140178685974/editor)

### DIFF
--- a/sections/main-cart-items.liquid
+++ b/sections/main-cart-items.liquid
@@ -203,15 +203,15 @@
                       {%- if item.variant.available and item.unit_price_measurement -%}
                         <div class="unit-price caption">
                           <span class="visually-hidden">{{ 'products.product.price.unit_price' | t }}</span>
-                          {{ item.variant.unit_price | money }}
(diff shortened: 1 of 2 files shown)


eugenekasimov (approved):

KaichenWang (commented):
Code changes look good! 

Will this cause a change in the unit price value displayed in the storefront? If so, is there an example where we can observe a change in the cart (we could also add this to scenarios we need to test for)?

KaichenWang (approved):

PR #2747, opened 2023-06-27
Add vol pricing to prod card

sofiamatulis:
### PR Summary: 

Add a note to products that contain vol pricing (prod card)

[Ticket](https://github.com/orgs/Shopify/projects/6397/views/5?pane=issue&itemId=30826314)

### Testing steps/scenarios

- [ ] Product with image
- [ ] Product without image
- [ ] Product with vol pricing
- [ ] Product without vol pricing

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/140178817046/editor)

### DIFF
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -200,6 +200,11 @@
             {%- endif -%}
 
             {% render 'price', product: card_product, price_class: '' %}
+            {%- if card_product.quantity_price_breaks_configured? -%}
+              <div class="card__information--volume-pricing-note">
+                <span class="caption">{{ 'products.product.volume_pricing.note' | t }}</span>
+              </div>
+            {%- endif -%}
           </div>
         </div>
         {%- if show_quick_add -%}
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -201,7 +201,7 @@
 
             {% render 'price', product: card_product, price_class: '' %}
             {%- if card_product.quantity_price_breaks_configured? -%}
-              <div class="card__information--volume-pricing-note">
+              <div class="card__information-volume-pricing-note">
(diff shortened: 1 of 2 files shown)


snippets/card-product.liquid
```diff
@@ -200,6 +200,11 @@
             {%- endif -%}
 
             {% render 'price', product: card_product, price_class: '' %}
+            {%- if card_product.quantity_price_breaks_configured? -%}
+              <div class="card__information--volume-pricing-note">
+                <span class="caption">{{ 'products.product.volume_pricing.note' | t }}</span>
```
eugenekasimov, line 205: this locales was added in different PR, is it right?

eugenekasimov (approved):

snippets/card-product.liquid
sofiamatulis: Yes I branched off of that :)

scottgmeadows (approved):

snippets/card-product.liquid
```diff
@@ -200,6 +200,11 @@
             {%- endif -%}
 
             {% render 'price', product: card_product, price_class: '' %}
+            {%- if card_product.quantity_price_breaks_configured? -%}
+              <div class="card__information--volume-pricing-note">
```
KaichenWang, line 204: Regarding the selector naming, why is this selector a modifier of `card__information`? What is the relationship between this element and the other `card__information` elements?

KaichenWang (approved):
Looks good!

One UX question: we may want to reduce the line-height?
Screenshot 2023-06-29 at 9 41 34 AM

snippets/card-product.liquid
KaichenWang: If we're using BEM convention here, a modifier (`card__information--volume-pricing-note`) should not be applied to a child of `card__information`

PR #2754, opened 2023-06-28
[Variant List] Display error messages

### LINKED ISSUE #2750: FE - Add errors
Error should show below the variant row like cart: https://screenshot.click/02-48-tzmcv-fmvn2.png

Error should come from the BE and not be hardcoded (like it currently is in cart - cart to be updated next)

Clamping should not occur (should be updated in Cart Next)

Generic errors should be where the success message is: https://screenshot.click/26-11-d65vi-l1kq8.png

KaichenWang:

Fixes #2750.

Enable errors returned by the server to be displayed for variant list. 

Most of the JS logic for variant list was adapted from the cart. 

Changes:
1. For variant list item errors, some selectors needed to be updated.
2. For global variant list errors, they've been moved in the DOM.
3. Some additional refactoring.

### Decision log

|---|---|---|---|---|
| 1 | There are some existing inconsistencies in the codebase regarding `id` selectors. Decided to keep the existing conventions for each respective id. | Refactor all ids to the same convention. | Out of scope for this PR and would increase scope of tophatting dramatically. | Inconsistencies in the codebase. |
| 2 | When the cart is updated due to a variant list quantity change, it’s possible that the “X items added” message is inaccurate (for example, when the desired quantity exceeds max available). Decision is to hide the “X items added” message altogether in those cases. | Find a way to calculate the actual amount added to cart. | Out of scope for this PR. This should be fixed in the main `variantlist` branch. | No message to notify the buyer of the amount of items added in some situations. (This may be ok, because an error is shown) |

### Testing steps/scenarios

| Test | Steps |
|---|---|
| Global variant list error | Simulate error scenario by blocking cart requests |
| Variant list item error | Try to add quantity beyond the maximum available. Or enter blank in the quantity input. |

https://github.com/Shopify/dawn/assets/3710548/764ad4db-c184-435a-bfd8-cc98ff9059c4

### Demo links

- [Store](https://os2-demo.myshopify.com/?preview_theme_id=140188188694)
- [Editor](https://os2-demo.myshopify.com/admin/themes/140188188694/editor)

### Checklist

### DIFF
--- a/assets/variant-list.css
+++ b/assets/variant-list.css
@@ -218,7 +218,7 @@ variant-list-remove-button {
   }
 
   .variant-list__message,
-  .variant-list-errors {
+  .variant-list-error {
     align-self: flex-end;
     padding-bottom: 1rem;
   }
@@ -268,10 +268,20 @@ variant-list-remove-button .icon-remove {
   width: 1.3rem;
 }
 
-.variant-list__message,
-.variant-list-errors {
+.variant-list__message {
   margin-top: 1rem;
   display: block;
+}
+
+.variant-list-error {
+  margin-top: 1rem;
+  display: flex;
+}
+
+.variant-list-error svg {
+  flex-shrink: 0;
+  width: 1.2rem;
+  margin-right: 0.7rem;
 }
 
 @media screen and (min-width: 750px) {
--- a/assets/variant-list.css
+++ b/assets/variant-list.css
@@ -45,10 +45,10 @@ variant-list .quantity__button {
     overflow-y: auto;
   }
 
-  .variant-item__quantity-wrapper--no-info {
+  .variant-item__quantity-wrapper--no-info,
+  .variant-item__error {
     padding-left: 4.9rem;
   }
-
 }
 
 .variant-list__table td,
@@ -215,6 +215,7 @@ variant-list-remove-button {
 
   .variant-list-total__column {
     display: flex;
+    flex-wrap: wrap;
   }
 
   .variant-list__message,
@@ -486,6 +487,10 @@ variant-list-remove-button:hover .icon-remove {
     display: flex;
     padding: 0.5rem;
   }
+
+  .variant-list-total__column {
+    width: 37%;
+  }
 }
 
 variant-list-remove-button {
@@ -504,10 +509,6 @@ variant-list-remove-button {
   .variant-list-buttons {
     display: flex;
     align-self: flex-start;
-  }
-
-  .variant-list-total__column {
-    width: 37%;
   }
 
   variant-list-remove-all-button {
--- a/assets/variant-list.css
+++ b/assets/variant-list.css
@@ -411,6 +411,12 @@ variant-list-remove-button:hover .icon-remove {
   .variant-item__inner--no-media {
     display: none;
   }
+  
+  .variant-list-total__column {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
 }
 
 .variant-item__error-text+svg {
--- a/assets/variant-list.css
+++ b/assets/variant-list.css
@@ -283,6 +283,8 @@ variant-list-remove-button .icon-remove {
   flex-shrink: 0;
   width: 1.2rem;
   margin-right: 0.7rem;
+  margin-top: 0.3rem;
+  align-self: flex-start;
 }
 
 @media screen and (min-width: 750px) {
--- a/assets/variant-list.css
+++ b/assets/variant-list.css
@@ -217,7 +217,8 @@ variant-list-remove-button {
     display: flex;
   }
 
-  .variant-list__message {
+  .variant-list__message,
+  .variant-list-errors {
     align-self: flex-end;
     padding-bottom: 1rem;
   }
@@ -267,7 +268,8 @@ variant-list-remove-button .icon-remove {
   width: 1.3rem;
 }
 
-.variant-list__message {
+.variant-list__message,
+.variant-list-errors {
   margin-top: 1rem;
   display: block;
 }
--- a/assets/variant-list.js
+++ b/assets/variant-list.js
@@ -234,23 +234,13 @@ class VariantList extends HTMLElement {
             this.getSectionInnerHTML(parsedState.sections[section.section], section.selector);
         }));
 
-        let message = '';
         let hasError = false;
 
-        if (action === this.actions.add) {
-          const updatedValue = parsedState.quantity ? parsedState.quantity : undefined;
-          if (parsedState.quantity !== parseInt(quantityElement.value)) {
-            this.updateError(updatedValue, id, message)
-            hasError = true;
-          }
-        } else {
-          const currentItem = parsedState.items.find((item) => item.variant_id === parseInt(id));
-          const updatedValue = currentItem ? currentItem.quantity : undefined;
-
-          if (items.length === parsedState.items.length && updatedValue !== parseInt(quantityElement.value)) {
-            this.updateError(updatedValue, id)
-            hasError = true;
-          }
+        const currentItem = parsedState.items.find((item) => item.variant_id === parseInt(id));
+        const updatedValue = currentItem ? currentItem.quantity : undefined;
+        if (items.length === parsedState.items.length && updatedValue !== parseInt(quantityElement.value)) {
+          this.updateError(updatedValue, id)
+          hasError = true;
         }
 
         const variantItem = document.getElementById(`Variant-${id}`);
@@ -315,8 +305,8 @@ class VariantList extends HTMLElement {
 
   }
 
-
-  updateError(updatedValue, id, message) {
+  updateError(updatedValue, id) {
+    let message = '';
     if (typeof updatedValue === 'undefined') {
       message = window.cartStrings.error;
     } else {
--- a/assets/variant-list.js
+++ b/assets/variant-list.js
@@ -216,7 +216,7 @@ class VariantList extends HTMLElement {
         const items = document.querySelectorAll('.variant-item');
 
         if (parsedState.description || parsedState.errors) {
-          quantityElement.value = quantityElement.getAttribute('value');
+          this.resetQuantityInput(id, quantityElement);
           if (parsedState.errors) {
             this.updateLiveRegions(id, parsedState.errors);
           } else {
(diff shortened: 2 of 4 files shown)


KaichenWang:
Known issue: Variant list update sometimes does not receive the same backend errors that are surfaced in the cart.

https://github.com/Shopify/dawn/assets/3710548/c8c86b1c-a82c-4bc3-abed-c2fc3281bfb5

KaichenWang:
Known issue: Theme check is failing due to missing translation string `products.product.product_variants`. This should be addressed in the base branch, not this PR.

edmund-teh:
> Here is a bug but it's likely coming from the PR you used as a base. I'll just leave it here anyway.
> 
> 1. In your test when you tried to add a crazy number for a variant in the variant list you got a message `You can only add 1000000 of this item to your cart.` I'm not sure where this number comes from but it has nothing in common with the available qty in stock. The real qty of the variant in your test is 1000. It seems that if you just input a number into the input field without touching "+"/"-" buttons the qty rule is not applied.
> 2. If you start interacting with qty by pressing "+"/"-" and go over the max available qty the qty will be reset to the max and you won't see any errors. Is it expected behaviour?
> 
> https://screenshot.click/28-09-1ydr1-ryv12.mp4

Nice catch with #1, @eugenekasimov. 

#2, I could be wrong, but I think this is where our clamping comes into play. I believe this is the expected behaviour... it should clamp to the max value you can input. @sofiamatulis, maybe you can confirm here?

assets/variant-list.css
```diff
@@ -201,7 +201,8 @@ variant-list-remove-button {
     display: flex;
```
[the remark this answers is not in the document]
sofiamatulis: cc @edmund-teh

assets/variant-list.js
```diff
@@ -283,12 +300,12 @@ class VariantList extends HTMLElement {
...
+    const variantItemError = document.getElementById(`Variant-list-item-error-${id}`);
+    if (variantItemError) variantItemError.querySelector('.variant-item__error-text').innerHTML = message;
 
     this.variantItemStatusElement.setAttribute('aria-hidden', true);
 
-    const cartStatus = document.getElementById('cart-live-region-text');
+    const cartStatus = document.getElementById('variant-list-live-region-text');
     cartStatus.setAttribute('aria-hidden', false);
 
     setTimeout(() => {
```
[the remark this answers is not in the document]
edmund-teh: @sofiamatulis @KaichenWang 

So, with this solution we would be moving away from inline error messaging? 

Do you think this would make for a bad experience on Mobile because the error messaging would be all the way at the bottom of the section (because the footer isn't sticky on Mobile)? 

Buyers would have to scroll down to the see the error message and then scroll back up to the variant they were interacting with.

Sorry, please correct me if I'm wrong on this one!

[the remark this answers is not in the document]
edmund-teh: @sofiamatulis my bad! I'm a bit slow this morning, sorry. 🤦‍♂️ 

Okay! This sounds good and all makes sense now. 
Can we match the same error styling with the red icon though? 
https://screenshot.click/Same-error-styling.png

cc: @KaichenWang

sections/variant-list.liquid
```diff
@@ -183,6 +184,7 @@
               <span class="variant-list__message-icon hidden">{%- render 'icon-checkmark' -%}</span>
               <span class="variant-list__message-text"></span>
             </span>
+            <span class="variant-list-errors caption-large"></span>
```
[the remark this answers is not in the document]
KaichenWang: Good point, screenreaders weren't announcing this. I've pushed an update that ensures the message is announced (even if the message is updated, but the text hasn't changed) - https://github.com/Shopify/dawn/pull/2754/commits/f7b2de4d9c4a83ef974d2cdb9411b3f1dc5f1b6c

KaichenWang:
> Are we able to change the color of the input border when we see an inline error message?
> https://screenshot.click/Error-Border-highlight.png
> [Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3453-154523&mode=design&t=9JMzRkOUJiY1f3Ss-4)

@edmund-teh afaik, this red error border isn't currently implemented in the cart? Ideally we'd do this for both the variant list items and cart items for consistency. So probably better to tackle this separately.

sofiamatulis:
> @edmund-teh afaik, this red error border isn't currently implemented in the cart? Ideally we'd do this for both the variant list items and cart items for consistency. So probably better to tackle this separately.

We also dont have this for the PDP add to cart

sofiamatulis:
> > @edmund-teh afaik, this red error border isn't currently implemented in the cart? Ideally we'd do this for both the variant list items and cart items for consistency. So probably better to tackle this separately.
> 
> We also dont have this for the PDP add to cart

Although we do have a different design for the qty picker in variant list.

edmund-teh:
> > Are we able to change the color of the input border when we see an inline error message?
> > https://screenshot.click/Error-Border-highlight.png
> > [Figma file here](https://www.figma.com/file/ov2FHR3uvfVzKqbVV33IPz/Variant-list-on-PDP?type=design&node-id=3453-154523&mode=design&t=9JMzRkOUJiY1f3Ss-4)
> 
> @edmund-teh afaik, this red error border isn't currently implemented in the cart? Ideally we'd do this for both the variant list items and cart items for consistency. So probably better to tackle this separately.

Fair— I'm okay with leaving this out for now and making it a fast-follow addition.

In speaking with UX stakeholders, I was told we should design variant list to act as a reference for cart (which should get some additional changes in the near future).

> "If we keep using old patterns as references, we’ll always be designing to a lowered quality floor."

edmund-teh (approved):

KaichenWang:
> In your test when you tried to add a crazy number for a variant in the variant list you got a message You can only add 1000000 of this item to your cart. I'm not sure where this number comes from but it has nothing in common with the available qty in stock. The real qty of the variant in your test is 1000. It seems that if you just input a number into the input field without touching "+"/"-" buttons the qty rule is not applied.

@eugenekasimov This should be fixed now

eugenekasimov (approved):
All works as expected. 

The only thing that grabbed my attention was if you try to enter blank in the quantity input you won't see any error and the min qty will be added to the cart. I'm not sure if it's a good behaviour, but it's not a blocker for this PR 👍

assets/variant-list.css
```diff
@@ -256,6 +257,17 @@ variant-list-remove-button .icon-remove {
...
+
+.variant-list-error svg {
+  flex-shrink: 0;
+  width: 1.2rem;
+  margin-right: 0.7rem;
+}
+
 @media screen and (min-width: 750px) {
   .variant-item .loading-overlay {
     padding-top: 2.5rem;
```
sofiamatulis, line 273: When there is the generic error, it still lets me add items to cart. Or at least it looks like it (qty isnt reset like the in line error)

Is that expected? https://screenshot.click/30-53-v9bx4-g7bt7.mp4

assets/variant-list.js
```diff
@@ -133,7 +133,7 @@ class VariantList extends HTMLElement {
         selector: '.shopify-section'
```
sofiamatulis, line 133: Buttons jump to the side on mobile: https://screenshot.click/30-53-hew2h-r4xmq.mp4

assets/variant-list.js
```diff
@@ -154,6 +154,9 @@ class VariantList extends HTMLElement {
       sections_url: window.location.pathname
     });
```
sofiamatulis, line 155: Some of the elements are moving once the error appears

https://screenshot.click/30-54-xbyxe-x6svz.mp4

assets/variant-list.css
```diff
@@ -256,6 +257,17 @@ variant-list-remove-button .icon-remove {
   display: block;
 }
 
+.variant-list-error {
```
sofiamatulis, line 260: Some screensizes the alignment is a bit off but I fixed that in the variantlist PR. Do have the latest changes? Want to ensure we are not introducing this
https://screenshot.click/30-05-a3yyt-dtyzg.png

sections/variant-list.liquid
```diff
@@ -134,6 +134,22 @@
               <span class="variant-list__message-icon hidden">{%- render 'icon-checkmark' -%}</span>
               <span class="variant-list__message-text"></span>
             </span>
+            <span class="variant-list-error">
+              <template>
```
sofiamatulis, line 138: Why are we adding template here? Not familiar with it 👀

sofiamatulis (commented):
Looks great 🎉 Added a few notes

assets/variant-list.css
KaichenWang: Updated - https://github.com/Shopify/dawn/pull/2754/commits/a67043158380dacbde9b74958ff61f7f29fd51c3

assets/variant-list.css
KaichenWang: Updated - https://github.com/Shopify/dawn/pull/2754/commits/3581fc3a3835934ae225fbb3f706cd5cb0bbb88e

assets/variant-list.css
```diff
@@ -45,10 +45,10 @@ variant-list .quantity__button {
     overflow-y: auto;
   }
```
sofiamatulis, line 47: Products with 1 variant, no variants that dont have the bottom bar are getting this error: https://screenshot.click/04-46-9dpo9-ad5e8.png

assets/variant-list.css
```diff
@@ -215,9 +215,11 @@ variant-list-remove-button {
 
   .variant-list-total__column {
     display: flex;
```
sofiamatulis, line 217: Also for generic errors (at the bottom) well have to ensure they show even when there is no bottom bar

sofiamatulis (commented):
Looks great! Added one final comment

assets/variant-list.css
```diff
@@ -45,10 +45,10 @@ variant-list .quantity__button {
     overflow-y: auto;
   }
 
-  .variant-item__quantity-wrapper--no-info {
+  .variant-item__quantity-wrapper--no-info,
+  .variant-item__error {
     padding-left: 4.9rem;
   }
```
sofiamatulis, line 51: Something ive noted to is that since the generic error is at the bottom, on mobile you may not see it (desktop is sticky). but im guessing this is an issue for cart too? 🤔 
https://screenshot.click/04-52-y26dw-v6wgh.mp4

assets/variant-list.css
KaichenWang: Updated - https://github.com/Shopify/dawn/pull/2754/commits/5dd7a331416bb68a8d5eb0f07e8f8888609f4aef

assets/variant-list.css
KaichenWang: Good point, maybe we'll need to revisit the usefulness of showing a generic error at the bottom of the variant list (and in the cart). Especially in cases where the list is long.

sofiamatulis: I think it's not an issue when you are on desktop (sticky bottom), but mobile and mobile _and_ desktop on cart have this issue 🤔 

cc @edmund-teh 

https://github.com/Shopify/dawn/issues/2767

sofiamatulis (approved):
Looks great 🎉

PR #2807, opened 2023-07-10
Slim down the height of the announcement bar on desktop

kjellr:
### PR Summary: 

Slims down the height of the announcement bar on desktop.

Addresses some feedback we've received about the general size of the announcements bar. This PR:
 
- Changes the height of the announcements bar from 48px to 38px on desktop. This is roughly the same size it was before. 
- Updates the hit area for all buttons on the announcement bar to be the entire height of the announcement bar. The hit area for all announcement bar buttons on desktop is at least 38px by 38px here. 
- Sets a min-height for the announcement bar message so that the height of the utility bar doesn't fluctuate as you turn on/off settings. 

It also fixes the following small bug: 

- It sets the language and currency picker font sizes to be 13px instead of 14px, to match the rest of the text in the announcement bar + header.  
- It adjusts the width of the announcement bar so that it matches the width of the rest of the header. Currently it was matching the width of the page below, which felt odd. 

The PR does **NOT** make any adjustments to the mobile state of the announcement bar. 

Screenshots:

New height: 

Before|After
---|---
Screenshot 2023-07-10 at 12 43 06 PM|after-b

Note the left/right arrows alignment: 

Before|After
---|---
Screenshot 2023-07-10 at 12 43 20 PM|after-a

### Testing steps/scenarios

- [Editor](https://os2-demo.myshopify.com/admin/themes/140249923606/editor?category=gid%3A%2F%2Fshopify%2FOnlineStoreThemeSettingsCategory%2FSocial%2Bmedia%3Ftheme_id%3D140249923606%26first_setting_id%3Dsocial_facebook_link&section=sections--17309029924886__announcement-bar)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -117,7 +117,8 @@ body:has(.section-header .drawer-menu) .announcement-bar-section .page-width {
     padding: 0 5rem;
   }
 
-  .header.page-width {
+  .header.page-width,
+  .utility-bar__grid.page-width {
     padding-left: 3.2rem;
     padding-right: 3.2rem;
   }
@@ -2196,6 +2197,10 @@ product-info .loading-overlay:not(.hidden) ~ *,
   width: 100%;
 }
 
+.announcement-bar .slider-button {
+  height: 3.8rem;
+}
+
 .announcement-bar .slider-button--next {
   margin-right: -1.5rem;
   min-width: 44px;
@@ -2266,6 +2271,7 @@ product-info .loading-overlay:not(.hidden) ~ *,
   padding: 1rem 0;
   margin: 0;
   letter-spacing: 0.1rem;
+  min-height: 3.8rem;
 }
 
 .announcement-bar-slider--fade-in-next .announcement-bar__message,
--- a/assets/base.css
+++ b/assets/base.css
@@ -2197,10 +2197,6 @@ product-info .loading-overlay:not(.hidden) ~ *,
   width: 100%;
 }
 
-.announcement-bar .slider-button {
-  height: 3.8rem;
-}
-
 .announcement-bar .slider-button--next {
   margin-right: -1.5rem;
   min-width: 44px;
@@ -2237,6 +2233,10 @@ product-info .loading-overlay:not(.hidden) ~ *,
 
   .announcement-bar-slider {
     width: 60%;
+  }
+
+  .announcement-bar .slider-button {
+    height: 3.8rem;
   }
 }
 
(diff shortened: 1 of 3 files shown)


assets/base.css
```diff
@@ -2266,6 +2276,7 @@ product-info .loading-overlay:not(.hidden) ~ *,
   padding: 1rem 0;
   margin: 0;
   letter-spacing: 0.1rem;
+  min-height: 3.8rem;
```
kmeleta, line 2279: If you're ok with this, I probably am, but historically we do make a best effort to keep things 44px. Just want to double check we don't want to add in the additional 6px. In theory a tablet touch device could display at the desktop breakpoint and still benefit from the 44px min.

kjellr:
Thanks, @lougoncharenko! That should be fixed in 310b916.

stephanie-shopify (approved):

assets/component-localization-form.css
```diff
@@ -234,11 +236,10 @@ noscript .localization-selector.link {
 
 .header-localization:not(.menu-drawer__localization) {
   align-items: center;
-  min-height: 4.4rem;
 }
 
 .js .header-localization:not(.menu-drawer__localization) .localization-form__select {
-  padding: 0 2.7rem 0 1.2rem;
+  padding: 0.85rem 2.7rem 0.85rem 1.2rem;
```
kmeleta, line 242: Not a big fan of the kinda magic numbers. I'd rather set the desired height of 38px more transparently and align things with flex box, but if that would require additional refactoring to accommodate than this works for now.

kmeleta (approved):

lougoncharenko:
Full screen mode is adding extra margins to both the announcement bar and header
Current in Dawn:
Screenshot 2023-07-10 at 1 39 37 PM

This PR:
Screenshot 2023-07-10 at 1 43 24 PM

assets/component-localization-form.css
kjellr: Deeeeefinitely. I don't know why I bothered trying to figure out these weird `rem` values. 😂 Fixed in dff48f2.

kmeleta:
@lougoncharenko I think you're probably seeing the difference when the desktop drawer menu is used, where we allow everything to go more full width. If using the dropdown or megamenu types everything should be restricted to page width.

kjellr:
Ah yes, that's got to be it. I just checked, and the announcement bar respects that setting too — when the menu is set to Drawer, the announcement bar goes full width alongside the rest of the header. And when it's set to Dropdown or Mega Menu, it continues mirroring the header by maxing out at the page width. 👍 

I think we're good in that case. Thanks everyone for the quick testing of this today!

assets/base.css
kjellr: Yeah, I'm ok with 38px for desktop. 👍

PR #2818, opened 2023-07-11
Variant List P1 feedback

sofiamatulis:
This PR fixes:

- Hover state for quantity input stays visible when cursor is away from quantity input area. (from @scottgmeadows) [Video](https://screenshot.click/10-41-01e03-dsz6c.mp4), The hover gets triggered even when your mouse is far away from the qty selector (from @vernalkick) https://screenshot.click/Current-PR-Kevin-Hover-trigger.png
- Nit: Spinner should be vertically centered quantity input. [Screenshot](https://screenshot.click/10-24-qx63m-cxcjz.png) (from @scottgmeadows) - Video: https://screenshot.click/11-14-kd7tv-5cz0w.mp4

Editor: https://os2-demo.myshopify.com/admin/themes/140258639894/editor

### DIFF
--- a/assets/variant-list.js
+++ b/assets/variant-list.js
@@ -224,7 +224,7 @@ class VariantList extends HTMLElement {
 
         if (parsedState.description || parsedState.errors) {
           const variantItem = document.querySelector(`[id^="Variant-${id}"] .variant-item__totals.small-hide .loading-overlay`);
-          variantItem.classList.add('loading-overlay--error')
+          variantItem.classList.add('loading-overlay--error');
           this.resetQuantityInput(id, quantityElement);
           if (parsedState.errors) {
             this.updateLiveRegions(id, parsedState.errors);


sofiamatulis:
> * General misalignment with the spinner. It seems like the spinner element is bottom-aligned to the `Variant total` value. It should be center-aligned.
>   https://screenshot.click/Bottom-aligned.png
>   https://screenshot.click/Center-aligned.png

@edmund-teh We were following what we were doing for cart. I wanted to confirm that

https://screenshot.click/11-06-7mfq4-7cyoq.mp4

scottgmeadows (approved):

assets/variant-list.js
```diff
@@ -223,6 +223,8 @@ class VariantList extends HTMLElement {
         const items = document.querySelectorAll('.variant-item');
 
         if (parsedState.description || parsedState.errors) {
+          const variantItem = document.querySelector(`[id^="Variant-${id}"] .variant-item__totals.small-hide .loading-overlay`);
+          variantItem.classList.add('loading-overlay--error')
```
KaichenWang, line 227:
```suggestion
          variantItem.classList.add('loading-overlay--error');
```

KaichenWang, line 227: We add `loading-overlay--error` class here, but it's never removed. Is that ok?

KaichenWang (approved):

assets/variant-list.js
sofiamatulis: We dont need to because when something gets added/removed to cart the whole row gets re-rendered (no error class by default). And if there is no error, no error class is added

PR #2828, opened 2023-07-12
Use consistent font sizes in the Utility Bar

kjellr:
### PR Summary: 

Change the font size of the the language and currency picker to be 13px instead of 14px to match the rest of the text in the announcement bar.

For consistency. Currently, the font size for the language and currency picker is 1px larger than the font size for the announcement bar messages.

I'd meant to take care of this in #2807 (it's even mentioned in the description!), but it slipped out at some point. 

### Screenshots

Before:
Screenshot 2023-07-12 at 7 50 43 AM

After:
Screenshot 2023-07-12 at 7 44 10 AM

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/140267618326/editor)

### Checklist

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -2230,6 +2230,11 @@ product-info .loading-overlay:not(.hidden) ~ *,
   animation: animateMenuOpen var(--duration-default) ease;
 }
 
+.utility-bar .localization-wrapper .disclosure .localization-form__select,
+.utility-bar .localization-wrapper .disclosure__link {
+  font-size: calc(var(--font-heading-scale) * 1.3rem);
+}
+
 @media screen and (min-width: 990px) {
   body:has(.section-header .header:not(.drawer-menu)) .utility-bar .page-width {
     padding-left: 5rem;


assets/base.css
```diff
@@ -2230,6 +2230,11 @@ product-info .loading-overlay:not(.hidden) ~ *,
   animation: animateMenuOpen var(--duration-default) ease;
 }
 
+.utility-bar .localization-wrapper .disclosure .localization-form__select,
+.utility-bar .localization-wrapper .disclosure__link {
+  font-size: calc(var(--font-heading-scale) * 1.3rem);
+}
+
 @media screen and (min-width: 990px) {
```
lougoncharenko, line 2238: The changes look good! Just out of curiosity and so that I can grow better as a developer, what is the purpose of add var(--font-heading-scale)*1.3rem as opposed to just using 1.3rem? Is this a consistent pattern I should be using for font sizes as I noticed that other sections use this calculation for sizing as well.

kjellr: Sure! I wanted to match [the `.h5` font size](https://github.com/Shopify/dawn/blob/1d336409bfbd159f45f198ffb4366cb01c2ca66f/assets/base.css#L358) we use for the announcement bar text, so that they won't go out of sync. I didn't need to wrap it in the media query because the localization pickers in the header only show up above the `750px` breakpoint anyway. 

We use this calc function to generate the font size so that when merchants adjust the font size scale in Theme Settings > Typography, that `var(--font-heading-scale)` variable will update and scale up all the text that uses these rules proportionally. 

(Ideally we should just be using a variable here, e.g. `var(--heading-font-size-h5)` rather than repeating that calc function over and over, but we haven't implemented something like that in Dawn.)

kmeleta (approved):

Issue #2845, opened 2023-07-14
Some of the volume pricing popover buttons should not inherit global button border styles


KaichenWang:
Some of the volume pricing popover buttons will inherit global button border styles. This should not be the case

<img width="285" alt="Screenshot 2023-07-12 at 5 57 34 PM" src="https://github.com/Shopify/dawn/assets/3710548/224e5803-135c-46f1-97c5-c09d6d01c031">

<img width="389" alt="Screenshot 2023-07-12 at 5 58 05 PM" src="https://github.com/Shopify/dawn/assets/3710548/0989dda6-16ae-4b95-b6df-1ccea225e2c8">

_Originally posted by @KaichenWang in https://github.com/Shopify/dawn/issues/2791#issuecomment-1633265225_

PR #2851, opened 2023-07-17
Price per item, Popover and global style bugs

### LINKED ISSUE #2849: Quick order list: Price per item does not show when product contains mixed pricing

sofiamatulis:
### PR Summary: 

- There is a bug where on Variant List, a product with variants that have vol pricing show the price per item, but variants without don't.
- There is a bug where Vol pricing popover breaks when it changes
- There is a bug where price per item doesnt show if a product has other variants with vol pricing (PDP and feat prod) https://screenshot.click/17-35-wiw36-oj1e9.png
- Remove global border styles: https://github.com/Shopify/dawn/issues/2845
- Remove currency from variant list: https://screenshot.click/17-52-25i6y-sga9i.png
- Add missing translations

Fixes #2849 and #2847 and #2845

### Editor

https://admin.shopify.com/store/os2-demo/themes/140305629206/editor


assets/quantity-popover.css
```diff
@@ -22,12 +23,19 @@ quantity-popover volume-pricing li {
   max-width: 36rem;
 }
 
+.quantity-popover__info .button-close,
+.variant-remove-total quick-order-list-remove-all-button,
+.quick-order-list-total__confirmation quick-order-list-remove-all-button {
+  --shadow-opacity: 0;
```
KaichenWang, line 29: We'll need to prevent borders also. If we apply the [exact settings shown in the issue](https://github.com/Shopify/dawn/issues/2845#issue-1805570855), the borders will be visible.

Also for a11y, I noticed that `quick-order-list-remove-all-button` is not focusable via keyboard navigation.

KaichenWang, line 29: We'll need to prevent borders also. If we apply the [exact settings shown in the issue](https://github.com/Shopify/dawn/issues/2845#issue-1805570855), the borders will be visible.

Also for a11y, I noticed that `quick-order-list-remove-all-button` is not focusable via keyboard navigation.

sections/featured-product.liquid
```diff
@@ -282,35 +282,72 @@
...
+                                {%- endif -%}
+                              {%- endfor -%}
+                            {%- endif -%}
+                          </div>
+                        {%- else -%}
+                          {%- assign variant_price = product.selected_or_first_available_variant.price
+                            | money_with_currency
+                          -%}
+                          {%- assign variant_price_compare = product.selected_or_first_available_variant.compare_at_price -%}
+                          {{ product.selected_or_first_available_variant.compare_at_price }}
```
KaichenWang, line 321: Is this supposed to be displayed here?
```suggestion

```

sections/main-product.liquid
```diff
@@ -275,24 +275,52 @@
...
-                              {%- break -%}
-                            {%- endif -%}
-                          {%- endfor -%}
-                        {%- endif -%}
-                      </div>
-                    {%- endif -%}
-                  </price-per-item>
+                        {%- assign variant_price = product.selected_or_first_available_variant.price | money_with_currency -%}
+                        {%- assign variant_price_compare = product.selected_or_first_available_variant.compare_at_price -%}
+                        {{ product.selected_or_first_available_variant.compare_at_price }}
```
KaichenWang, line 298: Is this supposed to be displayed here?
```suggestion

```

sections/featured-product.liquid
```diff
@@ -282,35 +282,72 @@
...
+                                </dt>
+                                <dd>
+                                  <s class="variant-item__old-price">
+                                    {{ variant_price_compare | money }}
+                                  </s>
+                                </dd>
+                                <dt class="visually-hidden">
+                                  {{ 'products.product.price.sale_price' | t }}
+                                </dt>
+                                <dd>
```
KaichenWang, line 336: Do we want the additional left margin introduced by `<dd>`?

Screenshot 2023-07-17 at 6 48 12 PM

eugenekasimov (approved):
looks good 👍

sections/featured-product.liquid
```diff
@@ -282,35 +282,72 @@
...
+                                      'products.product.volume_pricing.price_at_each'
+                                      | t: price: price_break_price
+                                    -}}
+                                  </span>
+                                  {%- break -%}
+                                {%- endif -%}
+                              {%- endfor -%}
+                            {%- endif -%}
+                          </div>
+                        {%- else -%}
```
KaichenWang, line 316: If the variant does not have price breaks, `<div class="price-per-item">` does not need to be nested inside `<price-per-item>` custom element. Otherwise, the custom element JS logic still runs on every quantity input change even though we don't need to calculate a new value.

sofiamatulis: Yes, because if you change variant youd want the element to be replaced. Does that make sense? 

Unless we add some custom logic to replace it in differently elements depending on where it lives.

 Perhaps an enhancement?

sections/quick-order-list.liquid
```diff
@@ -140,7 +140,11 @@
               </a>
               <div class="variant-remove-total">
                 {%- render 'loading-overlay' -%}
-                <quick-order-list-remove-all-button class="button button--tertiary no-js-hidden" data-action="confirm">
+                <quick-order-list-remove-all-button
```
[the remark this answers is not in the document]
sofiamatulis: Makes sense! Thanks for the suggestion! :)

sections/featured-product.liquid
sofiamatulis: Yes, sorry I meant that the wed have to target different elements (I thought global.js was targeting the custom element) but the logic is small enough 👍 Included in this PR! Great suggestion

sections/quick-order-list.liquid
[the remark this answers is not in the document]
KaichenWang: What i meant here is that we need to add a `<button>` element to the HTML, only adding `type="button"` doesn't seem to work.

sections/featured-product.liquid
```diff
@@ -282,35 +282,78 @@
...
-                                {%- assign price_break_price = price_break.price | money_with_currency -%}
-                                <span class="price-per-item--current">
-                                  {{- 'products.product.volume_pricing.price_at_each' | t: price: price_break_price -}}
-                                </span>
-                                {%- break -%}
-                              {%- endif -%}
-                            {%- endfor -%}
-                          {%- endif -%}
+                    {%- if product.quantity_price_breaks_configured? -%}
+                      {%- if product.selected_or_first_available_variant.quantity_price_breaks.size > 0 -%}
```
KaichenWang, line 286: Sorry i think my previous suggestion had a bug because it seems the variant change logic is only replacing the innerHTML portion of the element.

We can replace this in `global.js`:
```JS
pricePerItemDestination.innerHTML = pricePerItemSource.innerHTML;
```
With:
```
pricePerItemDestination.replaceWith(pricePerItemSource);
```


To be safe, we can also revert [this] (https://github.com/Shopify/dawn/pull/2851/commits/6c81fb718021c93c6b0715371b28d919e1969c2b), if we don't want to introduce any regressions.

nicklepine (approved):

sections/quick-order-list.liquid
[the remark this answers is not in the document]
sofiamatulis: Updated it! I was testing out making the custom element of type button but went with this approach

KaichenWang (approved):

PR #2864, opened 2023-07-18
Utility Bar: Try a fix for social icon wrapping and slider arrow overlap

### LINKED ISSUE #2862: Utility Bar: On middle breakpoints, social icons wrap to two lines

On middle breakpoints (depending on the number of social icons shown), it's possible that some social icons in the Announcement Bar will wrap to a second line. We should avoid this. 

wrapping

If there's visible space on the screen (as in the example above), the icons should fill that space. If there is no space, it would be preferable to hide the icons that can't fit rather than wrap them.

### LINKED ISSUE #2863: Utility Bar: Announcement bar left arrow overlaps social icon

When many social icons are present, the left arrow for the Announcement bar slider appears just a little too close to the icons. When you inspect the element, it's clear that the slider arrow element is actually overlapping the social icon. 

Screenshot 2023-07-18 at 8 07 49 AM

Screenshot 2023-07-18 at 8 07 43 AM

We should adjust the position of the arrow so that it won't overlap the social icons. 

## Possible solution

This is due to a negative left margin being applied on the button (the right button has a negative right margin too). If we can rework the button to remove that, we should be good.

kjellr:
### PR Summary: 

- Ensure that social icons don't wrap, even if you have all of them shown. 
- Ensure that the slider arrows do not overlap the social icons if you have all of the social icons shown. 

Fixes #2862.
Fixes #2863. 

1. I added a new, manual breakpoint (`1080px`) instead of the usual `980px` one to help aid against wrapping. 
2. I also adjusted the grid layout on that larger breakpoint to give a little more space to the icons. It's now `33% / 33% / 33%` instead of `30% / 40% / 30%`. 
3. To fix the arrows, I minimized the negative margin on the large breakpoint. As far as I can tell, it's not necessary there in the first place. 

https://github.com/Shopify/dawn/assets/1202812/7d804fa3-df61-454d-a546-9015d7c34dc4

### Other considerations

Positive: 
- Fixes the bugs, doesn't ever look broken. 
- The grid update feels more natural — equal thirds is more likely to line up with other columns on the page. 

Negative:
- Uses a non-standard breakpoint for us (I could argue that this doesn't matter, but I'm putting it under negative for now. 😉)
- For most merchants who have just a few social icons, we'll be hiding all icons far earlier than necessary. 
- Makes the announcement bar area slightly less wide on larger screens.

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/140304810006/editor)

### Checklist

### DIFF
--- a/assets/component-menu-drawer.css
+++ b/assets/component-menu-drawer.css
@@ -261,5 +261,5 @@ details[open].menu-opening > .menu-drawer__submenu {
 }
 
 .menu-drawer .list-social__link {
-  padding: 1.3rem 1.3rem;
+  padding: 1.1rem 1.1rem;
 }


assets/base.css
```diff
@@ -2154,9 +2154,9 @@ product-info .loading-overlay:not(.hidden) ~ *,
   scale: 0.9;
 }
 
-@media screen and (min-width: 990px) {
+@media screen and (min-width: 1090px) {
   .utility-bar__grid--3-col {
-    grid-template-columns: 3fr 4fr 3fr;
+    grid-template-columns: 1fr 1fr 1fr;
     grid-template-areas: 'social-icons announcements language-currency';
```
lougoncharenko, line 2160: This seems like a good fix. With all the icons present on the same line and not wrapping, it does seem a bit crowded however, I believe the number of merchants who will use all of the social icons is relatively small. The other point I wanted to bring up is that our previous break point coincided with when the mobile drawer would appear, the social icons and locales would become hidden and the drawer would appear at the same breakpoint. Now with the social icons/ locales becoming hidden and the mobile drawer appearing at different break points seems a bit odd to me.  This is not a blocker for the release but I did want to point this out and am wondering how other people feel about this change.

lougoncharenko (approved):
I tested it in the browser with different dimensions it worked as expected. It did look visually appealing to not have the social icons wrap onto the second line. I couldn't find a way to break anything with these changes and in my opinion, this fix is ready to go. I did leave some comments for future considerations as the different breakpoints between the social icons/locales from the media drawer do seem a bit odd, but that is just my opinion .

assets/base.css
eugenekasimov: I agree with Louisa that now we have a gap between 990px and 1090px when you can't reach social icons 🤔.

melissaperreault:
@eugenekasimov I tried to reproduce but I couldn't ([Video reference](https://screenshot.click/19-22-oh413-2g2t7.mp4))

melissaperreault:
Not related to this PR but an observation, it seems that the social icons are not showing in the same order from different place. I assumed the one in the announcement bar was following the order they appear in the theme settings, [but 2 are interchanged](https://screenshot.click/19-24-89biq-4j66w.png).

In the drawer, Twitter is first and [the rest is random](https://screenshot.click/19-27-i9xm5-jtukl.png)

In the footer, it follows the order from the announcement bar.

🤷‍♀️

kmeleta (changes requested):
I'm also seeing some disprencies around wrapping icons depending on, page width setting, menu type, and/or browser window size https://screenshot.click/20-25-hemfe-adg1z.mp4. Also just my opinion, but I also feel odd taking away additional space from the announcement bar for the social icons worst case scenario where they're all visible.

PR #2868, opened 2023-07-18
Fix cart submission on Quick Order List

### LINKED ISSUE #2857: Quick order list: quantity that doesn't not respect quantity limit can be added to cart
Meli's video: https://screenshot.click/17-43-5mgav-rzrde.mp4

Slack thread:  https://shopify.slack.com/archives/C04NPNBGU87/p1689619584764189

There are 2 bugs - probably related:
1. Hitting enter after inputing quantity on quick order list on a product that has 1 or zero variants redirects to /cart
2. I am able to add submit quantity that does not respect quantity limits ([bug video](https://screenshot.click/17-40-7ztpg-7y6pt.mp4))

I think the fix should include::

- Let's prevent being able to submit a quantity that doesn't respect quantity limits.
- ~Let's only suppress cart notification actions on quick order list when there are 2+ variants on a product. For products with zero or 1 variant, let's respect the cart notification behaviour selected in theme settings on quick order list~ (removed this requirement, [see this comment. ](https://github.com/Shopify/dawn/pull/2868#issuecomment-1640680929)
- Hitting "Return/Enter" keyboard key when a product has zero or 1 variant should not bring you to /cart (unless that's your selected cart notification behaviour)

Video recap: https://screenshot.click/17-29-afrh6-clay1.mp4

sofiamatulis:
Fixes #2857 

**Bug:** Hitting enter on a quick order list with 1 variant or no variants takes you to the cart page and allows for wrong amounts

I am preventing the form submission in Javascript to not allow this behaviour

Editor: https://admin.shopify.com/store/os2-demo/themes/140305694742/editor

Product with no variants: `product-with-no-variants`
Product with 1 variant: `one-variant-product`


nicklepine:
In the [issue](https://github.com/Shopify/dawn/issues/2857), I suggested the following, which was not implemented: 

> Let's only suppress cart notification actions on quick order list when there are 2+ variants on a product. For products with zero or 1 variant, let's respect the cart notification behaviour selected in theme settings on quick order list

👍 I think we did great not implementing this change given, going from product to product when quick order list is implemented, this could open the door to different cart behaviours depending on if a product has 1 or more variants. 

Video explanation: https://screenshot.click/18-37-xs08q-pn2pu.mp4

Overall, this fix looks great and works as expected. 

cc: @edmund-teh @melissaperreault  @sofiamatulis

sofiamatulis:
> In the [issue](https://github.com/Shopify/dawn/issues/2857), I suggested the following, which was not implemented:
> 
> > Let's only suppress cart notification actions on quick order list when there are 2+ variants on a product. For products with zero or 1 variant, let's respect the cart notification behaviour selected in theme settings on quick order list
> 
> The cart notification behaviour is suppressed on quick order list, even if the product has zero or 1 variant only. 👍 I think we did great not implementing this change given, going from product to product when quick order list is implemented, this could open the door to different cart behaviours depending on if a product has 1 or more variants.
> 
> Video explanation: https://screenshot.click/18-37-xs08q-pn2pu.mp4
> 
> Overall, this fix looks great and works as expected.
> 
> cc: @edmund-teh @melissaperreault @sofiamatulis

@nicklepine I agree this would be a nice behaviour when you only have 1 variant (show the drawer or the popup). However, we wouldnt be doing that when the cart type is type of page (which we do for Add to cart). So i think we need to talk a bit more about the expected behaviour. 

We can treat this as a fast follow enhancement!

eugenekasimov:
Input and buy button disappeared 🤔 Is it just me? 

Screenshot 2023-07-18 at 1 03 46 PM

eugenekasimov (approved):
Works for me.

I'm wondering if we could just add `onsubmit="return false;"` to `<form>` in `quick-order-list.liquid`? 🤔 I did a quick test and it seems working as well but maybe I missed something. I just wanted to avoid using JS for our solution.

PR #2888, opened 2023-07-21
Fix font family for quick order list

### LINKED ISSUE #2873: Quick order list: All transactional ($) values should use the same font
This occurs on the Quick order list section. 

All transactional ($) values  and the `Sold out` string should use the same font, preferably the `body` one, instead of `headings`. 

These two values and one string should use the Body font, but use Headings font…  `3`, `$495.00`, and `Sold out`…

Image
https://screenshot.click/Quick-order-list-body-font.png

See cart page for reference: https://screenshot.click/Cart-reference.png
See typography global settings for reference: https://screenshot.click/Global-settings-typography.png

sofiamatulis:
### PR Summary: 

Fixes font family to folllow body for
- sold out 
- total items
- total price

Fixes #2873.

Editor: https://admin.shopify.com/store/os2-demo/themes/140326141974/editor

### DIFF
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -603,6 +603,7 @@ quick-order-list-remove-button {
 .totals__subtotal-value {
   margin-top: 0;
   margin-bottom: 0;
+  color: rgb(var(--color-foreground));
 }
 
 .quick-order-list__total-items p,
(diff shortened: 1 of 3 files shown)


edmund-teh (commented):
@sofiamatulis 

I think the `Total items` and `Product subtotal` number values should be opacity: `100%` or `1.0.` With your recent change, they are showing as `0.75`. 

This should only be reflected with the # values themeselves, ie. 11 and $2,446.62 should be `100%`. Total items and Product subtotal, the headings themselves should be `75%`.

https://screenshot.click/Product-subtotal-color-foreground-opacity.png
https://screenshot.click/Total-items-color-foreground-opacity.png

Product-subtotal-color-foreground-opacity
Total-items-color-foreground-opacity

PR #2889, opened 2023-07-21
Hide vol pricing and qty rules when variant is unavailable

### LINKED ISSUE #2870: Volume Pricing: Variant option combination that doesn't exist 'Unavailable' product still shows 'Volume pricing available', etc
I believe we originally have this issue to avoid layout shift, but it might not make sense to display the `Volume pricing available` string and the `Volume pricing` table on a variant combination that doesn't exist. 

The information shouldn't display here, as the product is not available and therefore wouldn’t have or shouldn’t have volume pricing available, since the product is unavailable. 

See video recording for more info: https://screenshot.click/Volume-pricing-bug-Fast-follow-Jul-18-2023.mp4

https://screenshot.click/Current-PR-Desktop-Unavailable-variant-combo-v2.png
https://screenshot.click/Current-PR-Mobile-Unavailable-variant-combo-v2.png

We are presently doing this in our latest/current version of Dawn, by hiding the `Price` block on unavailable products/ variant combinations that don't exist: https://screenshot.click/Current-Dawn-Price-block-hidden.png

Image
Image

sofiamatulis:
### PR Summary: 

Ensure qty rules and vol pricing dont show for variants that dont exist.

Fixes #2870 and https://github.com/Shopify/dawn/issues/2932

Test:
- Mobile
- Desktop
- Feat Prod 
- PDP

[Editor](https://admin.shopify.com/store/os2-demo/themes/140326273046/editor)

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -798,7 +798,7 @@ class SlideshowComponent extends SliderComponent {
 
   setSlidePosition(position) {
     if (this.setPositionTimeout) clearTimeout(this.setPositionTimeout);
-    this.setPositionTimeout = setTimeout (() => {
+    this.setPositionTimeout = setTimeout(() => {
       this.slider.scrollTo({
         left: position,
       });
@@ -1110,6 +1110,12 @@ class VariantSelects extends HTMLElement {
         const pricePerItemSource = html.getElementById(`Price-Per-Item-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`);
 
         const volumePricingDestination = document.getElementById(`Volume-${this.dataset.section}`);
+        const qtyRules = document.getElementById(`Quantity-Rules-${this.dataset.section}`);
+        const volumeNote = document.getElementById(`Volume-Note-${this.dataset.section}`);
+
+        if (volumeNote) volumeNote.classList.remove('visibility-hidden');
+        if (volumePricingDestination) volumePricingDestination.classList.remove('visibility-hidden');
+        if (qtyRules) qtyRules.classList.remove('visibility-hidden');
 
         if (source && destination) destination.innerHTML = source.innerHTML;
         if (inventorySource && inventoryDestination) inventoryDestination.innerHTML = inventorySource.innerHTML;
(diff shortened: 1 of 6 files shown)


assets/global.js
```diff
@@ -1110,6 +1110,12 @@ class VariantSelects extends HTMLElement {
         const pricePerItemSource = html.getElementById(`Price-Per-Item-${this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section}`);
 
         const volumePricingDestination = document.getElementById(`Volume-${this.dataset.section}`);
+        const qtyRules = document.getElementById(`Quantity-Rules-${this.dataset.section}`);
+        const volumeNote = document.getElementById(`Volume-Note-${this.dataset.section}`);
+
+        if (volumeNote) volumeNote.classList.remove('visibility-hidden');
+        if (volumePricingDestination) volumePricingDestination.classList.remove('visibility-hidden');
```
eugenekasimov, line 1117: I think it's better to replace `visibility-hidden` with `hidden`. As far as I understand we want to keep layout only above input picker, so I would prefer to hide volume pricing table and not to keep the space. 
Screenshot 2023-07-21 at 9 27 44 AM

cc: @edmund-teh

eugenekasimov (commented):
Works well, I left one small comment.

eugenekasimov (approved):

assets/global.js
edmund-teh: > I think it's better to replace `visibility-hidden` with `hidden`. As far as I understand we want to keep layout only above input picker, so I would prefer to hide volume pricing table and not to keep the space. Screenshot 2023-07-21 at 9 27 44 AM
> 
> cc: @edmund-teh

You make a good point @eugenekasimov! Plus the `Unavailable` button is disabled so a buyer won't be able to take any action here. We also wanted to avoid the jumping with the different variant options, which we still would be doing.

sofiamatulis: Since `sku` and `price per item` are using `visibility hidden` I think we should follow the same for now for consistency.

We can address this issue when we address the existing spacing issue cc @edmund-teh - i believe we have an issue for this? cant seem to find it

edmund-teh: This issue? @sofiamatulis #2870?

sofiamatulis: @edmund-teh I thought we had an issue for the extra spacing? 🤔

sofiamatulis: Let's track it separately 👍

edmund-teh (approved):
Looks solid @sofiamatulis! Thanks for implementing the changes we outlined on Slack:

- [x] Hiding the price block because the selected variant does not exist
- [ ] Hiding any gaps (from quantity rules/vp/anything else) because the variant does not exist

Pills video:
https://screenshot.click/Pills-no-price-no-gaps.mp4

Dropdowns video:
https://screenshot.click/Dropdowns-no-price-no-gaps.mp4

Figma prototype [here](https://www.figma.com/proto/AxH2jkpnLbZXPR8NbgaoaK/Volume-pricing-in-flagship-themes?page-id=1257%3A309&type=design&node-id=1257-594&viewport=-146%2C484%2C0.35&t=jnQzYJelO5oakp45-1&scaling=scale-down-width&starting-point-node-id=1257%3A594&show-proto-sidebar=1&mode=design).

PR #2918, opened 2023-07-26
Bump theme_version in config/schema_settings.json to 10.0.0

burke:
The real action here isn't to merge, but to bump the v10.0.0 tag to this commit -- it was tagged prior to the version bump.


burke:
Fixed published tag

PR #2923, opened 2023-07-28
Fix misalignment of total items in quick order list

### LINKED ISSUE #2921: Quick order list: `total` misaligned when font size is larger
https://screenshot.click/27-58-fkzwz-2a11e.png

Quick video:  https://screenshot.click/24-10-fmlvp-by2az.mp4

eugenekasimov:
### PR Summary: 

This PR fixes:

- misalignment of `total items` element in quick order list. 
- the issue when images were cut in the quick order list if the size of typography was increased higher than 100% . 

Fixes #2921.

- Set a width for the `.quick-order-list__total-items`element same as the input component and add same margin as padding we use for the input in the table. 
- Changed a bit the way we calculate the padding of the `input` element to account for `icon-info` which is always `15px`.  
- Added `width: 100%` to .`variant-item__image`

### Testing steps/scenarios

- [ ] Go to a product page and add Quick Order List section. 
- [ ] Make sure that `total items` is aligned with the inputs in the table. 
- [ ] Play with different theme settings such as typography (heading and body font sizes), input border and make sure the `total-items` element is still aligned.
- [ ] Make sure that images in the Quick Order List are not cut. 

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/140381519894/editor)

### Checklist

### DIFF
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -123,6 +123,7 @@ quick-order-list .quantity__button {
 
 .variant-item__image {
   height: 100%;
+  width: 100%;
   object-fit: cover;
 }
 
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -562,7 +562,8 @@ quick-order-list-remove-button:hover .icon-remove {
   }
 
   .quick-order-list__total-items {
-    padding-left: calc(((11rem / var(--font-body-scale) + var(--inputs-border-width) * 2)/ 4) + 4.5rem);
+    width: calc(((11rem / var(--font-body-scale) + var(--inputs-border-width) * 2)));
+    margin-left: 4.9rem;
     flex-direction: column;
   }
 }
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -615,6 +615,10 @@ quick-order-list-remove-button {
   margin-bottom: 0;
 }
 
+.quick-order-list__total-items p {
+  text-align: center;
+}
+
 .quick-order-list-total__info {
   display: flex;
 }
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -567,6 +567,10 @@ quick-order-list-remove-button:hover .icon-remove {
     margin-left: 4.9rem;
     flex-direction: column;
   }
+
+  .has-volume-pricing .quick-order-list__total-items {
+    margin-left: calc(15px + 3.4rem);
+  }
 }
 
 quick-order-list-remove-button {
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -603,6 +603,7 @@ quick-order-list-remove-button {
 .totals__subtotal-value {
   margin-top: 0;
   margin-bottom: 0;
+  color: rgb(var(--color-foreground));
 }
 
 .quick-order-list__total-items p,
--- a/assets/quick-order-list.css
+++ b/assets/quick-order-list.css
@@ -215,6 +215,7 @@ quick-order-list .quantity__button {
 .variant-item__sold-out {
   opacity: 0.7;
   font-size: 1.6rem;
+  color: rgb(var(--color-foreground));
 }
 
 quick-order-list-remove-button {
@@ -598,7 +599,7 @@ quick-order-list-remove-button {
   align-items: center;
 }
 
-.quick-order-list__total-items h3,
+.quick-order-list__total-items span,
 .totals__subtotal-value {
   margin-top: 0;
   margin-bottom: 0;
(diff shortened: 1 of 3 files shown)


assets/quick-order-list.css
```diff
@@ -562,7 +563,8 @@ quick-order-list-remove-button:hover .icon-remove {
   }
```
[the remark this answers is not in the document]
eugenekasimov: Good catch Sofia!  

Fixed [here](https://github.com/Shopify/dawn/pull/2923/commits/7fbcee3d59cb64e50f1807be88f5245dd77ddadd)

sections/quick-order-list.liquid
```diff
@@ -128,7 +128,10 @@
         {% comment %} Populated by JS {% endcomment %}
       </span>
     {%- else -%}
-      <div class="quick-order-list__total gradient" id="quick-order-list-total">
+      <div
+        class="quick-order-list__total{% if product.quantity_price_breaks_configured? %} has-volume-pricing{% endif %} gradient"
```
sofiamatulis, line 132: They could have volume pricing or qty rules. We should check for both

(qty popover will show if there are only qty rules as well)

assets/quick-order-list.css
```diff
@@ -562,9 +563,14 @@ quick-order-list-remove-button:hover .icon-remove {
...
 
   .quick-order-list__total-items {
-    padding-left: calc(((11rem / var(--font-body-scale) + var(--inputs-border-width) * 2)/ 4) + 4.5rem);
+    width: calc(((11rem / var(--font-body-scale) + var(--inputs-border-width) * 2)));
+    margin-left: 4.9rem;
     flex-direction: column;
   }
+
+  .has-volume-pricing .quick-order-list__total-items {
+    margin-left: calc(15px + 3.4rem);
```
sofiamatulis, line 572: Why ae we using `px` here? Do we need this calculation?

assets/quantity-popover.css
```diff
@@ -160,4 +160,4 @@ quantity-popover .quantity__rules .divider:nth-child(2)::before {
 quantity-popover .quantity__button:not(:focus-visible):not(.focused),
 quantity-popover .quantity__input:not(:focus-visible):not(.focused) {
   background-color: initial;
-}
+}
```
sofiamatulis, line 163: Let's revert this

assets/quick-order-list.css
```diff
@@ -598,7 +605,7 @@ quick-order-list-remove-button {
   align-items: center;
 }
 
-.quick-order-list__total-items h3,
+.quick-order-list__total-items span,
```
sofiamatulis, line 608: Let's rebase `font` to ensure the changes stay on that PR

sections/quick-order-list.liquid
```diff
@@ -128,7 +128,18 @@
         {% comment %} Populated by JS {% endcomment %}
       </span>
     {%- else -%}
-      <div class="quick-order-list__total gradient" id="quick-order-list-total">
+      {% liquid
```
sofiamatulis, line 131: This makes sense for now but ideally we want to check for this on the product level so we dont have to loop.

We can also add `break` when it becomes true and can stop looping 
https://shopify.dev/docs/api/liquid/tags/break

Another idea is:

What I suggest you do is
In the row file you can check if there has_popover https://github.com/Shopify/dawn/blob/main/snippets/quick-order-list-row.liquid#L196 and add a class

In the quick order list file, you can use the `:has` property https://developer.mozilla.org/en-US/docs/Web/CSS/:has in CSS to see if it includes a row with popover.

This would avoid the loop. It's not available in all browsers but since this doesnt break the experience (just changes the spacing), I think it's a fair trade off

Ill let the 2nd reviewer weigh in as well

Ideally we dont want to loop through all of the variants to find out if the product has qty. This gets re-rendered every time the section re-renders and is re-calculated

eugenekasimov: @sofiamatulis I looked again into this issue and I don't think we need to avoid iteration because we already do the same iteration [here](https://github.com/Shopify/dawn/pull/2923/files#diff-b1e86f8491d31b61e045b54da8bb8ca6d0cc4dbb55f3560aace18423de6e176aR99). So, I just moved my logic to the existing `for` loop and removed the one I added before. 

What do you think?

edmund-teh (changes requested):
Sorry, @eugenekasimov... I'm still seeing misalignment on products that have Volume pricing/quantity rules on some variants and not on others.

See reference here: https://screenshot.click/Still-misalignment.png
This product: `A/ Image sizing test – 3D models and video`
https://os2-demo.myshopify.com/products/test-3d-models-video

FYI, I'm testing and playing with font-scaling and different font styles too in the global settings.

edmund-teh (approved):
This is working really well now @eugenekasimov! Thanks so much.

sofiamatulis (approved):
Works well! I think well still need to address this: https://github.com/Shopify/dawn/pull/2923#discussion_r1281936452

PR #2934, opened 2023-08-02
Fix double-taping on iPad to open popover in Quick Order List 

### LINKED ISSUE #2890: 🐞 **_Bug!_** To open the popover on iPad Pro (and larger tablets), it requires a double-tap: https://screenshot.click/RPReplay_Final1689019861.MP4

eugenekasimov:
### PR Summary: 

This PR fixes the issue when you need to tap twice on `info-icon` on iPad(tablets) in order to open the popover.

Fixes #2890.

There was a conflict between two event listeners `mouseenter` and `click` that fired the same function. On iPad when you tap an element both events are fired, first `mouseenter` and then `click`.

We still need to keep them both because we have to cover cases for mouse and keyboard users and iPad and Desktop may have the same screen size. 

I added a variable `eventMouseEnterHappened` to track a state of the `mouseenter` event. Now, if the `mouseenter` happens the 'click' event won't be fired until `eventMouseEnterHappened` is reset. 

### Testing steps/scenarios

- [ ] Add the quick order list section to the product page.
- [ ] Login as a b2b customer and go to a product that has volume-pricing or qty rule. 
- [ ] Hovering mouse over `info-icon` test that popover element appears as expected. 
- [ ] Test the same on a tablet and make sure the popover element opens after first tap on the `info-icon`. 

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/quantity-popover.js
+++ b/assets/quantity-popover.js
@@ -11,6 +11,7 @@ if (!customElements.get('quantity-popover')) {
         this.popoverInfo = this.querySelector('.quantity-popover__info');
         this.closeButton = this.querySelector('.button-close');
         this.variantInfo = this.querySelector('.quantity-popover-container');
+        this.eventMouseEnterHappened = false;
 
         if (this.closeButton) {
           this.closeButton.addEventListener('click', this.closePopover.bind(this));
@@ -29,6 +30,7 @@ if (!customElements.get('quantity-popover')) {
           this.infoButtonMobile.addEventListener('click', this.togglePopover.bind(this));
           this.infoButtonMobile.addEventListener('focusout', this.closePopover.bind(this));
         }
+
         if (this.infoButtonDesktop && this.mqlTablet.matches) {
           this.variantInfo.addEventListener('mouseenter', this.togglePopover.bind(this));
           this.variantInfo.addEventListener('mouseleave', this.closePopover.bind(this));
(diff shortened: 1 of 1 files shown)


assets/quantity-popover.js
```diff
@@ -37,6 +39,11 @@ if (!customElements.get('quantity-popover')) {
 
       togglePopover(event) {
         event.preventDefault();
+        if (event.type === 'mouseenter') {
```
sofiamatulis, line 42: Small suggestion: Maybe we can convert these to variables

    ```
this.actions = {
      mouseEnter: 'mouseenter',
      click: 'click'
    }
```


And refer to it as `this.actions.click` and `this.actions.mouseEnter` throughout the file

Comment (eugenekasimov) [reply assets/quantity-popover.js:42]: I don't see we use it a lot in the file, we use it only in one function. Am I getting you right that you suggest we do it this way? I'm not sure if it's worth it but I'm okay if everyone agrees. 

```suggestion
       this.actions = {
          mouseEnter: 'mouseenter',
          click: 'click'
        }

        if (event.type === this.actions.mouseEnter) {
          this.eventMouseEnterHappened = true;
        }

        if (event.type === this.actions.click && this.eventMouseEnterHappened) return;

        const button = this.infoButtonDesktop && this.mql.matches ? this.infoButtonDesktop : this.infoButtonMobile;
        const isExpanded = button.getAttribute('aria-expanded') === 'true';

        button.setAttribute('aria-expanded', !isExpanded);

        this.popoverInfo.toggleAttribute('hidden');

        const isOpen = button.getAttribute('aria-expanded') === 'true';

        button.classList.toggle('quantity-popover__info-button--open');

        if (isOpen && event.type !== this.actions.mouseEnter) {
          button.focus();
        }
```

PR #2980, opened 2023-08-30
Add liquid to assign uses_comma_decimals on mobile

Roi-Arthur:
### PR Summary: 

Adding a liquid check bring mobile facets filters to parity with desktop filters. In their current state they break (at least visually) when using a currency that uses commas as decimal separator.

Another small addition is the removal of the Turkish Lyra (TRY) from the list of such currencies, as Turkish Lyra does not in fact use a comma for decimals.
Edit: This is already being requested [here](https://github.com/Shopify/dawn/pull/2971)

Fixes #0.

Duplicated the existing liquid block that can be found around line 305 of the facets.liquid file and added it under the second instance of `{% when 'price_range' %}`
I debated adding the check further up to have it in the code only once but this approach ensures we only run that check when there *is* a price range filter.

No impact other than fixing value of filters inputs

### Testing steps/scenarios
1. Create a product with a price higher than 1000
2. Make sure your store uses a currency that uses commas as decimal separator (Euro for example).
3. Make sure that price is an available filter (use Search & discovery app to set that up)
4. In any of our themes, go to a collection that includes your product priced >1000
5. Set your filter layout to "Drawer"
6. Inspect the inputs for the price range filter and notice the issue.

### Demo links

Uploading Drawer input max error fix.mp4…

### Checklist


ludoboludo (approved):
Looks good 👍

ludoboludo:
Thanks for sharing @willbroderick !

PR #3016, opened 2023-09-27
[Image with text] Add color scheme picker for section wrapper

ludoboludo:
### PR Summary: 

Add color picker for the image with text section 

The image with text section has historically only offered a color picker for the content container and the media but not for the whole section. 
This PR aim to change that and match the experience you're getting on the Multirow section where you have two color pickers.

Had to add some element to wrap the content that doesn't get the `page-width` class applied to it. 

This enables something new where the whole image with text section can get a background color scheme. 

### Testing steps/scenarios

- [ ] Use the image with text section on different templates
- [ ] With different color scheme backgrounds and color schemes for the content container
- [ ] Test multirow as well with a gradient background on the section with a png image (the CSS styling impacts it)

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/157723656214/editor)

### Checklist

### DIFF
--- a/sections/image-with-text.liquid
+++ b/sections/image-with-text.liquid
@@ -19,8 +19,8 @@
   if section.index == 1
     assign fetch_priority = 'high'
   endif
-  if section.settings.color_scheme != section.settings.section_color_scheme
-    assign add_color_classes = true
+  if section.settings.color_scheme == section.settings.section_color_scheme and section.settings.content_layout == 'no-overlap'
+    assign remove_color_classes = true
   endif
 -%}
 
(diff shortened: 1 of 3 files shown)


sections/image-with-text.liquid
```diff
@@ -15,87 +15,93 @@
 {%- endstyle -%}
 
 {%- liquid
-  assign scheme1 = settings.color_schemes | first
   assign fetch_priority = 'auto'
   if section.index == 1
     assign fetch_priority = 'high'
   endif
+  if section.settings.color_scheme != section.settings.section_color_scheme
+    assign add_color_classes = true
```
kmeleta, line 23: The trouble with excluding the color classes entirely, rather than relying on the "fake" transparent effect of using the same color scheme classes is that is can't hide overlaps and shadows when needed. I know with the motion side-effects we want to need to rely on true transparency more though.

This one with the overlap layout I think we probably need to address.


Some others involving shadows like this have always been more edge casey so I wouldn't push back too hard on these regressions if our primary use cases are behaving correctly.

kmeleta (commented):
Looks pretty good. Found a couple cases so far that we should try and clean up if we can, but most of the main use cases are accounted for I think.

kmeleta (approved):
lgtm. I'm also in favor keeping the setting labels consistent with what we do on multirow, but would of course defer to Meli/Katy.

katycobb:
Hey friends! Agree the labeling here isn't a dealbreaker to shipping. 

My quick POV is that if there's just one color picker, call it `Color scheme`. But if we have more than one, we might want to add differentiators to both to describe what each applies to. 

Here I suggest updating `Color scheme` to `Background color scheme` to be more specific about the difference between the two pickers. `Container color scheme` works, and is consistent with how we label other settings that apply to content containers. 

Happy to do this (and multirow) in a followup though 👍

eugenekasimov (approved):
Everything looks good to me. I like a name suggestion Katy mentioned [here](https://github.com/Shopify/dawn/pull/3016#issuecomment-1747377227), but it's not a blocker for me 👍

danielvan:
> Hey friends! Agree the labeling here isn't a dealbreaker to shipping.
> 
> My quick POV is that if there's just one color picker, call it `Color scheme`. But if we have more than one, we might want to add differentiators to both to describe what each applies to.
> 
> Here I suggest updating `Color scheme` to `Background color scheme` to be more specific about the difference between the two pickers. `Container color scheme` works, and is consistent with how we label other settings that apply to content containers.
> 
> Happy to do this (and multirow) in a followup though 👍

+1 here.

PR #3019, opened 2023-09-28
[Colors] Add default scheme picker in theme settings

ludoboludo:
### PR Summary: 

There isn't a clear way to tell or change which scheme is applied by default to your theme templates (where no color picker option exist today). 
Here we're adding this as an option instead of always using the first scheme as the default, which made it hard to edit. 

we're adding this as an option instead of always using the first scheme as the default, which made it hard to edit. 

Instead of applying the first scheme to the `body` and `:root` we're clearly choosing to apply a color scheme based on a new color scheme picker in theme settings: 

### Other considerations
I will need to some help with the copy to make it clear what it does. As many parts of the theme have color picker on them so it won't update for those sections. 

There shouldn't be any real impact but it does provide a way for you to choose a new scheme to apply to your `body` element and apply to area where no color pickers exist (eg: page template, collection list template, etc)

### Testing steps/scenarios

- [ ] Pick a different color scheme than the first one
- [ ] Test all templates and sections to make sure it applied as expected (drawers, modals, sections, blocks)

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/157726474262/editor)

### Checklist

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -98,6 +98,13 @@
           "on_secondary_button": "secondary_button_label",
           "secondary_button_border": "secondary_button_label"
         }
+      },
+      {
+        "type": "color_scheme",
+        "id": "default_color_scheme",
+        "label": "t:settings_schema.colors.settings.default_color_scheme.label",
+        "default": "background-1",
+        "info": "Color scheme applied to places where no color picker is available (eg: templates)"
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


danielvan:
Agree to scrap it – this is a very risky one, since it applies to a vast amount of surfaces without a clear relationship to the merchant. Thanks for trying it out @ludoboludo, helped us align.

bakura10:
Just a quick note to let you know that I have done it this way for Prestige v8. Merchants do not seem confused by it. I don’t know the frequency of confused merchants in Dawn, but this approach seems to work well, from a support perspective.

PR #3045, opened 2023-10-05
Add visual representation for filters

jas7457:
### PR Summary: 

You can now use color swatches and thumbnails for your store's metaobject filters.

This will allow merchants using the Search & Discovery app a way to choose between a swatch or thumbnail visual style for certain filter types (to date: only metafield references to metaobjects).

I followed the [figma design](https://www.figma.com/file/DfFkTjUypdH1oDZSBfByNt/Storefront-Filtering-enhancements?node-id=6842%3A105934&mode=dev), and tried to componentize the visual representation so it can be used outside of the context of filters as well.

When wrapped in a `.visual-display-parent` and either `.visual-display-parent--swatch` or `.visual-display-parent--thumbnail`, hover, active, disabled, and focus states can be applied as well.

### Other considerations
Ensure that this can be reused outside of the context of filters.

It's worth noting that there is a TON of unnecessary duplication between the mobile filters and the non-mobile filters. I added a few classes to each to cut down on the duplication of css rules (`.facets__label`, `.facets-layout-grid`, `.facets-layout-list`, `.facet-checkbox__text` to name a few), but I think that we should take the time to refactor filters from the ground up at a later date to get rid of all/most of the duplication of HTML and CSS.

### Decision log

|---|---|---|---|---|
| 1 | Keep visual representation generic and not only filter-aware | Have less css/code and only allow it to work inside of the context of filters | Other teams or consumers may want to show a visual representation anywhere else in the theme. | The HTML structure and CSS gets a bit bloated with this approach, but allows for better componentization and reusability. |

In the image below, I am showing the different possible states. In order:
1. Default
2. Active
3. Hover
4. Focus
5. Disabled
6. Empty

image

### Testing steps/scenarios

Testing directions are listed [here](https://github.com/Shopify/dawn-private/issues/251).

[Theme upload](https://github.com/Shopify/dawn/files/13047474/Dawn-11.0.0.zip)

### Demo links

- [Editor](https://admin.shopify.com/store/jsonwebdev/themes/157816717334/editor?previewPath=%2Fcollections%2Fall)

### Checklist
- [x] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)

### DIFF
--- a/assets/component-facets.css
+++ b/assets/component-facets.css
@@ -272,9 +272,40 @@
   padding: 0.5rem 2rem;
 }
 
+.facets-layout-grid {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+  text-align: center;
+  padding: 2rem 2.4rem;
+  gap: 3rem 1rem;
+}
+
+.facets-layout-grid.facets__list--vertical {
+  padding: 1rem 0;
+}
+
 .facets__item {
   display: flex;
   align-items: center;
+}
+
+/* Hover/focus state */
+.facets-layout-list .facets__label:hover .facet-checkbox__text,
+.facets-layout-list input:focus ~ .facet-checkbox__text {
+  text-decoration: underline;
+}
+
+.facets-layout-grid > * {
+  align-items: flex-start;
+}
+
+.facets-layout-grid .visual-display-parent {
+  display: flex;
+  flex-direction: column;
+  gap: 0.8rem;
+  padding: 0;
+  height: 100%;
+  font-size: 1.3rem;
 }
 
 .facets__item label,
@@ -291,6 +322,18 @@
   word-break: break-word;
 }
 
+/* Hover, active, and focus states */
+:is(.facets__label:hover, .facets__label.active, .facets__label:has(:focus-visible)) {
+  color: rgba(var(--color-foreground), 1);
+}
+
+/* Focus states for older browsers */
+@supports not selector(:has(a, b)) {
+  .facets__label:focus-within {
+    color: rgba(var(--color-foreground), 1);
+  }
+}
+
 .facet-checkbox input[type='checkbox'] {
   position: absolute;
   opacity: 1;
@@ -301,6 +344,23 @@
   z-index: -1;
   appearance: none;
   -webkit-appearance: none;
+}
+
+.facets-layout-grid input[type='checkbox'] {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  height: 100%;
+  opacity: 0;
+}
+
+.facets__visual-display-wrapper {
+  display: flex;
+  justify-content: center;
+  flex-shrink: 0;
 }
 
 .no-js .facet-checkbox input[type='checkbox'] {
@@ -336,8 +396,10 @@
   }
 }
 
-.facet-checkbox--disabled {
-  color: rgba(var(--color-foreground), 0.4);
+.facet-checkbox--disabled,
+.mobile-facets__label--disabled {
+  opacity: 0.4;
+  pointer-events: none;
 }
 
 .facets__price {
@@ -744,13 +806,6 @@ details.menu-opening .mobile-facets__close svg {
 }
 
 .mobile-facets__highlight {
-  opacity: 0;
-  visibility: hidden;
-}
-
-.mobile-facets__checkbox:checked + .mobile-facets__highlight {
-  visibility: visible;
-  opacity: 1;
   position: absolute;
   top: 0px;
   left: 0px;
@@ -758,6 +813,13 @@ details.menu-opening .mobile-facets__close svg {
   bottom: 0px;
   display: block;
   background-color: rgba(var(--color-foreground), 0.04);
+  opacity: 0;
+  visibility: hidden;
+}
+
+.mobile-facets__checkbox:checked + .mobile-facets__highlight {
+  opacity: 1;
+  visibility: visible;
 }
 
 .mobile-facets__summary {
@@ -852,10 +914,6 @@ input.mobile-facets__checkbox {
 .mobile-facets__summary .icon-caret {
   margin-left: auto;
   display: block;
-}
-
-.mobile-facets__label--disabled {
-  opacity: 0.5;
 }
 
 .mobile-facets__footer {
--- a/assets/component-visual-display.css
+++ b/assets/component-visual-display.css
@@ -1,0 +1,89 @@
+.visual-display {
+  --visual-display__size: min(2.4rem, 100%);
+  position: relative;
+  width: var(--visual-display__size);
+  max-width: 100%;
+  border: 0.1rem solid rgba(var(--color-foreground), 0.2);
+  aspect-ratio: 1/1;
+}
+
+.visual-display.empty {
+  border-style: dashed;
+}
+
+.visual-display--presentation-swatch {
+  --visual-display__size: min(2.4rem, 100%);
+
+  border-radius: 100%;
+  overflow: hidden;
+}
+
+.visual-display-parent .visual-display--presentation-swatch {
+  outline-offset: 0.2rem;
+}
+
+/* Hover, active, and focus states */
+:is(
+    .visual-display-parent:hover .visual-display--presentation-swatch,
+    .visual-display-parent.active .visual-display--presentation-swatch,
+    .visual-display-parent:has(:focus-visible) .visual-display--presentation-swatch
+  ) {
+  outline-style: solid;
+}
+
+/* Active state */
+.visual-display-parent.active .visual-display--presentation-swatch {
+  outline-width: 0.1rem;
+  outline-color: rgb(var(--color-foreground), 1);
+}
+
+/* Hover state */
+.visual-display-parent:hover .visual-display--presentation-swatch {
+  outline-width: 0.2rem;
+  outline-color: rgb(var(--color-foreground), 0.4);
+}
+
+/* Focus state */
+.visual-display-parent:has(:focus-visible) .visual-display--presentation-swatch {
+  outline-width: 0.2rem;
+  outline-color: rgb(var(--color-foreground), 0.4);
+  box-shadow: 0 0 0 0.6rem rgb(var(--color-background)), 0 0 0 0.8rem rgba(var(--color-foreground), 0.5),
+    0 0 1.2rem 0.4rem rgba(var(--color-foreground), 0.3);
+}
+
+/* Focus state for older browsers */
+@supports not selector(:has(a, b)) {
+  .visual-display-parent:focus-within .visual-display--presentation-swatch {
+    outline-offset: 0.2rem;
+    outline: 0.2rem solid rgb(var(--color-foreground), 0.4);
+    box-shadow: 0 0 0 0.6rem rgb(var(--color-background)), 0 0 0 0.8rem rgba(var(--color-foreground), 0.5),
+      0 0 1.2rem 0.4rem rgba(var(--color-foreground), 0.3);
+  }
+}
+
+.visual-display-parent.disabled {
+  opacity: 0.4;
+  pointer-events: none;
+}
+
+/* Used to display the disabled dash */
+.visual-display-parent.disabled .visual-display::after {
+  display: block;
+  content: '';
+
+  /* 1.414 is not a magic number, it's the square root of 2, or the length of the diagonal */
+  width: calc(var(--visual-display__size) * 1.414);
+  border-bottom: 0.1rem solid rgb(var(--color-background-contrast));
+  transform: rotate(-45deg);
+  transform-origin: left;
+}
+
+.visual-display .visual-display__child {
+  display: block;
+  height: 100%;
+  width: 100%;
+}
+
+.visual-display--presentation-swatch .visual-display__image {
+  object-fit: cover;
+}
--- a/snippets/facets.liquid
+++ b/snippets/facets.liquid
@@ -13,6 +13,7 @@
 {% endcomment %}
 
 {{ 'component-show-more.css' | asset_url | stylesheet_tag }}
+{{ 'component-visual-display.css' | asset_url | stylesheet_tag }}
 
 {%- liquid
   assign sort_by = results.sort_by | default: results.default_sort_by
@@ -111,7 +112,20 @@
             <script src="{{ 'show-more.js' | asset_url }}" defer="defer"></script>
             {% comment %} Filters for both horizontal and vertical filter {% endcomment %}
             {%- for filter in results.filters -%}
-              {%- assign total_active_values = total_active_values | plus: filter.active_values.size -%}
+              {% liquid
+                assign total_active_values = total_active_values | plus: filter.active_values.size
+                case filter.presentation
+                  when 'swatch'
+                    assign has_visual_display = true
+                    assign show_more_number = 15
+                    assign visual_layout_class = 'facets-layout-grid facets-layout-grid--' | append: filter.presentation
+                  else
+                    assign has_visual_display = false
+                    assign visual_layout_class = 'facets-layout-list'
+                    assign show_more_number = 10
+                endcase
+              %}
+
               {% case filter.type %}
                 {% when 'boolean', 'list' %}
                   <details
@@ -157,14 +171,21 @@
                       <fieldset class="facets-wrap parent-wrap {% if filter_type == 'vertical' %} facets-wrap-vertical{% endif %}">
                         <legend class="visually-hidden">{{ filter.label | escape }}</legend>
                         <ul
-                          class="{% if filter_type != 'vertical' %} facets__list{% endif %} list-unstyled no-js-hidden"
+                          class="{{ visual_layout_class }}{% if filter_type == 'vertical' %} facets__list--vertical{% else %} facets__list{% endif %} list-unstyled no-js-hidden"
                           role="list"
                         >
                           {%- for value in filter.values -%}
-                            <li class="list-menu__item facets__item{% if forloop.index > 10 and filter_type == 'vertical' %} show-more-item hidden{% endif %}">
+                            {% liquid
+                              assign is_disabled = false
+                              if value.count == 0 and value.active == false
+                                assign is_disabled = true
+                              endif
+                            %}
+
+                            <li class="list-menu__item facets__item{% if forloop.index > show_more_number and filter_type == 'vertical' %} show-more-item hidden{% endif %}">
                               <label
                                 for="Filter-{{ filter.param_name | escape }}-{{ forloop.index }}"
-                                class="facet-checkbox{% if value.count == 0 and value.active == false %} facet-checkbox--disabled{% endif %}"
+                                class="facets__label facet-checkbox{% if is_disabled %} facet-checkbox--disabled disabled{% endif %}{% if has_visual_display %} visual-display-parent visual-display-parent--{{ filter.presentation }}{% endif %}{% if value.active %} active{% endif %}"
                               >
                                 <input
                                   type="checkbox"
@@ -174,38 +195,50 @@
                                   {% if value.active %}
                                     checked
                                   {% endif %}
-                                  {% if value.count == 0 and value.active == false %}
+                                  {% if is_disabled %}
                                     disabled
                                   {% endif %}
                                 >
 
-                                <svg
-                                  width="1.6rem"
-                                  height="1.6rem"
-                                  viewBox="0 0 16 16"
-                                  aria-hidden="true"
-                                  focusable="false"
-                                >
-                                  <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
-                                </svg>
+                                {% if has_visual_display %}
+                                  <div class="facets__visual-display-wrapper">
+                                    {% render 'visual-display',
+                                      type: value.display.type,
+                                      value: value.display.value,
+                                      presentation: filter.presentation
+                                    %}
+                                  </div>
+                                {% else %}
+                                  <svg
+                                    width="1.6rem"
+                                    height="1.6rem"
+                                    viewBox="0 0 16 16"
+                                    aria-hidden="true"
+                                    focusable="false"
+                                  >
+                                    <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
+                                  </svg>
 
-                                <svg
-                                  aria-hidden="true"
-                                  class="icon icon-checkmark"
-                                  width="1.1rem"
-                                  height="0.7rem"
-                                  viewBox="0 0 11 7"
-                                  fill="none"
-                                  xmlns="http://www.w3.org/2000/svg"
-                                >
-                                  <path d="M1.5 3.5L2.83333 4.75L4.16667 6L9.5 1"
-                                    stroke="currentColor"
-                                    stroke-width="1.75"
-                                    stroke-linecap="round"
-                                    stroke-linejoin="round" />
-                                </svg>
+                                  <svg
+                                    aria-hidden="true"
+                                    class="icon icon-checkmark"
+                                    width="1.1rem"
+                                    height="0.7rem"
+                                    viewBox="0 0 11 7"
+                                    fill="none"
+                                    xmlns="http://www.w3.org/2000/svg"
+                                  >
+                                    <path d="M1.5 3.5L2.83333 4.75L4.16667 6L9.5 1"
+                                      stroke="currentColor"
+                                      stroke-width="1.75"
+                                      stroke-linecap="round"
+                                      stroke-linejoin="round" />
+                                  </svg>
+                                {% endif %}
 
-                                <span aria-hidden="true">{{ value.label | escape }} ({{ value.count }})</span>
+                                <span class="facet-checkbox__text" aria-hidden="true">
+                                  {{- value.label | escape }} ({{ value.count }})</span
+                                >
                                 <span class="visually-hidden">
                                   {{- value.label | escape }} (
                                   {%- if value.count == 1 -%}
@@ -284,7 +317,7 @@
                           {%- endfor -%}
                         </ul>
                       </fieldset>
-                      {%- if filter.values.size > 10 and filter_type == 'vertical' -%}
+                      {%- if filter.values.size > show_more_number and filter_type == 'vertical' -%}
                         <show-more-button>
                           <button
                             class="button-show-more link underlined-link no-js-hidden"
@@ -631,6 +664,17 @@
             <div class="mobile-facets__main has-submenu gradient">
               {%- if enable_filtering -%}
                 {%- for filter in results.filters -%}
+                  {% liquid
+                    case filter.presentation
+                      when 'swatch'
+                        assign has_visual_display = true
+                        assign visual_layout_class = 'facets-layout-grid facets-layout-grid--' | append: filter.presentation
+                      else
+                        assign has_visual_display = false
+                        assign visual_layout_class = 'facets-layout-list'
+                    endcase
+                  %}
+
                   {% case filter.type %}
                     {% when 'boolean', 'list' %}
                       <details
@@ -657,12 +701,19 @@
                             {% render 'icon-arrow' %}
                             {{ filter.label | escape }}
                           </button>
-                          <ul class="mobile-facets__list list-unstyled" role="list">
+                          <ul class="{{ visual_layout_class }} mobile-facets__list list-unstyled" role="list">
                             {%- for value in filter.values -%}
+                              {% liquid
+                                assign is_disabled = false
+                                if value.count == 0 and value.active == false
+                                  assign is_disabled = true
+                                endif
+                              %}
+
                               <li class="mobile-facets__item list-menu__item">
                                 <label
                                   for="Filter-{{ filter.param_name | escape }}-mobile-{{ forloop.index }}"
-                                  class="mobile-facets__label{% if value.count == 0 and value.active == false %} mobile-facets__label--disabled{% endif %}"
+                                  class="facets__label mobile-facets__label{% if is_disabled %} mobile-facets__label--disabled disabled{% endif %}{% if has_visual_display %} visual-display-parent visual-display-parent--{{ filter.presentation }}{% endif %}{% if value.active %} active{% endif %}"
                                 >
                                   <input
                                     class="mobile-facets__checkbox"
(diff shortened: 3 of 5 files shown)


snippets/visual-display.liquid
```diff
@@ -0,0 +1,46 @@
...
+          assign rotation = '45deg'
+        else
+          assign rotation = '0deg'
+        endif
+
+        assign angle_increment = 360 | divided_by: size_limit
+        assign angle = 0
+      %}
+      {%- capture conic_gradient -%}
+				{%- for color in value limit: size_limit -%}
```
carolineschnapp, line 36: indenting too big here

patrickracicot (commented):
This is some good work.  I've only done a high level code read and quick test.  I'll test some more and deep dive the code tomorrow morning.

assets/component-facets.css
```diff
@@ -291,6 +320,18 @@
...
+:where(.facets__label:hover, .facets__label.active, .facets__label:has(:focus-visible)) {
+  color: rgba(var(--color-foreground), 1);
+}
+
+/* Focus states for older browsers */
+@supports not selector(:has(a, b)) {
+  .facets__label:focus-within {
+    color: rgba(var(--color-foreground), 1);
+  }
+}
```
patrickracicot, line 333: (question) Couldn't we just put the rule as one of the rules in the `:where` since where is a [forgiving selector parsing](https://developer.mozilla.org/en-US/docs/Web/CSS/:where#forgiving_selector_parsing)

assets/component-facets.css
```diff
@@ -272,11 +272,40 @@
...
 .facets__item {
   display: flex;
   align-items: center;
 }
 
+/* Hover/focus state */
+.facets-layout-list .facets__label:hover .facet-checkbox__text,
+.facets-layout-list input:focus ~ .facet-checkbox__text {
+  text-decoration: underline;
+}
```
patrickracicot, line 303: (remark) Although this is fine IMO, I just wanted to point out that before, the hover state on the `drawer`/`mobile` mode wouldn't apply and underline to the filter values.  Now this new selectors does have the side-effect of adding an underline style to thoses displays as well.

I do think this is a good change thought.

snippets/visual-display.liquid
```diff
@@ -0,0 +1,46 @@
...
+      {%- capture conic_gradient -%}
+        {%- for color in value limit: size_limit -%}
+          {{ color }} {{ angle }}deg{%- assign angle = angle | plus: angle_increment %} {{ angle }}deg{%- unless forloop.last %}, {%- endunless -%}
+        {%- endfor -%}
+      {%- endcapture -%}
+      <div style="background: conic-gradient({{ conic_gradient }}); transform: rotateZ({{ rotation }});"></div>
+    {%- when 'image' -%}
+      {{ value | image_url: width: 100 | image_tag: class: 'visual-display__image', alt: value.alt }}
+    {%- else -%}
+      <div></div>
```
patrickracicot, line 44: (question) Any reason you need to return this empty div ?

snippets/facets.liquid
```diff
@@ -157,55 +169,74 @@
...
+
                             <li class="list-menu__item facets__item{% if forloop.index > 10 and filter_type == 'vertical' %} show-more-item hidden{% endif %}">
                               <label
                                 for="Filter-{{ filter.param_name | escape }}-{{ forloop.index }}"
-                                class="facet-checkbox{% if value.count == 0 and value.active == false %} facet-checkbox--disabled{% endif %}"
+                                class="facets__label facet-checkbox{% if is_disabled %} facet-checkbox--disabled disabled{% endif %}{% if has_visual_representation %} visual-display-parent visual-display-parent--{{ filter.display_style }}{% endif %}{% if value.active %} active{% endif %}"
                               >
                                 <input
                                   type="checkbox"
+                                  class="{% if has_visual_representation %}visually-hidden{% endif %}"
```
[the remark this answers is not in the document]
jas7457: Weird that only happens in voiceover. It's as if it creates its own border around the element, which is likely what it's doing. To get around this, I just decided to match the height/width of the container. The actual element would go outside of this focus area because of using outline, but this is supposed to be for non-sighted uses anyways so I can't really think of a better (non-hacky) way.

image

assets/component-facets.css
jas7457: Yeah the non-mobile didn't have an underline before either. Al wanted to add it, he didn't specify if it should only be applied to desktop, but I don't think we should have different styles between the two (or else we'll continue to have spaghetti code in the future)

jas7457:
> (concern) I feel like the active/selected style is not strong enough for thumbnail displays https://screenshot.click/06-48-0c026-zjx41.png

I had a similar concern and brought it to Al and we updated it to what we have now. @al-mcquiston can you weigh in if we should make further adjustments?

assets/component-visual-display.css
```diff
@@ -0,0 +1,100 @@
...
+
+  /* 1.414 is not a magic number, it's the square root of 2, or the length of the diagonal */
+  width: calc(var(--visual-display__size) * 1.414);
+  border-bottom: 0.1rem solid rgba(var(--color-foreground), 0.4);
+  transform: rotate(-45deg);
+  transform-origin: left;
+}
+
+.visual-display > * {
+  display: block !important;
```
dan-menard, line 93: Why do you need `!important` here?

snippets/facets.liquid
```diff
@@ -157,14 +169,21 @@
...
+                          class="{{ visual_layout_class }}{% if filter_type == 'vertical' %} facets__list--vertical{% else %} facets__list{% endif %} list-unstyled no-js-hidden"
                           role="list"
                         >
                           {%- for value in filter.values -%}
+                            {% liquid
+                              assign is_disabled = false
+                              if value.count == 0 and value.active == false
+                                assign is_disabled = true
+                              endif
+                            %}
```
dan-menard, line 181: I think you can just do:

```liquid
assign is_disabled = value.count == 0 and value.active == false
```

(and if so, same thing down on line ~715)

snippets/visual-display.liquid
```diff
@@ -0,0 +1,44 @@
...
+        endif
+
+        assign rotation = '0deg'
+        if value.size == 2
+          assign rotation = '45deg'
+        endif
+
+        assign angle_increment = 360 | divided_by: size_limit
+        assign angle = 0
+      %}
```
dan-menard, line 32: Love how clear this is 👍

dan-menard (commented):
Code and 🎩 look pretty good to me.

@sofiamatulis can you take a look as well, and let me know what I missed? 😅

snippets/facets.liquid
dan-menard: Ahh right, liquid doesn't support assigning via ternaries, so I guess that extends to assigning _any_ expression, including a boolean one.

Sorry, my liquid is pretty rusty 🙈

snippets/facets.liquid
```diff
@@ -111,7 +112,18 @@
...
+                case filter.display_style
+                  when 'swatch', 'thumbnail'
+                    assign has_visual_representation = true
+                    assign visual_layout_class = 'facets-layout-grid facets-layout-grid--' | append: filter.display_style
+                  else
+                    assign has_visual_representation = false
+                    assign visual_layout_class = 'facets-layout-list'
+                endcase
+              %}
+
```
andy-schneidr, line 126: (I think the right place to put this comment is around line 145 but I can't reach ✋)
It could be just something wrong with my store, but when 🎩 ing, the filter `selected` count doesn't seem to work?
image
It was working on Dawn-11.0.0

snippets/facets.liquid
```diff
@@ -174,38 +193,50 @@
...
+                                      stroke="currentColor"
+                                      stroke-width="1.75"
+                                      stroke-linecap="round"
+                                      stroke-linejoin="round" />
+                                  </svg>
+                                {% endif %}
 
-                                <span aria-hidden="true">{{ value.label | escape }} ({{ value.count }})</span>
+                                <span class="facet-checkbox__text" aria-hidden="true">
+                                  {{- value.label | escape }} ({{ value.count }})</span
```
andy-schneidr, line 238: (question) what does the `-` in `- value.label` do?

andy-schneidr (commented):
Did some 🎩  and looked through. Just some comments from me, this is looking really cool! 😄 
A metaobject entry missing text or color/color list/image seems to be handled well!

assets/component-facets.css
```diff
@@ -744,20 +809,20 @@ details.menu-opening .mobile-facets__close svg {
...
   bottom: 0px;
   display: block;
   background-color: rgba(var(--color-foreground), 0.04);
+  opacity: 0;
+  visibility: hidden;
+}
+
+.mobile-facets__checkbox:checked + .mobile-facets__highlight {
+  opacity: 1;
+  visibility: visible;
```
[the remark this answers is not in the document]
jas7457: Oh I actually forgot that I got rid of highlight completely from the markup for the grid, but what it _was_doing before is it was an invisible element that had `gap` applied because it's now in a display flex with gap. Like I said I got rid of it from my markup later, but I still think it's beneficial to do this because we never want it to take up space, which is what this was doing before.

snippets/facets.liquid
jas7457: Yeah, blame @patrickracicot :D https://github.com/Shopify/dawn/pull/3048 - it's fixed in main, I'll rebase and squash in a bit.

snippets/facets.liquid
jas7457: It removes whitespace. Adding the class had the formatter push it to the next line, and it actually automatically added the `-` so no new white space was introduced.

andy-schneidr: Ohh, thanks! 👍

snippets/facets.liquid
andy-schneidr: Ohh, missed that whole situation, I'm glad it wasn't Just Me 😆

snippets/facets.liquid
```diff
@@ -111,7 +112,18 @@
             <script src="{{ 'show-more.js' | asset_url }}" defer="defer"></script>
             {% comment %} Filters for both horizontal and vertical filter {% endcomment %}
             {%- for filter in results.filters -%}
-              {%- assign total_active_values = total_active_values | plus: filter.active_values.size -%}
+              {% liquid
+                assign total_active_values = total_active_values | plus: filter.active_values.size
+                case filter.display_style
+                  when 'swatch', 'thumbnail'
+                    assign has_visual_representation = true
```
NathanPJF, line 119:
```suggestion
                    assign has_visual_display = true
```

snippets/facets.liquid
```diff
@@ -111,7 +112,18 @@
...
+                    assign visual_layout_class = 'facets-layout-grid facets-layout-grid--' | append: filter.display_style
+                  else
+                    assign has_visual_representation = false
```
NathanPJF, line 122:
```suggestion
                    assign has_visual_display = false
```

snippets/facets.liquid
```diff
@@ -174,38 +193,50 @@
...
-                                <svg
-                                  width="1.6rem"
-                                  height="1.6rem"
-                                  viewBox="0 0 16 16"
-                                  aria-hidden="true"
-                                  focusable="false"
-                                >
-                                  <rect width="16" height="16" stroke="currentColor" fill="none" stroke-width="1"></rect>
-                                </svg>
+                                {% if has_visual_representation %}
```
NathanPJF, line 201:
```suggestion
                                {% if has_visual_display %}
```

snippets/facets.liquid
```diff
@@ -641,6 +672,17 @@
             <div class="mobile-facets__main has-submenu gradient">
               {%- if enable_filtering -%}
                 {%- for filter in results.filters -%}
+                  {% liquid
+                    case filter.display_style
+                      when 'swatch', 'thumbnail'
+                        assign has_visual_representation = true
```
NathanPJF, line 678:
```suggestion
                        assign has_visual_display = true
```

snippets/facets.liquid
```diff
@@ -641,6 +672,17 @@
...
+                        assign visual_layout_class = 'facets-layout-grid facets-layout-grid--' | append: filter.display_style
+                      else
+                        assign has_visual_representation = false
```
NathanPJF, line 681:
```suggestion
                        assign has_visual_display = false
```

snippets/facets.liquid
```diff
@@ -667,12 +709,19 @@
...
+                                if value.count == 0 and value.active == false
+                                  assign is_disabled = true
+                                endif
+                              %}
+
                               <li class="mobile-facets__item list-menu__item">
                                 <label
                                   for="Filter-{{ filter.param_name | escape }}-mobile-{{ forloop.index }}"
-                                  class="mobile-facets__label{% if value.count == 0 and value.active == false %} mobile-facets__label--disabled{% endif %}"
+                                  class="facets__label mobile-facets__label{% if is_disabled %} mobile-facets__label--disabled disabled{% endif %}{% if has_visual_representation %} visual-display-parent visual-display-parent--{{ filter.display_style }}{% endif %}{% if value.active %} active{% endif %}"
```
NathanPJF, line 724:
```suggestion
                                  class="facets__label mobile-facets__label{% if is_disabled %} mobile-facets__label--disabled disabled{% endif %}{% if has_visual_display %} visual-display-parent visual-display-parent--{{ filter.display_style }}{% endif %}{% if value.active %} active{% endif %}"
```

snippets/facets.liquid
```diff
@@ -683,36 +732,48 @@
...
                                       checked
                                     {% endif %}
-                                    {% if value.count == 0 and value.active == false %}
+                                    {% if is_disabled %}
                                       disabled
                                     {% endif %}
                                   >
 
-                                  <span class="mobile-facets__highlight"></span>
+                                  {% if has_visual_representation %}
```
NathanPJF, line 740:
```suggestion
                                  {% if has_visual_display %}
```

snippets/facets.liquid
```diff
@@ -157,14 +169,21 @@
...
+                              if value.count == 0 and value.active == false
+                                assign is_disabled = true
+                              endif
+                            %}
+
                             <li class="list-menu__item facets__item{% if forloop.index > 10 and filter_type == 'vertical' %} show-more-item hidden{% endif %}">
                               <label
                                 for="Filter-{{ filter.param_name | escape }}-{{ forloop.index }}"
-                                class="facet-checkbox{% if value.count == 0 and value.active == false %} facet-checkbox--disabled{% endif %}"
+                                class="facets__label facet-checkbox{% if is_disabled %} facet-checkbox--disabled disabled{% endif %}{% if has_visual_representation %} visual-display-parent visual-display-parent--{{ filter.display_style }}{% endif %}{% if value.active %} active{% endif %}"
```
NathanPJF, line 186:
```suggestion
                                class="facets__label facet-checkbox{% if is_disabled %} facet-checkbox--disabled disabled{% endif %}{% if has_visual_display %} visual-display-parent visual-display-parent--{{ filter.display_style }}{% endif %}{% if value.active %} active{% endif %}"
```

snippets/visual-display.liquid
```diff
@@ -0,0 +1,44 @@
...
+      {%- capture conic_gradient -%}
+        {%- for color in value limit: size_limit -%}
+          {{ color }} {{ angle }}deg{%- assign angle = angle | plus: angle_increment %} {{ angle }}deg{%- unless forloop.last %}, {%- endunless -%}
+        {%- endfor -%}
+      {%- endcapture -%}
+      <div style="background: conic-gradient({{ conic_gradient }}); transform: rotateZ({{ rotation }});"></div>
+    {%- when 'image' -%}
+      {{ value | image_url: width: 100 | image_tag: class: 'visual-display__image', alt: value.alt }}
```
NathanPJF, line 40:
```suggestion
      {{ value | image_url: width: 300 | image_tag: class: 'visual-display__image', alt: value.alt }}
```
From our sync: The display size is ~`90`, but in order for this not to look blurry on devices with higher pixel density, (like 3.0x), you'll want to request a higher res image.  

The [Liquid `srcset` attribute](https://shopify.dev/docs/api/liquid/filters/image_tag#image_tag-srcset) doesn't allow us to specify an output like `srcset="imageurl-for-100px 1x, imageurl-for-200px 2x"`, so I recommend hardcoding `300` over `100`.

snippets/visual-display.liquid
NathanPJF: Idea from our sync: Could use a different element like `span` if you feel it gives slightly less weird CSS declarations on your end. Only reason to do that is to avoid a CSS selector from the screenshot above.

Note: We looked into `figure` + `figurecaption` instead, but that isn't the correct usage of the HTML attribute; especially in the scenario where there's a missing image.

assets/component-visual-display.css
```diff
@@ -0,0 +1,100 @@
...
+  height: 100%;
+  width: 100%;
+}
+
+.visual-display__image {
+  object-fit: cover;
+}
```
NathanPJF, line 100:
```suggestion
.visual-display__image--swatch {
  object-fit: cover;
}

.visual-display__image--thumbnail {
  object-fit: contain;
}
```

^ These classes don't exist yet.


From our sync: **Thumbnail** Images that are much taller than wide (or vice versa) will get a zoomed in look that may make it difficult to tell what the image is. Screenshot example below is showing a Fridge.  I think `contain` is better for thumbnails, even if it results in some filter images having different dimensions. However, when the images are used in a **swatch**, I agree that `cover` is the right implementation


Screenshot 2023-10-06 at 5 39 48 PM


Screenshot 2023-10-06 at 5 39 33 PM
different proportions - the merchant uploaded them that way.

NathanPJF (changes requested):
Summary of requests:
- Consistently calling it `display` over `representation`
- Image size requested from CDN to consider devices with higher pixel density
- Always show full thumbnail image when display type is `thumbnail`

assets/component-visual-display.css
jas7457: I already have top-level classes signifying the style, so I can simply do:
```css
.visual-display--style-swatch .visual-display__image {
  object-fit: cover;
}

.visual-display--style-thumbnail .visual-display__image {
  object-fit: contain;
}
```

snippets/visual-display.liquid
```diff
@@ -0,0 +1,44 @@
...
+<div class="visual-display visual-display--style-{{ style }}{% unless type == "color" or type == "colors" or type == "image" %} empty{% endunless %}">
+  {%- case type -%}
+    {%- when 'color' -%}
+      <div style="background-color: rgba({{ value.rgba }})"></div>
+    {%- when 'colors' -%}
+      {% liquid
+        assign size_limit = value.size
+        if value.size > 4
+          assign size_limit = 4
+        endif
```
NathanPJF, line 23:
```suggestion
        assign size_limit = value.size | at_most: 4
```

Galaxy brain moment: https://shopify.dev/docs/api/liquid/filters/at_most

al-mcquiston:
> > (concern) I feel like the active/selected style is not strong enough for thumbnail displays https://screenshot.click/06-48-0c026-zjx41.png
> 
> I had a similar concern and brought it to Al and we updated it to what we have now. @al-mcquiston can you weigh in if we should make further adjustments?

Weird I'm seeing something slightly  different (I think) where the contrast is a little more clear IMO. Are you still seeing the same thing @patrickracicot I know we had made some tweaks to things recently.
image

carolineschnapp (approved):

patrickracicot (approved):
Works as intended !  I do have some concerns about how subtle the selected/active state is, but I won't block the PR on this.

assets/component-facets.css
```diff
@@ -272,11 +272,40 @@
...
+.facets-layout-list .facets__label:hover .facet-checkbox__text,
+.facets-layout-list input:focus ~ .facet-checkbox__text {
+  text-decoration: underline;
+}
+
+.facets-layout-grid > * {
+  align-items: flex-start;
+}
+
 .facets__item label,
```
sofiamatulis, line 309: When showing multiple colors, it's showing as white. Is that expected? https://screenshot.click/12-00-qb4vh-ppeei.mp4

assets/component-facets.css
```diff
@@ -291,6 +320,18 @@
   word-break: break-word;
 }
```
sofiamatulis, line 322: I wonder if we have a different threshold for colors in terms of show more? For text, 10 makes sense because the list is long, but for color I feel like we could have more rows before showing `show more`?

https://screenshot.click/12-02-15hgf-y47n7.mp4

assets/component-facets.css
```diff
@@ -744,20 +810,20 @@ details.menu-opening .mobile-facets__close svg {
...
```
sofiamatulis, line 810: This is happening on mobile: https://screenshot.click/12-04-wo88b-bk2tr.mp4

assets/component-facets.css
```diff
@@ -744,20 +810,20 @@ details.menu-opening .mobile-facets__close svg {
...
   display: block;
   background-color: rgba(var(--color-foreground), 0.04);
+  opacity: 0;
+  visibility: hidden;
+}
+
+.mobile-facets__checkbox:checked + .mobile-facets__highlight {
+  opacity: 1;
+  visibility: visible;
 }
```
sofiamatulis, line 828: Is there a reason why certain colors show the opacity different when unavailable while some have the `/`?

https://screenshot.click/12-06-17qnp-wwrsu.png

assets/component-visual-display.css
sofiamatulis: Can we find a different approach instead of using `important`? 👀 

We can add a class and with specificity, itll ensure you dont need `important` and it displays as `block`

https://screenshot.click/12-14-icau3-783u9.png

assets/component-visual-display.css
```diff
@@ -0,0 +1,107 @@
...
+/* Focus state */
+.visual-display-parent:has(:focus-visible) .visual-display--style-swatch,
+.visual-display-parent--thumbnail:has(:focus-visible) {
+  outline-width: 0.15rem;
+  box-shadow: 0 0 0 0.55rem rgb(var(--color-background)), 0 0 0 0.85rem rgba(var(--color-foreground), 0.5),
+    0 0 1.15rem 0.1rem rgba(var(--color-foreground));
+}
+
+/* Focus state for older browsers */
+@supports not selector(:has(a, b)) {
```
sofiamatulis, line 68: What is `a` and `b` here? 👀

snippets/facets.liquid
```diff
@@ -157,14 +169,21 @@
                       <fieldset class="facets-wrap parent-wrap {% if filter_type == 'vertical' %} facets-wrap-vertical{% endif %}">
                         <legend class="visually-hidden">{{ filter.label | escape }}</legend>
                         <ul
-                          class="{% if filter_type != 'vertical' %} facets__list{% endif %} list-unstyled no-js-hidden"
+                          class="{{ visual_layout_class }}{% if filter_type == 'vertical' %} facets__list--vertical{% else %} facets__list{% endif %} list-unstyled no-js-hidden"
                           role="list"
                         >
```
sofiamatulis, line 174: Since this is `media` I wonder if it should follow the global styles

https://screenshot.click/12-18-7qjvv-eo3le.mp4

assets/component-facets.css
```diff
@@ -272,11 +272,40 @@
...
+
+.facets-layout-grid.facets-layout-grid--swatch {
+  gap: 3rem 1rem;
+}
+
+.facets-layout-grid.facets-layout-grid--thumbnail {
+  gap: 1.2rem;
+}
+
 .facets__item {
```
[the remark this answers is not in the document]
sofiamatulis: I see we have the special case for white, but we should ensure it works for whatever the background color is 👀 
https://screenshot.click/12-20-93gyq-vif08.mp4

sofiamatulis (commented):
This looks good 🎉 Added a few comments/questions

assets/component-facets.css
```diff
@@ -272,11 +272,40 @@
   padding: 0.5rem 2rem;
```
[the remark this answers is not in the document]
jas7457: You're looking at the wrong "Color" filter, remember you had two set up? Since you've searched for "bag", that doesn't match the product you attached the visual filter to, so it won't show up.

You could search for "*" instead and you should see it.

assets/component-facets.css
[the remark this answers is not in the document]
jas7457: Hmm we don't have any special casing for white, it should work the same no matter the background color. We put a border around it like we can see in your screenshot, and the hover/focus/active states will make it stand out as well. Do you have any suggestions for any changes? I'd think the border around the swatch/thumbnail is all we need. https://screenshot.click/12-58-s8ep1-tb2lw.mp4

assets/component-facets.css
jas7457: That's because when we set up the filter, we set the override to Image (which doesn't look like you're using).

image

Instead, you can click the pencil/edit icon and change Swatch override to Colors and save.

snippets/facets.liquid
jas7457: We talked about it inheriting from global styles, but decided for the first release, to not have it be based on anything except the button radius for thumbnails. @al-mcquiston may be able to better give background, but it was a conscious decision to not attach to global styles.

assets/component-facets.css
jas7457: Good call out. I had added some styles that are no longer necessary. Instead, I got rid of those styles and set the width to 100% (to match the grid cell) in the case that we're inside of the `.facets-layout-grid`. That way, if it's used by itself it should have the 9rem, otherwise it fills the full width.

Note: the reason this is necessary is our "default" should be 9rem (if this is used outside of filters), but that doesn't actually fill the width for our desktop filters. If we don't force it to 100% there, we will get extra spacing on the left/right, which we don't want, so I can just force it to 100% width.

image

andy-schneidr (approved):
lgtm, I think it's a little hard for me to [tell which thumbnails I've selected](https://screenshot.click/12-16-umlz2-vucu6.png) (Only thing I can notice is bolded text)
Even if it's not Just Me, I wouldn't hold the review up over it 👍

assets/component-facets.css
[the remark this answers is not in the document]
sofiamatulis: What I mean is that the border around the white is dotted while the others arent. That works when the background is white, but not when you change it to a different color 🤔 

@melissaperreault thoughts?

assets/component-facets.css
jas7457: We use the current foreground color at 40% opacity, along with setting the opacity of the image itself to 40%. Perhaps we could boost the opacity to make it more apparent, but the opacity itself also signifies it's disabled. Any suggestions?

al-mcquiston: I was noticing this too and think it's something we should fix. Is there a way we can just not apply the decreased opacity to the /?

assets/component-visual-display.css
sofiamatulis: Thanks for the clarification. Im not sure if we need this here 🤔 Is there a reason why we are using `focus-within`

In Dawn, we have used `focus` and `focus-visible` historically

https://github.com/search?q=repo%3AShopify%2Fdawn+%3Afocus&type=code

jas7457: So this is a pretty unique scenario, but I essentially want to know if _anything_ inside of `visual-display-parent` has the `:focus-visible` (thus only visible if you use your keyboard to focus, not clicking, and then style the `.visual-display` inside of it. There isn't a css pseudo class that allows this, and there's a pretty good [article](https://larsmagnus.co/blog/focus-visible-within-the-missing-pseudo-class) about it.

Essentially I have this class `.visual-display-parent:has(:focus-visible) .visual-display--style-swatch` giving us the preferred css selector, then this `@supports not` style will be the fallback for browsers that don't yet support `:has`, giving them the less ideal `:focus-within` experience.

Here's a [sandbox](https://codesandbox.io/s/holy-fast-mdnhsq?file=/src/styles.css) and a [video](https://screenshot.click/12-40-j67lj-2tk2h.mp4) showing the difference!

assets/component-facets.css
melissaperreault: - +1 for using a multiple of 3, it's great to have some rules here to structure the scannability.
- For the limit I'll let @al-mcquiston provide the guidance. 24 seems a lot considering the color labels length and how they wrap on multiple lines.

snippets/facets.liquid
jas7457: > Have we explored not inheriting at all until we have a better home for such settings?

I am aligned not doing _any_ global settings for this first iteration. I can remove the border radius.

> I would recommend increasing the bottom padding of those thumbnails to improve the white balance.

Updated to 8px, good call out.

> Have we considered reducing the label size as well to help make more room for more letters?

I updated to 1.3rem instead of 1.4rem. It's not a huge difference, but 1.2 looked way too tiny compared to the other filter text.

> Have we explored making the selected state thicker?

We've heard this feedback a few times. @al-mcquiston do you think we should make any changes since it's pretty common feedback?

mkoenig-shopify:
I know it's a bit late in the game here but I do really wonder if for v1 here we can just ship swatch and text - and then iterate if we truly need thumbnail? i know we could rename it as "croppable" or "not-croppable" but still, feels like we don't know for sure we need that yet and the complexity it adds is high

NathanPJF (approved):

sofiamatulis (approved):
Looks good! thanks for addressing the feedback 🎉 Let's just make sure we address these before merging

- https://github.com/Shopify/dawn/pull/3045#discussion_r1358395991
- https://github.com/Shopify/dawn/pull/3045#discussion_r1357329830
- https://github.com/Shopify/dawn/pull/3045#discussion_r1358511327

snippets/facets.liquid
jas7457: > We've heard this feedback a few times. @al-mcquiston do you think we should make any changes since it's pretty common feedback?

To follow up on this: I've made some changes that Al and I talked about and this has been pushed. We're chatting about the `/` a bit now and might make a couple slight tweaks to the focus, but then we should be good!

assets/component-facets.css
jas7457: Changed to use 15 for swatch (5 rows), 12 for thumbnail (4 rows)

melissaperreault:
Thank you so much @jas7457 for all the updates! 🙌 

Capturing notes from our DM conversations here:
- Contrast between the default and selected state looks good! 🎉 💪 
- I didn't get the chance to over stress test the limit before we truncate the thumbnails or swatches to reveal the show more link but I trust this was heavily tested 🙏 

### Not a blocker, but worth investigating
**Hard coded color**: Small hesitation for using a hard coded color for the disabled diagonal line. Why? Let me explain!

_**Summary:**_ Have we explored changing the brightness of the text (foreground color of a scheme) to inform the result over the thumbnails and color swatches? If we observed that text color can't be dark/light enough to be noticeable used at a certain opacity, it may be a safer solution rather than relying on a hard coded color that merchants can't change easily. This way, the result with some opacity may appear as strong as you are expecting it to be initially.

**Potential issues**
- When text color is colorful, that hard coded color can become unexpected. ([Dramatic example](https://screenshot.click/18-34-0rs9h-5a5lk.png))
- When building a theme, when we use a hard coded color we ensure it will be contrasting enough. e.g. The Pickup location checkmark white outline.

**[Debut](https://debut-theme-lightroom.myshopify.com) use case**
Depending on the background color brightness of the primary button we made the hover lighter or darker identifying a specific threshold for contrast.

See other examples: see example 10 and 11 https://www.w3.org/TR/css-color-5/#relative-colors

NathanPJF (changes requested):
Hold off as we discuss API changes

NathanPJF (approved):
🎩 'd and new API useage 👍🏻

PR #3052, opened 2023-10-09
fix(animations.js): Only set `--animation-order` css variable. Fixes #3051

### LINKED ISSUE #3051: Inline HTML styles gets overridden on data-cascade attribute when animation(reveal on scroll) is enabled.
The inline style gets overridden by js on elements with `data-cascade` attribute when reveal on scroll is enabled in theme settings.

To reproduce add some styles in  https://github.com/Shopify/dawn/blob/000a4695cada47f315b0ac0c038ea9be2589e260/sections/multicolumn.liquid#L75 anything with `data-cascade` will do.

` style="--animation-order: {{ forloop.index }};padding:1rem;"` for example,

get's overridden by  ` style="--animation-order: 1"` once the page loads.

Only the css variable, `--animation-order`, in this case should change. 
Other styles should not be affected.

- Dawn Version: main branch
- Chrome Version: 119.0.6045.10
 
## Possible solution

Do not set the whole style property in https://github.com/Shopify/dawn/blob/000a4695cada47f315b0ac0c038ea9be2589e260/assets/animations.js#L14

rawalplawit:
### PR Summary: 

Only sets `--animation-order` css variable instead of the whole style property.

Currently, the file `animations.js` sets the whole `style` property, removing all other inline styles in the element.
#3051 for more information about the issue.

### DIFF
--- a/assets/animations.js
+++ b/assets/animations.js
@@ -11,7 +11,7 @@ function onIntersection(elements, observer) {
       if (elementTarget.classList.contains(SCROLL_ANIMATION_OFFSCREEN_CLASSNAME)) {
         elementTarget.classList.remove(SCROLL_ANIMATION_OFFSCREEN_CLASSNAME);
         if (elementTarget.hasAttribute('data-cascade'))
-          elementTarget.setAttribute('style', `--animation-order: ${index};`);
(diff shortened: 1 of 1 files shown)


rawalplawit:
I have signed the CLA!

PR #3061, opened 2023-10-11
[Facets] update visual representation of facets operators

patrickracicot:
### PR Summary: 

Added visual changes to filters that have the AND operator.  The goal of this change is to add a subtitle to filters that have the `AND` operator associated to them as well as keeping the selected values for these filters grouped.

Following the introduction of the AND operator to certain filter types it was deemed to subtle to only use the dynamic counts to inform buyers when a filter is using the new logical operator.  After various tries and prototypes, the final decision was to add a subtitle as well as group the selected filter values.

### Testing steps/scenarios

- [ ] Install the [1P app Search & Discovery ](https://apps.shopify.com/search-and-discovery?search_id=f8922d74-7041-40da-92db-41957c14d188&surface_detail=search+%26+discovery&surface_inter_position=1&surface_intra_position=4&surface_type=search)
- [ ] Create a new filter using the `tags` filter type
- [ ] Change the logical operator to `AND`
- [ ] On the online store, apply the tags filter
- [ ] Ensure that the selected filter values are grouped and kept at the start of the list

### Demo links

- [Store](https://collect-payment-a11y.myshopify.com/)
- [Editor](https://admin.shopify.com/store/collect-payment-a11y/themes/157789454358/editor)

### Checklist
- [-] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)

### DIFF
--- a/assets/component-facets.css
+++ b/assets/component-facets.css
@@ -223,9 +223,32 @@
   transition: text-decoration var(--duration-short) ease;
 }
 
-.facets__summary:hover span {
+.facets__summary:hover .facets__summary-label {
   text-decoration: underline;
   text-underline-offset: 0.3rem;
+}
+
+.facets__and-helptext {
+  color: rgba(var(--color-foreground), 0.5);
+  font-size: calc(var(--font-heading-scale) * 1.2rem);
+  line-height: calc(var(--font-heading-scale) * 1.2rem);
+}
+
+@media only screen and (min-width: 750px) {
+  .facets__and-helptext {
+    font-size: calc(var(--font-heading-scale) * 1.3rem);
+    line-height: calc(var(--font-heading-scale) * 1.3rem);
+  }
+}
+
+.facets__disclosure .facets__and-helptext,
+.facets__disclosure-vertical .facets__and-helptext {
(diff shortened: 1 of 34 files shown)


dan-menard:
One thing I noticed while 🎩, this spinner in the top-right corner never stops spinning:

image

Seems to work well otherwise tho, so if that's a known issue we're good 👍

dan-menard (approved):

jas7457:
Just noticed this, seems likely it happened when you changed from div to span. There's a css rule that puts a text-decoration: underline of any span in that summary.

I think the proper fix is to not target any span, but a specific class, for this css.

image

patrickracicot:
> Just noticed this, seems likely it happened when you changed from div to span. There's a css rule that puts a text-decoration: underline of any span in that summary.
> 
> I think the proper fix is to not target any span, but a specific class, for this css.
> 

Fixed it by adding a css class to the label span and making the CSS selector use that class instead of element selectors.  This should also prevent similar issues in the future.

jas7457 (approved):
LGTM!

PR #3072, opened 2023-10-16
Quick-Order-Form Scaffolding

eugenekasimov:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


sections/quick-order-form.liquid
```diff
@@ -32,6 +32,46 @@
...
+            <use xlink:href="#icon-reset">
+          </svg>
+        </button>
+        <button type="submit" class="search__button field__button" aria-label="{{ 'general.search.search' | t }}">
+          <svg class="icon icon-search" aria-hidden="true" focusable="false">
+            <use xlink:href="#icon-search">
+          </svg>
+        </button>
+      </div>
+      {%- if section.settings.enable_filtering -%}
```
sofiamatulis, line 63: Let's remove filtering

sections/quick-order-form.liquid
```diff
@@ -67,13 +99,15 @@
             </thead>
 
             <tbody>
-              {%- render 'quick-order-list-row',
-                item: section.settings.collection.products.first,
-                image: section.settings.collection.products.first.featured_media,
-                sku: section.settings.collection.products.first.selected_or_first_available_variant.sku,
-                variant: section.settings.collection.products.first.selected_or_first_available_variant
+              {%-
```
sofiamatulis, line 102: Thinking more about it and the design we have, I dont think the product row needs to use `quick-order-list-row`. Can we create a separate snippet for it? We only need the product title and image and the quick order list row loads way more information than that

sections/quick-order-form.liquid
```diff
@@ -67,13 +99,15 @@
...
+                render 'quick-order-list-row',
+                item: collections['shop-all'].products.first,
+                image: collections['shop-all']products.first.featured_media,
+                sku: collections['shop-all']products.first.selected_or_first_available_variant.sku,
+                variant: collections['shop-all']products.first.selected_or_first_available_variant
               -%}
-              {%- for variant in section.settings.collection.products.first.variants -%}
+              <tr class="empty-row"></tr>
```
sofiamatulis, line 109: Why do we need an empty row?

locales/en.default.schema.json
```diff
@@ -2742,6 +2742,20 @@
...
+      "settings": {
+        "show_variant_image": {
+          "label": "Show variant images"
+        },
+        "show_sku": {
+          "label": "Show SKUs"
+        }
+      },
+      "presets": {
+        "name": "Quick order list"
```
sofiamatulis, line 2756:
```suggestion
        "name": "Quick order form"
```

sections/quick-order-form.liquid
eugenekasimov: that makes sense to me 👍

assets/quick-order-form.css
```diff
@@ -6,6 +6,22 @@
   background-color: rgba(var(--color-foreground), 0.02);
```
sofiamatulis, line 6: Is this hover behaviour supposed to be here? 🤔 I dont think so

https://screenshot.click/02-09-eju1a-xyvsh.mp4

assets/quick-order-form.css
```diff
@@ -6,6 +6,22 @@
...
+.quick-order-form thead th:last-child {
+  padding-right: 1rem;
+}
+
+.quick-order-form tr.empty-row {
+  height: 1rem;
+}
+
+.quick-order-form .quick-order-list__total {
+  box-shadow: 0px -15px 30px 10px rgb(var(--color-background), 0.7);
```
sofiamatulis, line 22:
```suggestion
  box-shadow: 0px -1.5rem 3rem 1rem rgb(var(--color-background), 0.7);
```

assets/quick-order-form.css
```diff
@@ -20,4 +36,95 @@
...
+  font-size: 1.4rem;
+  margin-left: 1.6rem;
+  margin-bottom: 1.8rem;
+  display: flex;
+  align-items: center;
+}
+
+.seach__filters span {
+  margin-left: 0.8rem;
+}
```
sofiamatulis, line 130: The loader isnt aligned https://screenshot.click/02-11-6vtd9-qbhu1.mp4

assets/quick-order-form.css
```diff
@@ -20,4 +36,95 @@
...
+.seach__filters {
+  white-space: nowrap;
+  line-height: calc(1 + .5 / var(--font-body-scale));
+  font-size: 1.4rem;
+  margin-left: 1.6rem;
+  margin-bottom: 1.8rem;
+  display: flex;
+  align-items: center;
+}
+
```
sofiamatulis, line 127: Is mobile part of this PR? Spacing is still there when image setting is off, search placeholder is coming outside of its box etc

https://screenshot.click/02-12-urw0v-tcf5q.mp4

sections/quick-order-form.liquid
eugenekasimov: I don't like this solution much either; however, I couldn't find a better way to achieve the desired design. According to Figma, there is a space between the product and variant elements. 
Screenshot 2023-11-02 at 1 15 57 PM

We can add the separator line by applying a `box-shadow` to the product element. However, there is no way to add space after that line. Since we have a `table` with `display: table`, we cannot use margin. We can only use `padding` for elements inside the table. Unfortunately, applying `padding` to variant elements is not working for us because we have a `hover` effect. Let me know what you think or if you have any workarounds.

As an alternative, we can simply remove that space, and it will look like this.
Screenshot 2023-11-02 at 1 13 40 PM

assets/quick-order-form.css
eugenekasimov: I didn't do anything for mobile yet. I was thinking it would be a different PR.

assets/quick-order-form.css
eugenekasimov: There is no hover effect in here, this is a background for thead.

assets/quick-order-form.css
eugenekasimov: Hm 🤔, I don't know why you got this misalignment. I noticed there is a bit of misalignment but not that huge that you see. 

https://screenshot.click/02-08-zlq1t-e2383.mp4

assets/quick-order-form.css
sofiamatulis: When I hover in the video, the background color changes

assets/quick-order-form.css
sofiamatulis: Got it! Can you update the PR description so we know the scope of testing?

assets/quick-order-form.css
eugenekasimov: Oh, but this is what we have on Figma. According to the design all variant rows should have hover effect.

assets/quick-order-form.css
eugenekasimov: done.

assets/quick-order-form.css
sofiamatulis: Given the rows aren't clickable, Im not sure if they should have a hover 🤔 cc @melissaperreault

eugenekasimov: I think it's a common approach and in this case it's juts for visual convenience, that way it's easy to see when you for example want to walk through variants and see what have you added. 

Let's @melissaperreault decide but I personally like it 🙂

melissaperreault: Let's keep the highlight! It REALLy helps to identify which variant line yo are interacting with even though it's not clickable. It helps the way finding a lot.

We'll have to adjust the padding on both sides though, something I can take a look closely to provide better guidance.

sections/quick-order-form.liquid
melissaperreault: The alternative looks good to me @eugenekasimov without the space. 👍

melissaperreault (commented):
Thank you @eugenekasimov ! Starting to look good

Here are few observations:
- [ ] Gradient support: Unsure where [it takes the shadow](https://screenshot.click/03-59-892t4-04w29.png), but shouldn't be visible there.
- [x] Media theme settings: We should inherit the same settings we do from the Quick order list section for the media. We just don't support shadow on media. [We currently don't](https://screenshot.click/03-02-cst38-ymvh8.png)
- Not sure if it was intended but we are aligned to [drop the price](https://screenshot.click/03-04-udonp-d4ski.png) on the product row. (Spoke with @nicklepine)

I'll continue other investigations! 🙌

sections/quick-order-form.liquid
eugenekasimov: Done [here](https://github.com/Shopify/dawn/pull/3072/commits/b633c12de2d2a74e739e815663328232d5f39131)

melissaperreault (commented):
- [ ] I would recommend to keep the link [on the product name like we have in cart](https://screenshot.click/03-31-k89hh-5eq8e.png) where you can link back to the more detailed product if needed. Unless it's not a recommended pattern for such a feature. 🤔

eugenekasimov:
> * [ ]  I would recommend to keep the link [on the product name like we have in cart](https://screenshot.click/03-31-k89hh-5eq8e.png) where you can link back to the more detailed product if needed. Unless it's not a recommended pattern for such a feature. 🤔

So, the link should lead us to the product page, is it right?

sections/quick-order-form.liquid
eugenekasimov: Done here [835d2a9](https://github.com/Shopify/dawn/pull/3072/commits/835d2a9e4846eb23eb34a3e656a357da0bf6724e)

melissaperreault (commented):
Thank you Eugene! We will fine tune the UI details in a follow up issue when I meet with the team. 🙏 

- [ ] I noticed the highlight on gradient background is leaving an empty container impression [shown in this video.](https://screenshot.click/07-34-bkzir-9rowh.mp4)
- [ ] May be related, the outline is shown [even if there are no variant image available](https://screenshot.click/07-29-eex96-yxtk9.png). On the Quick order list section, [we just don't display the outline.](https://screenshot.click/07-30-vzxy8-qrokw.png)

assets/quick-order-form.css
```diff
@@ -20,4 +28,115 @@
...
+.quick-order-form .field {
+  margin-bottom: 1.8rem;
+}
+
+.quick-order-form .field label {
+  opacity: 0.3;
+}
+
+.search--wrapper {
```
sofiamatulis, line 120: If this is specific only to quick order form, let's call it something like `quick-order-form-search--wrapper` to avoid confusion with the existing search

sections/quick-order-form.liquid
```diff
@@ -32,6 +32,38 @@
     class="quick-order-form page-width section-{{ section.id }}-padding"
     data-id="{{ section.id }}"
   >
+    <div class="search--wrapper">
```
sofiamatulis, line 35: Same here

sections/quick-order-form.liquid
```diff
@@ -32,6 +32,38 @@
...
+          </svg>
+        </button>
+        <button type="submit" class="search__button field__button" aria-label="{{ 'general.search.search' | t }}">
+          <svg class="icon icon-search" aria-hidden="true" focusable="false">
+            <use xlink:href="#icon-search">
+          </svg>
+        </button>
+      </div>
+      <div class="total-items">
+        <span>{{ collections['shop-all'].products.first.variants.size }} items</span>
```
sofiamatulis, line 64: **items** is hardcoded here

sections/quick-order-form.liquid
```diff
@@ -193,12 +224,29 @@
 
 {% schema %}
 {
-  "name": "Quick Order form",
+  "name": "t:sections.quick-order-form.name",
+  "limit": 1,
+  "enabled_on": {
+    "templates": ["index", "list-collections", "page", "blog", "article"]
```
sofiamatulis, line 230: Let's confirm with @nicklepine to ensure these are the correct templates

sections/quick-order-list.liquid
```diff
@@ -87,15 +87,17 @@
                   item: product,
                   image: product.featured_media,
                   sku: product.selected_or_first_available_variant.sku,
-                  variant: product.selected_or_first_available_variant
+                  variant: product.selected_or_first_available_variant,
+                  is_variant: true
```
sofiamatulis, line 91: Let's remove this

sections/quick-order-list.liquid
```diff
@@ -87,15 +87,17 @@
...
                 -%}
               {%- else -%}
                 {%- for variant in product.variants -%}
                   {%- render 'quick-order-list-row',
                     item: variant,
                     image: variant.image,
                     sku: variant.sku,
-                    variant: variant
+                    variant: variant,
+                    is_variant: true
```
sofiamatulis, line 100: Let's remove this 
```suggestion
```

snippets/quick-order-product-row.liquid
```diff
@@ -0,0 +1,40 @@
...
+  Usage:
+  {% render 'quick-order-product-row' %}
+{% endcomment %}
+
+<tr
+  class="product"
+>
+  <td class="product-item__inner">
+    <div class="product-item__title--wrapper">
+      <div class="variant-item__media">
```
sofiamatulis, line 16: Since it's a product row, unsure about the naming being `variant`. But we can change this later too 🤔

sofiamatulis (commented):
A few notes but looking good! I will do one final pass

assets/quick-order-list.js
```diff
@@ -123,7 +122,7 @@ class QuickOrderList extends HTMLElement {
       .then((response) => response.text())
       .then((responseText) => {
         const html = new DOMParser().parseFromString(responseText, 'text/html');
-        const sourceQty = html.querySelector(this.quickOrderListId);
+        const sourceQty = html.querySelector(`#${this.quickOrderListId}`);
```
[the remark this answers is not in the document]
eugenekasimov: Hey Sofia, good question. We need this changes because otherwise it's not working. When we use `querySelector` we pass an `ID` and an `ID` should be wrapped with `' '`. It's hard to see when this affects the form. If change the cart to drawer and try to change qty in the drawer you will see that the form won't be updated because `html.querySelector(this.quickOrderListId);` will be `null`.

assets/quick-order-form.css
eugenekasimov: that makes sense, thanks @sofiamatulis

sections/quick-order-form.liquid
nicklepine: Confirmed!

locales/en.default.json
```diff
@@ -359,6 +359,11 @@
       "remove_all_items_confirmation": "Remove all {{ quantity }} items from your cart?",
       "remove_all": "Remove all",
       "cancel": "Cancel"
+    },
+    "quick_order_form": {
+      "search": "Search by product, variant or SKU",
+      "search_number_items": "{{ count }} items",
+      "line_total": "Line total"
```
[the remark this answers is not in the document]
sofiamatulis: @eugenekasimov  `{{ count }} items` already exists in the Dawn repo 👀 can we refer to that line to avoid duplicating translations. Also a note that translations arent approved so lets hold off on requesting them :)

locales/en.default.json
```diff
@@ -359,6 +359,11 @@
       "remove_all_items_confirmation": "Remove all {{ quantity }} items from your cart?",
       "remove_all": "Remove all",
       "cancel": "Cancel"
+    },
+    "quick_order_form": {
+      "search": "Search by product, variant or SKU",
+      "search_number_items": "{{ count }} items",
```
sofiamatulis, line 365: @melissaperreault @eugenekasimov Im not sure if `line total` is the right language 👀 Each line is more of the variant total 

Line total is what we use in cart where discounts are applied etc. Here they arent. That may lead to confusion. Should we use something similar to quick order form?

assets/quick-order-list.js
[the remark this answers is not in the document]
sofiamatulis: Interesting 🤔 I havent looked into this yet, but doesnt that mean the existing quick order list would be broken?

assets/quick-order-form.css
```diff
@@ -6,6 +6,14 @@
   background-color: rgba(var(--color-foreground), 0.02);
```
sofiamatulis, line 6: Subtotal and total items arent updating 👀 https://screenshot.click/09-26-z1ygp-tp5ua.mp4

eugenekasimov: I was aware of it but honestly I was sure that this would be another task. I think I saw previously a task related to the total bar. But that's okay, I can fix it here.

assets/quick-order-list.js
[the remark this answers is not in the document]
eugenekasimov: It doesn't because we changed it everywhere accordingly. 

I leave a short video and I'm sorry for this complication. https://screenshot.click/09-14-ckarz-cba3c.mp4

locales/en.default.json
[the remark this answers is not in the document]
eugenekasimov: Honestly, I agree with the idea of not duplicating the same translation. However, I'm not a big fan of using translations related to different sections, even if they are the same. As you mentioned, we already have `{{ count }} items` and it's living under the header section.

I'm okay with using the existing translation, but I just want to make sure that we are comfortable with that. Thoughts? 

cc: @melissaperreault @andrewetchen

assets/quick-order-form.css
eugenekasimov: Fixed here [914d27c](https://github.com/Shopify/dawn/pull/3072/commits/914d27c8f40d8b1571ef4917e44ef52a23958466).

eugenekasimov: @sofiamatulis @andrewetchen @melissaperreault 

I'd like to clarify one thing and get your approval on it. Since the Quick Order Form is going to represent all products, the Total Bar, unlike the Quick Order List, reflects the quantity and sum of all items added to the cart. In other words, even if you add an item via the product page, it will be reflected in the Total Bar of the Quick Order Form. Please remember that we currently iterate only through one product, but the Total Bar already reflects all items in the cart. Don't be confused if you see that the data from the current form doesn't match the Total Bar.

melissaperreault:
We got a UX team jam session today and we would like to **remove the currency code from each price (it's too repetitive)** but keeping it only on the subtotal price. I see we don't have it on the subtotal price in this PR. I'll follow up in anotehr issue for the more visual changes.

- [ ] It would always be there regardless if the [Theme setting/Currency](https://screenshot.click/10-03-vq0lt-nr9vj.png) is checked. This is because that subtotal is acting like the cart and we always display currency codes in Cart and Checkout.

🙏 Thank you!

eugenekasimov:
> We got a UX team jam session today and we would like to **remove the currency code from each price (it's too repetitive)** but keeping it only on the subtotal price. I see we don't have it on the subtotal price in this PR. I'll follow up in another issue for the more visual changes.
> 
> * [ ]  It would always be there regardless if the [Theme setting/Currency](https://screenshot.click/10-03-vq0lt-nr9vj.png) is checked. This is because that subtotal is acting like the cart and we always display currency codes in Cart and Checkout.

@melissaperreault I don't see that we have a currency code either on the form or the list. 🤔 We also don't have it in the total bar. I'm a bit confused. Do we want to add the currency code to the total bar?

locales/en.default.json
katycobb: Could it just be `total`?

katycobb: Related question re: `Product subtotal` at the bottom. My understanding is that the form can list more than one product - in this case, should that just say `Subtotal`?

melissaperreault:
Let's recap the final details to ship this PR, did I get everything?! They could also be follow up if it's easier! 🙏 

- [ ] `Product subtotal` [becomes](https://screenshot.click/12-41-380w7-ayypw.png) `Subtotal`
- [ ] Display the currency code on the `Subtotal` [price](https://screenshot.click/12-48-ka9g3-tofta.png)

melissaperreault (approved):
Thank you Eugene! 🚢 
Looking forward to iterate next! 🙌

sections/quick-order-form.liquid
```diff
@@ -21,17 +21,49 @@
...
   <quick-order-list
     id="{{ section.id }}"
     class="quick-order-form page-width section-{{ section.id }}-padding"
     data-id="{{ section.id }}"
   >
+    <div class="quick-order-form__search--wrapper">
+      <div class="field">
+        <input
+          class="search__input field__input"
+          id="Search-In-Template"
```
sofiamatulis, line 34: This is an existing id for the Search template (in `main-search`). We shouldnt really be using the same id twice 👀 I know we dont allow this section in the search template, but it could be a problem in the future.

Can we rename it?

sections/quick-order-form.liquid
```diff
@@ -193,12 +225,29 @@
 
 {% schema %}
 {
-  "name": "Quick Order form",
+  "name": "t:sections.quick-order-form.name",
+  "limit": 1,
+  "enabled_on": {
+    "templates": ["index", "list-collections", "page", "blog", "article"]
```
sofiamatulis, line 231: Small nitpick. Wondering if we should have this in alphabetical order?

sofiamatulis (commented):
Looks great :) Final nitpicks and should be good to merge

sections/quick-order-form.liquid
eugenekasimov: oh, thats a good catch, I definitely missed it while I was coping it from the search. Thanks Sofia.

eugenekasimov: Fixed 👌

locales/en.default.json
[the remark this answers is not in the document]
sofiamatulis: I agree it doesnt make sense since its under header. However, we should avoid creating duplicate translations. Can we move it from the `header` to `general`?

```
      "cart_count": {
        "one": "{{ count }} item",
        "other": "{{ count }} items"
      }
```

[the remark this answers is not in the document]
eugenekasimov: Since we changed "Product subtotal" to "Subtotal" in the total bar, I'm going to use the existing "Subtotal" from {{ 'customer.order.subtotal' | t }}. This makes sense to me because, from a certain perspective, the total bar of the quick order form reflects the customer's order.

Unfortunately, I cannot say the same about {{ count }} items. The existing string is located under the header > cart_count. So the whole path looks like {{ 'sections.header.cart_count' | t }}, and even if we move it under the general section, it will still have a reference to "cart_count". This is confusing because we are going to use it for the quick order form search, and in our case, it will represent the quantity of items we render in the quick order form. What I'm trying to say is, if we want to reuse the existing {{ count }} items, I would move only "other": "{{ count }} items" to the general section, but not the whole "cart_count": { }.

It will look something like 

```
"general": {
   ...
   "other" : {
      "count_items" : "{{ count }} items"
    }
}

```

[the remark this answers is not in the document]
sofiamatulis: I think there are two things here 👀 

1. I dont think customers makes the most sense since it's a separate section and you have to be logged in. 🤔 I think in that case it also makes sense to move it.
2. I would move `"count_items" : "{{ count }} items"` to a general section in the translations and remove it from the header. This way we can use it for both the header and this section.

I think this is something we may run into again :) But I think the general rule is to
1. Avoid duplication
2. Move it to a general translation object if the string is reused somewhere else

We can take this to the channel to get other developers povs too

PR #3135, opened 2023-11-30
Improved country selector

hamideha:
### PR Summary: 

**Designer responsible:** @designrichly 
**Developer responsible:**: @hamideha 

Improves the country selectors in Dawn.

1. Adds basic search functionality
2. Shows store's popular countries at the top of the list. This uses `popular?` on the Country object in liquid which is true for at most 4 countries on a shop.
3. Improves mobile responsiveness of country selector. On mobile instead of having a dropdown, the selector is essentially a modal dialog.
4. Some UI improvements
   - Move the country to show up before the currency
   - Use a checkmark instead of underline to indicate that a country is selected

As part of the mission of increasing cross-border conversion, we decided to improve the first piece of the online store that a cross-border buyer interacts with.

User-testing was done with a prototype of the new selectors and they outperformed the current selectors. The users used the popular countries, they often searched for countries, and they commented on how the current selectors are less user-friendly than the prototype. 

The code that was added, for the most part, follows patterns already established in Dawn. Nothing unique or out of the ordinary. Possibly the only thing that is different is the search functionality (it's more of a filter than search); it uses a very naive and basic filtering approach with string matching (`.includes(string)`) 

The no-js approach to country selectors is exactly the same as it is today; using a browser-native select element.

### Decision log

Some UX decisions that were made:
|---|---|---|---|---|
| 1 | By default we only popular countries if there are more than 9 total countries and if there is more than 1 popular country |  | If there are less than 9 countries, the popular country would appear very close to its identical copy in the regular list |   |
| 2 | We only show the search box if there are more than 9 countries |  | The list would be short enough that users don't have to search |   |

1. New search box in the country selector
2. New mobile design
3. Small UI nits

### Testing steps/scenarios
Just play around with the country selector and the search functionality. There are 3 places that the country selector could show up: the header, the footer, and the announcement bar (only on desktop).

### Demo links
Store without popular countries:
- [Editor](https://admin.shopify.com/store/hamidehs-store/themes/162597994518?key=snippets%2Fcountry-localization.liquid)

Store with popular countries (Note that the popular countries are hardcoded):
- [Store](https://mylittleearthling.myshopify.com/)
- [Editor](https://potteryplants.myshopify.com/admin/themes/137609019605?key=snippets%2Fcountry-localization.liquid)

All 1st party themes:
[Sense](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694168598)
[Crave](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694201366)
[Craft](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694266902)
[Studio](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694299670)
[Taste](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694332438)
[Colorblock](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694365206)
[Ride](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694397974)
[Refresh](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694430742)
[Origin](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694463510)
[Publisher](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694496278)
[Spotlight](https://hamidehs-store.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=162694529046)

### Checklist

### DIFF
--- a/assets/component-localization-form.css
+++ b/assets/component-localization-form.css
@@ -51,12 +51,19 @@
 }
 
 .localization-form__currency {
-  display: inline-block;
+  opacity: 0;
+  transition: opacity 0.15s cubic-bezier(0.25, 0.46, 0.45, 0.94);
+  white-space: nowrap;
+  padding-right: 1.7rem;
 }
 
 @media screen and (max-width: 749px) {
   .localization-form .button {
     word-break: break-all;
+  }
+
+  .localization-form__currency {
+    opacity: 1;
   }
 }
 
@@ -149,6 +156,7 @@ noscript .localization-selector.link {
   border-radius: var(--popup-corner-radius);
   box-shadow: var(--popup-shadow-horizontal-offset) var(--popup-shadow-vertical-offset) var(--popup-shadow-blur-radius)
     rgba(var(--color-shadow), var(--popup-shadow-opacity));
+  max-height: 27.5rem;
 }
 
 .disclosure__list {
@@ -158,11 +166,153 @@ noscript .localization-selector.link {
   padding-bottom: 0.5rem;
   padding-top: 0.5rem;
   scroll-padding: 0.5rem 0;
-  min-height: 8.2rem;
-  max-height: 20.65rem;
-  max-width: 22rem;
+  max-height: 20.5rem;
+  max-width: 25.5rem;
   min-width: 12rem;
   width: max-content;
+}
+
+.country-selector__list {
+  width: 25.5rem;
+  padding-bottom: 0.95rem;
+  padding-top: 0;
+}
+
+.country-selector__close-button {
+  display: none;
+}
+
+.country-filter {
+  display: flex;
+  justify-content: end;
+  align-items: center;
+  gap: 0.5rem;
+  padding: 1.5rem 1.5rem 0.6rem;
+  position: sticky;
+  top: -0.02rem;
+  background-color: rgb(var(--color-background));
+  z-index: 6;
+}
+
+.country-filter__reset-button,
+.country-filter__search-icon {
+  right: calc(var(--inputs-border-width));
+  top: var(--inputs-border-width);
+}
+
+.country-filter__reset-button:not(:focus-visible)::after,
+.country-filter__reset-button:not(:focus)::after {
+  display: block;
+  height: calc(100% - 1.8rem);
+  content: '';
+  position: absolute;
+  right: 0;
+}
+
+.country-filter__reset-button:focus,
+.country-filter__reset-button:focus-visible {
+  background-color: rgb(var(--color-background));
+  z-index: 4;
+}
+
+.country-filter__reset-button:not(:focus-visible):not(.focused) {
+  box-shadow: inherit;
+  background-color: inherit;
+}
+
+.country-filter__reset-button:hover .icon {
+  transform: scale(1.07);
+}
+
+.country-filter__reset-button .icon.icon-close,
+.country-filter__search-icon .icon {
+  height: 1.8rem;
+  width: 1.8rem;
+  stroke-width: 0.1rem;
+}
+
+.country-filter__search-icon {
+  transition: opacity var(--duration-short) ease, visibility var(--duration-short) ease;
+  opacity: 1;
+  visibility: visible;
+  pointer-events: none;
+  cursor: text;
+}
+
+.country-filter__search-icon--hidden {
+  opacity: 0;
+  visibility: hidden;
+}
+
+@media screen and (max-width: 749px) {
+  .disclosure__list-wrapper.country-selector:not([hidden]) + .country-selector__overlay:empty {
+    display: block;
+  }
+
+  .country-selector__overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: rgba(var(--color-foreground), 0.5);
+    z-index: 3;
+  }
+
+  .disclosure__list-wrapper.country-selector {
+    position: fixed;
+    bottom: -1rem;
+    left: 0;
+    width: 100%;
+    height: 80%;
+    max-height: 80vh;
+    border-radius: 0;
+    border: none;
+    box-shadow: none;
+    z-index: 4;
+  }
+
+  .disclosure__list.country-selector__list {
+    max-height: 85%;
+    min-width: 100%;
+    margin-bottom: 0;
+    z-index: 5;
+  }
+
+  .country-selector__close-button.link {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    padding: 0rem;
+    background-color: transparent;
+    height: 4.4rem;
+    width: 4.4rem;
+  }
+
+  .country-selector__close-button .icon {
+    width: 2rem;
+    height: 2rem;
+  }
+
+  .country-filter {
+    padding: 1.5rem 2rem 0.6rem 1.5rem;
+  }
+}
+
+@media screen and (min-width: 750px) {
+  .country-filter--no-padding {
+    padding: 0;
+  }
+}
+
+.countries {
+  padding-top: 0.6rem;
+}
+
+.popular-countries {
+  border-bottom: 1px solid rgba(var(--color-foreground), 0.2);
+  padding-bottom: 0.6rem;
+  padding-top: 0.6rem;
 }
 
 .disclosure__item {
@@ -170,8 +320,10 @@ noscript .localization-selector.link {
 }
 
 .disclosure__link {
-  display: block;
-  padding: 0.95rem 3.5rem 0.95rem 2rem;
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  gap: 0.5rem;
+  padding: 0.9rem 1.5rem;
   text-decoration: none;
   line-height: calc(1 + 0.8 / var(--font-body-scale));
   color: rgba(var(--color-foreground), 0.75);
@@ -182,8 +334,16 @@ noscript .localization-selector.link {
   color: rgb(var(--color-foreground));
 }
 
-.disclosure__link--active {
-  text-decoration: underline;
+.disclosure__link .icon-checkmark {
+  width: 1rem;
+  margin-right: 0.2rem;
+}
+
+.disclosure__link:hover .localization-form__currency,
+.disclosure__link:focus-visible .localization-form__currency,
+.disclosure__link:focus .localization-form__currency {
+  display: inline-block;
+  opacity: 1;
 }
 
 .disclosure__button[aria-expanded='true'] .icon-caret {
--- a/assets/component-menu-drawer.css
+++ b/assets/component-menu-drawer.css
@@ -263,3 +263,11 @@ details[open].menu-opening > .menu-drawer__submenu {
 .menu-drawer .list-social__link {
   padding: 1.1rem 1.1rem;
 }
+
+@media screen and (max-width: 749px) {
+  .menu-drawer.country-selector-open {
+    transform: none !important;
+    filter: none !important;
+    transition: none !important;
+  }
+}
--- a/assets/localization-form.js
+++ b/assets/localization-form.js
@@ -6,12 +6,32 @@ if (!customElements.get('localization-form')) {
         super();
         this.elements = {
           input: this.querySelector('input[name="locale_code"], input[name="country_code"]'),
-          button: this.querySelector('button'),
+          button: this.querySelector('button.localization-form__select'),
           panel: this.querySelector('.disclosure__list-wrapper'),
+          search: this.querySelector('input[name="country_filter"]'),
+          closeButton: this.querySelector('.country-selector__close-button'),
+          resetButton: this.querySelector('.country-filter__reset-button'),
+          searchIcon: this.querySelector('.country-filter__search-icon'),
+          liveRegion: this.querySelector('#sr-country-search-results'),
         };
-        this.elements.button.addEventListener('click', this.openSelector.bind(this));
-        this.elements.button.addEventListener('focusout', this.closeSelector.bind(this));
         this.addEventListener('keyup', this.onContainerKeyUp.bind(this));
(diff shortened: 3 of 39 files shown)


designrichly:
I'll hold off my review until I've synced with @kjellr , but one thing I'd like to test is seeing the search icon on the unfocused state.

hamideha:
I have signed the CLA!

sections/footer.liquid
```diff
@@ -247,7 +248,7 @@
   {%- endif -%}
 
   <div
-    class="footer__content-bottom{% if settings.animations_reveal_on_scroll %} scroll-trigger animate--slide-in{% endif %}"
+    class="footer__content-bottom"
```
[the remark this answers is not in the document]
hamideha: @Roi-Arthur thanks for the suggestion! I explored your method and the issue is that the country selector would still be bound by the height of the footer. A screen with this resolution is very unlikely, but in the image below I set the height so that more content other than the footer is visible and setting the overlay's height to 100vh doesn't actually take 100% of the viewport but rather 100% of the footer's height.

Screenshot 2023-12-11 at 12 53 01 PM

hamideha:
@ludoboludo @eugenekasimov @andrewetchen Is there any way to expedite the review on this PR? I'm really sorry for the quick notice but we'd like to merge this by the Editions cutoff and I head off for the holidays on Wednesday.

sections/footer.liquid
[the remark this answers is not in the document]
Roi-Arthur: Uh, interesting 🤔
In my example the footer didn't take the whole screen (you can see the Subscribe panel but also the end of the section above, with the Shop all button.

In order to preserve the animation for desktop, would you then instead consider changing the CSS?
Currently [Dawn has this on line 3289](https://github.com/hamideha/dawn/blob/hamideha/selector/assets/base.css#L3289)
```css
.scroll-trigger.animate--slide-in {
    transform: translateY(2rem);
  }
```

We could add something to disable that transform in the footer and on mobile? Using the !important doesn't feel great but is necessary because the animation css kicks in and overrides it otherwise 😞
```css
.scroll-trigger.animate--slide-in {
    transform: translateY(2rem);
  }

  .footer__content-bottom.scroll-trigger.animate--slide-in {
    transform: none !important;
  }
}
```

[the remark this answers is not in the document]
hamideha: Ah yes that works and is a cleaner solution. Thank you!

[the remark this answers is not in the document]
ludoboludo: So am I understanding correctly the comments and the animation class should be re added ?

assets/component-localization-form.css
```diff
@@ -182,8 +329,16 @@ noscript .localization-selector.link {
   color: rgb(var(--color-foreground));
 }
 
-.disclosure__link--active {
-  text-decoration: underline;
+.disclosure__link .icon-checkmark {
+  width: 1rem;
+  margin-right: 0.2rem;
```
ludoboludo, line 334: This is working for the country selection but not language. We should be adding the same icon to the language picker otherwise it's hard to tell what is currently selected: 
<details>
<summary>Screenshot</summary>


</details>

assets/component-menu-drawer.css
```diff
@@ -263,3 +263,11 @@ details[open].menu-opening > .menu-drawer__submenu {
...
   padding: 1.1rem 1.1rem;
 }
+
+@media screen and (max-width: 749px) {
+  .menu-drawer.country-selector-open {
+    transform: none !important;
+    filter: none !important;
+    transition: none !important;
+  }
+}
```
ludoboludo, line 273: I don't think this was meant to be applied for the menu drawer country/language selectors but for the footer one 🤔

snippets/country-localization.liquid
```diff
@@ -5,41 +5,129 @@
...
+        {% render 'icon-close' %}
+      </button>
+    </div>
+    <div
+      class="disclosure__list country-selector__list"
+      id="{{ localPosition }}-country-results"
+      aria-live="polite"
+    >
+      {% if localization.available_countries.size > 9 and popular_countries.size > 1 %}
+        <ul role="list" class="list-unstyled popular-countries">
```
ludoboludo, line 80: I wonder if there would be some value in having a clear label as to what this list is compared to the second one. When I use a screen reader, I could see it being useful to understand the difference 🤔

snippets/country-localization.liquid
```diff
@@ -5,41 +5,129 @@
...
+        type="button"
+        aria-label="{{ 'accessibility.close' | t }}"
+      >
+        {% render 'icon-close' %}
+      </button>
+    </div>
+    <div
+      class="disclosure__list country-selector__list"
+      id="{{ localPosition }}-country-results"
+      aria-live="polite"
```
ludoboludo, line 77: Something that I notice as well is that when I open the country selector, it reads all the current list elements in there. 
So I open it, then it starts telling me all the currencies 😅 
The `aria-live="polite"` seem to be the issue. I think we could have it worked differently. Right now if I type as well something in the search, it's reading out all the list to me... I wonder if that is useful really to a user. Maybe if it's just a hidden div we update with a text saying `list updated` that isn't as invasive as mentioning all the items in the list. 

Not too sure, what's your take on it ?

assets/component-menu-drawer.css
hamideha: One of the UX feedbacks we received regarding the country selector in the menu drawer is that we should also overlay the header so that we don't see 2 close buttons, since clicking the menu drawer's close button would also close the country selector and this might be a source of confusion. These styles allow us to do that. 

Screenshot 2023-12-14 at 8 45 30 AM

assets/component-localization-form.css
hamideha: Good catch. Thank you!

sections/footer.liquid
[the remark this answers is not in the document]
hamideha: I haven't added them back because testing on other WebKit browsers, specifically on iOS versions lower than 17 
```css
.footer__content-bottom.scroll-trigger.animate--slide-in {
    transform: none !important;
}
```
doesn't work. I'm still trying to figure out so I haven't added them back yet.

eugenekasimov (commented):
In Safari the desktop selectors in the header and footer look different and not elegant in my opinion, especially in the footer because the search is located on the top and you cannot see it until you scroll all the way up.
https://screenshot.click/14-46-m19ev-tw4pe.mp4

snippets/country-localization.liquid
hamideha: @ludoboludo which screen reader are you using? Using Mac's voice over I can't reproduce it reading all the countries and currencies in the list when typing.

hamideha: I agree, though, that it's not a friendly experience to have all the countries and currencies read out and will change it.

ludoboludo: Yeah it was with Voice Over on Safari. Were you doing voice over and chrome ? 🤔 Ill try again. 
Here is [a recording](https://screenshot.click/15-13-l3qaq-hpewi.mp4)

eugenekasimov:
I found one more small issue that only occurs in Safari.

When you open the country selector, whether in the header or footer, and start scrolling, you may notice that the list of countries extends slightly beyond the top edge of the main container. 
https://screenshot.click/22-52-ekjde-uhvkj.mp4

snippets/country-localization.liquid
```diff
@@ -5,41 +5,143 @@
...
-        <li class="disclosure__item" tabindex="-1">
-          <a
-            class="link link--text disclosure__link caption-large focus-inset{% if country.iso_code == localization.country.iso_code %} disclosure__link--active{% endif %}"
-            href="#"
-            {% if country.iso_code == localization.country.iso_code %}
-              aria-current="true"
-            {% endif %}
-            data-value="{{ country.iso_code }}"
+  <div class="disclosure__list-wrapper country-selector" hidden>
+    <div class="country-selector__actions {% unless show_country_filter %}no-padding{% endunless %}">
```
ludoboludo, line 39: I think a class name that would be a "modifier" following BEM convention would be better and less generic. 

```suggestion
    <div class="country-selector__actions {% unless show_country_filter %}country-selector__actions--no-padding{% endunless %}">
```

It's definitely longer but at least not as generic. If it was a class we use in different spots for the same outcome it would make sense but for now I think it's only used here.

assets/section-footer.css
```diff
@@ -68,6 +68,12 @@
     row-gap: 1.5rem;
     justify-content: center;
   }
+
+  .footer__content-bottom.scroll-trigger.animate--slide-in {
+    animation: none;
+    opacity: 1;
+    transform: inherit;
+  }
```
ludoboludo, line 76: Why are we removing the animation on the mobile footer again?

snippets/language-localization.liquid
```diff
@@ -30,7 +30,12 @@
             {% endif %}
             data-value="{{ language.iso_code }}"
           >
-            {{ language.endonym_name | capitalize }}
+            <span class="{% if language.iso_code != localization.language.iso_code %}visibility-hidden{% endif %}">
```
ludoboludo, line 33: Nitpick. So we don't output an empty class attribute. 
```suggestion
            <span {% if language.iso_code != localization.language.iso_code %}class="visibility-hidden"{% endif %}>
```

assets/component-localization-form.css
```diff
@@ -140,7 +147,8 @@ noscript .localization-selector.link {
   border-width: var(--popup-border-width);
   border-style: solid;
   border-color: rgba(var(--color-foreground), var(--popup-border-opacity));
-  overflow: hidden;
+  overflow-x: hidden;
+  overflow-y: auto;
```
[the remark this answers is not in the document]
hamideha: Ahh thanks for pointing this out! The solution here was to simply use the `keydown` event instead of `keyup` for the up and down arrows.

hamideha:
@ludoboludo PR is ready for review again. I've addressed your comments but let me know if there's anything else that should be fixed.

assets/component-localization-form.css
```diff
@@ -158,20 +167,164 @@ noscript .localization-selector.link {
...
   min-width: 12rem;
+  max-width: 25.5rem;
   width: max-content;
 }
 
+.disclosure__list.country-selector__list {
+  width: 25.5rem;
+  padding-bottom: 0.95rem;
+  padding-top: 0;
+}
```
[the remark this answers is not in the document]
hamideha: @ludoboludo both these issues should be fixed now.

assets/localization-form.js
```diff
@@ -6,28 +6,91 @@ if (!customElements.get('localization-form')) {
...
 
       hidePanel() {
         this.elements.button.setAttribute('aria-expanded', 'false');
         this.elements.panel.setAttribute('hidden', true);
+        if (this.elements.search) {
+          this.elements.search.value = '';
+          this.filterCountries();
+          this.elements.search.setAttribute('aria-activedescendant', '');
+        }
+        document.body.classList.remove('overflow-hidden-mobile');
```
ludoboludo, line 47: You will probably need to update this one as well to make sure it's added when needed and not on tablet sizes.

assets/base.css
```diff
@@ -3125,7 +3129,8 @@ details-disclosure > details {
   .select__select:focus,
   .customer .field input:focus,
   .customer select:focus,
-  .localization-form__select:focus.localization-form__select:after {
+  .localization-form__select:focus.localization-form__select:after,
+  .country-filter__input:focus {
```
ludoboludo, line 3133: Here I feel like we don't really need any of those `country-filter__input` added to those CSS declarations since it's already getting it from `field__input` no ? Or am I missing something 🤔

ludoboludo (commented):
Looks like we might need to check in with the team that works on the translation platform. As we need to make sure the translations are back in before merging the PR.

assets/base.css
hamideha: Yep you're 100% right. Initially I wasn't using `field__input` and I just never thought of removing these once I did start using them.

assets/localization-form.js
```diff
@@ -44,15 +107,76 @@ if (!customElements.get('localization-form')) {
           'aria-expanded',
           (this.elements.button.getAttribute('aria-expanded') === 'false').toString()
         );
+        document.body.classList.add('overflow-hidden-mobile');
```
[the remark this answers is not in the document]
hamideha: The way I ended up fixing this is by only adding `overflow-hidden-mobile` if `overflow-hidden-tablet` isn't already a class on the body.

assets/localization-form.js
hamideha: This line removes the class so I think it's fine as is

tyleralsbury (commented):
There are two things that I noticed.  I wouldn't say that either are blockers, but I would at least create issues to follow up on them.

---

When you have sticky header open and open up the dropdown in the footer on mobile, the header is over the background.  It can still be clicked to close the modal, but if you click the logo it will go to the homepage.  Ideally clicking out of the modal would just close it.

Screenshot 2024-01-04 at 8 58 53 AM

---

Secondly, because of its similarity in behaviour to the predictive search bar, I believe we should add support for arrow keys to go through the results.  For an example, make a search in the predictive search bar and then use your up and down arrow keys.  You'll see a highlight and can hit enter to go to the corresponding page.

This is nice for UX and for accessibility.

ludoboludo (approved):
I think we're looking good 👍 
Thanks for addressing the feedback

PR #3151, opened 2023-12-13
Remove empty space from inventory status

kjellr:
### PR Summary: 

Removes empty space created when the inventory status block is present, but a product has untracked inventory. 

Fixes Shopify/temp-project-mover-Archetypically-20250409144733#155

I removed the CSS that was preventing this space from collapsing. This was initially there to present layout shift, but @vernalkick and I feel that the layout shift is a reasonable tradeoff in this case. 

~I also added some info text as recommended by @sampede. This could use some wordsmithing if we'd like to include it (cc @katycobb!).~ _Edit: We decided to remove this for now](https://github.com/Shopify/dawn/pull/3151#issuecomment-1854489466)._ 👍 

When the space collapses, it would be great to animate this so it's not so sudden. But I think that'll be way easier to accomplish with view transitions, so I've left it out for now. 

Before:

https://github.com/Shopify/dawn/assets/1202812/7f27e5f4-5020-4a24-9b46-e26f3d3ccb46

After:

https://github.com/Shopify/dawn/assets/1202812/dab43f57-8d75-4a00-90eb-c0c128f65862

### Testing steps/scenarios

1. Turn off track inventory on product/variant page and click save
2. Open the editor
3. Open the product template and view the product that you aren't tracking inventory for
4. Add the Inventory Status block, note that it takes up no space. 
5. If the product has another variant that _does_ have inventory tracked, note that there's a layout shift when switching between the two variants.

### Demo links

- [Editor](https://os2-demo.myshopify.com/admin/themes/162667528214/editor?previewPath=%2Fproducts%2F3-variants%3Fview%3D__DEFAULT__)

### Checklist
- [] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)


kjellr:
> At a glance, I’m hesitant to add this help text. It’s quite abstract to describe specific setting combinations that live elsewhere in admin, plus it seems logical that this would only appear if you’ve opted to track inventory? I think my preference would be to not add it.

Yeah, I think you're right. In any case, I think it's ok to not do it in this PR and see how it goes. I'll remove for now — Thanks for the quick response!

sampede:
Thanks for looking into this further, @kjellr! 👏

In [this issue](https://github.com/Shopify/dawn/issues/3149) you mentioned: 
> This is to prevent layout shift a situation like this, where Continue selling when out stock is set, and one variant has inventory in stock, and another does not:

From what I was observing, `Continue selling when out of stock` shows the `In stock` message without a number. So if you switched between a variant that has that turned on to one that is out of stock, it should keep that space filled with content. It'd just go from `In stock` to `Out of stock`.

The same solution could be leveraged for `Track quantity` being turned off as that'd also cause infinite selling. So we could leverage the `In stock` message without a number. Then if another variant of that product has `Track quantity` turned on, it could display the messages, `In stock`, `Out of stock`, and `Low stock` with the number as appropriate.

sampede:
@kjellr Great questions! 

> * Is it misleading to specifically say a product is "In Stock" if we don't actually know for sure that it's in stock? This is a minor point, but if I see a bright green dot that says something's in stock when I buy it, I think I'm going to be slightly more frustrated if that turns out to be untrue afterwards.

If the merchant isn't tracking quantities then the they are infinitely selling. Shopify assumes it's in stock infinitely. So the buyer can keep buying. Making it show `In stock` is just making that clearer with how that functionality works. It shouldn't turn out to be unavailable to buyers unless merchants misunderstand the `Track quantity` functionality. 

> * If we make this change, any merchants who are currently not tracking inventory on their products, but _do_ have the Inventory Status block active will suddenly have a green "In stock" message on all those products. Are we ok with this?

This is really just aligning with the functionality that is actually happening, so I don't see a big risk in surfacing this `In stock` information. 

Let me know if you'd like to hop on a quick sync and chat this out to align!

tyleralsbury:
@kjellr this looks like a nice quick change to add to an upcoming release.  Are you looking to follow up on it and finalize things?

kjellr:
Yeah  — I originally left this PR and tried the suggestion from @sampede above, but the experience still felt a little weird to me for [the same reasons I mentioned above](https://github.com/Shopify/dawn/pull/3151#issuecomment-1856195735). I think in general, this quick change of collapsing the space is a reasonable step for now, and we can always decide to modify the behavior again in the future. I'll sync this up with `main` and let's review it. 👍

PR #3165, opened 2023-12-20
[Visual Display] Display accurate filter colors when high contrast is enabled

alisterdev:
### PR Summary: 

Display accurate filter swatch colors when high contrast mode is enabled

@NathanPJF did some research to determine if it makes sense to override high contrast settings specifically to display colors on swatches.

Since swatches are non-text content for our filter controls, overriding contrast settings does not hurt readability and hence is acceptable in terms of a11y while providing better UX in general.

[Source](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/)

Apply `forced-color-adjust: none;`

### Other considerations

Safari does not support `forced-color-adjust: none;`, but iOS/macOS provides different a11y tools to improve contrast, so this change should not have any negative effects on iOS/macOS users.
  
Filters that represent colors will now show colors accurately

| Before | After |
| :----: | :---: |
|  image |  image.  |

### Testing steps/scenarios

- [ ] Try out color filters with high contrast mode enabled in different browsers

### Demo links

- [Store](https://bandzsupply.tk/collections/all)

### Checklist

### DIFF
--- a/assets/component-visual-display.css
+++ b/assets/component-visual-display.css
@@ -82,6 +82,7 @@
   display: block;
   height: 100%;
   width: 100%;
+  forced-color-adjust: none;
 }
 
 .visual-display--presentation-swatch .visual-display__image {


jas7457 (approved):
Thanks for the change. Just making sure - where you added this change it will affect both colors and images (they both use this class). Since we aren't setting a background color for images, I assume it's fine, but just wanted to make sure we didn't want this to specifically apply to colors for any reason.

alisterdev:
>I assume it's fine, but just wanted to make sure we didn't want this to specifically apply to colors for any reason.

Yeah, the images are already unaffected by the high contrast settings, so adding this property on an image simply doesn't have any effect.

ludoboludo (approved):
Without having too much context on it I had no idea what `visual-display` is for 😅 it's pretty vague at first glance. 
Though not the topic here. 
Tested it and looked good 👍 

You can invert colors in Mac OS's accessibility settings and that impacts the color of the swatch 🤔 I wonder if it's something we should worry about or not 🤷 

I saw this mentioned on [stack overflow](https://stackoverflow.com/questions/38104388/how-to-detect-mac-os-inverted-color-mode-in-javascript-css): 
```
@media (inverted-colors) {
    img.cancel-inversion {
        filter: invert(1);
    }
}
```

alisterdev:
>You can invert colors in Mac OS's accessibility settings and that impacts the color of the swatch 🤔 I wonder if it's something we should worry about or not 🤷

This is probably another can of worms that would require its own research and potential PR 😅

assets/component-visual-display.css
```diff
@@ -82,6 +82,7 @@
   display: block;
   height: 100%;
   width: 100%;
+  forced-color-adjust: none;
```
NathanPJF, line 85: I was surprised this was working outside of a `@media` query for `forced-colors: active`, but I tophatted in Chrome and Edge and it all works. Thanks for the update.

NathanPJF (approved):
Thanks for the change @alisterdev. For posertity, this is the relevant section of [the linked blog post](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/#:~:text=In%20this%20case%2C%20these%20swatches%20have%20meaning%20to%20the%20user%20and%20their%20original%20colors%20should%20be%20preserved.)

> However, the palette swatches were adjusted to use the user’s purple theme, as well. Usually this is desirable for legibility. **In this case, these swatches have meaning to the user and their original colors should be preserved.**

Issue #2483, opened 2023-04-02
[Request] The need for a "unique_id" generation filter in Liquid


bakura10:
[Disclaimer] This issue is not directly related to Dawn, but as always I don't have any better place to do platform suggestion, and this one would benefit to Dawn as well.

Hi,

Michael from Maestrooo.

As you may have heard, our theme Impact suddenly became unusable after Shopify rolled on April 1st an internal change to the platform. I would like to use this issue to give some details and suggest an improvement that should make it easier for Dawn as well to have a more robust code, and avoid the same trouble that we had to face.

## The problem of ID generation

In a lot of place in a Shopify theme, we have to deal with ID generation. Such examples include form input ID, popover or drawer ID...

To do that, one typical approach in Liquid is to generate an ID like this and hoping to make it unique:

```
{% capture id %}input-{{ section.id }}-{{ product.id }}{% endcapture %}
```

However, this uniqueness can be hard to achieve under some context. For instance, a merchant may be using a featured collections section featuring two collections, with the two collection having the same product. In such cases, we would end up with the same ID.

Another issue is for generating ID in the filters. Let's take for instance the use case of collection filters in Dawn: https://github.com/Shopify/dawn/blob/main/snippets/facets.liquid#L173

Dawn uses the following pattern: `Filter-{{ filter.param_name | escape }}-{{ forloop.index }}`

In our theme for instance, we have a more complicated design where the facets are outputted twice: one in a sidebar, and once in a drawer. We therefore have isolated the code inside a "facets.liquid" snippet that gets re-used. So here, the Dawn approach for instance would result in having duplicate ID, which would cause problem in the case of checkboxes, for instance.

Of course, we could add a more complex ID generation rule, but it can gets very cumbersome to achieve uniqueness. Even more if, as in our theme, we abstract things in more granular snippet (for instance, Dawn encourages copy-pasting by always having to re-type the whole input with all its classes, while our theme abstract this into a snippet for cleaner consumption):

```
{%- render 'input', name: filter_param.name, value: filter_param.value -%}
```

You can't also rely on things like value, as value may contained non ASCII character, space... or whatever that are not usable in ID. Handleizing them is not a possibility neither as Japanese kanji for instance cannot be converted to handle.

When using this granularity, providing a unique ID becomes extremely hard. Dawn encourage copy-pasting over composition, which I do not feel is the best approach, and Dawn would face the same problems that we are facing if they change their architecture in the future.

## The approach I found

[...]

PR #3180, opened 2024-01-08
[Variant Picker] Add swatch display type

alisterdev:
### PR Summary: 

Update `variant-picker` with support for two additional display types: a swatch and a dropdown with a swatch. These will render a swatch only for options that support swatches.

Fixes https://github.com/Shopify/custom-data-issues/issues/1847

- Extract product variant CSS from main-product into its own CSS file
- Unify `variant-select` and `variant-radio` under a single custom element that supports both use cases
- Created a small `swatch` snippet that renders a swatch-drop
- Created a `swatch-input` snippet to represent an input that renders a swatch
- Update dropdown to render a swatch when supported
- Updated variant-picker to use swatch settings for options that support swatches

- A new settings block is added under Variant Picker section
- Product options that support swatches would be configured using this new block

### Testing steps/scenarios

- [ ] Ensure that variant picker works as before and that new swatch block has no effect on products not configured with swatches
- [ ] Ensure that swatch settings work correctly for products configure with swatches.

### Demo links

- [Store](https://shop1.shopify.d1.alex-ilea.us.spin.dev/?preview_theme_id=8)
- [Editor](https://admin.web.d1.alex-ilea.us.spin.dev/store/shop1/themes/8/editor)

### Checklist

### DIFF
--- a/assets/component-swatch.css
+++ b/assets/component-swatch.css
@@ -7,7 +7,7 @@
   aspect-ratio: 1 / 1;
   background: var(--swatch--background);
   background-size: cover;
-  border: 0.1rem solid rgba(var(--color-foreground), 0.2);
+  border: 0.1rem solid rgba(var(--color-background-contrast), 0.5);
   border-radius: var(--swatch--border-radius);
 }
 
@@ -16,6 +16,5 @@
 }
 
 .swatch--unavailable {
-  --swatch--background: #ffffff;
   border-style: dashed;
 }
--- a/assets/component-swatch.css
+++ b/assets/component-swatch.css
@@ -1,0 +1,21 @@
+.swatch {
+  --swatch--size: var(--swatch-input--size, 4.4rem);
+  --swatch--border-radius: var(--swatch-input--border-radius, 50%);
+
+  display: block;
+  width: var(--swatch--size);
+  aspect-ratio: 1 / 1;
+  background: var(--swatch--background);
+  background-size: cover;
+  border: 0.1rem solid rgba(var(--color-foreground), 0.2);
+  border-radius: var(--swatch--border-radius);
+}
+
+.swatch--square {
+  --swatch--border-radius: var(--swatch-input--border-radius, 0.2rem);
+}
+
+.swatch--unavailable {
+  --swatch--background: #ffffff;
+  border-style: dashed;
+}
--- a/assets/component-swatch.css
+++ b/assets/component-swatch.css
@@ -1,3 +1,4 @@
+/* swatch lives in its own file for reusability of the swatch in swatch-input and dropdown */
 .swatch {
   --swatch--size: var(--swatch-input--size, 4.4rem);
   --swatch--border-radius: var(--swatch-input--border-radius, 50%);
@@ -17,4 +18,5 @@
 
 .swatch--unavailable {
   border-style: dashed;
+  border-color: rgba(var(--color-foreground), 0.5);
 }
--- a/sections/featured-product.liquid
+++ b/sections/featured-product.liquid
@@ -6,6 +6,7 @@
 {{ 'component-rating.css' | asset_url | stylesheet_tag }}
 {{ 'component-volume-pricing.css' | asset_url | stylesheet_tag }}
 {{ 'component-product-variant-picker.css' | asset_url | stylesheet_tag }}
+{{ 'component-swatch.css' | asset_url | stylesheet_tag }}
 
 {%- style -%}
   .section-{{ section.id }}-padding {
--- a/sections/featured-product.liquid
+++ b/sections/featured-product.liquid
@@ -7,6 +7,7 @@
 {{ 'component-volume-pricing.css' | asset_url | stylesheet_tag }}
 {{ 'component-product-variant-picker.css' | asset_url | stylesheet_tag }}
 {{ 'component-swatch.css' | asset_url | stylesheet_tag }}
+{{ 'component-swatch-input.css' | asset_url | stylesheet_tag }}
 
 {%- style -%}
   .section-{{ section.id }}-padding {
--- a/sections/featured-product.liquid
+++ b/sections/featured-product.liquid
@@ -804,6 +804,10 @@
       "limit": 1,
       "settings": [
         {
+          "type": "header",
+          "content": "t:sections.featured-product.blocks.variant_picker.settings.options.header"
+        },
+        {
           "type": "select",
           "id": "picker_type",
           "options": [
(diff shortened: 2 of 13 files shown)


snippets/product-variant-options.liquid
```diff
@@ -45,10 +47,25 @@
...
+    else
+      assign swatch_value = nil
+    endif
+  -%}
+
+  {%- capture input_id -%}
+    {{ section.id }}-{{ option.position }}-{{ forloop.index0 }}
+  {%- endcapture -%}
+
+  {%- capture radio_input -%}
```
lhoffbeck, line 65: nit -- if radio_input is only used once below, do we need to wrap it in a `capture`?

sections/featured-product.liquid
```diff
@@ -5,6 +5,9 @@
 {{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
 {{ 'component-rating.css' | asset_url | stylesheet_tag }}
 {{ 'component-volume-pricing.css' | asset_url | stylesheet_tag }}
+{{ 'component-product-variant-picker.css' | asset_url | stylesheet_tag }}
+{{ 'component-swatch.css' | asset_url | stylesheet_tag }}
+{{ 'component-swatch-input.css' | asset_url | stylesheet_tag }}
```
ludoboludo, line 10: Might be worth wrapping in a conditional if the product doesn't have variants 🤔 Or something to track for later follow ups. 
On the product object there is `has_online_default_variant` which we could use so we don't load unused styling ([docs](https://shopify.dev/docs/api/liquid/objects/product#product-has_only_default_variant)).

snippets/product-variant-options.liquid
```diff
@@ -45,10 +47,47 @@
...
+  {%- liquid
+    if value.swatch.image
+      assign image_url = value.swatch.image | image_url: width: 50
+      assign swatch_value = 'url(' | append: image_url | append: ')'
+    elsif value.swatch.color
+      assign swatch_value = 'rgb(' | append: value.swatch.color.rgb | append: ')'
+    else
+      assign swatch_value = nil
+    endif
+  -%}
```
ludoboludo, line 59: I wonder if that should just be part of the previous `{%- liquid ...` content 🤔 

Though maybe it's nice to differentiate the two goals.

sections/main-product.liquid
```diff
@@ -9,6 +9,9 @@
   {{ 'component-slider.css' | asset_url | stylesheet_tag }}
   {{ 'component-rating.css' | asset_url | stylesheet_tag }}
   {{ 'component-deferred-media.css' | asset_url | stylesheet_tag }}
+  {{ 'component-product-variant-picker.css' | asset_url | stylesheet_tag }}
+  {{ 'component-swatch-input.css' | asset_url | stylesheet_tag }}
+  {{ 'component-swatch.css' | asset_url | stylesheet_tag }}
```
ludoboludo, line 14: I'm wondering if we should actually load this CSS within the snippet itself 🤔  the `component-swatch.css`. Since it's specific to that one snippet if I'm not mistaking.

eugenekasimov (approved):
Apart from those minor feedback/suggestions that have been left, everything works as expected. 👍

snippets/product-variant-options.liquid
```diff
@@ -45,10 +47,47 @@
...
+
+  {%- capture input_id -%}
+    {{ section.id }}-{{ option.position }}-{{ forloop.index0 }}
```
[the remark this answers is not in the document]
alisterdev: Moving it to assign above makes it a really verbose declaration (because we need to use `append` to build the output). It's a good point about the newline that gets appended, fixing it using rstrip. (i.e `...forloop.index0 -}}`

snippets/product-variant-options.liquid
alisterdev: Updated!

sections/main-product.liquid
alisterdev: As discussed on Slack, we don't want to move it inside the snippet because it will try to fetch this CSS file for every instance of a swatch-input (or swatch) instead of once per page

sections/featured-product.liquid
tyleralsbury: When there is more than one Featured Product section, we will likely render the file reference more than once anyway.  But that said, I think it's okay to not check for whether or not it's configured.

snippets/swatch.liquid
```diff
@@ -0,0 +1,31 @@
...
+
+{%- liquid
+  assign swatch_value = nil
+  if swatch.image
+    assign image_url = swatch.image | image_url: width: 50
+    assign swatch_value = 'url(' | append: image_url | append: ')'
+  elsif swatch.color
+    assign swatch_value = 'rgb(' | append: swatch.color.rgb | append: ')'
+  endif
+-%}
```
ludoboludo, line 22: And this we needed here so it can be easily reusable right ? Cause we also have the similar declaration in `product-variant-options.liquid`

assets/component-swatch.css
```diff
@@ -0,0 +1,22 @@
+/* swatch lives in its own file for reusability of the swatch in swatch-input and dropdown */
+.swatch {
+  --swatch--size: var(--swatch-input--size, 4.4rem);
+  --swatch--border-radius: var(--swatch-input--border-radius, 50%);
+
+  display: block;
+  width: var(--swatch--size);
+  aspect-ratio: 1 / 1;
+  background: var(--swatch--background);
```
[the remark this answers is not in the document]
alisterdev: I used `--` to separate the variable "main" name from CSS property it represents. I think this is a pretty common approach, but I didn't notice if this was done differently in the rest of Dawn

snippets/swatch.liquid
alisterdev: The declaration in `product-variant-options.liquid` is needed specifically for the dropdown to work, because we need to dynamically update the swatch value on every dropdown change.

The block in `swatch.liquid` is reusable, but it's static

snippets/product-variant-options.liquid
```diff
@@ -43,12 +45,47 @@
...
+    elsif value.swatch.color
+      assign swatch_value = 'rgb(' | append: value.swatch.color.rgb | append: ')'
+    else
+      assign swatch_value = nil
+    endif
   -%}
 
-  {%- if block.settings.picker_type == 'button' -%}
+  {%- capture input_id -%}
+    {{ section.id }}-{{ option.position }}-{{ forloop.index0 -}}
```
bakura10, line 60: This will break if you embed multiple form in a section (think of a listing of products with quick view, a shop the look section allowing to highlight multiple products)... This emphasize the importance of my proposal here: https://github.com/Shopify/dawn/issues/2483

PR #3175, opened 2024-01-04
Improved country selectors

hamideha:
# Original PR with reviews and discussion https://github.com/Shopify/dawn/pull/3135
### PR Summary: 

**Designer responsible:** @designrichly 
**Developer responsible:** @hamideha 

Improves the country selectors in Dawn.

1. Adds basic search functionality
2. Shows store's popular countries at the top of the list. This uses `popular?` on the Country object in liquid which is true for at most 4 countries on a shop.
3. Improves mobile responsiveness of country selector. On mobile instead of having a dropdown, the selector is essentially a modal dialog.
4. Some UI improvements
   - Move the country to show up before the currency
   - Use a checkmark instead of underline to indicate that a country is selected

### Why are these changes introduced?

As part of the mission of increasing cross-border conversion, we decided to improve the first piece of the online store that a cross-border buyer interacts with.

User-testing was done with a prototype of the new selectors and they outperformed the current selectors. The users used the popular countries, they often searched for countries, and they commented on how the current selectors are less user-friendly than the prototype. 

### What approach did you take?

The code that was added, for the most part, follows patterns already established in Dawn. Nothing unique or out of the ordinary. Possibly the only thing that is different is the search functionality (it's more of a filter than search); it uses a very naive and basic filtering approach with string matching (`.includes(string)`) 

The no-js approach to country selectors is exactly the same as it is today; using a browser-native select element.

### Decision log

Some UX decisions that were made:
| # | Decision | Alternatives | Rationale | Downsides |
|---|---|---|---|---|
| 1 | By default we only popular countries if there are more than 9 total countries and if there is more than 1 popular country |  | If there are less than 9 countries, the popular country would appear very close to its identical copy in the regular list |   |
| 2 | We only show the search box if there are more than 9 countries |  | The list would be short enough that users don't have to search |   |

### Visual impact on existing themes
1. New search box in the country selector
2. New mobile design
3. Small UI nits

### Testing steps/scenarios
Just play around with the country selector and the search functionality. There are 3 places that the country selector could show up: the header, the footer, and the announcement bar (only on desktop).

### Demo links
Store without popular countries:
- [Store](https://hamidehs-store.myshopify.com/)
- [Editor](https://admin.shopify.com/store/hamidehs-store/themes/162597994518?key=snippets%2Fcountry-localization.liquid)

Store with popular countries (Note that the popular countries are hardcoded):
- [Store](https://mylittleearthling.myshopify.com/)
[...]

PR #3183, opened 2024-01-09
Focus search on country selector open and fix iOS bug

hamideha:
### PR Summary: 

Follow up to https://github.com/Shopify/dawn/pull/3175

Sets focus to search field when country selector is open on desktop screens only and adds stacking context to banner

In order to match the behaviour of the predictive search of focusing the field when opened.

The stacking context is added because on Webkit/iOS browsers there's a bug (maybe feature? it looks similar to [this](https://bugs.webkit.org/show_bug.cgi?id=153852) which seems to be resolved) that when the keyboard is open overflow: hidden on document.body doesn't work. Since the banner in Dawn didn't belong to a stacking context, you would get this bug:

https://github.com/Shopify/dawn/assets/54807699/9c36fa79-0a38-4096-b96f-f917209a3dde

By adding a z-index this is the result:

https://github.com/Shopify/dawn/assets/54807699/68f2c810-39af-425a-a3f5-c18522d357c2

### Demo links

- [Editor](https://admin.shopify.com/store/hamidehs-store/themes/162662645782?key=assets%2Flocalization-form.js)

### DIFF
--- a/assets/localization-form.js
+++ b/assets/localization-form.js
@@ -117,7 +117,7 @@ if (!customElements.get('localization-form')) {
         if (!document.body.classList.contains('overflow-hidden-tablet')) {
           document.body.classList.add('overflow-hidden-mobile');
         }
-        if (this.mql.matches) {
+        if (this.elements.search && this.mql.matches) {
           this.elements.search.focus();
         }
         document.querySelector('.menu-drawer').classList.add('country-selector-open');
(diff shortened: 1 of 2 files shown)


eugenekasimov (approved):
works as expected 🚀

tyleralsbury (approved):
Looks good. 👍

PR #3188, opened 2024-01-11
Prevent sticky header from hiding when localization selector is open

hamideha:
### PR Summary: 

We've received feedback that the localization selectors (country and language selectors) do not behave the same way as other disclosures in the sticky header. Specifically, when scrolling, the selector cuts off and and the sticky header becomes hidden.

https://github.com/Shopify/dawn/assets/54807699/e650b9cc-e703-4bfb-b697-059a7398f255

Since we only want to prevent the header from hiding when the selector in the header (not in the footer or announcement bar) is open, I added a data attribute `data-prevent-hide` to the `localization-form` that indicates it's in the header and thus prevent it from hiding.

### Demo links

- [Editor](https://admin.shopify.com/store/hamidehs-store/themes/162662645782?key=sections%2Fheader.liquid)

The header scroll behaviour when the country or language selector is open should be exactly the same when a menu item is open.

### DIFF
--- a/assets/component-localization-form.css
+++ b/assets/component-localization-form.css
@@ -257,6 +257,7 @@ noscript .localization-selector.link {
     height: 100%;
     background-color: rgba(var(--color-foreground), 0.5);
     z-index: 3;
+    animation: animateLocalization var(--duration-default) ease;
   }
 
   .disclosure__list-wrapper.country-selector {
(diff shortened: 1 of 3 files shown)


melissaperreault (approved):
Thank you! 🚢 

Different subject, I noticed a small detail, not a blocker though; the popover flashes on mobile when you interact with it, is it just me? [Video reference](https://screenshot.click/11-12-bs8tj-u8e21.mp4)

tyleralsbury (approved):
Code looks good.   A quick test and it works as expected. 👍

PR #3198, opened 2024-01-17
Fix translation path

ludoboludo:
A translation path was changed when it shouldn't have. Reverting it back to the right thing.


tyleralsbury (approved):
Verified against path in the en schema in locales folder.

Issue #3207, opened 2024-01-20
Update Dawn to use the Cart Ajax API instead of undocumented endpoints


bakura10:
Hi,

For a few months, we have been experiencing **a ton** of issues and regressions, happening at the Cart Ajax API level, sometimes on super basic features like "removing a line item from cart" which fails. We are have found today such a regression on a merchant store. The Cart Ajax API has been extremely unreliable to use lately.

Those issues are extremely frustrating to handle on our end because despite being Shopify issues, merchants say that they reached Shopify, that Shopify support tried on Dawn, and because it works on Dawn, then it must be a theme issue (and, of course, merchants will always tend to believe Shopify rather than us).

One thing that exhibits the problem is that Dawn itself uses **undocumented APIs** that appears to follow different paths on Shopify backend (and that seems to work while the documented API does not). If Dawn theme would use the documented API, it would help a lot to demonstrate to merchants that the issue is indeed due to Shopify and not the theme. Shopify eating its own food would also ensure that such cart issues are spotted and fixed faster.

For instance, when removing a product from the cart will do a POST Ajax request to the endpoint `/cart/change`, with a JSON payload:

<img width="573" alt="image" src="https://github.com/Shopify/dawn/assets/1198915/75b9c108-02ee-468c-badf-df19a9c2b575">

<img width="703" alt="image" src="https://github.com/Shopify/dawn/assets/1198915/48390383-98f2-46af-a8c8-7cd816d1f132">

However, this API is undocumented. The only documented one is `/cart/change.js` (notice the .js at the end): https://shopify.dev/docs/api/ajax/reference/cart#post-locale-cart-change-js

Same thing when changing line item quantity, where Dawn uses undocumented `/cart/change` with JSON payload instead of `cart/change.js`:

<img width="559" alt="image" src="https://github.com/Shopify/dawn/assets/1198915/c43399ca-388f-4eab-97ba-1abf24aef119">

When adding to the cart, Dawn also uses an Ajax call to `/cart/add` instead of the documented `cart/add.js` endpoint (although, contrary to the line item removal, for this one Dawn uses a form data so it would be the same as submitting the form without Ajax ;but still, as soon as you're doing Ajax call Dawn should use the documented APIs for that work).

resistorsoftware:
Did you ever ask yourself what the dotjs means at the end of the url? Are you sure there is a difference between a POST to /cart/change and /cart/change.js? 

[...]

PR #3208, opened 2024-01-20
Update cart API usage

bakura10:
### PR Summary: 

This PR addresses #3207 and changes the usage of line to key.

There are several reasons why I am doing those changes.

#### Changing of the endpoint

As I explained in #3207, the usage of `/cart/change` with JSON payload is undocumented. The only reference is the `/cart/change.js` endpoint (https://shopify.dev/docs/api/ajax). I remember having spotted some inconsistencies between the .js and non .js endpoint in the past, although this might have been changed. In all cases, using a behavior not documented anywhere is unpredictable, can lead to confusion, and makes our life harder when we want to compare it with Dawn behavior to confirm if an issue is coming from the platform or the theme.

If Shopify decides to reject this PR, then the documentation of the Ajax API should be updated first to indicate that .js and non .js endpoints are exactly equivalent (which is why I never used them on our themes).

#### Using key instead of line

Shopify is using the `url_to_remove` Liquid variable for the non-JS mode (https://github.com/Shopify/dawn/blob/main/sections/main-cart-items.liquid#L281). This Liquid feature generates a URL using the line item key (rather than the line position), so this PR unifies it to ensure that the key is used everywhere.

One main advantage of the line key over the line ID is that it no longer depends on the order of line items in the cart. One of the reasons we switched to `key` on our themes a long time ago is merchants install apps or write custom code that dynamically adds items to the cart without necessarily refreshing the drawer/cart page. Doing so shifts the indexes (as the newly added product in Ajax becomes first) and can cause the incorrect items to be updated/removed. Using the `key` instead ensures that the code is entirely independent of the order and makes the theme more robust when interacting with the cart.

The other reason (the main one, actually) is that in the last few months, the Ajax API in Shopify has been **extremely unreliable**. A high number of regressions and bugs for super basic features has been introduced, such as the impossibility of removing a bundle product from the cart, the impossibility of removing a product that has been split with automatic discounts, and yesterday I found another issue where Shopify generates _different line item keys_ when the user changes pages, even though the cart content has not changed. Those issues were mainly only reproducible using the `key` (but not for the simpler line position that Dawn is using currently).

As all our themes at Maestrooo use the line item `key` to interact with the cart API (which is, by the way, a documented feature in the Ajax API that we should be able to rely on), we end up in a complex situation where it seems that Shopify simply does not properly test the Cart Ajax API when new features are released, simply because Dawn is not using it.
Unfortunately, when a merchant reaches Shopify support to report such problems, the debugging process of Shopify support is as basic as "it works on Dawn, so the theme is faulty; reach your theme developer"... And then we have to struggle sometimes for weeks to try to escalate the issue and have Shopify recognizing the issue is indeed on their end.

Considering how critical the cart Ajax API is, I am hoping that updating Dawn to also use the key approach (which is by itself superior and more flexible than the line position) will force Shopify to better test, identify, and fix the cart Ajax API sooner, and stop hiding themselves behind the very frustrating support position that "it works on Dawn, so it's not our responsibility." Ultimately, the merchants always suffer the most from those situations.

### Checklist

### DIFF
--- a/sections/main-cart-items.liquid
+++ b/sections/main-cart-items.liquid
@@ -79,7 +79,7 @@
 
               <tbody>
                 {%- for item in cart.items -%}
-                  <tr class="cart-item" id="CartItem-{{ item.index | plus: 1 }}">
+                  <tr class="cart-item" id="CartItem-{{ item.key | handle }}">
                     <td class="cart-item__media">
                       {% if item.image %}
                         {% comment %} Leave empty space due to a:empty CSS display: none rule {% endcomment %}
@@ -262,8 +262,9 @@
                                 step="{{ item.variant.quantity_rule.increment }}"
                                 {% # theme-check-enable %}
                                 aria-label="{{ 'products.product.quantity.input_label' | t: product: item.product.title | escape }}"
-                                id="Quantity-{{ item.index | plus: 1 }}"
-                                data-index="{{ item.index | plus: 1 }}"
+                                id="Quantity-{{ item.key | handle }}"
+                                data-handleized-key="{{ item.key | handle }}"
+                                data-key="{{ item.key }}"
                               >
                               <button class="quantity__button no-js-hidden" name="plus" type="button">
                                 <span class="visually-hidden">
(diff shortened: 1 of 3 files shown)


sections/main-cart-items.liquid
```diff
@@ -274,8 +275,9 @@
                             </quantity-input>
                           </div>
                           <cart-remove-button
-                            id="Remove-{{ item.index | plus: 1 }}"
-                            data-index="{{ item.index | plus: 1 }}"
+                            id="Remove-{{ item.key | handle }}"
+                            data-handleized-key="{{ item.key | handle }}"
```
bakura10, line 279: The reason it is handleized here is because the key contains the ":" character, which is invalid character when used as a CSS selector.

PR #3217, opened 2024-01-23
Quick Add bulk

sofiamatulis:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


assets/quick-add-bulk.js
```diff
@@ -0,0 +1,159 @@
...
+      })
+  }
+
+  getSectionsToRender() {
+    return [
+      {
+        id: `quick-add-bulk-${this.dataset.id}`,
+        section: document.getElementById('product-grid').dataset.id,
+        selector: `#quick-add-bulk-${this.dataset.id}`
+      },
```
eugenekasimov, line 119: Working on the progress bar I faced a few difficulties that makes my code more complex because of we re render these elements. That made me questioned the sense of it. 
These elements contain input and buttons. Do we really need to re-render them? At a quick glance all that should be updated there is qty but it is updated anyway without re-rendering.

Do I missing something?

sofiamatulis: Thanks for sharing Eugene! Yes we do, one of reasons for the re-render is because we need the updated cart qty in the qty input `data-cart-quantity`. Otherwise you are updating the value, but not getting the updated element with the updated items in cart

What difficulties are you facing? Could you share a bit more?

dan-menard:
A couple questions/notes from testing:

(1) When we open the quick add modal, should we auto-focus the first input? I feel like it would give a faster input experience.

(2) If I pick a product with lots of variants, like "Puppy", then enter a quantity in the first line and hit enter every half-second or so to skip down the list, my keyboard loses focus. I think this is happening when the cart amount is rendered. (Happens on quick order list as well.)

sofiamatulis:
> (2) If I pick a product with lots of variants, like "Puppy", then enter a quantity in the first line and hit enter every half-second or so to skip down the list, my keyboard loses focus. I think this is happening when the cart amount is rendered. (Happens on quick order list as well.)

What do you mean exactly? I add something to cart and move quickly and cant seem to reproduce https://screenshot.click/26-07-810lt-h16zf.mp4

sofiamatulis:
> (1) When we open the quick add modal, should we auto-focus the first input? I feel like it would give a faster input experience.

We could! Good idea 😄  I would consider this an enhancement. @scottgmeadows thoughts?

assets/quantity-popover.js
```diff
@@ -48,32 +46,45 @@ if (!customElements.get('quantity-popover')) {
...
+        if (!isButtonChild && !isPopoverChild) {
           button.setAttribute('aria-expanded', 'false');
           button.classList.remove('quantity-popover__info-button--open');
           this.popoverInfo.setAttribute('hidden', '');
+          this.infoButtonDesktop.classList.remove('quantity-popover__info-button--icon-only--animation')
         }
 
         this.eventMouseEnterHappened = false;
+
+
```
dan-menard, line 87: Could 🔥 these extra new lines.

dan-menard (approved):
No surprises on the code side 👍 

Agree that we can tackle the UX improvements as enhancements.

dan-menard:
> @dan-menard What do you mean exactly? I add something to cart and move quickly and cant seem to reproduce https://screenshot.click/26-07-810lt-h16zf.mp4

Hmmm I'm having trouble reproducing it now too. I noticed I had a _lot_ of stuff in my cart, it might've been related to the cart quantity bug or something. Let's not worry about it for now.

scottgmeadows (approved):

PR #3246, opened 2024-02-07
Support Combined Listings and 2k Variants

lhoffbeck:
## Purpose

This PR adds support for [Combined Listings](https://www.shopify.com/editions/winter2024#shopify-combined-listings) and [products with 2000 variants](https://www.shopify.com/editions/winter2024#apis-for-2000-variants) to the Dawn theme.

## What are we trying to solve?

Both Combined Listings and 2k variants products are surfaced via liquid as normal products. However, because of performance concerns, [product.variants](https://shopify.dev/docs/api/liquid/objects/product#product-variants) only returns a max variant count of the [shop's variant limit](https://shopify.dev/docs/api/admin-graphql/latest/objects/ShopResourceLimits#field-shopresourcelimits-maxproductvariants). So for example, if a combined listing product has 2k variants because the parent product has 20 variants but each child has 100 variants, but the shop only supports a max of 100 variants per product, calling product.variants in the storefront would only return 100 of the 2k total variants.

See the [shopify docs](https://shopify.dev/docs/themes/product-merchandising/variants/combined-listings) for more background on updating themes to support combined listings.

### Overall requirements
We need to build support for:
- Products with 2000 variants
- Products that have more than 3 options
- Switching between different products on a PDP

## How are we solving it?

The primary issue is that we need to load **_way_** too much data for our current architecture to support. Dawn currently loads all variants for a product and dumps them as JSON into the page, which the theme JS later uses to help with switching between selected variants. Even if we scale down what variant info is dumped to JSON, we run into performance issues once there's a significant number of variants (Shopify devs -- reach out and I can share internal perf metrics). Among a number of other approaches, we also explored a variant availability hash that would contain variantID and availability for a combination of option values. While this is better than dumping full variants, you still end up with way too much data, in addition to creating a parallel data structure for themes to read option value information from.

Given that we can't use the existing architecture, we came up with a few principles from a Dawn perspective:
- Page must be performant, not be significantly data bloated or slow to load
- Keep the behavior as similar to the existing behavior as possible
- Minimize the number of theme changes
- Handle all products (normal or 2k variants) as similarly as possible in the theme

We can't load all variant information and dump it all into the page. What we _really_ need is some form of pagination. However, normal page- or cursor-based pagination doesn't really make sense in a PDP context because the buyer doesn't need to proceed through variants sequentially. We need something else...

### Proposal

When you're on a PDP that has multiple options, you typically can't access all variants from the current state. For example:

<img width="1106" alt="Screenshot 2024-01-31 at 10 50 51 AM" src="https://github.com/Shopify/dawn/assets/3619012/72f73d2c-6d36-4d2f-a7b7-2aea40d02514">

In the example above, the buyer has  Orange / s / Simple / Cotton  selected. The set of variants a buyer is shown depends on which option value they select. In this example, there are eight variants a buyer could choose next:

- Color is changed: [2 other colors] / s / Simple / Cotton 
- Size is changed: Orange / [2 other sizes] / Simple / Cotton 
Style is changed: Orange / s / [2 other styles] / Liner
Material is changed: Orange / s / Simple / [2 other materials]

In other words, the total number of variants that a merchant can "paginate" to next is determined by the currently selected option values. 


#### Liquid changes

We have made the following changes to the Liquid API to support combined listings and 2k variants:

[product\_option](https://shopify.dev/docs/api/liquid/objects/product_option) (ProductOptionDrop)

|       |           |                                    |                                                           |
| ----- | --------- | ---------------------------------- | --------------------------------------------------------- |
| **#** | Attribute | Relevant API information and Notes | More informations                                         |
| **1** | values    | `product_option.values`            | Returns a `product_option_value` (ProductOptionValueDrop) |

[product\_option\_value](https://shopify.dev/docs/api/liquid/objects/product_option_value) (ProductOptionValueDrop) – NEW

|       |              |                                    |                                                                                                                                                                                                                             |
| ----- | ------------ | ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **#** | Attribute    | Relevant API information and Notes | More informations                                                                                                                                                                                                           |
| **1** | id           | `value.id`                         | Id of the option value.                                                                                                                                                                                                     |
| **2** | name         | `value.name`                       | Name of the option value. For instance, we have a T-shirt with a given size of small, medium and large. `value.name `would return “S”, “M”, or “L”                                                                          |
| **3** | selected     | `value.selected`                   | To validate whether the current option value is the selected variant.Previously, it was used as [product\_option.selected\_value](https://shopify.dev/docs/api/liquid/objects/product_option#product_option-selected_value) |
| **4** | swatch       | `value.swatch`                     | To be used by swatches.                                                                                                                                                                                                     |
| **5** | variant      | `value.variant`                    | The variant that’s associated with this option value, taking all of the other selected option values into consideration.                                                                                                    |
| **6** | product\_url | `value.product_url`                | The product URL for the option value. This value may change for different option values if the product is a combined listing.                                                                                               |

[product](https://shopify.dev/docs/api/liquid/objects/product) (ProductDrop)

|       |           |                                    |                                                                                                                                                                                         |
| ----- | --------- | ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **#** | Attribute | Relevant API information and Notes | More informations                                                                                                                                                                       |
| **1** | variants  | `product.variants`                 | [Same behaviour](https://shopify.dev/docs/api/liquid/objects/product#product-variants) as today, but we will cap the number up to a maximum length of the shop's product variant limit. |



#### Dawn changes

There are a few main changes in this PR:

**global.js**
- Instead of dumping all of a product's variants to JSON, availability is determined using the Section API.
- When switching between a product (Combined Listings case), the entire product info section is replaced. This is done because the ProductInfo template contains conditionally included content ([example](https://github.com/Shopify/dawn-private/blob/975ac8a9c3fd17625cdbb6bb833c35825c647a00/sections/main-product.liquid#L44-L61)) that may not be present on the newly loaded product. We could do this replacement piecemeal, but it is a brittle approach because it would require anyone making changes to main-product.liquid to make the requisite "replacement destination with source" change in global.js. We will continue to investigate how we can minimize layout shift, preload new product images, and a11y concerns.
- ProductRecommendations was updated to refresh if the associated main product is swapped to a combined listing sibling.

**product-variant-options.liquid**
- Option values are rendered using only options_with_values and the new information on the ProductOptionValueDrop. We no longer need to iterate through all variants to determine availability and enabledness.
- Inputs have 3 new data attributes for `product-url`, `available`, and `option-value-id`
- Variant info is dumped to JSON for each option value input

**product-variant-picker.liquid**
- Remove dumping all product.variants to JSON


### UX changes from the current implementation

While we've tried to make the UX change as little as possible, by adding pagination there is _some_ divergence.

<a id="availability"></a>

**Availability is no longer top-down**
Availability in Dawn is currently top-down. In other words, an option value is available if there is at least one subsequent option value that is available, recursively.

In the new approach, availability is based on the currently selected variant, meaning that option value availability responds when a new option value is selected. For more information on this approach, see [the dev docs](/docs/themes/product-merchandising/variants/combined-listings#ux-considerations-availability).

**Latency**
Because all product variants aren’t being dumped into JSON anymore, the theme JS no longer has the ability to [set availability](https://github.com/Shopify/dawn/blob/07c1828d978d8a655545596c2a95f90468a42323/assets/global.js#L967) of the next set of option values prior to the Section API call. There will be a brief latency while the section API call resolves.

`Update 2023-12-07`: The team caught up with @tyleralsbury and are aligned that this latency is not a showstopper.

### Impact of limiting product.variants
Product.variants in liquid has been limited to a max count of the [shop's variant limit](https://shopify.dev/docs/api/admin-graphql/latest/objects/ShopResourceLimits#field-shopresourcelimits-maxproductvariants). This impacts our themes in a few areas:

**Search result relevancy**
To support better [search results relevancy](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data), we currently dump all of a product's variants to JSON in [featured-product.liquid](https://github.com/Shopify/dawn/blob/2c847245ea3dfec6f248dcbb2dd76c4ceb65c437/sections/featured-product.liquid#L636-L657) and [main-product.liquid](https://github.com/Shopify/dawn/blob/2c847245ea3dfec6f248dcbb2dd76c4ceb65c437/sections/main-product.liquid#L646-L666). This will now only contain ONLY the first 100 variants.

**Quick order list**
[link](https://github.com/Shopify/dawn/blob/2c847245ea3dfec6f248dcbb2dd76c4ceb65c437/sections/quick-order-list.liquid#L93-L100)

We will need to update `variants` to be paginate-able and update this section to use pagination. For the moment, it will only display the first 100 variants. We are coordinating with @sofiamatulis and @dan-menard for this work.

**Featured product**
[link](https://github.com/Shopify/dawn/blob/2c847245ea3dfec6f248dcbb2dd76c4ceb65c437/sections/featured-product.liquid#L92-L94)

This is currently broken because the product displayed on featured_product comes from the section settings instead of being dynamically returned via the section API. Our team is working on a solution to support this.

### Demo video



https://github.com/Shopify/dawn/assets/3619012/8d25f880-da8e-4382-9ec4-3ebaa5e29353


### Tophatting

- https://combined-listings-test.myshopify.com/products/chinos

### DIFF
--- a/sections/main-product.liquid
+++ b/sections/main-product.liquid
@@ -500,7 +500,7 @@
                 {%- endif -%}
               {%- when 'complementary' -%}
                 <product-recommendations
-                  class="complementary-products quick-add-hidden no-js-hidden{% if block.settings.make_collapsible_row %} is-accordion{% endif %}{% if block.settings.enable_quick_add %} complementary-products-contains-quick-add{% endif %}"
+                  class="complementary-products quick-add-hidden{% if block.settings.make_collapsible_row %} is-accordion{% endif %}{% if block.settings.enable_quick_add %} complementary-products-contains-quick-add{% endif %}"
                   data-url="{{ routes.product_recommendations_url }}?limit={{ block.settings.product_list_limit }}&intent=complementary"
                   data-section-id="{{ section.id }}"
                   data-product-id="{{ product.id }}"


snippets/product-variant-options.liquid
```diff
@@ -54,68 +30,81 @@
...
+    endif
   -%}
 
   {%- capture input_id -%}
     {{ section.id }}-{{ option.position }}-{{ forloop.index0 -}}
   {%- endcapture -%}
 
+  {%- capture input_name -%}
+    {{ option.name }}-{{ option.position }}
+  {%- endcapture -%}
```
lhoffbeck, line 46: In a combined listing, parent and child options are surfaced together, so it's technically possible for a product to have multiple options with the same `option.name`. So that the form doesn't break in this case, we not differentiate with position.

snippets/quick-order-list-row.liquid
```diff
@@ -13,6 +13,12 @@
 {% assign cart_qty = cart | item_count_for_variant: variant.id %}
 {% # theme-check-enable %}
 
+{%- capture escaped_options -%}
+  {% for option in variant.options -%}
+    {{ option | escape }}{%- unless forloop.last -%}{{ ' / ' }}{%- endunless -%}
+  {%- endfor %}
+{%- endcapture -%}
+
```
lhoffbeck, line 21: This change has no visual impact for normal products, but for a combined listing the result is that both the parent and child option values will be displayed in the quick order list.

For example, for a Pants CL product with a child `Pants - black` and option `size`, this will display `Black / Large` instead of just `Large`.

MiniLuke (approved):
Looks good to me overall. 

I'm not sure if this is new or related -- the fourth image (and beyond, I assume) in a gallery doesn't get its full size image preloaded, so it ends up being really slow if you click on it to open the modal. I came across it organically and then had trouble reproducing it on my normal connection; but it is definitely noticeable if you throttle to fast 3g using dev tools for example. The end result is being focused on a giant white modal. I was looking at https://combined-listings-test.myshopify.com/products/chinos and clicking onto the blue chinos image at the end.

slowModalLoad

bakura10:
Hi everyone,

Michael from Maestrooo (developer of Prestige, Impact...) here. I'm cross-posting here what I have reported to some Shopify engineers, as it might benefit everyone.

I had a first look at the API, and could get my hand on a dev store. I also reviewed the Dawn implementation and, unfortunately, you overlook some important details that make the implementation broken.

To recap, we now have two use possibilities:

* A buyer changes an option from the combined listing: this will re-render the main product section using the `option_values`.
* A buyer changes an option from a standard product: this will re-render just the variant.

However, an important feature has been overlook here: because each product of a combined listing is implemented as its own product, all the sections (and not only the main product template) might be different from product to product (they might have a different template, or have sections content bound to a product metafield).

It is therefore necessary for Dawn to re-render not only the product section, but the whole page. Actually, while I was looking at the code I wondered why you went through this whole complexity instead of just using a plain old <a>:

```
<a href="{{ variant.product_url }}">Swatch</a>
```

And then let the browser re-render the whole page, and ensure that all sections are contextualized properly.

But then I realized the API is actually missing a way to detect if an option has values that are just variants or full product, especially since the `option_value.product_url` always return something, so we can't use this variable to detect it.

It will therefore be needed to introduce a new property in the `option` object to know if the values inside this option should are part of a combined listing or not.

This will already allow you to remove all the code related to section rendering using `option_values`, which won't be necessary anymore.

lhoffbeck:
That is pretty quick! We want to avoid this approach because of the [full page flash](https://screenshot.click/28-32-1l248-kcleg.mp4) ([here's a demo of the dawn implementation for reference](https://screenshot.click/28-35-6v668-bdiam.mp4))

> This is too hacky. An option_value that does not redirect to a different product should have the product_url set to null. To give you more context, here is how we render all option values in our themes. We basically...

Interesting... could you push the comparison down into your option value template by passing in product?

```
  if value.swatch != nil
       render 'option-value', type: 'swatch', url: value.product_url, product: product, ...
  else
       render 'option-value', type: 'block', url: value.product_url, product: product, ...
endif

// option-value.liquid
<a href="{{ url }}" data-should-redirect="{{ url != product.url }}" ... />
  
// js
// capture option value link click, prevent default if dataset.shouldRedirect is false
```

bakura10:
Is it that much of an issue? We had this feature for years in our theme, and merchants never complained. And the good thing is that View Transitions API will allow to solve this issue cleanly in the future. Build features on the standard!

`Interesting... could you push the comparison down into your option value template by passing in product?`

No. An option_value snippet does not have to deal with a product. Actually, we use the exact same snippet for collection pages to render swathces, or filter option. An option-value is just an option-value (so a swatch, a block option...). Passing a product is adding a useless dependency.

Really, the solution here is really simple: `option_value.product_url` should return null for standard variant, and a URL for combined listing product. If someone want to retrieve the product URL of the variant associated to an option, they can do it with `option_value.variant.product.url`. Here the `product_url` is just an alias that adds nothing in its current form, and could actually just be removed if it is just staying this way.

bakura10:
Actually, this whole set of hardcoded things to update: https://github.com/Shopify/dawn/pull/3246/files#diff-47a8b2f1172e03da15783b379f0e007d8e3dfd27cbd9142adf9e32c4b9591079R1227 seems like a big no-go to me.

What about:

* App blocks that would require to update on variant change? By explicitly selecting which element should be re-rendered inside the section, you might miss important elements to update.
* What if merchant decide to update their Liquid code to add a new information that also need to be re-rendered on combined product change? They will need to remember to also update the JS, follow the same convention as Dawn.
* What if Shopify continues that add new features? You will end up adding more and more and more hardcoded lines? This won't scale.

The small trade-off of having a full page flash (that, once again, can be completely eliminated with the upcoming View Transition API that is already available in Chrome) completely overweights the complexity of the approach taken by Dawn imo.

PR #3280, opened 2024-02-15
Changed price filter to use text input

tyleralsbury:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/facets.js
+++ b/assets/facets.js
@@ -300,9 +300,11 @@ FacetFiltersForm.setListeners();
 class PriceRange extends HTMLElement {
   constructor() {
     super();
-    this.querySelectorAll('input').forEach((element) =>
-      element.addEventListener('change', this.onRangeChange.bind(this))
-    );
+    this.querySelectorAll('input').forEach((element) => {
+      element.addEventListener('change', this.onRangeChange.bind(this));
+      element.addEventListener('keydown', this.onKeyDown.bind(this));
+      element.addEventListener('keydown', this.onKeyUp.bind(this));
+    });
     this.setMinAndMaxValues();
   }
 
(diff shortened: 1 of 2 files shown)


snippets/facets.liquid
```diff
@@ -357,13 +357,6 @@
                     </div>
                   </details>
                 {% when 'price_range' %}
-                  {% liquid
-                    assign currencies_using_comma_decimals = 'ANG,ARS,BRL,BYN,BYR,CLF,CLP,COP,CRC,CZK,DKK,EUR,HRK,HUF,IDR,ISK,MZN,NOK,PLN,RON,RUB,SEK,UYU,VES,VND' | split: ','
```
tyleralsbury, line 361: A weight off my shoulders for this to be gone. 😄

snippets/facets.liquid
```diff
@@ -1051,7 +928,7 @@
                 {%- else -%}
                   {{ 0 | money }}
                 {%- endif -%}
-                -
+                -&nbsp;
                 {%- if filter.max_value.value -%}
                   {{ filter.max_value.value | money }}
                 {%- else -%}
```
NathanPJF, line 934: These if-statements could be replaced with:
```
{{ filter.min_value.value | default: 0 | money }} - {{ filter.max_value.value | default: filter.range_max | money }}
```

I know it's longer, but I think it's clear what it does. And you don't need to manage the whitespace with `&nbsp;`

snippets/facets.liquid
```diff
@@ -1168,7 +1045,7 @@
                       {%- else -%}
                         {{ 0 | money }}
                       {%- endif -%}
-                      -
+                      -&nbsp;
```
NathanPJF, line 1048: Same comment as above; I recommend replacing the if-statements with this:

```
{{ filter.min_value.value | default: 0 | money }} - {{ filter.max_value.value | default: filter.range_max | money }}
```

snippets/facets.liquid
```diff
@@ -487,7 +426,7 @@
                           {%- else -%}
                             {{ 0 | money }}
                           {%- endif -%}
-                          -
+                          -&nbsp;
```
NathanPJF, line 429: Same comment as elsewhere, I recommend: 

```
{{ filter.min_value.value | default: 0 | money }} - {{ filter.max_value.value | default: filter.range_max | money }}
```

assets/facets.js
```diff
@@ -311,20 +312,27 @@ class PriceRange extends HTMLElement {
     this.setMinAndMaxValues();
   }
 
+  onKeyDown(event) {
+    if (event.metaKey) return;
+
+    const pattern = /[0-9]|\.|,| |Tab|Backspace|Enter|ArrowUp|ArrowDown|ArrowLeft|ArrowRight|Delete/;
```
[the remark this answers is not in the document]
NathanPJF: Should add apostrophe `'`, it's one of the supported currency formats https://help.shopify.com/en/manual/markets/pricing/currency-formatting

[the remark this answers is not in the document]
NathanPJF: Interesting that I was able to press `Esc` without issue even though `Escape` is it's own value. 🤔  [MDN link](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#:~:text=%22-,Escape,-%22%20%5B2%5D). I wonder if some other event listener on the modal caught this, and the event was allowed to propagate up. Regardless, worth adding `Escape` to the list.

NathanPJF (approved):
Approved with comments. I tested the suggested change

🎩 Tophatted on a store with Chilean Peso and another with Euros using various currencies. Nice work!

Comment outside the scope of this PR: It's pretty difficult to use this input IMO because of how quickly a `change` event is registed and submits the new price. I have to type super quickly to avoid `5` being submitted when I really wanted `50` - same applies for the mobile experience. Feels like there should be a "submit" button of some sort.  The Mobile drawer already has an ["apply" button](https://screenshot.click/19-00-mqumi-r3s0y.png), which doesn't seem to need to be used.  Examples seen in the wild:
- Best Buy CA : https://screenshot.click/19-55-mu7u7-j02sr.png
- Amazon.com : https://screenshot.click/19-48-43ffd-sbuhm.png
- Home Depot CA: https://screenshot.click/19-57-bunf4-x7f8o.png


A range slider could be another approach, but the UX there is too finicky IMO

ludoboludo (commented):
I'm wondering if there is a11y announcement of the inputs that should be tweaked to be clearer as there are no clues about the fact that it's only supporting numbers. Except the summary that mentions it's a price filter. 

Other wise other than addressing Nathan's comment ideally before merge, it seems good 👍 

I'm curious about one thing. When I change to different markets/currency it removes the price filter. Is that something coming from markets/search & disco app?

ludoboludo (approved):
Seems good to me 👍

tyleralsbury:
> I had similar concerns about submitting the form this quickly in this new form, there is a difference in the experience for sure with this change, even with the latest delay change. A better UX would be to implement a submit button to avoid distraction for buyers using mouse/keyboard or just keyboard. [Reference video](https://screenshot.click/29-37-ivr0e-nqora.mp4) walking through the pain.
> 
> cc. @kjellr @Oliviammarcello FYI for you scope of work.

I don't really want to increase the scope for this PR in particular to include a pattern like that, but something I do notice is that the focus doesn't return to the specific element you're typing in now, where in the past it seems like it did.  I will see if I can resolve that because it would alleviate the issue a bit.

PR #3289, opened 2024-02-21
Add product puppeteer class and simplify component update behavior

lhoffbeck:
⚠️  This PR builds on the work to support combined listings and 2k variants https://github.com/Shopify/dawn/pull/3378

tl;dr - this PR primarily copies code out of the variant picker class in `global.js` into `product-info.js`, which has been promoted to act as a wrapper for Product.

This PR came out of a conversation with @tyleralsbury earlier this year while working on [adding support for combined listings / 2k variants](https://github.com/Shopify/dawn/pull/3246). In that PR, the `VariantSelects` class needs to swap out the page content for its product when the selected variant is associated with a different Combined Listing child product. `VariantSelects` can be embedded in a few different product contexts (main product, featured product, quick add modal), and I was running into an issue where the class needed some pretty janky logic to determine its wrapping context ([here](https://github.com/Shopify/dawn/pull/3289/files#diff-47a8b2f1172e03da15783b379f0e007d8e3dfd27cbd9142adf9e32c4b9591079L1152-L1158) and [here](https://github.com/Shopify/dawn/pull/3289/files#diff-47a8b2f1172e03da15783b379f0e007d8e3dfd27cbd9142adf9e32c4b9591079L1170-L1172)). For a more stable architecture, child components shouldn't care what context they are rendered in.

An additional problem is that if components need to be updated when an event in a sibling happens, there wasn't a great "namespaced" context to manage the update. We had two primary patterns for handling this:

**Directly update sibling content** ([link](https://github.com/Shopify/dawn/blob/05e61a33d45358b09b85a089e0dd1b19109320c9/assets/global.js#L1092-L1102))
```js
class VariantSelects extends HTMLElement {
  
  onVariantChange(event) {
    ...
    this.updatePickupAvailability();
    ...
  }

  updatePickupAvailability() {
    const pickUpAvailability = document.querySelector('pickup-availability');
    if (!pickUpAvailability) return;

    if (this.currentVariant && this.currentVariant.available) {
      pickUpAvailability.fetchAvailability(this.currentVariant.id);
    } else {
      pickUpAvailability.removeAttribute('available');
      pickUpAvailability.innerHTML = '';
    }
  }
}
```

**Publish an event with the section, sibling subscription triggers update** ([link](https://github.com/Shopify/dawn/blob/05e61a33d45358b09b85a089e0dd1b19109320c9/assets/global.js#L1183-L1189))
```js
class VariantSelects extends HTMLElement {
    ...
    publish(PUB_SUB_EVENTS.variantChange, {
      data: {
        sectionId,
        html,
        variant: this.currentVariant,
      },
    });
  ...
}

class ProductInfo extends HTMLElement {
  connectedCallback() {
      this.variantChangeUnsubscriber = subscribe(PUB_SUB_EVENTS.variantChange, (event) => {
      const sectionId = this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section;

      // check whether the event came from this ProductInfo's section
      if (event.data.sectionId !== sectionId) return;
      this.updateQuantityRules(event.data.sectionId, event.data.html);
      this.setQuantityBoundries();
    });
  }
}
```

We have 2 primary issues: child components need to be context aware, and handling product component updates was complicated and inconsistent.

To solve both of these problems, I introduced a product class that wraps the entire rendered product and serves as an event handling puppeteer.

We already had a product wrapper class called `ProductInfo` that wrapped part of the rendered product; I moved it higher up the tree and extracted component update logic into it.

### Other considerations

I left a [comment](https://github.com/Shopify/dawn/pull/3289/files#r1499947293), but we possibly could decompose further by moving the old ProductInfo content into a QuantityForm class.

None

### Testing steps/scenarios

Here is a zip file for these theme changes: https://github.com/Shopify/dawn/archive/refs/heads/add-product-wrapper-class.zip

- [ ] Option value switching still works as expected
  - [ ] Switching between option values
  - [ ] Switching to an option value that has no associated variant
  - [ ] Switching between combined listings children
- [ ] Product recommendations load when switching between combined listings children
- [ ] Quick order list is updated when switching between combined listings children
- [ ] Quick add modal
  - [ ] works as expected for a normal product
  - [ ] works as expected for a combined listing product
- [ ] Quantity and volume price rules. (A catalog is set up on the shop, feel free to add yourself as a user or reach out to me for a test account)

### Demo links

- https://combined-listings-test.myshopify.com/products/chinos-orange

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -6,6 +6,25 @@ function getFocusableElements(container) {
   );
 }
 
+class SectionId {
+  static #separator = '__';
+
+  // for a qualified section id (e.g. 'template--22224696705326__main'), return just the section id (e.g. 'template--22224696705326')
+  static parseId(qualifiedSectionId) {
+    return qualifiedSectionId.split(SectionId.#separator)[0];
+  }
+
+  // for a qualified section id (e.g. 'template--22224696705326__main'), return just the section name (e.g. 'main')
+  static parseSectionName(qualifiedSectionId) {
+    return qualifiedSectionId.split(SectionId.#separator)[1];
+  }
+
+  // for a section id (e.g. 'template--22224696705326') and a section name (e.g. 'recommended-products'), return a qualified section id (e.g. 'template--22224696705326__recommended-products')
+  static getIdForSection(sectionId, sectionName) {
+    return `${sectionId}${SectionId.#separator}${sectionName}`;
+  }
+}
+
 class HTMLUpdateUtility {
   #preProcessCallbacks = [];
   #postProcessCallbacks = [];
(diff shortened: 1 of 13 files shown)


assets/global.js
```diff
@@ -995,75 +995,35 @@ customElements.define('slideshow-component', SlideshowComponent);
 class VariantSelects extends HTMLElement {
   constructor() {
     super();
-    this.addEventListener('change', this.handleProductUpdate);
-    this.initializeProductSwapUtility();
   }
```
lhoffbeck, line 999: This is the primary change of this PR--extracting product-update logic out of VariantSelects to a wrapper class. This also enables us to eliminate some gnarly selectors from [here](https://github.com/Shopify/dawn/pull/3289/files#diff-47a8b2f1172e03da15783b379f0e007d8e3dfd27cbd9142adf9e32c4b9591079L1152-L1158) and [here](https://github.com/Shopify/dawn/pull/3289/files#diff-47a8b2f1172e03da15783b379f0e007d8e3dfd27cbd9142adf9e32c4b9591079L1170-L1172).

assets/product-info.js
```diff
@@ -2,81 +2,324 @@ if (!customElements.get('product-info')) {
...
+      onVariantChangeUnsubscriber = undefined;
+      cartUpdateUnsubscriber = undefined;
+      swapProductUtility = undefined;
+      abortController = undefined;
+
       constructor() {
         super();
-        this.input = this.querySelector('.quantity__input');
-        this.currentVariant = this.querySelector('.product-variant-id');
-        this.submitButton = this.querySelector('[type="submit"]');
```
lhoffbeck, line 9: `currentVariant` can be grabbed from the product form class, `submitButton` wasn't used anywhere.

assets/quick-order-list.js
lhoffbeck, line 1: This content update now happens declaratively in ProductInfo

assets/global.js
```diff
@@ -1315,39 +1085,24 @@ class ProductRecommendations extends HTMLElement {
...
 
   connectedCallback() {
```
lhoffbeck, line 1087: Recommendation reinitialization now happens declaratively in the ProductInfo class

tauthomas01 (commented):
There are a few things I noticed while tophatting:

1) When I use tab navigation, navigating through options with regular variant switching works well (e.g. `?variant=47651697099054`), but when I load new child product (e.g. `chinos-black?variant=xxx`), the focus is set after the main product section. Since it's loaded asynchronously, it would be nice to set the focus before the main product section so we're not losing out sight when a new content is loaded.

https://github.com/Shopify/dawn/assets/658169/fa56a1a9-c29c-4566-9557-7813f38b4c09

2) There is a JS error in some case where the cart section id looks undefined when adding the product into the cart. It says the following:

```
TypeError: Cannot read properties of null (reading 'innerHTML')
    at CartNotification.getSectionInnerHTML(cart-notification.js:66:85)
```

I looked into this and it's related to this:

```
getSectionInnerHTML(html, selector = '.shopify-section') {
    return new DOMParser().parseFromString(html, 'text/html').querySelector(selector).innerHTML;
  }
```

Printing the `selector` param returns this: `'[id="cart-notification-product-undefined"]'`

https://github.com/Shopify/dawn/assets/658169/0ad053a0-610c-4154-80c7-b415e5f6b50e

assets/product-form.js
```diff
@@ -113,6 +114,20 @@ if (!customElements.get('product-form')) {
...
+
+      toggleSubmitButton(disable = true, text) {
+        if (disable) {
+          this.submitButton.setAttribute('disabled', 'disabled');
+          if (text) this.submitButtonText.textContent = text;
+        } else {
+          this.submitButton.removeAttribute('disabled');
+          this.submitButtonText.textContent = window.variantStrings.addToCart;
+        }
+      }
```
lhoffbeck, line 126: `toggleSubmitButton` logic was pulled out of the VariantSelects class

sections/main-product.liquid
```diff
@@ -1,7 +1,10 @@
-<section
+<product-info
   id="MainProduct-{{ section.id }}"
   class="section-{{ section.id }}-padding gradient color-{{ section.settings.color_scheme }}"
   data-section="{{ section.id }}"
+  data-product-id="{{ product.id }}"
+  data-update-url="true"
+  data-url="{{ product.url }}"
 >
```
lhoffbeck, line 8: Worth flagging, this node was changed from a `section` to a `product-info`

sections/main-product.liquid
```diff
@@ -746,7 +747,7 @@
       }
     </script>
   </div>
-</section>
+</product-info>
```
ludoboludo, line 750: This is somehow breaking the padding applied and the content is flush with the header at the top: 


Seems like it's missing a `display: block;` that it was getting from the user agent stylesheet:

ludoboludo: Which should also fix the color scheme issue not working atm:

ludoboludo (changes requested):

PR #3299, opened 2024-02-26
Bring back swatch-input snippet

jas7457:
### PR Summary: 

This brings back the `swatch-input` snippet which was removed in [this PR](https://github.com/Shopify/dawn/pull/3283). I will have a follow up PR which uses the swatch-input for filters, not just the PDP.

### Why are these changes introduced?

This is necessary for upcoming filter work.

### What approach did you take?

I copied/pasted the changes, along with a few small tweaks that the filters will use (namely, splitting the `.visually-disabled` class with the `disabled` attribute).

### Visual impact on existing themes

None: this just allows for code sharing.

### Testing steps/scenarios

- [ ] View the editor link (link below)
- [ ] Verify that the swatches render their different states properly (visually_disabled, active, hover, etc).

### Demo links

- [Store](https://shop1.shopify.dawn-public.jason-addleman.us.spin.dev/?preview_theme_id=3)
- [Editor](https://admin.web.dawn-public.jason-addleman.us.spin.dev/store/shop1/themes/3/editor?previewPath=%2Fproducts%2Ftaxonomy-t-shirt&block=template--599__main%2Fvariant_picker&section=template--599__main)

### Checklist
- [x] Added PR summary for [release notes](https://themes.shopify.com/themes/dawn/styles/default#ReleaseNotes)
- [ ] Requested review from UX (Only for changes that are affecting the experience or perceivable visual details)
- [ ] Created a ticket for the [help.shopify.com](https://help.shopify.com) documentation team about updates to theme settings. (Internal-only task)
- [x] Followed [theme code principles](https://github.com/Shopify/dawn/blob/main/.github/CONTRIBUTING.md#theme-code-principles)
- [ ] Linted with [Theme Check](https://github.com/Shopify/theme-check)
- [ ] Tested on [mobile](https://shopify.dev/themes/store/requirements#mobile-browser-requirements)
- [ ] Tested on [multiple browsers](https://shopify.dev/themes/store/requirements#desktop-browser-requirements)
- [ ] Tested for [accessibility](https://shopify.dev/themes/best-practices/accessibility)

PR #3303, opened 2024-02-26
Switch the filter swatches to use the swatch api

jas7457:
Blocked by #3299

### PR Summary: 

Fixes https://github.com/Shopify/discovery-app/issues/4287

This changes the filters in swatch to start using the swatch API rather than the display API it was previously using. There may be slight visual differences between the old swatches and the reusable swatches - this is intentional and preferred! The point is, the two should be the same.

Note: This is a copy of PR 294 in the private repo that was already reviewed.

I had to duplicate logic 3 different times: one for desktop, desktop-no-js, and mobile. I do think that this could be greatly simplified, but is outside the scope of these changes.

I put more things into variables, as things like the id, the classes, etc should be the same between swatch and label. This makes sure that the two don't drift too far

There are very slight color differences between the old way that Dawn rendered a filter swatch and the new way. Most of it should be transparent to the merchant/buyer.

### Testing steps/scenarios

- [ ] Test that color filters work
- [ ] Test that image filters work
- [ ] Test that image focal point works (although, it may/may not shift the image very much)

### Demo links

- [Store](https://shop1.shopify.dawn-public-swatch.jason-addleman.us.spin.dev/?preview_theme_id=3)
- [Editor](https://admin.web.dawn-public-swatch.jason-addleman.us.spin.dev/store/shop1/themes/3/editor)

### Checklist


assets/component-facets.css
```diff
@@ -326,7 +328,7 @@
   align-items: flex-start;
 }
 
-.facets-layout-grid .visual-display-parent {
+.facets-layout-grid .facets__label {
```
NathanPJF, line 331: This is overwriting the styles being applied by the `.facet-checkbox` declaration. That's fine to overwrite them... the weird part is that `.facet-checkbox` is setting the display, flex-direction, etc to begin with.  Two things to change that could fix the alignment of list items with a checkbox:

1. Change `.facet-checkbox` on line 345 to `.facets__label`.  This is in line with a number of other changes you made that removes the focus on `facet-chackbox` to the labels.
2. Add `align-items: center` to that declaration. 

This would fix a misalignment that's most visible in the mobile's drawer view



You can also see it in the vertical layout; this is what it looks like with the fix applied

ludoboludo: +1

andy-schneidr (approved):

assets/component-facets.css
jas7457: I talked with @NathanPJF about this. If the text wraps, then the checkbox (but not the checkmark) would get centered. We can fix the checkmark as well, but in either case, we decided having the checkbox centered looked worse, so we won't make this change. I can do the rename, but not the alignment.

image

PR #3332, opened 2024-03-08
[Quick Order List] Fix an issue for one-variant product

eugenekasimov:
### PR Summary: 

This PR is to fix an issue for one-variant product in the Quick Order List. After we introduced the Keyboard Interaction the Quick Order List stopped working for one-variant products.

[Video](https://screenshot.click/08-38-nlcwd-jgtl8.mp4) of the issue. 

In the Quick Order List products with one variant don't have the total bar as products with multiple variants do. It was not taken on account. When we try to calculate a position for the total bar we may face an error in JS because we use an element that doesn't exist for one-variant products. 

I added an `if` statement to check if the total bar exists in the `constructor`. I also added an `if` statement inside `swicthVariant()` to avoid scrolling on one-variant products.  

### Demo links
[Editor](https://admin.shopify.com/store/os2-demo/themes/163004579862/editor)

- [ ] Make sure that the Quick Order List works properly when there is only one variant. ([this is a good example of a product](https://admin.shopify.com/store/os2-demo/themes/163004579862/editor?previewPath=%2Fproducts%2Fhobo-l-latte))
- [ ] Make sure that products with many variants still work properly.

### DIFF
--- a/assets/quick-order-list.js
+++ b/assets/quick-order-list.js
@@ -80,12 +80,14 @@ class QuickOrderList extends HTMLElement {
       };
     }
     this.totalBar = this.querySelector('.quick-order-list__total');
-    this.totalBarPosition = this.totalBar ? window.innerHeight - this.totalBar.offsetHeight: null;
+    if (this.totalBar) {
+      this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
 
-    window.addEventListener('resize', () => {
-      this.totalBarPosition = this.totalBar ? window.innerHeight - this.totalBar.offsetHeight: null;
-      this.stickyHeader.height = this.stickyHeaderElement ? this.stickyHeaderElement.offsetHeight: null;
-    });
+      window.addEventListener('resize', () => {
+        this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
+        this.stickyHeader.height = this.stickyHeaderElement ? this.stickyHeaderElement.offsetHeight: 0;
+      });
+    }
 
     form.addEventListener('submit', this.onSubmit.bind(this));
     const debouncedOnChange = debounce((event) => {


eugenekasimov:
> 👋 @eugenekasimov , could you share what the error was beforehand?

Sure. If you go to main and open a product with one variant you will see the Quick Order List is not working at all. 
https://screenshot.click/08-38-nlcwd-jgtl8.mp4

On of the main reasons of that was I didn't account the case that for one-variant product we don't have the total bar when I built the Keyboard interaction feature.

sofiamatulis (approved):
Thanks Eugene! Good catch 👍

sofiamatulis (commented):
Added one suggestion!

assets/quick-order-list.js
```diff
@@ -209,34 +209,42 @@ class QuickOrderList extends HTMLElement {
...
-
-        if (!e.shiftKey) {
-          const nextIndex = currentIndex + 1;
-          const nextVariant = this.allInputsArray[nextIndex] || this.allInputsArray[0];
-          nextVariant.select();
-        } else {
-          const previousIndex = currentIndex - 1;
-          const previousVariant = this.allInputsArray[previousIndex] || this.allInputsArray[this.allInputsArray.length - 1];
-          previousVariant.select();
+    if (this.allInputsArray.length !== 1) {
```
utkarshsaxena93, line 212: is `this.allInputsArray.length <= 1` a better way to go?

also what do we expect to happen when a merchant presses `enter` on a single variant list? Should it do anything?

assets/quick-order-list.js
```diff
@@ -79,11 +79,11 @@ class QuickOrderList extends HTMLElement {
         type: `${this.stickyHeaderElement.getAttribute('data-sticky-type')}`
       };
     }
-
-    this.totalBarPosition = window.innerHeight - this.querySelector('.quick-order-list__total').offsetHeight;
+    this.totalBar = this.querySelector('.quick-order-list__total');
+    this.totalBarPosition = this.totalBar ? window.innerHeight - this.totalBar.offsetHeight: null;
```
utkarshsaxena93, line 83: why is the total bar position set to `null` and not `0` ? 

The only place we use it is `const totalBarCrossesInput = inputBottomBorder > this.totalBarPosition;` so is it doing `> null` check?

utkarshsaxena93 (commented):
PR description needs more details. Can you update the PR description with what the issue was before and why it was happening? and what you did to fix it?

assets/quick-order-list.js
eugenekasimov: > is this.allInputsArray.length <= 1 a better way to go?

My logic was that inside the Quick Order List there is always at least one input, the only case when there is no input would be when all variants are sold out but in this case we there is nothing to interact with. 

Is there other reasons to use `<= 1` instead of` !==1`? Or is it just a better principle in coding to avoid '!==" ?

>also what do we expect to happen when a merchant presses enter on a single variant list? Should it do anything?

We just want to remove focus from the input before updating started.

assets/quick-order-list.js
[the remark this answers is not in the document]
eugenekasimov: That's a good idea. Thanks.

assets/quick-order-list.js
```diff
@@ -209,34 +209,42 @@ class QuickOrderList extends HTMLElement {
...
+      const stickyHeaderCrossesInput = this.stickyHeaderElement && this.stickyHeader.type !== 'on-scroll-up' && this.stickyHeader.height > inputTopBorder;
+      const stickyHeaderScrollupCrossesInput = this.stickyHeaderElement && this.stickyHeader.type === 'on-scroll-up' && this.stickyHeader.height > inputTopBorder && stickyHeaderBottomBorder > 0;
 
-    if (totalBarCrossesInput || inputOutsideOfViewPort || stickyHeaderCrossesInput || stickyHeaderScrollupCrossesInput) {
-      this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
+      if (totalBarCrossesInput || inputOutsideOfViewPort || stickyHeaderCrossesInput || stickyHeaderScrollupCrossesInput) {
+        this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
+      }
+    } else {
+      this.variantListInput.addEventListener('keydown', (e) => {
```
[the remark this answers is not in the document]
eugenekasimov: @utkarshsaxena93 this helps us to remove focus from the input right after we hit Enter. If we remove this the focus will be removed anyway but only after we update the element after API call. I just wanted to keep consistency. In a multiple-variant product when we hit Enter we remove focus from the current input right away and jump to the next row. Does it make sense to you?

assets/quick-order-list.js
```diff
@@ -79,13 +79,15 @@ class QuickOrderList extends HTMLElement {
...
+      this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
 
-    this.totalBarPosition = window.innerHeight - this.querySelector('.quick-order-list__total').offsetHeight;
-
-    window.addEventListener('resize', () => {
-      this.totalBarPosition = window.innerHeight - this.querySelector('.quick-order-list__total').offsetHeight;
-      this.stickyHeader.height = this.stickyHeaderElement ? this.stickyHeaderElement.offsetHeight: null;
-    });
+      window.addEventListener('resize', () => {
+        this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
```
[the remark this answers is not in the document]
eugenekasimov: I do need it and I place it inside the same IF statement

```
if (this.totalBar) {
      this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;

      window.addEventListener('resize', () => {
        this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
        this.stickyHeader.height = this.stickyHeaderElement ? this.stickyHeaderElement.offsetHeight: 0;
      });
    }
```

sofiamatulis (approved):
Looks good 👍

eugenekasimov:
> Unrelated to this PR but another keyboard focus related issue:
> 
> * https://share.descript.com/view/5qXac8x0agI
> * https://share.descript.com/view/17uyfD8v6fR

I created issues for these. Thank you Utkarsh.

assets/quick-order-list.js
```diff
@@ -79,13 +79,15 @@ class QuickOrderList extends HTMLElement {
         type: `${this.stickyHeaderElement.getAttribute('data-sticky-type')}`
       };
     }
+    this.totalBar = this.querySelector('.quick-order-list__total');
+    if (this.totalBar) {
+      this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
```
utkarshsaxena93, line 84: I think it is worth updating the code on line 235 to`const totalBarCrossesInput = this.totalBarPosition ? inputBottomBorder > this.totalBarPosition : false`.

utkarshsaxena93 (approved):

assets/quick-order-list.js
utkarshsaxena93: Makes sense. Best solution is where we don't have completely different logic running based on the array length 1 v/s many but it is okay for now. We can look at improving this in the future.

PR #3333, opened 2024-03-11
[Quick Add Bulk] Quick fix for the Progress bar

eugenekasimov:
### PR Summary: 

There is an issue happened after refactoring in the Progress Bar PR https://github.com/Shopify/dawn/pull/3288/commits/9b49ae369183667f25a143420954884722817546

The issue is that before we moved `this.querySelector('quantity-input input')` into a variable the input element was re-selected every time the input changed and it worked good. Since we store it in a variable `this.input `we select the input once but for DOM it’s different element every time we change qty on it.

I added a function to set input every time after we change qty on the input. 

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163009855510/editor)

### Checklist

### DIFF
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -35,7 +35,6 @@ class QuickAddBulk extends HTMLElement {
 
   setInput() {
     this.input = this.querySelector('quantity-input input');
-    return this.input;
   }
 
   listenForActiveInput() {
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -10,7 +10,7 @@ class QuickAddBulk extends HTMLElement {
       }
     }, ON_CHANGE_DEBOUNCE_TIMER);
     this.addEventListener('change', debouncedOnChange.bind(this));
-    this.input = this.querySelector('quantity-input input');
+    this.setInput();
     this.listenForActiveInput();
     this.listenForKeydown();
     this.lastActiveInputId = null;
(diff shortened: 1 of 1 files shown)


assets/quick-add-bulk.js
```diff
@@ -33,6 +33,11 @@ class QuickAddBulk extends HTMLElement {
     }
   }
 
+  setInput() {
+    this.input = this.querySelector('quantity-input input');
+    return this.input;
```
sofiamatulis, line 38: Do we need to return this? We already saved it to a variable

assets/quick-add-bulk.js
```diff
@@ -10,7 +10,7 @@ class QuickAddBulk extends HTMLElement {
       }
     }, ON_CHANGE_DEBOUNCE_TIMER);
     this.addEventListener('change', debouncedOnChange.bind(this));
-    this.input = this.querySelector('quantity-input input');
+    this.setInput();
```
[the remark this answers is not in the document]
sofiamatulis: thanks for clarifying! I didnt remember if we had it in the 1st commit 👍

sofiamatulis (approved):

assets/quick-add-bulk.js
```diff
@@ -33,6 +33,10 @@ class QuickAddBulk extends HTMLElement {
     }
   }
 
+  setInput() {
+    this.input = this.querySelector('quantity-input input');
```
utkarshsaxena93, line 37: Can we do 
```
getInput() {
    return this.querySelector('quantity-input input');
}
```

and then do `this.getInput().addEvent.....` or wherever else it is used.

Why do we want to store it in a variable?

eugenekasimov: Thanks @utkarshsaxena93 I need a bit time to learn this approach and figure out how it works but looks interesting even for my development. I'll take a look :)

PR #3335, opened 2024-03-11
[Quick Add Bulk] Errors on the product card using browser's validation

eugenekasimov:

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -35,7 +35,6 @@ class QuickAddBulk extends HTMLElement {
 
   setInput() {
     this.input = this.querySelector('quantity-input input');
-    return this.input;
   }
 
   listenForActiveInput() {
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -10,7 +10,7 @@ class QuickAddBulk extends HTMLElement {
       }
     }, ON_CHANGE_DEBOUNCE_TIMER);
     this.addEventListener('change', debouncedOnChange.bind(this));
-    this.input = this.querySelector('quantity-input input');
+    this.setInput();
     this.listenForActiveInput();
     this.listenForKeydown();
     this.lastActiveInputId = null;
@@ -31,6 +31,11 @@ class QuickAddBulk extends HTMLElement {
     if (this.cartUpdateUnsubscriber) {
       this.cartUpdateUnsubscriber();
     }
+  }
+
+  setInput() {
+    this.input = this.querySelector('quantity-input input');
+    return this.input;
   }
 
   listenForActiveInput() {
@@ -177,7 +182,7 @@ class QuickAddBulk extends HTMLElement {
     if (this.isEnterPressed) {
       this.querySelector(`#Quantity-${this.lastActiveInputId}`).select();
     }
-
+    this.setInput();
     this.listenForActiveInput();
     this.listenForKeydown();
   }
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -3,6 +3,7 @@ class QuickAddBulk extends HTMLElement {
   constructor() {
     super();
     const debouncedOnChange = debounce((event) => {
+      this.cleanErrors();
       if (parseInt(event.target.dataset.cartQuantity) === 0) {
         this.addToCart(event);
       } else {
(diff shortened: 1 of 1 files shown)


eugenekasimov:
> 👋 @eugenekasimov , we are getting some changes in this PR that are not related to this PR. Can we ensure we only have the changes related to this PR here :)

Hey @sofiamatulis I did on purpose and I mentioned about it in description to this PR. 
>! Attantion
I pulled some code from PRs that haven't been merged yet to give you a better understanding of how it should look and work like when all is done. If you are looking at the code, please ignore this.cleanErrors() and this.setInput() reviewing this PR.

That way I make sure that all works as it should.

sofiamatulis:
👋 @eugenekasimov , can we resolve the conflicts? https://screenshot.click/12-33-ydba4-244uc.png

assets/quick-add-bulk.js
```diff
@@ -52,6 +57,18 @@ class QuickAddBulk extends HTMLElement {
...
+    const input = document.getElementById(id);
+    input.value = input.getAttribute('value');
+    this.isEnterPressed = false;
+  }
+
+  cleanErrorMessageOnType(event) {
+    event.target.addEventListener('keypress', () => {
+      event.target.setCustomValidity('');
+    }, { once: true });
+  }
```
eugenekasimov, line 70: I added this function to remove the tooltip as soon as you start typing another qty. Otherwise, the tooltip will keep being open and blinking after each type until you submit a new qty. 
That how it looked without cleaning the error message on type 👇

12-28-72hik-irzsv

assets/quick-add-bulk.js
```diff
@@ -52,6 +57,18 @@ class QuickAddBulk extends HTMLElement {
     });
   }
 
+  resetQuantityInput(id) {
```
sofiamatulis, line 60: Should we be resetting these ourselves? This should come back from the BE form what I understand

melissaperreault (commented):
[Noticed an issue](https://screenshot.click/13-04-bt8o4-p2r5a.mp4) with one of the product (Studio - Brown bag), it doesn't expose the error validation even though the inventory limit is reached 🤔 

_Re: I tried in preview and I did get the error message. Is this only due to the editor sporadically?_

assets/quick-add-bulk.js
eugenekasimov: That is before experience 👆

assets/quick-add-bulk.js
sofiamatulis: It seems like this was already done in cart for a while. Ill try to gain more context around this

eugenekasimov:
> [Noticed an issue](https://screenshot.click/13-04-bt8o4-p2r5a.mp4) with one of the product (Studio - Brown bag), it doesn't expose the error validation even though the inventory limit is reached 🤔
> 
> _Re: I tried in preview and I did get the error message. Is this only due to the editor sporadically?_

I checked it and the reason why this Studio - Brown product behaved different was in the Admin it was set to keep selling this product when out of stock. This means that it ignores the real stock and don't return an error. But at the same time it seems that BE has their own max for products - 1000000pcs. If you exceed this amount then the BE adds to the cart only 1000000 :)

melissaperreault (approved):
Thank you Eugene! 🚢 🎉

assets/quick-add-bulk.js
```diff
@@ -52,6 +57,18 @@ class QuickAddBulk extends HTMLElement {
...
+    const input = document.getElementById(id);
+    input.value = input.getAttribute('value');
+    this.isEnterPressed = false;
+  }
+
+  cleanErrorMessageOnType(event) {
+    event.target.addEventListener('keypress', () => {
```
utkarshsaxena93, line 67: is there an `onChange` on the `input` we can use instead of the `keypress` ?

eugenekasimov:
> Is it possible to center the tip of the tooltip with the input box? Not crucial but will look cleaner and more symmetric. Screenshot 2024-03-15 at 2 55 31 PM

We know about this issue and on different browsers it looks different, sometime there is no even a tip. Unfortunately, we cannot change a browser's default settings.

assets/quick-add-bulk.js
eugenekasimov: It won't work for us. Because `onChange` would be triggered for `input` only when you submit changes and that's not what we want. We literally want to clean the error message for a browser  as soon as we start typing.

dan-menard (approved):

PR #3338, opened 2024-03-12
If user added more than max to cart, don't show a negative number

rbarreca:
### PR Summary: 

If variant has `Continue selling when out of stock` enabled (or maybe even if not and available quantity changed), a user can potentially have more in cart than available. The current quick-add modal and PDP quantity input can show a negative number in these cases. This PR maxes it out at zero.


rbarreca:
I have signed the CLA!

PR #3343, opened 2024-03-13
[Quick Add Bulk] Adjust scrolling the Quick Order List to work inside modal

### LINKED ISSUE #3312: [Quick Add bulk] Quick Order List scrolling adjustment for Keyboard interaction
The Keyboard interaction scrolling in Quick order List needs to be adjusted for the case where we show the Quick Order List inside the Quick Add Bulk modal. 

Currently the scrolling doesn't happen when you reach a not fully shown row/input.

29-59-5nk6c-xyyg4

eugenekasimov:
### PR Summary: 

This PR makes sure that scrolling of the Quick Order List inside the modal of the Quick Add Bulk works properly. 

Fixes #3312.

[Befor](https://screenshot.click/13-57-pvzz6-3v6l6.mp4)

[After](https://screenshot.click/13-01-goa06-b1ez4.mp4)

-  I added a condition depending whether the Quick Order List is rendered inside the Quick Add Bulk modal or not. If yes, then we calculate positions for table head and total bar. 
- I also moved all logic related to scrolling to a new function`scrollQuickOrderListTable() ` I believe it makes sense and the code looks more readable.

### Other considerations
In the feature we would want to trigger scrolling effect on TABing buttons. This PR covers only a case when you move over the list via inputs by hitting Enter. 

### Testing steps/scenarios

- [ ] Go to Shop All collection. In the Product grid template make sure Quick Add set to Bulk. 
- [ ] On any product cards click on Choose options to open a modal with the Quick Order List inside. 
- [ ] Start moving down/up the inputs by hitting Enter/Shift+Enter. Make sure that as soon as you reach an input that is partially hidden under the total bar or table header the list is automatically scrolled. 
- [ ] Make sure that the same pattern in the Quick Order List on the PDP hasn't been affected and works the same way. 

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163020111894/editor?previewPath=%2Fcollections%2Fall&section=template--21573248581654__product-grid)

### Checklist

### DIFF
--- a/assets/quick-add.css
+++ b/assets/quick-add.css
@@ -88,7 +88,7 @@
 
   .quick-add-modal__content-info--bulk thead {
     position: sticky;
-    z-index: 2;
+    z-index: 3;
     top: 0;
     background-color: rgb(var(--color-background));
   }
--- a/assets/quick-order-list.js
+++ b/assets/quick-order-list.js
@@ -75,8 +75,8 @@ class QuickOrderList extends HTMLElement {
     const form = this.querySelector('form');
     this.inputFieldHeight = this.querySelector('.variant-item__quantity-wrapper').offsetHeight;
     this.isListInsideModal = document.querySelector('.quick-add-bulk');
-    this.tableHead = document.querySelector('.quick-order-list__table thead');
     this.stickyHeaderElement = document.querySelector('sticky-header');
+    this.getTableHead();
 
     if (this.stickyHeaderElement) {
       this.stickyHeader = {
@@ -212,13 +212,17 @@ class QuickOrderList extends HTMLElement {
     this.defineInputsAndQuickOrderTable();
   }
 
+  getTableHead() {
+    return this.tableHead = document.querySelector('.quick-order-list__table thead');
+  }
+
   scrollQuickOrderListTable() {
     const inputTopBorder = this.variantListInput.getBoundingClientRect().top;
     const inputBottomBorder = this.variantListInput.getBoundingClientRect().bottom;
 
     if (this.isListInsideModal) {
       const totalBarCrossesInput = inputBottomBorder > this.totalBar.getBoundingClientRect().top;
-      const tableHeadCrossesInput = inputTopBorder < this.tableHead.getBoundingClientRect().bottom;
+      const tableHeadCrossesInput = inputTopBorder < this.getTableHead().getBoundingClientRect().bottom;
 
       if (totalBarCrossesInput || tableHeadCrossesInput) {
         this.scrollToCenter();
--- a/assets/quick-order-list.js
+++ b/assets/quick-order-list.js
@@ -221,7 +221,7 @@ class QuickOrderList extends HTMLElement {
       const tableHeadCrossesInput = inputTopBorder < this.tableHead.getBoundingClientRect().bottom;
 
       if (totalBarCrossesInput || tableHeadCrossesInput) {
-        this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
+        this.scrollToCenter();
       }
     } else {
       const stickyHeaderBottomBorder = this.stickyHeaderElement && this.stickyHeaderElement.getBoundingClientRect().bottom;
@@ -231,9 +231,13 @@ class QuickOrderList extends HTMLElement {
       const stickyHeaderScrollupCrossesInput = this.stickyHeaderElement && this.stickyHeader.type === 'on-scroll-up' && this.stickyHeader.height > inputTopBorder && stickyHeaderBottomBorder > 0;
 
       if (totalBarCrossesInput || inputOutsideOfViewPort || stickyHeaderCrossesInput || stickyHeaderScrollupCrossesInput) {
-        this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
+        this.scrollToCenter();
       }
     }
+  }
+
+  scrollToCenter() {
+    this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
   }
 
   switchVariants(event) {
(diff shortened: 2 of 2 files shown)


assets/quick-add.css
```diff
@@ -88,7 +88,7 @@
 
   .quick-add-modal__content-info--bulk thead {
     position: sticky;
-    z-index: 2;
+    z-index: 3;
```
utkarshsaxena93, line 91: Why this change? Should we make this a sufficiently large value if we want this to appear on the top at all times?

utkarshsaxena93 (approved):
Scroll behaviour looks good.

Additional improvement to how we manage focus https://share.descript.com/view/MZeAfZs5K8s

This should be the default behaviour for all modals where it traps focus, not just this one.

melissaperreault (commented):
Aligned with @utkarshsaxena93 's comments! We do have the close X sticky on smaller screens, might be quick to add on larger screens!? 🤷 Agree it's a follow up!

**_Android and iPhone_**
- [ ] The input in the modal keeps selecting the value and reveals other UI. It's not the same behaviour from other input component elsewhere. It is annoying as a user. 

| Modal | Product template | Cart |
|--------|--------|--------|
| Screenshot_20240314-120539_Chrome | Screenshot_20240314-120555_Chrome | Screenshot_20240314-120609_Chrome |

eugenekasimov:
> Scroll behaviour looks good.
> 
> Additional improvement to how we manage focus https://share.descript.com/view/MZeAfZs5K8s
> 
> This should be the default behaviour for all modals where it traps focus, not just this one.

Thanks for the feedback Utkarsh. About the taping focus I absolutely agree with you. Since it's not related to this PR I'll create a ticket for that and also I'll check if we have this issue on other modals in the theme 👍

The comment about close button "X" I believe @sofiamatulis may have context on it because I noticed it in our base PR Quick Add Bulk and I think I even mentioned about it but she said it was changed that way on purpose.

sofiamatulis:
> > Scroll behaviour looks good.
> > Additional improvement to how we manage focus https://share.descript.com/view/MZeAfZs5K8s
> > This should be the default behaviour for all modals where it traps focus, not just this one.
> 
> Thanks for the feedback Utkarsh. About the taping focus I absolutely agree with you. Since it's not related to this PR I'll create a ticket for that and also I'll check if we have this issue on other modals in the theme 👍
> 
> The comment about close button "X" I believe @sofiamatulis may have context on it because I noticed it in our base PR Quick Add Bulk and I think I even mentioned about it but she said it was changed that way on purpose.

Yes, the x button being sticky only on mobile is a behavior we have for the existing modal. But we can challenge that experience and create a follow up ticket :) (although unrelated to this project since it also affect the existing modal)

assets/quick-add.css
eugenekasimov: Here is why https://share.descript.com/view/wOls6SoXhau

utkarshsaxena93:
Polaris modal is a good example to look at as a good example of some of the patterns around focus trapping and/or close buttons

assets/quick-order-list.js
```diff
@@ -208,12 +212,38 @@ class QuickOrderList extends HTMLElement {
...
+      }
+    } else {
+      const stickyHeaderBottomBorder = this.stickyHeaderElement && this.stickyHeaderElement.getBoundingClientRect().bottom;
+      const totalBarCrossesInput = inputBottomBorder > this.totalBarPosition;
+      const inputOutsideOfViewPort = inputBottomBorder < this.inputFieldHeight;
+      const stickyHeaderCrossesInput = this.stickyHeaderElement && this.stickyHeader.type !== 'on-scroll-up' && this.stickyHeader.height > inputTopBorder;
+      const stickyHeaderScrollupCrossesInput = this.stickyHeaderElement && this.stickyHeader.type === 'on-scroll-up' && this.stickyHeader.height > inputTopBorder && stickyHeaderBottomBorder > 0;
+
+      if (totalBarCrossesInput || inputOutsideOfViewPort || stickyHeaderCrossesInput || stickyHeaderScrollupCrossesInput) {
+        this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
```
sofiamatulis, line 234: Just an idea but if this and what youre doing in line 224 are the same we can always make it a function to avoid duplicating the same line

And just call the function in the two places `this.scrollIntoView()`

```
scrollIntoView() {
 this.variantListInput.scrollIntoView({ block: 'center', behavior: 'smooth' });
}
```

Thoughts?

eugenekasimov: Love this idea even though there are just two repetitive lines, still looks better. Thanks for the idea @sofiamatulis

sofiamatulis (commented):
Just to ensure I understand, we are not supposed to scroll back up correct?

https://screenshot.click/14-12-xmuox-88xps.mp4

Updated video: https://screenshot.click/14-32-lojam-j7bgz.mp4

eugenekasimov:
> Just to ensure I understand, we are not supposed to scroll back up correct?
> 
> https://screenshot.click/14-12-xmuox-88xps.mp4
> 
> Updated video: https://screenshot.click/14-32-lojam-j7bgz.mp4

Nice catch @sofiamatulis . There was a bug and I fixed it [e1c902d](https://github.com/Shopify/dawn/pull/3343/commits/e1c902da1203447c4caa6f88d47d5f0ac5ef4552). Every time we change qty on any input for DOM we actually change the table and the table head accordingly. That's why after changing qty the calculation for table head stops working because we never get it again after updating. I used the same approach we did in other recent PR and added `getTableHead()`. 
Just for context, it has never been a case for other elements we used for position calculation because those other elements such as the Header or Total Bar never change while we update the table.

sofiamatulis (approved):
Thanks for fixing it 🔥 Let's just update the branch before merging https://screenshot.click/15-37-rh9x4-n0qsr.png

PR #3357, opened 2024-03-18
Add Quick add bulk to feat collection

sofiamatulis:
### PR Summary: 

Add the Quick Add Bulk to the feat collection section https://screenshot.click/15-07-qmmm7-epzf7.png

Fixes https://github.com/orgs/Shopify/projects/6397/views/26?pane=issue&itemId=50891026

### Testing steps/scenarios

- [ ] Test the feat collection with bulk add on any template (try adding more than one of the same section)
- [ ] Test the existing collection page with bulk add to ensure it still works as expected
- [ ] Add feat collection with quick add bulk to the collection page so you have quick add bulk in both sections twice and test that  

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163032563734/editor)

### Checklist

### DIFF
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -14,7 +14,6 @@ class QuickAddBulk extends HTMLElement {
     this.listenForActiveInput();
     this.listenForKeydown();
     this.lastActiveInputId = null;
-    // this.sectionId = this.dataset.sectionId
   }
 
 
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -14,6 +14,7 @@ class QuickAddBulk extends HTMLElement {
     this.listenForActiveInput();
     this.listenForKeydown();
     this.lastActiveInputId = null;
+    // this.sectionId = this.dataset.sectionId
   }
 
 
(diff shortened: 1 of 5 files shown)


sofiamatulis:
> Loos good but we need to fix these errors, we had similar in the past. Screenshot 2024-03-18 at 10 31 56 AM

The reason this happens is because in this case im importing `quick-add-bulk.js` twice in the product grid liquid and in the feat collection liquid if the merchant has both turned on twice given the `if` statement is true twice that the setting is on 

Let me think how we can handle this

scottgmeadows (approved):

eugenekasimov (approved):

assets/quick-add-bulk.js
```diff
@@ -1,212 +1,216 @@
...
-    this.listenForActiveInput();
-    this.listenForKeydown();
-  }
+        this.listenForActiveInput();
+        this.listenForKeydown();
+      }
+    }
+);
 }
-customElements.define('quick-add-bulk', QuickAddBulk);
```
dan-menard, line 212: What's the value in writing the class inline? Like why not wrap this line with the `if (!customElements.get('quick-add-bulk'))` conditional?

sofiamatulis: @dan-menard Fair point! I am just following our existing standard in Dawn where we have done this before: https://github.com/Shopify/dawn/blob/main/assets/localization-form.js#L1

We can always change the approach

dan-menard (approved):

PR #3360, opened 2024-03-19
[Quick Add Bulk] Fix the issue with multiple quick order list on a page

eugenekasimov:
### PR Summary: 

This PR fixes issues:

1. If there is more than one instance of the Quick Order List on a page(i.e the Quick Add Bulk on a collection) the Keyboard interaction doesn't work after one of the instances is changed.
2. [Quick Add Bulk collection] If you interact with the Quick Order List, then the input on a one-variant product won't work properly. Same issue happens vice versa, if you first interact with the input on a one-variant product , then the Quick Order List Keyboard interaction won't work.
3. For both scenarios I mentioned above the  positions of elements calculation we use for scrolling the list doesn't work properly.

🍿 [Video describing the issues](https://share.descript.com/view/iBDRZNFa1Nt)
🍿 [Video explaining why it's happening](https://share.descript.com/view/JDMt0kaHa3z)

The reason of why this happens is because we have the function `onCartUpdate()` that changes innerHTML for an entire custom element if the cart was updated from outside (from different section). As a result, some important functions never run again for those instances that were changed from outside.

I added the existing functions that are responsible for proper functionality to onCartUpdate() right after we update `inner HTML` to make sure we update needed elements in all instances. 

### Testing steps/scenarios

- [ ] Open Shop All collection and make sure you the Quick Add set to Bulk in the section settings. 
- [ ] On any products click on Choose options to open the Quick order List inside the modal and change qty on any variant. Make sure that everything works as expected. Make sure the keyboard interaction still works. 
- [ ] Close the modal and open another product. Test it again and make sure that all mentioned in previous step works good.
- [ ] Close the modal. Find a one-variant product and click on the input. Make sure that the value is selected for default. 
- [ ] Test previous steps in different order. 

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163039969302/editor)

### Checklist

### DIFF
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -77,11 +77,11 @@ if (!customElements.get('quick-add-bulk')) {
 
       onCartUpdate() {
         return new Promise((resolve, reject) => {
-          fetch(`${window.location.pathname}?section_id=${document.getElementById('product-grid').dataset.id}`)
+          fetch(`${window.location.pathname}?section_id=${this.closest('.collection').dataset.id}`)
             .then((response) => response.text())
             .then((responseText) => {
               const html = new DOMParser().parseFromString(responseText, 'text/html');
-              const sourceQty = html.querySelector(`#quick-add-bulk-${this.dataset.id}`);
+              const sourceQty = html.querySelector(`#quick-add-bulk-${this.dataset.id}-${this.closest('.collection').dataset.id}`);
               this.innerHTML = sourceQty.innerHTML;
               resolve();
             })


assets/quick-order-list.js
```diff
@@ -154,6 +153,7 @@ class QuickOrderList extends HTMLElement {
         const html = new DOMParser().parseFromString(responseText, 'text/html');
         const sourceQty = html.querySelector(`#${this.quickOrderListId}`);
         this.innerHTML = sourceQty.innerHTML;
+        this.defineInputsAndQuickOrderTable();
```
sofiamatulis, line 156:
```suggestion
```

Can we add this to connected call back instead? Eventually well want to refactor this and take this function out since its reused in other `js` files like cart and quick add bulk

sofiamatulis (changes requested):
Left a comment

assets/quick-order-list.js
eugenekasimov: I tried to add it to `connectedCallback()` but it didn't work. The reason is because inside callback we need to run `this.onCartUpdate();` first and this takes time since we call API there. Even if I place `this.defineInputsAndQuickOrderTable();` after `this.onCartUpdate();` it won't wait until the `this.onCartUpdate();`done. The `this.defineInputsAndQuickOrderTable();` will run synchronously.

eugenekasimov: I can try to implement it with promises. Let me try it :)

eugenekasimov: Okay, that's how it looks now with another approach [f18f7b6](https://github.com/Shopify/dawn/pull/3360/commits/f18f7b643372a9dd2755e3353b79c759bfa2f60d)

It seems more complex but probably this approach is more appropriate. Thoughts?

dan-menard (approved):
Good idea to solve this with promises! 👍

sofiamatulis (commented):
Can we fix the conflicts?

https://screenshot.click/21-47-vw52x-fu10m.png

assets/quick-order-list.js
sofiamatulis: Yes this works :) thanks for updating it!

sofiamatulis (changes requested):
After something gets added to cart and 
- there is an error
- it works

Also it stops when it gets to an input with items in cart: https://screenshot.click/22-26-l24uj-seg15.png

the tabbing stops working

https://screenshot.click/22-21-et5yj-a76k1.mp4
https://screenshot.click/22-22-4rz5j-vgj2l.mp4
https://screenshot.click/22-23-i51tn-r7bp8.mp4


I think this may be related to this: https://github.com/Shopify/dawn/pull/3366

Let's merge that fix and pull it in to ensure it works for you

sofiamatulis (approved):
Thanks for changing this with the feedback :) Works well 🎉

assets/quick-add-bulk.js
```diff
@@ -73,16 +76,20 @@ if (!customElements.get('quick-add-bulk')) {
...
-          })
-          .catch(e => {
-            console.error(e);
-          });
+        return new Promise((resolve, reject) => {
+          fetch(`${window.location.pathname}?section_id=${document.getElementById('product-grid').dataset.id}`)
+            .then((response) => response.text())
+            .then((responseText) => {
+              const html = new DOMParser().parseFromString(responseText, 'text/html');
+              const sourceQty = html.querySelector(`#quick-add-bulk-${this.dataset.id}`);
```
sofiamatulis, line 84:
```suggestion
             const sourceQty = html.querySelector(`#quick-add-bulk-${this.dataset.id}-${this.closest('.collection').dataset.id}`);
```

assets/quick-add-bulk.js
```diff
@@ -73,16 +76,20 @@ if (!customElements.get('quick-add-bulk')) {
...
-          .then((responseText) => {
-            const html = new DOMParser().parseFromString(responseText, 'text/html');
-            const sourceQty = html.querySelector(`#quick-add-bulk-${this.dataset.id}-${this.closest('.collection').dataset.id}`);
-            this.innerHTML = sourceQty.innerHTML;
-          })
-          .catch(e => {
-            console.error(e);
-          });
+        return new Promise((resolve, reject) => {
+          fetch(`${window.location.pathname}?section_id=${document.getElementById('product-grid').dataset.id}`)
```
sofiamatulis, line 80:
```suggestion
fetch(`${window.location.pathname}?section_id=${this.closest('.collection').dataset.id}`)```

sofiamatulis (changes requested):

sofiamatulis (commented):
Let's ensure we are not overwriting the `quick-add-bulk` changes https://screenshot.click/22-40-2toih-t66ua.png

sofiamatulis (approved):

PR #3367, opened 2024-03-22
Increase max columns for collection gride to 6

dpasiliao:
### PR Summary: 

Increases the maximum columns for featured collections and the main collection product grid.

Based on the research we've done on current online stores, some merchants are starting to prefer a denser look to their shops where they can feature more products in their homepage.

Increased the maximum columns in the `featured-liquid

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist


dpasiliao:
> Thanks for this, @dpasiliao! It works well, no flags from a UX perspective. I think we should also increase the column limit for the Related Products block as well while we're doing this to keep that in sync. That way all the product rows/grids have the same column options.
> 
> We could also consider raising the limit from 5 to 6 for the Collection List and Collections List page if we'd like. Those also have a limit of 5 currently.

@kjellr Great idea! I've updated the PR to increase the column count for anything that maxes out at 5 to 6. I've updated the PR description to explicitly identify the changes, and I've updated the [Theme Editor demo](https://a7c44a-4.myshopify.com/admin/themes/163049930774/editor).

kjellr (approved):
Works well, thanks!

tyleralsbury (approved):
Looks good. 

This is a spot where having conditional settings would be so nice... so that we could have the limits per page be a multiple of each of the values as they are selected, which would avoid stragglers.

Alternatively, a "number of rows per page" setting might actually be better than a "total number of items per page" setting to accommodate this.  They could pick between say... 2 - 5 columns per page and always have full rows (for pages that are full).

@kjellr do you think this sort of thing would be worth the upgrade toil or am I overthinking it a bit?

kjellr:
Something we can think about for later, but I think this is ok for now. I think it's relatively natural for folks to think in terms of total items, instead of rows.

melissaperreault:
Love seeing this! 👏 
We could increase the max of it though for `Products per page` to accommodate for a wider column number selection. Collection pages would get faster to 24 and might impact the scrolling perception experience for buyers. I'll bring it up if this is an issue, it's easy to adjust, not saying this wouldn't impact performance though...

PR #3373, opened 2024-03-26
🐛 Quick order list and combined listings

lhoffbeck:
### PR Summary: 

Fixes JS error when reloading the quick order list section and titles for combined listings products

**JS error**
- We were throwing a JS error because the custom element was already defined. I wrapped the element definition in an existence check like we do [elsewhere in dawn](https://github.com/Shopify/dawn/blob/main/assets/product-form.js#L1-L3)

**Variant titles**
- Combined listings are a group of related products, surfaced as a single product. In this case, we can't accurately rely on `has_only_default_variant` to determine whether a variant is default or not, because `has_only_default_variant` still returns `false` if a combined listing is made up of child products that each only have a single variant. For example, if we have a combined listing "Pants" that has children "Pants - blue", "Pants - green" where both blue and green have no additional options, the product has 2 total variants where each is a default variant.
- To fix this, I am currently relying on a default option name and default variant name check, but IMO we should fix this by extending the variant drop to have a `is_default_variant` value. This also enabled us to simplify quick order list row input to just take a variant instead of a product or a variant. 
- My changes also fix a bug I introduced where products with only default variants show `Default title` instead of the product title.

#### Combined listing parent
**Before**

Screenshot 2024-03-26 at 2 57 31 PM

**After**

Screenshot 2024-03-26 at 2 56 35 PM   

#### Combined listing child

**Before**
Screenshot 2024-03-26 at 2 57 35 PM

**After**

 Screenshot 2024-03-26 at 2 56 39 PM

#### Product with options

**Before**
Screenshot 2024-03-26 at 2 57 40 PM

**After**

Screenshot 2024-03-26 at 2 56 43 PM

#### Product with only default variant

**Before**

Screenshot 2024-03-26 at 2 57 45 PM

**After**

Screenshot 2024-03-26 at 2 56 50 PM

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/quick-order-list.js
+++ b/assets/quick-order-list.js
@@ -1,467 +1,510 @@
-class QuickOrderListRemoveButton extends HTMLElement {
-  constructor() {
-    super();
-    this.addEventListener('click', (event) => {
-      event.preventDefault();
-      const quickOrderList = this.closest('quick-order-list');
-      quickOrderList.updateQuantity(this.dataset.index, 0);
-    });
+if (!customElements.get('quick-order-list-remove-button')) {
+  class QuickOrderListRemoveButton extends HTMLElement {
+    constructor() {
+      super();
+      this.addEventListener('click', (event) => {
+        event.preventDefault();
+        const quickOrderList = this.closest('quick-order-list');
+        quickOrderList.updateQuantity(this.dataset.index, 0);
+      });
+    }
   }
+
+  customElements.define('quick-order-list-remove-button', QuickOrderListRemoveButton);
 }
 
-customElements.define('quick-order-list-remove-button', QuickOrderListRemoveButton);
+if (!customElements.get('quick-order-list-remove-all-button')) {
+  class QuickOrderListRemoveAllButton extends HTMLElement {
+    constructor() {
+      super();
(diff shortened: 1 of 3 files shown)


assets/quick-order-list.js
lhoffbeck, line 1: I highly recommend [hiding whitespace](https://github.com/Shopify/dawn/pull/3373/files?diff=split&w=1) to make this diff easier to understand

assets/quick-order-list.js
```diff
@@ -1,467 +1,510 @@
...
-
-    if (!hasVariantsInCart) {
-      this.classList.add('hidden');
-    }
-
-    this.actions = {
-      confirm: 'confirm',
-      remove: 'remove',
-      cancel: 'cancel'
+if (!customElements.get('quick-order-list-remove-button')) {
```
sofiamatulis, line 1: We actually already have these changes on `main` . It was from the quick add bulk PR

https://github.com/Shopify/dawn/blob/main/assets/quick-order-list.js#L1

I recommend rebasing off of main

lhoffbeck: oh amazing 🙏 🙏 🙏 🙏

Issue #3340, opened 2024-03-12
"You can only add x of this item to your cart" error


keralab:
## Describe the current behaviour

https://github.com/Shopify/dawn/assets/14318515/e56ca056-c457-422d-848c-0fecc5a30346

The problem seems to be that when updating a line that's part of a bxgy, the line that was modified might not result in the new quantity and causes and error. For example, if I have b2g1;

line 1, product 1, quantity 2, discount applied
line 2, product 1, quantity 1, discount applied
line 3, product 1, quantity 1, no discount

And then i make a request to set line 2 to quantity 2, it results in this;

line 1, product 1, quantity 2, discount applied
line 2, product 1, quantity 1, discount applied "Error: You can only add x of this item to your cart"
line 3, product 1, quantity 2, no discount

## Describe the expected behavior

There should be no error.

## Version information (Dawn, browsers and operating systems)

- Dawn Version: 7.0.1
- Chrome Version 108.0.5359.124
- macOS Version 13.1

## Possible solution

[This condition](https://github.com/Shopify/dawn/blob/main/assets/cart.js#L149) checks that the inputted quantity for the line matches the quantity of the line in the cart api response, but it is possible and correct for those quantities to be different in some scenarios.

## Additional context/screenshots
Original cart issue: https://github.com/Shopify/shopify/issues/478497

PR #3395, opened 2024-04-04
[Fix] QOL Cart race conditions

sofiamatulis:
### PR Summary: 

Update Quick Order List to allow multiple variants to update at once using the `update.js` endpoint in the cart API. This is to avoid cart racing conditions.

Here is the **before**:
cart

Here is the **after**
30-22-ciuwu-99ay8

We were running into an issue before where if the buyer was going fast, due to cart racing some line items werent updated and being overwritten by the next calls

You also couldnt keep clicking +/- and update the cart items if a call was in progress

- Lump all the variants in a queue in 1 call together actually and call the **update.js** instead of **change.js** or **add.js** per variant. 
- Adding a timer/check to ensure we only call the next queue of variants when the first one is done

### Other considerations

I did notice there is clamping/errors dont necessarily show up when using the `update` endpoint. We have an issue for it and are investigating this problem https://github.com/Shopify/shopify/issues/440605

### Testing steps/scenarios

- [ ] Test QOL with qty rules
- [ ] Test QOL with +/-
- [ ] Test QOL by going fast from multiple inputs
- [ ] Do **not** test products with discounts for now. Issue here https://github.com/Shopify/dawn/issues/3340

### Demo links
- [Editor](https://admin.shopify.com/store/os2-demo/themes/163077783574/editor)


scottgmeadows:
Know it's early but here's a few issues I'm seeing.

1 - [Some requests are still dropped when you go fast](https://screenshot.click/04-39-tl9d7-uxzhe.mp4)
2 - [Whenever the cart call finishes, the input loses focus](https://screenshot.click/04-53-3vzfk-u1mxr.mp4)
3 - [Enter 0 doesn't update the cart](https://screenshot.click/04-54-0f229-yyjl1.mp4)
4 - [Quantities inputs flicker to different values while being added to cart](https://screenshot.click/04-56-0vwfy-ge4t3.mp4)

tyleralsbury:
> 2 - [Whenever the cart call finishes, the input loses focus](https://screenshot.click/04-53-3vzfk-u1mxr.mp4)

For this one, I think we'll need to check which input is currently focused (by variant ID) just before the rows re-render and re-focus that.  That way, even if a few renders happen in succession, the focus stays on the row the person was already on.

>  4 - [Quantities inputs flicker to different values while being added to cart](https://screenshot.click/04-56-0vwfy-ge4t3.mp4)

For this one, I didn't check what you did, Sofia, but if you're re-rendering all of the rows now you'll probably need to, instead, just re-render the ones that actually had a change.  You can use the variant IDs from the request that was made to know which ones need to be replaced.

assets/quick-order-list.js
```diff
@@ -592,9 +518,9 @@ if (!customElements.get('quick-order-list')) {
         this.updateLiveRegions(id, message);
       }
 
-      cleanErrors() {
-        this.querySelectorAll('.desktop-row-error').forEach((error) => error.classList.add('hidden'));
-        this.querySelectorAll(`.variant-item__error-text`).forEach((error) => (error.innerHTML = ''));
+      cleanErrors(id) {
+        // this.querySelectorAll('.desktop-row-error').forEach((error) => error.classList.add('hidden'));
```
sofiamatulis, line 522: This is commented out for now given that there is clamping happening + no errors returned from the BE. More info here: https://github.com/Shopify/shopify/issues/440605

assets/quick-order-list.js
```diff
@@ -262,60 +275,65 @@ if (!customElements.get('quick-order-list')) {
         this.querySelectorAll('quantity-input').forEach((qty) => {
           const debouncedOnChange = debounce((event) => {
             this.onChange(event);
-          }, ON_CHANGE_DEBOUNCE_TIMER);
+          }, 100);
```
sofiamatulis, line 278: Because we are assuming the buyer will now click very quickly (or could), it makes sense for us to have this be a shorter amount. I spoke to Tyler before too and the previous amount was arbitrary

assets/bulk-add.js
```diff
@@ -0,0 +1,37 @@
+class BulkAdd extends HTMLElement {
```
sofiamatulis, line 1: I will be re-using some of this functionality for this: https://github.com/orgs/Shopify/projects/6397/views/28?pane=issue&itemId=59790024

So I am creating an extendable component I will also use for Quick Add

assets/bulk-add.js
```diff
@@ -0,0 +1,37 @@
...
+  sendRequest(queue) {
+    this.requestStarted = true;
+    const items = {}
+    const ids = []
+    queue.forEach((q) => {
+      items[parseInt(q.id)] = q.quantity;
+    });
+    queue.forEach((q) => {
+      ids.push(q.id)
+    })
```
dan-menard, line 31: These don't need to be separate `forEach` loops, you can do the `items` assignment and the `ids` push in the same iteration.

assets/bulk-add.js
```diff
@@ -0,0 +1,37 @@
...
+    this.queue = this.queue.filter(q => !queue.includes(q));
+    this.updateMultipleQty(items, ids)
```
dan-menard, line 33: This is a little scary—that function isn't defined in this class, so there's no guarantee it will exist if this class is re-used in a different context. Does it need to be somewhere else, or could it move into this class?

assets/quick-order-list.js
```diff
@@ -178,13 +176,26 @@ if (!customElements.get('quick-order-list')) {
...
+      //   const int = setInterval(() => {
+      //     if (this.queue.length > 0) {
+      //       if (!this.requestStarted)  {
+      //         this.sendRequest(this.queue)
+      //       }
+      //     } else {
+      //       clearInterval(int)
+      //     }
+      //   }, 250)
+      // }
```
dan-menard, line 197: 🔥?

assets/quick-order-list.js
dan-menard: Should this match the timeout for the queue in the bulk-add component?

https://github.com/Shopify/dawn/pull/3395/files#diff-7b699c3e0ddf5492f7762bfb759ca470eba4f3e8ff2696351106ef84404fc5f6R18

assets/quick-order-list.js
```diff
@@ -262,60 +273,65 @@ if (!customElements.get('quick-order-list')) {
...
+      //   const ids = []
+      //   queue.forEach((q) => {
+      //     items[parseInt(q.id)] = q.quantity;
+      //   });
+      //   queue.forEach((q) => {
+      //     ids.push(q.id)
+      //   })
+      //   this.queue = this.queue.filter(q => !queue.includes(q));
+      //   this.updateMultipleQty(items, ids)
+      // }
```
dan-menard, line 333: 🔥?

assets/bulk-add.js
sofiamatulis: good idea :)

assets/bulk-add.js
sofiamatulis: So I wanted to move this to this class but because I havent done the refactoring for quick add bulk yet, i didnt want to have it on this PR yet 🤔 

I thought it was worth it to have this live in a reusable component given I will reuse this in quick add bulk

Also not all functions will be exactly the same for QOL and quick add bulk, but I figured I could call them the same - or do you think this would still feel a bit risky?

assets/quick-order-list.js
sofiamatulis: I havent gotten to doing the quick add bulk work yet, but that will be in a follow up here https://github.com/orgs/Shopify/projects/6397/views/28?pane=issue&itemId=59790024 :)

assets/quick-order-list.js
sofiamatulis: Yes :) just wanted to get some thoughts on the extendable component. But deleting for now

assets/bulk-add.js
dan-menard: ahhh that makes sense. And yeah, no problem with having them same name for both (because I was raised on Object Orientation, I call it [Polymorphism](https://stackify.com/oop-concept-polymorphism/) and consider it a good thing)

dan-menard: And fine to leave it as-is for now 👍

assets/bulk-add.js
```diff
@@ -0,0 +1,35 @@
+class BulkAdd extends HTMLElement {
+  constructor() {
+    super();
+    this.queue = []
+    this.requestStarted = false;
+  }
+
+  startQueue(id, quantity) {
+    this.queue.push({id, quantity})
+    const int = setInterval(() => {
```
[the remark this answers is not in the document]
sofiamatulis: Sure :)

assets/bulk-add.js
```diff
@@ -0,0 +1,35 @@
+class BulkAdd extends HTMLElement {
+  constructor() {
+    super();
+    this.queue = []
```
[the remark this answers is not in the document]
sofiamatulis: Updated it 🤔 but itll update the dawn repo which has different settings

https://github.com/Shopify/dawn/commit/9b482c80cf9f6424f747acddd73dcf049b648dbc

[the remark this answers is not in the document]
sofiamatulis: Let's handle this in a separate PR

assets/quick-order-list.js
```diff
@@ -1,13 +1,12 @@
...
     'quick-order-list-remove-button',
-    class QuickOrderListRemoveButton extends HTMLElement {
+    class QuickOrderListRemoveButton extends BulkAdd {
       constructor() {
         super();
         this.addEventListener('click', (event) => {
           event.preventDefault();
-          const quickOrderList = this.closest('quick-order-list');
-          quickOrderList.updateQuantity(this.dataset.index, 0);
+          this.startQueue(this.dataset.index, 0);
```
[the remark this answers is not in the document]
sofiamatulis: Good catch 👍 fixed

dan-menard (approved):

PR #3421, opened 2024-04-15
Fix selector for Quick Add Bulk

sofiamatulis:
### PR Summary: 

Originally we moved the `section.id` to account for Quick Add Bulk: https://github.com/Shopify/dawn/commit/889e2a08a6bdb452381a07cd9bcd5088e1018be6
https://screenshot.click/15-39-ew18j-2wwe4.png

This was causing an issue for Filtering: Clearing any filter throws this section_id undefined error 
https://screenshot.click/15-51-emwn5-ioys8.png https://github.com/Shopify/dawn/blob/main/assets/facets.js#L248

I have reverted where section.id is and created a different selector for Quick Add Bulk

Test
- Quick Add Bulk (in collection page and feat collection)
- Filtering

Editor: https://admin.shopify.com/store/os2-demo/themes/163124576278/editor

### DIFF
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -177,9 +177,9 @@ if (!customElements.get('quick-add-bulk')) {
       getSectionsToRender() {
         return [
           {
-            id: `quick-add-bulk-${this.dataset.id}-${this.closest('.collection').dataset.id}`,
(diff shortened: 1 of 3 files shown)


tyleralsbury (approved):
Tested on release branch and then here - problem was there and is now fixed. 👍

PR #3429, opened 2024-04-19
[Prod page] Add check to prevent errors if no feat. image exists

ludoboludo:
I wonder if this still works with combined listing or not? 

Does the media id needs to grab `product.featured_media.id` in its place when there is no variant id attached to the currently selected option to re render the new product page ?

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -1154,9 +1154,11 @@ class VariantSelects extends HTMLElement {
       });
     }
 
-    document
-      .querySelector(`[id^="MediaGallery-${this.dataset.section}"]`)
-      ?.setActiveMedia?.(`${this.dataset.section}-${this.currentVariant.featured_media?.id}`);
+    if (this.currentVariant.featured_media) {
+      document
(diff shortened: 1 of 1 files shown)


assets/global.js
```diff
@@ -1154,9 +1154,11 @@ class VariantSelects extends HTMLElement {
       });
     }
 
-    document
-      .querySelector(`[id^="MediaGallery-${this.dataset.section}"]`)
-      ?.setActiveMedia?.(`${this.dataset.section}-${this.currentVariant.featured_media?.id}`);
+    if (this.currentVariant.featured_media) {
+      document
+        .querySelector(`[id^="MediaGallery-${this.dataset.section}"]`)
+        ?.setActiveMedia?.(`${this.dataset.section}-${this.currentVariant.featured_media?.id}`);
```
lhoffbeck, line 1160: nit -- we should be able to remove the optional chain here `this.currentVariant.featured_media?.id` since we already check that featured_media is preesent

lhoffbeck (approved):

PR #3431, opened 2024-04-22
Fix quick add color scheme

Roi-Arthur:
### PR Summary: 
Moving the `color-scheme` class further up the DOM by a few elements fixes color issue in Quick Add modal.

When using a quick add modal in a product grid, the color-scheme setting of the section only partially applied to the modal.
[](https://screenshot.click/22-17-9opuw-brkns.png)

Moved the class a bit

Fixes color issue

### Testing steps/scenarios
- [ ] Install Dawn 14
- [ ] Navigate to collection page
- [ ] Enable Quick add (standard or bulk) – make sure to reload to test
- [ ] Change the color scheme of the section and check issue is fixed

Fix demo: https://screenshot.click/22-21-234hq-ru38a.gif

### DIFF
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -503,12 +503,7 @@
                           </div>
                         {%- endif -%}
                       </div>
-                      {%- render 'quick-order-list',
-                        product: card_product,
-                        show_image: true,
-                        show_sku: true,
-                        is_modal: true
-                      -%}
+                      {%- render 'quick-order-list', product: card_product, show_image: true, show_sku: true -%}
                     </div>
                   </div>
                 </div>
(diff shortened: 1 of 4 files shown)


snippets/card-product.liquid
```diff
@@ -330,7 +330,10 @@
                   >
                     {% render 'icon-close' %}
                   </button>
-                  <div id="QuickAddInfo-{{ card_product.id }}" class="quick-add-modal__content-info"></div>
+                  <div
+                    id="QuickAddInfo-{{ card_product.id }}"
+                    class="color-{{ section.settings.color_scheme }} quick-add-modal__content-info"
```
ludoboludo, line 335: I don't think it needs to be added here, does it ? This is for the `standard` quick add which doesn't have any color scheme issues at the moment. When I check on the storefront, I see two color schemes class applied:

PR #3450, opened 2024-05-02
Change h elememts to divs with associated classes for better seo

skati-apite:
### PR Summary: 

According to SEO, H-headings should only be used to structure texts and anything that is not continuous text should not receive H-formatting.

At the shop level, there are areas that have been stored as H formatting. For example, product-level phrases such as "Item has been added to the shopping cart", "Filter and sort", "Filter" were stored as H formatting, which contradict the actual benefit / use of H formatting in SEO."

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/sections/featured-blog.liquid
+++ b/sections/featured-blog.liquid
@@ -113,9 +113,9 @@
                     {%- endif -%}
                     <div class="card__content">
                       <div class="card__information">
-                        <h3 class="card__heading h2">
+                        <div class="card__heading h2">
                           {{ 'sections.featured_blog.onboarding_title' | t }}
-                        </h3>
+                        </div>
                         <p class="article-card__excerpt rte-width">
                           {{ 'sections.featured_blog.onboarding_content' | t }}
                         </p>
@@ -124,9 +124,9 @@
                   </div>
                   <div class="card__content">
                     <div class="card__information">
-                      <h3 class="card__heading h2">
+                      <div class="card__heading h2">
                         {{ 'sections.featured_blog.onboarding_title' | t }}
-                      </h3>
+                      </div>
                       <p class="article-card__excerpt rte-width">
                         {{ 'sections.featured_blog.onboarding_content' | t }}
                       </p>
--- a/sections/footer.liquid
+++ b/sections/footer.liquid
@@ -76,7 +76,7 @@
                 {% endif %}
               >
                 {%- if block.settings.heading != blank -%}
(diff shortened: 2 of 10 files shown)


skati-apite:
Signed the CLA!

skati-apite:
I have signed the CLA!

notkurt:
Would have to agree with this change in principal, but not in practice. 

Dawn is far too overzealous when it comes to its use of heading tags and does not follow a logical structure, H1 > H2 > H2 etc. We’ve worked with a number of SEO agencies who raise heading structure as an issue, on a number of themes - including Dawn.  

We’ve done a bit of internal research on this since SEO agencies all have their own approaches. Our research mainly involved insurance sites, of all places, due to the fact it’s a far more competitive from an SEO perspective than eCommerce. Ad costs are through the roof in those kind of verticals, which means SEO ends up being an area they invest in pretty heavily. 

What we found was that the majority of the larger insurance providers had a logical heading structure, did not use heading tags for footer headings, did not use heading tags for things such as “Country/region”, “Filter” etc.  

Even looking at a site like Amazon, they’re not using H3’s for product cards, and they are definitely not using heading tags for navigational elements.  

Are these critical ranking factors? Probably not. Amazon doesn’t even have a H1 on their homepage. 

Does it probably play a small part? Maybe. 

Has Dawn got it right? I don’t think so.

In saying this, I don’t think this is right for Dawn. An incremental update is not the time to make major changes to a heading structure on a site. This would be better geared towards a future reference theme.

tomatic:
> > According to SEO, H-headings should only be used to structure texts and anything that is not continuous text should not receive H-formatting.
> 
> I see where you're coming from, but there's still no solid reason to assume this would negatively effect SEO. Moving from h headings to only describe the contents of the page would negatively impact screen readers or other accessibility tools trying to describe the ui

The point here is that Dawn is using heading tags for things that are not headings, like the country/language selector, collection page filters, cart notification messages, etc. Why is Dawn using `h2` tags for these elements?

Headings are meant to structure page content and they do play a role in SEO:

https://www.searchenginejournal.com/on-page-seo/header-tags/

So the only elements where themes should use headings are

- `h1` for the page titles  
- `h2` or lower for section headings and subheadings
- `h2` or lower for product titles on collection pages 

There might me some more places where heading tags would make sense, but I don't see any good reason to use them in the way Dawn is using them.

Issue #3448, opened 2024-05-01
Quick add media isn't update on variant change


ludoboludo:
## Describe the current behavior
In the quick add modal, for a product that has different variants and images tied to them, if you change the variant it doesn't update the media it's showing you when it should and used to until the previous version `13.0.1`. 

## Describe the expected behavior
When changing variant, the image shown should be updated as well. 

## Version information (Dawn, browsers and operating systems)

- Dawn Version: 14.0.0

## Possible solution
Some refactor was done around product medias and this use case was missed. In `updateMedia(html) {...}` we're setting the `mediaGalleryDestination` based on an ID using `this.dataset.section` but because we've edited those IDs in `preventDuplicatedIDs()` an added `quickadd-` in front of it, it's not finding the media list in the new `html` we're querying. 
We would need to potentially update `this.dataset.section` value and remove the `quickadd-` or something like that. 

```javascript
const mediaGalleryDestination =
      html.querySelector(`[id^="MediaGallery-${quickAddTransformedId}"] ul`) ||
      html.querySelector(`[id^="MediaGallery-${this.dataset.section.replace('quickadd-', '')}"] ul`);
```

## Additional context/screenshots

[Gif of the issue](https://screenshot.click/30-35-kh127-x7z3q.gif)

PR #3451, opened 2024-05-03
Fix variant media not changing in the quick add modal

ludoboludo:
### PR Summary: 

In the quick add modal, on version 14, when you change variant it doesn't update media shown when it used to. 

Fixes https://github.com/Shopify/dawn/issues/3448

From what I see it's looking a `mediaGalleyDestination` but it can't grab any cause `this.dataset.section` can't be found in the new `html`. ([recording](https://screenshot.click/01-59-692om-f0isq.mp4))

With this new approach, it seems like the issue comes from the fact that we have `preventDuplicatedIDs()` to prevent duplicate IDs in the quick add modal by adding `quickadd-`. So when we do 
```javascript
const mediaGalleryDestination = html.querySelector(`[id^="MediaGallery-${this.dataset.section}"] ul`);
```
It's not finding that destination as it doesn't have a similar `ID`. 
~We could tweak that `data-section` and remove the `quickadd-`  within `updateMedia(html){...}` 🤷~

In the function that prevents the duplicate IDs we're setting a data attribute that keeps the original section ID. So we're using this in priority if it's present and we don't need to do the string manipulation 👍 

If I do this instead it seem to fix the issue: 
```javascript
const sectionId = this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section;
...
const mediaGalleryDestination =
      html.querySelector(`[id^="MediaGallery-${sectionId}"] ul`);
```

It now works when changing variant that has a feat. image

### Testing steps/scenarios

- [ ] Set the feat. collection on the homepage to use quick add standard
- [ ] Then open the modal for a product with variant images like Louise slide sandal or Puppy
- [ ] Change variant to see the image updated
- [ ] Also test and make sure it still works as expected on the product page and feat. product section

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163207184406/editor)

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -1103,10 +1103,9 @@ class VariantSelects extends HTMLElement {
   }
 
   updateMedia(html) {
+    const sectionId = this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section;
     const mediaGallerySource = document.querySelector(`[id^="MediaGallery-${this.dataset.section}"] ul`);
-    const mediaGalleryDestination = html.querySelector(
-      `[id^="MediaGallery-${this.dataset.section.replace('quickadd-', '')}"] ul`
-    );
+    const mediaGalleryDestination = html.querySelector(`[id^="MediaGallery-${sectionId}"] ul`);
 
     const refreshSourceData = () => {
       const mediaGallerySourceItems = Array.from(mediaGallerySource.querySelectorAll('li[data-media-id]'));


jonpennington:
Could we perhaps use this functionality instead? It would follow similar logic used elsewhere in global.js. The problem seems to originate with how Section Rendering API now works.

```
const sectionId = this.dataset.originalSection ? this.dataset.originalSection : this.dataset.section;
const mediaGallerySource = document.querySelector(`[id^="MediaGallery-${this.dataset.section}"] ul`);
const mediaGalleryDestination = html.querySelector(`[id^="MediaGallery-${sectionId}"] ul`);
```

sofiamatulis (approved):
Nice! Works as expected 🎉

PR #3484, opened 2024-05-22
Use translation keys for default values in liquid files

dpasiliao:
### PR Summary: 

Use translation keys as default values for sections

This updates Dawn to make use of the new update that allows default content to be translated and rendered. 

Manually updated all the default values in the liquid schemas with translation keys and moved those values to the `en.default.schema.json` 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

Any default value when a merchant adds a new section in the theme editor will now match the default language of their store

### Testing steps/scenarios

- [ ] While waiting for translations to arrive, add each section from the theme into your store and make sure that the strings that appear in the section are not translation keys (i.e. `t:some.translation.key`)
- [ ] Once translations arrive, change your store's default language to a language is not English and has available translations
- [ ] Add a new section and check if the values of the section match the language you've set your store's default language to

### Demo links

- [Store](https://19b2dc.myshopify.com/?preview_theme_id=136547795095)
- [Editor](https://19b2dc.myshopify.com/admin/themes/136547795095/editor)

### Checklist


locales/en.default.schema.json
```diff
@@ -992,10 +1007,12 @@
...
           "settings": {
             "heading": {
-              "label": "Heading"
+              "label": "Heading",
+              "default": "Heading"
             },
             "subtext": {
-              "label": "Subtext"
+              "label": "Subtext",
+              "default": "<p>Share contact information, store details, and brand content with your customers.</p>"
```
dan-menard, line 1015: It's generally an antipattern to put HTML in translation files. This has burned me before a few times, like when a mobile native app reuses our locale files, or translators mis-interpreting what the HTML is doing.

Why do we need to do it here?

sections/email-signup-banner.liquid
```diff
@@ -342,7 +342,7 @@
         {
           "type": "richtext",
           "id": "text",
-          "default": "<p>Be the first to know when we launch.</p>",
+          "default": "t:sections.email-signup-banner.blocks.paragraph.settings.paragraph.default",
```
dan-menard, line 345: Ah yeah, related to my previous comment, we should keep the `<p>` tags here in the liquid file, rather than the translations.

dan-menard (changes requested):

dan-menard (approved):

Al-Campuzano (approved):
LGTM :)

Issue #3445, opened 2024-04-30
[Quick Add Bulk] Disable button / quantity picker on card when all variants are out of stock


nicklepine:
When Quick Add is set to "Bulk" and a product's sole variant, or all of its variants are Out of Stock, we should disable the button / quantity input, like we do with Quick Add "Standard" 

![Image](https://github.com/Shopify/dawn/assets/8127145/f00fba95-c806-4d70-be63-9f61c9a50f30)

PR #3487, opened 2024-05-24
Add an out of stock note for quick add bulk, like the one for quick add standard

dan-menard:
Fixes https://github.com/Shopify/dawn/issues/3445

I copied the button code over from the quick-add-standard block ([here](https://github.com/Shopify/dawn/pull/3487/files#diff-167fc70beb2d8c5dbf3e81a213fd3808f6b2e1d1c60590a6863724dc0dca3ea6R354-R381)) and tweaked it a bit, so there's some duplication. I picked this approach because:

(1) Extracting the common code into a snippet is messy, because it uses so many local assigns and passing a huge thing of assigns is almost as messy and more indirect than duplicating some liquid markup.

(2) Modifying the quantity-input snippet to include out of stock would have consequences on other pages, I didn't want to bite off that much scope.

But let me know if there's another approach I should have considered, or a strong preference for one of the approaches above.

Here's how it looks before:
image

Here's how it looks in this PR:
image

- [Editor](https://admin.shopify.com/store/dan-theme-test-shop/themes/159008751672/editor)

### DIFF
--- a/snippets/card-product.liquid
+++ b/snippets/card-product.liquid
@@ -287,10 +287,10 @@
             {%- endif -%}
           </div>
         </div>
+        {% assign product_form_id = 'quick-add-' | append: section_id | append: card_product.id %}
         {% if quick_add == 'standard' %}
           <div class="quick-add no-js-hidden">
             {%- liquid
-              assign product_form_id = 'quick-add-' | append: section_id | append: card_product.id
               assign qty_rules = false
               if card_product.selected_or_first_available_variant.quantity_rule.min > 1 or card_product.selected_or_first_available_variant.quantity_rule.max != null or card_product.selected_or_first_available_variant.quantity_rule.increment > 1
                 assign qty_rules = true
@@ -391,7 +391,24 @@
               class="quick-add-bulk"
               data-index="{{ card_product.selected_or_first_available_variant.id }}"
             >
-              {% render 'quantity-input', variant: card_product.selected_or_first_available_variant, min: 0 %}
+              {% if card_product.selected_or_first_available_variant.available == false %}
+                <button
+                  id="{{ product_form_id }}-submit"
+                  type="submit"
+                  name="add"
+                  class="quick-add__submit button button--full-width button--secondary"
+                  aria-labelledby="{{ product_form_id }}-submit title-{{ section_id }}-{{ card_product.id }}"
+                  data-sold-out-message="true"
+                  disabled
+                >
+                  <span>{{ 'products.product.sold_out' | t }}</span>
+                  <span class="sold-out-message hidden">
+                    {{ 'products.product.sold_out' | t }}
+                  </span>
+                </button>
+              {% else %}
+                {% render 'quantity-input', variant: card_product.selected_or_first_available_variant, min: 0 %}
+              {% endif %}
             </quick-add-bulk>
           {% else %}
             <div class="quick-add no-js-hidden">


snippets/card-product.liquid
```diff
@@ -384,7 +384,31 @@
             {%- endif -%}
           </div>
         {% elsif quick_add == 'bulk' %}
-          {% if card_product.variants.size == 1 %}
+          {% if card_product.selected_or_first_available_variant.available == false %}
```
sofiamatulis, line 387: It looks like we have some extra logic here for buy buttons for when it's supposed to say sold out

https://github.com/Shopify/dawn/blob/main/snippets/buy-buttons.liquid#L77-L79

templates/index.json
```diff
@@ -21,7 +21,7 @@
           "type": "buttons",
```
sofiamatulis, line 21: Let's revert this. I think you may need to enable `one sync` with github

dan-menard: Actually how do you do this? I think I've always just connected them with two-way sync and can't find the button to change it.

snippets/card-product.liquid
dan-menard: ooh that's interesting. I copied this condition from the logic for quick add standard: https://github.com/Shopify/dawn/pull/3487/files#diff-167fc70beb2d8c5dbf3e81a213fd3808f6b2e1d1c60590a6863724dc0dca3ea6R368-R372

Do we want to keep this change consistent with quick add standard, or update both to copy what we're doing for buy-buttons? I'm not sure why the buy-button condition and quick add standard are currently different, is there risk in changing how we do it for quick add standard?

templates/index.json
dan-menard: That's where I figured it would be, but I don't have that button for some reason 😕 
image

sofiamatulis: Ah I have seen this before - let me message you

snippets/card-product.liquid
sofiamatulis: @ludoboludo do you have context on this?

snippets/card-product.liquid
```diff
@@ -384,7 +384,31 @@
...
+            <quick-add-bulk
```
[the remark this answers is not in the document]
sofiamatulis: Ah I see you have updated the logic, so it may not be worth it to change it to save some lines of code 👍

snippets/card-product.liquid
ludoboludo: Nope. I don't think there is a specific reason. My guess would be that when we added quantity rule logic to the theme we didn't apply to the quick add logic 🤷 Seems like it could have been an easy miss

ludoboludo: I could see how the logic for quantity rules could be useful. I haven't tested it though 🤔 
The rest of logic with `inventory_management` and `inventory_policy` isn't needed because it's part of the `.available` check:

snippets/card-product.liquid
```diff
@@ -391,7 +391,24 @@
               class="quick-add-bulk"
               data-index="{{ card_product.selected_or_first_available_variant.id }}"
             >
-              {% render 'quantity-input', variant: card_product.selected_or_first_available_variant, min: 0 %}
+              {% if card_product.selected_or_first_available_variant.available == false %}
+                <button
+                  id="{{ product_form_id }}-submit"
+                  type="submit"
+                  name="add"
+                  class="quick-add__submit button button--full-width button--secondary{% if horizontal_quick_add %} card--horizontal__quick-add{% endif %}"
```
ludoboludo, line 399: I think here we don't need the check for `horizontal_quick_add` as this is needed specifically for complementary products which doesn't support quick add bulk at this point 👍

ludoboludo (commented):
I think it might be good to use the quantity rules check that Sofia mentioned too

snippets/card-product.liquid
dan-menard: Yeah, +1 to do this as a follow-up. I think if we want to align quick add/quick order list with the buy button logic, we should look at both options (buy button using what QOL has) and consider what else we might need to add to both places, etc.

Logged an issue here: https://github.com/Shopify/dawn/issues/3499

ludoboludo (approved):

Issue #3406, opened 2024-04-08
Quick Add Bulk UX feedback


sofiamatulis:
## Describe the enhancement you'd like

Desktop
- [ ] Explore animation for Modal when it opens
- [ ] Revisit `X` close button and contrast to the modal and how it disappears as you scroll
- [ ] Explore animation of popover when it is exposed
- [ ] Extra spacing in the footer when you scroll down
- [ ] As you are clicking on +/-, it gets blocked when the spinner starts (API call starts). Can we not prevent the buyer from clicking on the button?

Mobile
- [ ] Explore making the modal full width (sheet?) - `x` button close to some of the text due to lack of spacing
- [ ] Footer is not sticky?

sofiamatulis:
cc @melissaperreault @scottgmeadows

sofiamatulis:
@nicklepine There are some P1 tickets here, can we move them to another ticket?

PR #3488, opened 2024-05-24
Quick add modal fixes for Desktop

dan-menard:
Fixes https://github.com/Shopify/dawn/issues/3406

This fixes the issues for the Quick Add modal at various larger browser sizes. Each fix is in its own commit, if you want to review the fixes individually, which is a good idea because they touch a lot of the same files.

There's more context and some screenshots etc for each fix in the little expandy-sections below.

And here's an [editor](https://admin.shopify.com/store/dan-theme-test-shop/themes/163275309112/editor) with these changes.

<details>
<summary>Center "View cart label" in the Quick add modal</summary>
[commit](https://github.com/Shopify/dawn/pull/3488/commits/cf36a20b742a6178413c46bfed9a5039b5a037db)

This was a quick one, just needed a "text-align: center" on the button. It doesn't impact quick order list, as the button wraps to a new line before it runs into this problem.

image
image
</details>

<details>
<summary>Quick add bulk modal header spacing is off</summary>
[commit](https://github.com/Shopify/dawn/pull/3488/commits/981afdcc03898e94c1bde09fdf46734b51af6323)

Ludo recorded a lovely [video](https://screenshot.click/28-17-cqjzi-3qqkb.mp4) demonstrating this bug. I aligned the breakpoints as he suggested in [this comment](https://github.com/Shopify/dawn/pull/3488#discussion_r1616976257).

image
image
</details>

<details>
<summary>Clicking on the product name should link to the product details page</summary>
[commit](https://github.com/Shopify/dawn/pull/3488/commits/682c02e55b95c0ebdde557123660da1b761c2d22)

Another quick one, copied the link styles for the hover state from the product cards.

image
image
</details>

<details>
<summary>Remove the extra underlined space from the "view details" link</summary>
[commit](https://github.com/Shopify/dawn/pull/3488/commits/da18752297b8dc31df7366159e5eb225c09e4e0a)

This was slightly maddening, but the fix was relatively elegant in the end. I targeted it specifically to the quick add modal, because I couldn't figure out where else this text was shown—if anyone knows, raise a flag, we might as well fix it for other cases too.

image
image
</details>

<details>
<summary>Extra spacing in the footer when you scroll down</summary>
[commit](https://github.com/Shopify/dawn/pull/3488/commits/7a373b9c48467705932ae00117c778b473c6c990)

Another easy one, just had to reduce the padding-bottom when we're running quick order list in the modal.

image
image
<details>

### DIFF
--- a/assets/quick-add.css
+++ b/assets/quick-add.css
@@ -177,7 +177,7 @@
   width: auto;
 }
 
-@media screen and (max-width: 990px) {
+@media screen and (max-width: 989px) {
   .quick-add-modal__content-info--bulk .quick-add__content-info__media {
     display: flex;
     margin: 0;
@@ -193,7 +193,7 @@
   }
 }
 
-@media screen and (min-width: 989px) {
+@media screen and (min-width: 990px) {
   .quick-add-modal__content-info--bulk .quick-add__info {
     flex-direction: column;
     position: sticky;
@@ -203,7 +203,7 @@
   }
 }
 
-@media screen and (max-width: 990px) {
+@media screen and (max-width: 989px) {
   .quick-add-modal__content-info--bulk {
     flex-direction: column;
   }


assets/quick-add.css
```diff
@@ -122,6 +134,11 @@
...
 }
 
+.quick-add-modal__content-info--bulk-details > a:hover {
+  color: rgb(var(--color-foreground));
+  text-decoration: underline;
+}
+
 @media screen and (min-width: 990px) {
   .quick-add-modal__content-info--bulk .quick-add__product-media,
   .quick-add-modal__content-info--bulk .quick-add__product-container,
```
ludoboludo, line 144: >Quick add bulk modal header spacing is off

This could be fixed in here. It's a tricky visual bug to replicate but it's due to a media query not using the right value in the file. 

We have: 
```css
@media screen and (min-width: 989px) {
```
and 
```css
@media screen and (max-width: 990px) {
```

We should be using those values the other way around like in the rest of the CSS files. `990` is always used with `min-width` and `989` with `max-width`. 

[Recording](https://screenshot.click/28-17-cqjzi-3qqkb.mp4).
We actually have another instance of this in the file as well that mess with the [padding applied](https://screenshot.click/28-23-suppx-vp3wk.png). 

You can search for `990px` and make sure none is tied to `max-width:` and same for `989` not being tied with `min-width:`

assets/quick-add.css
```diff
@@ -35,6 +35,18 @@
...
+  padding-bottom: 1.5rem;
+}
+
+.quick-add-modal .link.product__view-details {
+  text-decoration: none;
+}
+
+.quick-add-modal .link.product__view-details .link__text {
+  text-decoration: underline;
+}
```
ludoboludo, line 48: This seem a bit hacky to me 🤔 I think we can keep the original approach and add a margin-left to the svg arrow icon just like we do for the featured-product section: 
<details>
<summary>screenshot</summary>


</details>

assets/quick-add.css
```diff
@@ -122,6 +134,11 @@
   padding-left: 1rem;
 }
 
+.quick-add-modal__content-info--bulk-details > a:hover {
+  color: rgb(var(--color-foreground));
```
ludoboludo, line 138: I don't think we need to re declare the color here since it should already set to use the current color 🤔 
Maybe we can also add `text-underline-offset: 0.3rem;` though to match other places with underlined links 👍

snippets/card-product.liquid
```diff
@@ -479,7 +479,7 @@
                           {% endif %}
                           class="link product__view-details animate-arrow small-hide medium-hide"
                         >
-                          {{ 'products.product.view_full_details' | t }}
+                          <span class="link__text">{{ 'products.product.view_full_details' | t }}</span>
```
ludoboludo, line 482: Here I don't think we need the extra span element and class. We can keep the text as it was originally and use some whitespace control within the brackets: 
```
{{- 'products.product.view_full_details' | t -}}
```

snippets/card-product.liquid
```diff
@@ -495,8 +495,17 @@
...
+                          href="{{ card_product.url }}"
+                          id="StandardCardNoMediaLink-{{ section_id }}-{{ card_product.id }}"
+                          class="full-unstyled-link"
+                          aria-labelledby="StandardCardNoMediaLink-{{ section_id }}-{{ card_product.id }} NoMediaStandardBadge-{{ section_id }}-{{ card_product.id }}"
+                        >
+                          <h3 class="h2">
+                            {{ card_product.title | escape }}
+                          </h3>
+                          {% render 'price', product: card_product, price_class: '', show_compare_at_price: true %}
+                        </a>
```
ludoboludo, line 508: I think here we could move the link to wrap the `{{ card_product.title | escape }}` only rather than the h3 element as well as the price.

And the ID could be simpler? Or why does it mention `NoMedia` ?

assets/quick-add.css
```diff
@@ -35,6 +35,18 @@
   transform: none;
 }
 
+.quick-add-modal .quick-order-list__container {
+  padding-bottom: 1.5rem;
+}
+
+.quick-add-modal .link.product__view-details {
```
sofiamatulis, line 42: This will change the regular quick add modal too. I think we only want this only for the bulk modal?

assets/quick-add.css
```diff
@@ -35,6 +35,18 @@
   transform: none;
```
sofiamatulis, line 35: Im still noticing the extra spacing at the bottom

https://screenshot.click/29-05-atpsx-c5big.mp4

assets/quick-order-list.css
```diff
@@ -450,6 +450,10 @@ quick-order-list-remove-button:hover .icon-remove {
   }
 }
 
+.quick-order-list__button-text {
+  text-align: center;
```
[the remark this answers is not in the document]
dan-menard: For QoL it doesn't matter, the button always wraps to a new line before the text gets squished enough for the bug to appear. But I figure it's better to keep the CSS consistent for both in case we change the wrapping point someday. What do you think?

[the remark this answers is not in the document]
sofiamatulis: That is fair :) I am aligned to include it

assets/quick-add.css
ludoboludo: I think ideally the scrolling happening is only for the `.quick-order-list-container`. But that means quite a bit of rework of the CSS 🤔 
Cause the issue is the padding applied on `.quick-add-modal__content-info` I believe. 
You could keep the padding value applied for the left and right on `.quick-add-modal__content-info` and apply a top value on `.quick-add-modal__content-info--bulk-details` and then it might work as expected 👍

snippets/card-product.liquid
```diff
@@ -498,8 +498,16 @@
                     </div>
                     <div>
                       <div class="quick-add-modal__content-info--bulk-details small-hide medium-hide">
-                        <h3 class="h2">{{ card_product.title | escape }}</h3>
-                        {% render 'price', product: card_product, price_class: '', show_compare_at_price: true %}
+                        <a
+                          href="{{ card_product.url }}"
+                          id="ProductTitleLink-{{ section_id }}-{{ card_product.id }}"
```
ludoboludo, line 503: I wonder if this ID is needed for anything 🤔 I don't think so.

snippets/card-product.liquid
ludoboludo: On the standard quick add, only the product title is linked. Not the price part. So I think ill lean towards title only 👍

melissaperreault: Please just allow the product title to link to the product. It may feel natural on a product card due to the nature of that target space. Just the title is intuitive since it also gets the underline. This is matching how we do it in Cart. Thanks!
30-16-sllmr-x597f

snippets/card-product.liquid
dan-menard: That's a good point, if we're not using it I might as well not include it. I'll remove.

snippets/card-product.liquid
```diff
@@ -482,7 +482,7 @@
                           {% endif %}
                           class="link product__view-details animate-arrow small-hide medium-hide"
                         >
-                          {{ 'products.product.view_full_details' | t }}
+                          {{ 'products.product.view_full_details' | t -}}
                           {% render 'icon-arrow' %}
```
ludoboludo, line 486: Now I wonder if this is redundant 🤔 Since we can already click on the product title to get us there. 
What do you think ?

snippets/card-product.liquid
```diff
@@ -498,7 +498,11 @@
                     </div>
                     <div>
                       <div class="quick-add-modal__content-info--bulk-details small-hide medium-hide">
-                        <h3 class="h2">{{ card_product.title | escape }}</h3>
+                        <a href="{{ card_product.url }}" class="full-unstyled-link">
+                          <h3 class="h2">
+                            {{ card_product.title | escape }}
+                          </h3>
+                        </a>
```
ludoboludo, line 505: Let's also add this for the product title on mobile. Right above this on line `490`

ludoboludo (commented):
A couple comments, otherwise looking good 👍

assets/quick-add.css
sofiamatulis: @melissaperreault I thought it was the extra space when you scroll down?  🤔 

https://screenshot.click/29-05-atpsx-c5big.mp4

scottgmeadows: Seems like @dan-menard fixed it. Issue was [this extra bit of scrolling](https://screenshot.click/03-06-jbj96-wwg23.mp4) at the bottom of the modal. 

👍

scottgmeadows (approved):

assets/quick-add.css
```diff
@@ -78,6 +82,10 @@
   max-width: 100%;
```
sofiamatulis, line 82: Nitpick but spacing feels a bit wide here 🤔 Want to ensure it's intentional 
https://screenshot.click/03-20-ihglj-stnst.png

dan-menard: Yeah it feels like too much to me too, but I picked the value we were using already for Featured Products:
image

So it was intentional, but lemme know if you think we should run it by UX?

sofiamatulis: That is fair 👍 I was just comparing it to what we had before: https://screenshot.click/03-17-lla7m-blvm9.png

ludoboludo (approved):
🚢

PR #3495, opened 2024-05-30
🐛 Avoid invalid variant state after rapid option value switching

lhoffbeck:
### PR Summary: 

Fixes a bug where option value selection is incorrect if the buyer changes options while the new option state is loading

During an internal team bughunt on 2k variants and combined listings, we found 2 issues when switching rapidly between option values:
1) The option value selection would sometimes revert to previously selected option values ([demo](https://screenshot.click/22-35-4s9z1-v7i6g.mp4))
2) The variant added to the cart would sometimes be incorrect ([demo](https://screenshot.click/30-53-5vo00-7jtbo.mp4)) 

In both cases, the reason for this is for high-variant products to be performant, we _need_ to not serialize all the product's variants when rendering the page, and instead only include the variants associated with the currently displayed option values. See [the `Supporting high-high variant products` guide](https://shopify.dev/docs/storefronts/themes/product-merchandising/variants/support-high-variant-products) for more information.

On option value change, we were calling the section rendering api with the variant ID associated with the clicked option value. The issue with this, is that the variant associated with option values is inherently tied to the option value selection.

For example, say we have a combined listing product with color and size:

```
color: [red] green blue
size:   [small] medium large
```

If the buyer selected `medium` here, we would call the section rendering api with the variant ID associated with `red/medium`. However, we only have the variants a "render snapshot" at a time, so the variant IDs are not updated until the option value picker is rerendered. Going back to our example, the impact of this is that if the buyer selects `medium` and then selects `blue` _before_ the first section rendering api call resolves, the final network call would be for the variant `blue/small` and NOT `blue/medium`. When the response comes back, the option value selection state may be mismatched with the values the buyer most recently selected.

Instead of calling the section rendering API with a variant ID, we instead call it with the selected option value IDs. This enables us to always return the response based on the latest buyer selection. Once the API response returns, we know which variant (or no variant) is associated with the option value selection, and can then update the page accordingly (forms, urls, etc).

There should be no impact

### Testing steps/scenarios
- [ ] With your network panel open, change option values quickly and then add to your cart. It is easier to trigger cancelled API requests if you throttle to slow or fast 3g.
- [ ] Also verify that this change works as expected in the quick add modal

### Demo links

- store url: https://combined-listings-test.myshopify.com/products/chinos-blue

> This store has a few themes, please test with theme `pr_3495`

### Checklist

### DIFF
--- a/assets/constants.js
+++ b/assets/constants.js
@@ -3,7 +3,7 @@ const ON_CHANGE_DEBOUNCE_TIMER = 300;
 const PUB_SUB_EVENTS = {
   cartUpdate: 'cart-update',
   quantityUpdate: 'quantity-update',
-  variantChangeStart: 'variant-change-start',
+  optionValueSelectionChange: 'option-value-selection-change',
   variantChange: 'variant-change',
   cartError: 'cart-error',
   sectionRefreshed: 'section-refreshed',
--- a/assets/product-info.js
+++ b/assets/product-info.js
@@ -8,6 +8,7 @@ if (!customElements.get('product-info')) {
       cartUpdateUnsubscriber = undefined;
       swapProductUtility = undefined;
       abortController = undefined;
+      inflightRequestProductUrl = null;
 
       constructor() {
         super();
@@ -19,7 +20,7 @@ if (!customElements.get('product-info')) {
         this.initializeProductSwapUtility();
 
         this.onVariantChangeUnsubscriber = subscribe(
-          PUB_SUB_EVENTS.variantChangeStart,
+          PUB_SUB_EVENTS.optionValueSelectionChange,
           this.handleOptionValueChange.bind(this)
         );
 
@@ -68,37 +69,30 @@ if (!customElements.get('product-info')) {
         });
       }
 
-      handleOptionValueChange({ data: { event, target, variant } }) {
+      handleOptionValueChange({ data: { event, target, selectedOptionValues } }) {
         if (!this.contains(event.target)) return;
 
-        const targetUrl = target.dataset.productUrl || this.dataset.url;
+        const baseUrl = target.dataset.productUrl || this.inflightRequestProductUrl || this.dataset.url;
+        this.inflightRequestProductUrl = baseUrl;
 
         const productForm = this.productForm;
         productForm?.toggleSubmitButton(true);
         productForm?.handleErrorMessage();
 
-        let callback = () => {};
-        let productUrl = this.getProductInfoUrl(targetUrl, variant?.id);
-        if (this.dataset.url !== targetUrl) {
-          callback = this.handleSwapProduct(targetUrl, variant);
-          productUrl = this.getProductInfoUrl(targetUrl, variant?.id, true);
-        } else if (!variant) {
-          this.setUnavailable();
-          callback = (html) => {
-            this.pickupAvailability?.update(variant);
-            this.updateOptionValues(html);
-          };
-        } else {
-          this.updateVariantInputs(variant.id);
-          callback = this.handleUpdateProductInfo(targetUrl, variant);
-        }
-
-        this.renderProductInfo(productUrl, target.id, callback);
+        const shouldSwapProduct = this.dataset.url !== baseUrl;
+        this.renderProductInfo({
+          productUrl: this.getProductInfoUrl(baseUrl, selectedOptionValues, shouldSwapProduct),
+          targetId: target.id,
+          callback: shouldSwapProduct ? this.handleSwapProduct(baseUrl) : this.handleUpdateProductInfo(baseUrl),
+        });
       }
 
-      handleSwapProduct(baseUrl, variant) {
+      handleSwapProduct(baseUrl) {
         return (html) => {
           this.productModal?.remove();
+
+          // Grab the selected variant from the new product info
+          const variant = this.getSelectedVariant(html.querySelector(`product-info[data-section=${this.sectionId}]`));
           this.updateURL(baseUrl, variant?.id);
 
           // If we are in an embedded context (quick add, featured product, etc), only swap product info.
@@ -107,18 +101,20 @@ if (!customElements.get('product-info')) {
             this.swapProductUtility.viewTransition(this, html.querySelector('product-info'));
           } else {
             document.querySelector('head title').innerHTML = html.querySelector('head title').innerHTML;
+
             this.swapProductUtility.viewTransition(document.querySelector('main'), html.querySelector('main'));
           }
         };
       }
 
-      renderProductInfo(productUrl, targetId, callback) {
+      renderProductInfo({ productUrl, targetId, callback }) {
         this.abortController?.abort();
         this.abortController = new AbortController();
 
         fetch(productUrl, { signal: this.abortController.signal })
           .then((response) => response.text())
           .then((responseText) => {
+            this.inflightRequestProductUrl = null;
             const html = new DOMParser().parseFromString(responseText, 'text/html');
             callback(html);
           })
@@ -135,18 +131,18 @@ if (!customElements.get('product-info')) {
           });
       }
 
-      getProductInfoUrl(url, variantId, shouldFetchFullPage = false) {
+      getSelectedVariant(productInfoNode) {
+        const selectedVariant = productInfoNode.querySelector('variant-selects [data-selected-variant]')?.innerHTML;
+        return !!selectedVariant ? JSON.parse(selectedVariant) : null;
+      }
+
+      getProductInfoUrl(url, optionValues, shouldFetchFullPage = false) {
         const params = [];
 
         !shouldFetchFullPage && params.push(`section_id=${this.sectionId}`);
 
-        if (variantId) {
-          params.push(`variant=${variantId}`);
-        } else {
-          const optionValues = this.variantSelectors.selectedOptionValues;
-          if (optionValues.length) {
-            params.push(`option_values=${optionValues.join(',')}`);
-          }
+        if (optionValues.length) {
+          params.push(`option_values=${optionValues.join(',')}`);
         }
 
         return `${url}?${params.join('&')}`;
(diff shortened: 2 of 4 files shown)


assets/constants.js
```diff
@@ -3,7 +3,7 @@ const ON_CHANGE_DEBOUNCE_TIMER = 300;
 const PUB_SUB_EVENTS = {
   cartUpdate: 'cart-update',
   quantityUpdate: 'quantity-update',
-  variantChangeStart: 'variant-change-start',
+  optionValueSelectionChange: 'option-value-selection-change',
```
lhoffbeck, line 6: This event was updated to be focused on option value selection instead of variant selection. Because the theme only loads the variants associated with the [currently displayed option values](https://shopify.dev/docs/storefronts/themes/product-merchandising/variants/support-high-variant-products#option-value-availability), if a buyer clicks through option value selections very quickly, we do not know which variant is being switched to until the section rendering api request resolves.

assets/product-info.js
```diff
@@ -8,6 +8,7 @@ if (!customElements.get('product-info')) {
       cartUpdateUnsubscriber = undefined;
       swapProductUtility = undefined;
       abortController = undefined;
+      inflightRequestProductUrl = null;
```
lhoffbeck, line 11: This variable is used to track which product url should be requested when a buyer is clicking between options before the section renderer request resolves.

For example, we have a combined listing Pants with parent option color and child option size.

If the buyer is on `black/small` and clicks `orange` and then `medium` before the `orange` request resolves, the final request should be for `orange/medium`.

assets/product-info.js
```diff
@@ -68,37 +69,30 @@ if (!customElements.get('product-info')) {
...
-          callback = this.handleUpdateProductInfo(targetUrl, variant);
-        }
-
-        this.renderProductInfo(productUrl, target.id, callback);
+        const shouldSwapProduct = this.dataset.url !== baseUrl;
+        this.renderProductInfo({
+          productUrl: this.getProductInfoUrl(baseUrl, selectedOptionValues, shouldSwapProduct),
+          targetId: target.id,
+          callback: shouldSwapProduct ? this.handleSwapProduct(baseUrl) : this.handleUpdateProductInfo(baseUrl),
+        });
```
lhoffbeck, line 87: This logic was refactored because we don't want to / can't access the `variant` entity until the section rendering request resolves.

Primary changes are:
- [Unavailable state](https://github.com/Shopify/dawn/pull/3495/files#diff-1ab5b2ed05e2f2ba4adfb4b617297e5d95e49a89d944f6e3ed33ad39d44f69ddL86-L90) is now deferred and [set in the request callback](https://github.com/Shopify/dawn/pull/3495/files#diff-1ab5b2ed05e2f2ba4adfb4b617297e5d95e49a89d944f6e3ed33ad39d44f69ddR166-R169)
- [Variant inputs](https://github.com/Shopify/dawn/pull/3495/files#diff-1ab5b2ed05e2f2ba4adfb4b617297e5d95e49a89d944f6e3ed33ad39d44f69ddL92) on the product form are now set [in the request callback](https://github.com/Shopify/dawn/pull/3495/files#diff-1ab5b2ed05e2f2ba4adfb4b617297e5d95e49a89d944f6e3ed33ad39d44f69ddR164)

assets/product-info.js
```diff
@@ -135,18 +131,18 @@ if (!customElements.get('product-info')) {
...
 
-        if (variantId) {
-          params.push(`variant=${variantId}`);
-        } else {
-          const optionValues = this.variantSelectors.selectedOptionValues;
-          if (optionValues.length) {
-            params.push(`option_values=${optionValues.join(',')}`);
-          }
+        if (optionValues.length) {
+          params.push(`option_values=${optionValues.join(',')}`);
```
lhoffbeck, line 145: This change isn't really meaningful. You can request a product and specify the selection either with a variant or with option value IDs, you get the same response either way. The reason for this change is because we don't want to rely on a variant here anymore.

assets/product-info.js
```diff
@@ -157,12 +153,20 @@ if (!customElements.get('product-info')) {
...
+      handleUpdateProductInfo(baseUrl) {
         return (html) => {
+          const variant = this.getSelectedVariant(html);
+
           this.pickupAvailability?.update(variant);
-          this.updateMedia(html, variant?.featured_media?.id);
           this.updateOptionValues(html);
           this.updateURL(baseUrl, variant?.id);
+          this.updateMedia(html, variant?.featured_media?.id);
+          this.updateVariantInputs(variant?.id);
```
lhoffbeck, line 164: Note that these 2 lines are a meaningful behavioral change when switching to an option value combination that has no associated variant!

The impact here is that:
1) We clear any previous variant's featured media, the same as we do when switching to a variant without featured media.
2) We clear the variant ID value in the product form

assets/product-info.js
```diff
@@ -203,7 +207,7 @@ if (!customElements.get('product-info')) {
           `#product-form-${this.dataset.section}, #product-form-installment-${this.dataset.section}`
         ).forEach((productForm) => {
           const input = productForm.querySelector('input[name="id"]');
-          input.value = variantId;
+          input.value = variantId ?? '';
```
lhoffbeck, line 210: This clears the input value if the selection has no associated variant

snippets/product-variant-picker.liquid
```diff
@@ -84,5 +84,7 @@
         </div>
       {%- endif -%}
     {%- endfor -%}
+
+    <script type="application/json" data-selected-variant>{{ product.selected_or_first_available_variant | json }}</script>
```
lhoffbeck, line 88: This is a convenience to enable us to easily grab the JSON for the variant that is selected.

assets/product-info.js
```diff
@@ -68,37 +69,30 @@ if (!customElements.get('product-info')) {
         });
       }
 
-      handleOptionValueChange({ data: { event, target, variant } }) {
+      handleOptionValueChange({ data: { event, target, selectedOptionValues } }) {
         if (!this.contains(event.target)) return;
 
-        const targetUrl = target.dataset.productUrl || this.dataset.url;
+        const baseUrl = target.dataset.productUrl || this.inflightRequestProductUrl || this.dataset.url;
+        this.inflightRequestProductUrl = baseUrl;
```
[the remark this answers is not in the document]
ludoboludo: I think I'm missing something. Why do we need the `inflightRequestProductUrl` ?

- Its value is defined on variant selection change. 
- We then run the fetch request based on the `baseUrl`
- When the fetch runs, it will set the `inflightRequestProductUrl` back to `null` if it has completed. 
- If you happen to change the variant selection while a fetch is running, it will go through `handleOptionValueChange` and update the `baseUrl` in case it has a product url associated with it, if not it will use the `inflightRequestProductUrl`. But why is that needed ? Maybe because I'm testing on my store without combined listing items. 

As a user I click on a variant which happens to have a `data-product-url` associated to it, so it: 
- sets the `baseUrl` and `inflightRequestProductUrl` to the same value
- runs the fetch...
- but I change variant selection while doing so, and re sets the base url to either the `data-product-url` of the option I clicked on or the previous baseUrl value via `this.inflightRequestProductUrl`. But what if that option I clicked on doesn't have a `data-product-url` (null) ? Wouldn't we be using the wrong `baseUrl` then ?

assets/product-info.js
ludoboludo: I didn't realize before but I think some of the updateMedia changes (done recently) create an issue where if I have a variant with a media attached and other variants with no media attached, it gives me nothing as the main media shown: [recording](https://screenshot.click/04-31-3j6ex-u2clb.mp4)

assets/product-info.js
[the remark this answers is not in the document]
lhoffbeck: > But what if that option I clicked on doesn't have a data-product-url (null) ? Wouldn't we be using the wrong baseUrl then

Yeah exactly, this is the case that `inflightRequestProductUrl` is meant to account for. Only parent options have product urls:

04-31-4oqnl-q7v6u

Without this change, the bug that presents is that the final request uses stale information, and can return the wrong product or the wrong option value selection. For example:
- Buyer starts with a selection of `black/m/simple`
- 👍  Buyer clicks Orange, sfapi is called with `/pants-orange?option_values=<orange_id>,<m_id>,<simple_id>`
- 👎  _Before_ the first request resolves, buyer clicks `style: fancy`. Because there is no product-url, the code used to assume this meant the option value change was within the currently active product. Also, because option values haven't been updated yet from the first request response, the variant associated with the option value is stale. So here, having `orange/m/simple` selected and clicking `fancy` fires a request for `/pants-black?variant=<black/m/fancy>` but it should request for `/pants-orange?option_values=<orange_id>,<m_id>,<fancy_id>`

With the changes, we do 2 things to make sure we request the right product:
1) Cache `inflightRequestProductUrl` after each request so we can fallback to the correct parent product selection if product-url is null
2) Call the api with option_values instead of variant

So with the same example:
- Buyer starts with a selection of `black/m/simple`
- 👍  Buyer clicks Orange, sfapi is called with `/pants-orange?option_values=<orange_id>,<m_id>,<simple_id>`
- 👍  Set `this.inflightRequestProductUrl = /pants-orange`
- _Before_ the first request resolves, buyer clicks `style: fancy`
  - 👍 This fallback `const baseUrl = target.dataset.productUrl || this.inflightRequestProductUrl || ...` returns the correct parent product url of `/pants-orange`
  - 👍 [We read the selected option values](https://github.com/Shopify/dawn/pull/3495/files#r1621561196) to get `/pants-orange?option_values=<orange_id>,<m_id>,<fancy_id>`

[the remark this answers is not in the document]
lhoffbeck: > Maybe because I'm testing on my store without combined listing items.

Yeah that's definitely it, this weird case only comes up with combined listings 👍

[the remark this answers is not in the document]
ludoboludo: I'm noticing an issue with the complementary block within the product info on `main-product.liquid`. 
When switching variants in that modal, the url it's using for the fetch request is sometimes wrong and breaks the experience: [recording](https://screenshot.click/05-55-9v5s6-diow0.mp4)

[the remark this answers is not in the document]
lhoffbeck: PR here: https://github.com/Shopify/dawn/pull/3504

dan-menard (approved):
Changes seem reasonable to me 👍 thanks for all the comments to explain the changes, they helped a bunch 😄

assets/product-info.js
ludoboludo: PR with that fix here: Shopify/dawn/pull/3505

ludoboludo (approved):
🚀

PR #3509, opened 2024-06-07
Limit the number of stylesheets on the page

krzksz:
### PR Summary: 

Limit the number of separate stylesheets added to the page to improve rendering and interaction performance.

I got a report from one of the merchants that their store is suffering from high Interaction To Next Paint values. They use the Trade theme, but from what I can see all of the flagship themes are affected, although to a smaller degree.

When I profiled the page, I noticed that the browser takes a very long time when it needs to recalculate the styles (opening a mobile menu on Trade theme demo collection page with 4x CPU throttling):
Screenshot 2024-06-07 at 09 58 34

Style recalculations happen throughout the entire page lifecycle (HTML parsing, script execution and interaction handling) so we need to make sure they stay fast.

I managed to pinpoint the issue to the fact that there are cases where we include a lot of separate stylesheets to the page. Yes, each URL will only be downloaded once, but the browser still has to parse and apply each one of them. This is where the slowdown is coming from.

The problem is especially visible in every place where a lot of product cards are used (carousels, grids, etc.), especially where the quick add to card option is enabled. Each one of those snippets will include their styles slowing the recalculations down. This is why it's more prominent in the Trade theme then Dawn.

There are three main offenders that I found during the analysis:
- The stylesheets added by the `card-product.liquid` snippet.
- The stylesheet added by the `loading-spinner.liquid` snippet.
- The stylesheet added by the `progress-bar.liquid` snippet.

For the product card snippet I added a new optional `skip_styles` option. It can be used in all of the sections where the snippet is rendered multiple times so there is no need to include stylesheets for the 2nd iteration onwards.

Because both `loading-spinner` and `progress-bar` snippets are used globally and their stylesheets are quite small I decided the best approach would be to just move them to the `base.css` file instead.

For Dawn demo collection page, this change brings the number of stylesheets from more then a 100 to around 30.
For Trade demo collection page, the number goes down from more then 1150 to around 30 as well 🤯

Here's the impact on the same menu open interaction with the 4x CPU throttling:
Screenshot 2024-06-07 at 10 18 36

That's a 98% improvement 🎉

Oh and as a bonus, there is way less spam in the devtools styles panel.

### Other considerations

It hard to keep a balance between including all styles in a single CSS file vs serving what users actually need based on which sections are used.

Still, in the future we should take into the consideration how and where certain stylesheets will be used to make sure they don't get duplicated a lot.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

There should be no visual impact on existing themes. Any customisations should still work as before.

### Testing steps/scenarios

- [ ] This branch is attached to the [preview theme on Dawn demo store](https://theme-dawn-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=129239679065).
- [ ] The theme should look and work exactly the same.
- [ ] You can open devtools on [one of the collection pages](https://theme-dawn-demo.myshopify.com/collections/bags) and execute `document.querySelectorAll('link[rel=stylesheet]').length` in the console to get the number of stylesheets.
- [ ] The number of stylesheets in the preview theme on collection page should be significantly smaller compared to the current release.

### Demo links

- [Store](https://theme-dawn-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&preview_theme_id=129239679065)
- [Editor](https://admin.shopify.com/store/theme-dawn-demo/themes/129239679065/editor)

### Checklist

### DIFF
--- a/sections/collage.liquid
+++ b/sections/collage.liquid
@@ -26,6 +26,7 @@
       </h2>
     {%- endif -%}
     <div class="collage{% if section.settings.mobile_layout == 'collage' %} collage--mobile{% endif %}">
+      {% assign skip_card_product_styles = false %}
       {%- for block in section.blocks -%}
         <div
           class="collage__item collage__item--{{ block.type }} collage__item--{{ section.settings.desktop_layout }}{% if settings.animations_reveal_on_scroll %} scroll-trigger animate--slide-in{% endif %}"
@@ -97,8 +98,10 @@
                 media_aspect_ratio: 'adapt',
                 show_secondary_image: block.settings.second_image,
                 extend_height: true,
-                placeholder_image: placeholder_image
+                placeholder_image: placeholder_image,
+                skip_styles: skip_card_product_styles
               %}
+              {% assign skip_card_product_styles = true %}
             {%- when 'collection' -%}
               {%- assign placeholder_image = 'collection-apparel-' | append: placeholder_image_index -%}
               {% render 'card-collection',
--- a/sections/featured-collection.liquid
+++ b/sections/featured-collection.liquid
@@ -94,11 +94,8 @@
         role="list"
         aria-label="{{ 'general.slider.name' | t }}"
       >
+        {% assign skip_card_product_styles = false %}
         {%- for product in section.settings.collection.products limit: section.settings.products_to_show -%}
-          {% assign skip_styles = false %}
-          {%- if forloop.index > 1 -%}
-            {%- assign skip_styles = true -%}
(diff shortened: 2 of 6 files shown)


dan-menard (approved):
That's a huge performance gain 😍 let's get this merged.

I'd also like to iterate on the approach at some point. I don't like that we require templates to opt-in to performance gains, it would be neat if we could fix this for all templates somehow, or even better, fix it in SFR somewhere so the benefits go to all themes 😄

PR #3513, opened 2024-06-11
Fixed templates not allowed, Section limit

khkesav:
To fix the templates not allowed issue.

Replace "templates": ["password"] with 
"enabled_on": {
    "templates": ["password"]
  }


skmohotasin:
I've done it already 
https://github.com/Shopify/dawn/pull/3457/files

PR #3515, opened 2024-06-11
Update updateQuantityRules function

### LINKED ISSUE #3512: [Product] When hiding/removing the quantity selector block, variant switch is broken
If you hide the quantity selector block on the main product page or feat. prod section, the variant switch won't happen and error out because it's not able to run some logic as it's assuming that it should find a quantity selector input. p

Whether it's hidden or not, the form should work and let you add to cart. 

- Dawn Version: 14
- macOS Version 13.1

## Possible solution
Early return in `updateQuantityRyles(sectionId, html){}`. 
```javascript
if (!this.quantityInput) return;
```

[Recording](https://screenshot.click/10-12-4empn-dnc72.mp4)

Roi-Arthur:
### PR Summary: 

Adding an early return to the updateQuantityRules() function to avoid breaking variant selection.

Fixes #3512 

We could probably rework this differently to avoid this function running altogether; we do however already use the same early return earlier in the same file, so it fits in with our existing code.

### Testing steps/scenarios
- Go to a product template or add a featured product section. Make sure that you pick a product that has variants
- Hide/Remove the quantity selector block
- Switch variants

### Demo links
- [Example product to test with](https://admin.shopify.com/store/os2-demo/products/6895499935766)

### Checklist

### DIFF
--- a/assets/product-info.js
+++ b/assets/product-info.js
@@ -364,6 +364,7 @@ if (!customElements.get('product-info')) {
       }
 
       updateQuantityRules(sectionId, html) {
+        if (!this.quantityInput) return;
         this.setQuantityBoundries();
 
         const quantityFormUpdated = html.getElementById(`Quantity-Form-${sectionId}`);


ludoboludo (approved):
Tested by setting up my account as B2B to see the quantity rules be applied. If I remove the quantity selector I don't get weird errors but expected ones 👍 

It doesn't let you add to cart as it requires you to add the min quantity the product is set for with the quantity rules. 
So if you have to buy a minimum of `5` units, you will not be able to add it to the cart but I wouldn't expect it differently. Just don't remove the quantity selector if you're using quantity rules 😅

ludoboludo:
Adding @sofiamatulis as she has great context on those features 🕵️‍♀️

assets/product-info.js
```diff
@@ -364,6 +364,7 @@ if (!customElements.get('product-info')) {
       }
 
       updateQuantityRules(sectionId, html) {
+        if (!this.quantityInput) return;
```
sofiamatulis, line 367: Connected it to the store and it looks good 👍 https://admin.shopify.com/store/os2-demo/themes/163387473942/editor

However I think we should do this check in even before running the `updateQuantityRules` function 🤔  

What do you think? cc @ludoboludo

ludoboludo: I think it's fine as is 🤔 We do the same approach within the `initQuantityHandlers()` method 👍 return.

And in a few other places in the theme. There is definitely a mix of approaches but I think it's good here since the method is called a couple times. Instead of adding this check twice having it part of the method seem good because it takes care of all the logic itself.

sofiamatulis (approved):

PR #3516, opened 2024-06-11
Media gallery bugfixes

lhoffbeck:
### PR Summary: 

1. Fixes bug where media flashes after switching to variant with the same featured image
2. Fixes bug where featured media is removed if switching from a variant with featured media to a variant without featured media

1. Add check that the featured media is not already the first child of the parent before prepending it
2. Abort the media update if the variant has no featured media

In both cases, reverts us to the behavior that was present in V14.0.0

### Testing steps/scenarios

Use theme [lh-media-gallery-bugfixes](https://os2-demo.myshopify.com/?_ab=0&_fd=0&_sc=1&key=426fb1f420e906418fa2448926c74e347c9832a999315511ee1be9880525803a&preview_theme_id=163385212950).

- [ ] [On this product](https://os2-demo.myshopify.com/products/louise-slide-sandal)
  - [ ] Check that switching size does not cause the main image to flash
  - [ ] Check that the slider is reset after switching variants.
  - [ ] Check that switching color updates the featured image and thumbnail
- [ ] On [this product](https://os2-demo.myshopify.com/products/combine-listing-test-orange)
  - [ ] Check that switching from small to medium does not remove the variant featured media

### Checklist

### DIFF
--- a/assets/media-gallery.js
+++ b/assets/media-gallery.js
@@ -41,11 +41,13 @@ if (!customElements.get('media-gallery')) {
         activeMedia?.classList?.add('is-active');
 
         if (prepend) {
-          activeMedia.parentElement.prepend(activeMedia);
+          activeMedia.parentElement.firstChild !== activeMedia && activeMedia.parentElement.prepend(activeMedia);
+
           if (this.elements.thumbnails) {
             const activeThumbnail = this.elements.thumbnails.querySelector(`[data-target="${mediaId}"]`);
-            activeThumbnail.parentElement.prepend(activeThumbnail);
+            activeThumbnail.parentElement.firstChild !== activeThumbnail && activeThumbnail.parentElement.prepend(activeThumbnail);
           }
+
           if (this.elements.viewer.slider) this.elements.viewer.resetPages();
         }
 
--- a/assets/product-info.js
+++ b/assets/product-info.js
@@ -253,6 +253,8 @@ if (!customElements.get('product-info')) {
       }
 
       updateMedia(html, variantFeaturedMediaId) {
+        if (!variantFeaturedMediaId) return;
+
         const mediaGallerySource = this.querySelector('media-gallery ul');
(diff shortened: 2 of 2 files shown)


assets/media-gallery.js
```diff
@@ -41,11 +41,13 @@ if (!customElements.get('media-gallery')) {
...
+          activeMedia.parentElement.firstChild !== activeMedia && activeMedia.parentElement.prepend(activeMedia);
+
           if (this.elements.thumbnails) {
             const activeThumbnail = this.elements.thumbnails.querySelector(`[data-target="${mediaId}"]`);
-            activeThumbnail.parentElement.prepend(activeThumbnail);
+            activeThumbnail.parentElement.firstChild !== activeThumbnail && activeThumbnail.parentElement.prepend(activeThumbnail);
           }
+
           if (this.elements.viewer.slider) this.elements.viewer.resetPages();
         }
```
ludoboludo, line 52: I think we should do one other edit. I don't want the theme to scroll me back up to the active media if that active media is the same one as before. 
Below there is a `setTimeout()` in which we're doing the `window.scrollTo()`. We should add a check to not run that in the scenario that we're still showing the same media. 

Here is what I mean: [recording](https://screenshot.click/12-21-p8q9w-lxqsb.mp4)

Though to be fair it was also on previous versions... Maybe something we can double check with UX but I personally find it a bit annoying 🤔

ludoboludo (approved):
Looking good 👍 

Noted a potential extra tweak but it's a behaviour that's existing in previous version so not a high priority but a nice fix 🤷

assets/media-gallery.js
lhoffbeck: Yeah that seems valuable, maybe we roll it out as a separate pr?

Roi-Arthur (commented):
The comment I Ieft is mostly me not knowing the syntax of the proposed change.
Seems good to go otherwise

PR #3526, opened 2024-06-13
Combined listings with multiple product templates

lhoffbeck:
### PR Summary: 

Adds support for combined listings where the products use different templates

If a combined listing parent or children products used different templates, the query selector for reading the selected variant from the section rendering api response was returning null. This PR enables multiple product templates by explicitly selecting for the `MainProduct` if there may be multiple `<product-info>` elements in the response, or `<product-info>` if there is only one.

### Testing steps/scenarios

Please use this product for testing: https://combined-listings-test.myshopify.com/products/multiple-product-templates

The parent product is set up with the `default product` template, black and orange child products use `product - 1`, and blue uses `product - 2`. There is a banner that indicates whether the product is the main product or a featured product.

Scenarios to test:
- Switching between `color` options on the main product loads the appropriate combined listing child product, and returns sections for the template.
- When on the `blue` product, the first product on the page is a featured product. Verify that switching between option values loads the correct combined listing child product. (*note -- static featured products still don't work for combined listings, this case _does_ work because it is set up dynamically using a metafield)
- On the collection page, open the `multiple product templates` product's quick-add modal. Verify that changing color options loads the correct combined listing child product.

### Checklist

### DIFF
--- a/assets/product-info.js
+++ b/assets/product-info.js
@@ -57,15 +57,6 @@ if (!customElements.get('product-info')) {
         this.postProcessHtmlCallbacks.push((newNode) => {
           window?.Shopify?.PaymentButton?.init();
           window?.ProductModel?.loadShopifyXR();
-          publish(PUB_SUB_EVENTS.sectionRefreshed, {
-            data: {
-              sectionId: this.sectionId,
-              resource: {
-                type: SECTION_REFRESH_RESOURCE_TYPE.product,
-                id: newNode.dataset.productId,
-              },
-            },
-          });
         });
       }
 
@@ -77,12 +68,14 @@ if (!customElements.get('product-info')) {
         const productUrl = target.dataset.productUrl || this.pendingRequestUrl || this.dataset.url;
         this.pendingRequestUrl = productUrl;
         const shouldSwapProduct = this.dataset.url !== productUrl;
-        const shouldFetchFullPage = !this.isFeaturedProduct && shouldSwapProduct;
+        const shouldFetchFullPage = this.dataset.updateUrl === 'true' && shouldSwapProduct;
 
         this.renderProductInfo({
           requestUrl: this.buildRequestUrlWithParams(productUrl, selectedOptionValues, shouldFetchFullPage),
           targetId: target.id,
-          callback: shouldSwapProduct ? this.handleSwapProduct(productUrl) : this.handleUpdateProductInfo(productUrl),
+          callback: shouldSwapProduct
+            ? this.handleSwapProduct(productUrl, shouldFetchFullPage)
+            : this.handleUpdateProductInfo(productUrl),
         });
       }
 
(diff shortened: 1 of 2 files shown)


assets/product-info.js
```diff
@@ -77,12 +68,14 @@ if (!customElements.get('product-info')) {
         const productUrl = target.dataset.productUrl || this.pendingRequestUrl || this.dataset.url;
         this.pendingRequestUrl = productUrl;
         const shouldSwapProduct = this.dataset.url !== productUrl;
-        const shouldFetchFullPage = !this.isFeaturedProduct && shouldSwapProduct;
+        const shouldFetchFullPage = this.dataset.updateUrl === 'true' && shouldSwapProduct;
```
lhoffbeck, line 71: Prior to this change, we were requesting a full PDP from both main-product and quick add modal, and just requesting the section for featured product. However, for the quick add modal we only need the main product section.

This change makes it so that we request the full page when swapping between combined listings in a main product section, and request just the product section for both featured product and quick add modal.

assets/product-info.js
```diff
@@ -77,12 +68,14 @@ if (!customElements.get('product-info')) {
...
 
         this.renderProductInfo({
           requestUrl: this.buildRequestUrlWithParams(productUrl, selectedOptionValues, shouldFetchFullPage),
           targetId: target.id,
-          callback: shouldSwapProduct ? this.handleSwapProduct(productUrl) : this.handleUpdateProductInfo(productUrl),
+          callback: shouldSwapProduct
+            ? this.handleSwapProduct(productUrl, shouldFetchFullPage)
+            : this.handleUpdateProductInfo(productUrl),
```
lhoffbeck, line 78: Explicitly pass whether we're fetching the full page to the handleSwapProduct callback

assets/product-info.js
```diff
@@ -92,33 +85,27 @@ if (!customElements.get('product-info')) {
...
-          // Grab the selected variant from the new product info
-          const variant = this.getSelectedVariant(html.querySelector(`product-info[data-section=${this.sectionId}]`));
+          const selector = updateFullPage ? "product-info[id^='MainProduct']" : 'product-info';
+          const variant = this.getSelectedVariant(html.querySelector(selector));
           this.updateURL(productUrl, variant?.id);
 
-          // If we are in an embedded context (quick add, featured product, etc), only swap product info.
-          // Otherwise, refresh the entire page content and sibling sections.
-          if (this.dataset.updateUrl === 'false') {
+          if (updateFullPage) {
```
lhoffbeck, line 96: I reversed this conditional so that we can rely on the `updateFullPage` variable instead of re-querying the dom

dan-menard:
Sometimes when I click back and forth between variants, the DOM order does something weird, and I get a view like this:
combined-listings-test myshopify com_products_blue_variant=48994097594670

assets/product-info.js
```diff
@@ -92,33 +85,27 @@ if (!customElements.get('product-info')) {
...
-
-      handleSwapProduct(productUrl) {
+      handleSwapProduct(productUrl, updateFullPage) {
         return (html) => {
           this.productModal?.remove();
 
-          // Grab the selected variant from the new product info
-          const variant = this.getSelectedVariant(html.querySelector(`product-info[data-section=${this.sectionId}]`));
+          const selector = updateFullPage ? "product-info[id^='MainProduct']" : 'product-info';
+          const variant = this.getSelectedVariant(html.querySelector(selector));
```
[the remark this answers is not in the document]
lhoffbeck: Not really, the issue is the id is [MainProduct-{{ section.id}}](https://github.com/Shopify/dawn/blob/main/sections/main-product.liquid#L2), so we'd have the same issue of not knowing the new section ID if the product template switches. `.querySelector` also lets us use the generic `product-info` selector when updateFullPage is false (line 92)

dan-menard (approved):
Alright, thanks for explaining the ordering 😅 I think this is good to go.

Issue #2715, opened 2023-06-14
Incorrect Pagination Button Classes


amoser67:
## Describe the current behavior

The "next" and "prev" buttons have their classes swapped, both in _pagination.liquid_ and _component-pagination.css_.

```
{%- if paginate.parts.size > 0 -%}
  <div class="pagination-wrapper">
    ...
        {%- if paginate.previous -%}
          <li>
            <a
              href="{{ paginate.previous.url }}{{ anchor }}"
              class="pagination__item pagination__item--next pagination__item-arrow link motion-reduce"
              aria-label="{{ 'general.pagination.previous' | t }}"
            >
              {% render 'icon-caret' %}
            </a>
          </li>
        {%- endif -%}
        
        ...
     
        {%- if paginate.next -%}
          <li>
            <a
              href="{{ paginate.next.url }}{{ anchor }}"
              class="pagination__item pagination__item--prev pagination__item-arrow link motion-reduce"
              aria-label="{{ 'general.pagination.next' | t }}"
            >
              {%- render 'icon-caret' -%}
            </a>
          </li>
        {%- endif -%}
```

You can see that pagination__item--next is on the previous button, and pagination__item--prev is on the next button.

Similarly, in the pagination CSS file, you can see that the rotation rules are swapped as well.

```
.pagination__item--next .icon {
  margin-left: -0.2rem;
  transform: rotate(90deg);
}

.pagination__item--next:hover .icon {
  transform: rotate(90deg) scale(1.07);
}

.pagination__item--prev .icon {
  margin-right: -0.2rem;
  transform: rotate(-90deg);
}

.pagination__item--prev:hover .icon {
  transform: rotate(-90deg) scale(1.07);
}
```

## Describe the expected behavior

The expected behavior is for the "...--prev" and "...--next" classes to be on the appropriate links, and for the rotation rules to correspond properly to those links.

## Version information (Dawn, browsers and operating systems)

- Dawn Version: 10.0.0
- Chrome Version 108.0.5359.124
- Ubuntu

## Possible solution

The solution is very easy, just swap the classes in pagination.liquid, and swap the rules in component-pagination.css.

kmishal:
@kimberlyoleiro @amoser67  I have created PR for this bug. Can you help me understand how it be reviewed and merged?

PR #3533, opened 2024-06-19
fix pagination button classes

AyoubAloui94:
This change resolves issue #2715


AyoubAloui94:
I have signed the CLA!

haroldao:
It's not an issue. It's meant to be "next" -> "navigate to the next item". "prev" -> "navigate to the previous item".

haroldao:
It's not an issue. "next" is meant to navigate to the next item, and "prev" is meant to navigate to the previous item.

PR #3545, opened 2024-07-03
Fixed issues when header is hidden

tyleralsbury:
Five issues are fixed here.

1. Share fails to initialize when the header is hidden because DetailsDisclosure is in the header.  Moved the details related JS files to `theme.liquid` so that there aren't dependencies on the header in other sections.
2. `SearchForm` missing causes predictive search to throw an exception.  `SearchForm` was also loaded in the header, so I moved it as well.
3. When the `drawer` cart style was selected in theme settings and the header was hidden, the drawer would render with `visibility: hidden` but the rest of its CSS was missing.  So it created a huge space above the announcement bar.  Moved the CSS to `theme.liquid`.
4. The search page predictive search was looking for `section-header`.  Added optional chaining to that.
5. The search page predictive search `alt` was not being escaped, so HTML in it broke the layout

### DIFF
--- a/layout/theme.liquid
+++ b/layout/theme.liquid
@@ -31,6 +31,10 @@
     <script src="{{ 'constants.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'details-disclosure.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'details-modal.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'search-form.js' | asset_url }}" defer="defer"></script>
+
     {%- if settings.animations_reveal_on_scroll -%}
       <script src="{{ 'animations.js' | asset_url }}" defer="defer"></script>
     {%- endif -%}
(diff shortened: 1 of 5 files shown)


dan-menard (approved):

layout/theme.liquid
```diff
@@ -31,6 +31,10 @@
     <script src="{{ 'constants.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'pubsub.js' | asset_url }}" defer="defer"></script>
     <script src="{{ 'global.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'details-disclosure.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'details-modal.js' | asset_url }}" defer="defer"></script>
+    <script src="{{ 'search-form.js' | asset_url }}" defer="defer"></script>
+
```
ludoboludo, line 37: I wonder here if we should also add: 
```liquid
{%- if settings.cart_type == "drawer" -%}
  <script src="{{ 'cart-drawer.js' | asset_url }}" defer="defer"></script>
{%- endif -%}
```
Which is in the header right now. So that if you click on `Add to cart` then the drawer is shown anyway?
If we're already including the styles assets, we might as well do the script too.

ludoboludo (commented):
Just one piece of feedback the rest looks fixed to me 👍

layout/theme.liquid
tyleralsbury: Goooood point.  I just updated it - took a bit of a change to the code in drawer (since it has an awareness of the header), but wasn't too tricky.

ludoboludo (approved):

PR #3572, opened 2024-07-24
Show default price for placeholder content

ludoboludo:
### PR Summary: 

We recently made changes to output the price in certain conditions. 
Though there was an impact that we didn't notice, the default price shown for placeholder content is gone. 
In this PR we're adding an extra check to render the price anyway if it's a placeholder context.

To make sure placeholder prices are shown. Right now it looks like this: 

Added a placeholder property to the price snippet. If it's not added, it will be `null`, and if it's there we will still render the price. 

Show the default price again

### Testing steps/scenarios

- [ ] Test adding a feat. collection section
- [ ] Test adding a feat. product section
- [ ] Test adding a collage section with a product card
- [ ] The product cards and price for feat. product section is the only places where I thought the change was needed. 🤔 If you think of any other use case make sure to comment on the PR!
- [ ] Test an unavailable variant and make sure the price is hidden still. Eg: Louise Slide Sandal -> Buttermilk - 46

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/163591553046/editor)

### Checklist

### DIFF
--- a/sections/featured-product.liquid
+++ b/sections/featured-product.liquid
@@ -142,14 +142,14 @@
                       {%- endif -%}
                     </div>
                   {%- endif -%}
-                  {%- if product != blank -%}
+                  {%- unless placeholder -%}
                     <div {{ block.shopify_attributes }}>
                       {%- form 'product', product -%}
                         <input type="hidden" name="id" value="{{ product.selected_or_first_available_variant.id }}">
                         {{ form | payment_terms }}
                       {%- endform -%}
                     </div>
-                  {%- endif -%}
+                  {%- endunless -%}
                 {%- when 'sku' -%}
                   <p
                     class="product__sku{% if block.settings.text_style == 'uppercase' %} caption-with-letter-spacing{% elsif block.settings.text_style == 'subtitle' %} subtitle{% endif %}{% if product.selected_or_first_available_variant.sku.size == 0 %} visibility-hidden{% endif %}"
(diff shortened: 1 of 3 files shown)


snippets/price.liquid
```diff
@@ -37,94 +38,97 @@
...
-              <s class="price-item price-item--regular variant-item__old-price">
-                {% if settings.currency_code_enabled %}
-                  {{ compare_at_price | money_with_currency }}
-                {% else %}
-                  {{ compare_at_price | money }}
-                {% endif %}
-              </s>
-            </span>
-          {%- endunless -%}
+{%- unless target == null and placeholder == null -%}
```
ludoboludo, line 41: One thing I actually noticed which is minimal. But curious to know how we want to go about it. 

There is an issue where if you add a feat. collection to the product page, the placeholder product cards get a price that matches the currently selected variant. 
So with the `placeholder` prop I added, we can actually check if it is true and assign `null` to the `target` to keep our default `19.99` price as expected. 

So I might keep the `placeholder` approach for that reason 🤔 

The other solution could be to keep your approach with `product != null` but change it so it's not referring to the actual `product` in the snippet props. Where right now it says in the comment at the top of the file:
```
- product: {Object} Product Liquid object (optional)
```
Cause when on a product template, it will always return the product object of the template as a fallback. 
It should maybe be named instead `product_resource` so we can actually know whether that prop was passed when rendering the snippet or not 🤔 
And the file also points to `product.` a few times instead of `target.`

[Recording](https://screenshot.click/25-45-llp2s-hs7y0.mp4)

Let me know what you think. cc: @Roi-Arthur @emma-boardman @lhoffbeck

sections/featured-product.liquid
```diff
@@ -114,6 +117,7 @@
                   <div id="price-{{ section.id }}" role="status" {{ block.shopify_attributes }}>
                     {%- render 'price',
                       product: product,
+                      placeholder: placeholder,
```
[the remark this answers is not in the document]
ludoboludo: Done! Updated a couple instances to use the placeholder check.
We shall if that's the approach we take based on Arthur's feedback 👍

snippets/price.liquid
Roi-Arthur: Yep I can see the theoretical improvement as it preserves the same example price 👍🏼

On the other hand, I think the important piece is to show _any_ price (rather than none), and ultimately the example shouldn't stick around for too long. I also somewhat doubt that Merchants would go around changing the product to see how it affects the example.
And lastly, if we're going back to the initial point that this messes up the AI generated themes, I think that this is mostly relevant to the front page.

Overall I think I would go for the least amount of code/complexity but happy to go either way.

lhoffbeck (approved):

PR #3575, opened 2024-08-01
Prefetch Relevant Pages for Faster Navigation

joshistoast:
### PR Summary: 

Prefetch link urls for faster page navigation.
For desktop, hovered links will be prefetched. For mobile, links within view will be prefetched.

> Why not just use any other pre-made js page prefetching solutions?

Because they consistently break the cart. Also, who doesn't want a native implementation?

Faster page loads = better customer experience = better sales

Created 3 functions: `addPrefetchLink`, `initPagePrefetching` and `getPrefetchMethod`.

1. `addPrefetchLink` takes arguments `href` and `priority`, and appends a `<link>` element to the document head. If an href is already present in document head `<link>` then skip.
2. `initPagePrefetching` takes a single `method` argument which can be either `mouseover` or `intersection`. The function is responsible for indexing all relevant links and- dependent on the method argument, runs `addPrefetchLink` on hover or intersection.
3. `getprefetchMethod` will return `mouseover` or `intersection` dependent on browser size.

`initPagePrefetching` will run on dom content loaded using the method provided by `getPrefetchMethod`. When browser is resized, the init is run again w/ debounce so the correct prefetch method is always correct.

### Decision log

|---|---|---|---|---|
| 1 | Add support for speculation api | rely entirely on link prefetches | It's a better and newer approach | Not very widespread browser support |
| 2 | Move speculation api usage to prefetching function | | Better consolidation and better for mobile prefetching | None apparent |

Faster page load, less content flash.

### Testing steps/scenarios

- [x] Within Application -> Speculative Loads -> Speculations, links successfully prefetch to the best of the browser's ability.
- [x] Desktop hovering internal links adds to head and present in network requests.
- [x] Mobile link visibility in intersection adds to head and present in network requests.
- [x] Resizing browser changes link prefetch method. Try hovering/scrolling between screen sizes and analyze network requests for behaviors mentioned above.

### Demo links

- [Store](https://dawn-prefetching.myshopify.com/)
	Password: `prefetch`

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -276,9 +276,9 @@ function initPagePrefetching(deviceMethod) {
   const prefetchLinkRegex = new RegExp(`^(\\/|(\\/(?:${prefetchRegions.join('|')})(\\/.*)?))$`);
   const prefetchLinks = document.querySelectorAll('a[href]')
 
-  const handleMouseOver = (event) => {
+  const handleMouseOver = (event) =>
     addPrefetchLink(event.currentTarget.href, 'high', prefetchMethod)
-  }
+
   const observer = new IntersectionObserver((entries, observer) => {
     entries.forEach((entry) => {
       if (entry.isIntersecting) {
@@ -300,11 +300,27 @@ function initPagePrefetching(deviceMethod) {
   })
 }
 
-// If mobile, prefetch on intersection, otherwise prefetch on mouseover
+
+/**
+ * Returns the preferred method for prefetching based on the window size.
+ * @returns {'intersection' | 'mouseover'} The preferred method for prefetching. Possible values are 'intersection' or 'mouseover'.
+ */
 const getPrefetchMethod = () => {
   return window.matchMedia('(max-width: 768px)').matches
     ? 'intersection'
     : 'mouseover'
+}
+
+/**
+ * Determine if devices should even consider prefetching
+ * Low power mode, data saver, etc.
+ * @returns {boolean} Whether or not prefetching should be used.
+ */
+const shouldUsePrefetching = () => {
+  // Data saver mode
+  if (navigator.connection?.saveData) return false
+
+  // TODO: Low power mode
 }
 
 document.addEventListener('DOMContentLoaded', () => initPagePrefetching(getPrefetchMethod()))
--- a/assets/global.js
+++ b/assets/global.js
@@ -264,6 +264,7 @@ function addPrefetchLink(href, priority = 'auto', method = 'link') {
  * @param {'mouseover' | 'intersection'} method - The method to use for prefetching pages.
  */
 function initPagePrefetching(deviceMethod) {
+  if (!shouldUsePrefetching()) return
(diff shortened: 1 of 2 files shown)


joshistoast:
> This is a great addition to Dawn and i think it can benefit many merchants. I would love to check the demo but the url's are not working.
> 
> Anyway, there is a native JS API for this, that serves the same purpose, called [Speculation Rules ](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API). [More here](https://developer.chrome.com/docs/web-platform/prerender-pages) We can argue that it does not have full browser compatibility (yet) but it is functional on any modern browser, yet.

Updated the PR description with a development store link :)

I'll definitely need to look into the `speculation api`, maybe could update this pr with both methods dependent on browser :eyes:

krzksz:
Hi @joshistoast, thank you for the contribution! Unfortunately, I'm aligned with that @bakura10 mentioned above.

I'm exploring a platform-wide solution that would output Speculation Rules as a part of `content_for_header`. This will allow us to potentially cover all of the stores, no matter if they're based on the new version of Dawn or not.

We'll also be able to enable/adjust/disable the configuration and logic any time, without having to deal with all of the logic that's already spread across merchants theme code.

I don't have an ETA yet, but it's on top of my priority list right now.

joshistoast:
I like @krzksz platform-wide solution idea, however depending on if that eta is a few weeks/months or even years, I see no reason to delay having an initial implementation until then. Much like how we initially used structured data and removed lots of it in favor of the new `structured_data` filter.

Also, shouldn't speculation rules go at the end of the body instead of the head?

To address @bakura10's suggestions:
1. I've tried several prefetching libraries already including instantpage, all of them reliably cause cart issues within dawn specifically... without configuring what routes to disallow/allow. Ushering merchants to a non-native solution and having them configure the js to get it working sounds less intuitive.
2. Reading the instantpage source, my approach is just as aggressive, but pre-configured for only relevant resources and natively implemented.
3. Power saving can be accommodated for with a commit, slow connections sounds like an interesting case- since prefetching would probably help them the most.

joshistoast:
If the concern is in speculation api, I can simply remove it in favor of the tried and tested `<link>` method.

joshistoast:
@krzksz

joshistoast:
Disappointing

krzksz:
@joshistoast We actually ended up implementing global prefetching across the entire platform using Speculation Rules.

You can learn more about them in the talk that I gave: https://www.youtube.com/watch?v=8k2cGNjwJ3c
We also have a blog post coming as well.

joshistoast:
> @joshistoast We actually ended up implementing global prefetching across the entire platform using Speculation Rules.
> 
> 
> 
> You can learn more about them in the talk that I gave: https://www.youtube.com/watch?v=8k2cGNjwJ3c
> 
> We also have a blog post coming as well.

Ahh, that's good context to have. Thanks.

PR #3595, opened 2024-08-27
Release 15.1.0

tyleralsbury:
Version `15.1.0` of Dawn

Dawn 15.1.0 changes the way SVGs are rendered in Dawn
### Fixes and improvements
- Moves SVGs from Liquid snippets to `.svg` files in the `/assets` folder
- Uses new `inline_asset_content` Liquid filter to render SVGs

### DIFF
--- a/assets/section-main-product.css
+++ b/assets/section-main-product.css
@@ -185,6 +185,10 @@ product-info {
   max-width: 100%;
 }
 
+.product-form__input .svg-wrapper {
+  right: 1rem;
+}
+
 .product-form__submit {
   margin-bottom: 1rem;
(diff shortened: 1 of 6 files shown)


eugenekasimov (approved):

Roi-Arthur (changes requested):
I think our default color schemes are no bueno?
1. By default we have a sticky header using a transparent background (as the background gradient in Color scheme is not set!
  

2. Can debate whether Color schemes 5 is okay, but schemes 3 & 4 are bad.
[](https://screenshot.click/09-07-x3yz4-plpea.png)

PR #3629, opened 2024-09-26
Adjust CSS for new Portable Wallet

### LINKED ISSUE #3613: Adopt new Portable Wallets styling
We are currently using the fallback to payment sheets in Dawn, but need to migrate to use the new Variables.  See [portable wallet docs](https://shopify.dev/docs/storefronts/themes/pricing-payments/accelerated-checkout/upgrade-accelerated-checkout) for details.

Design-wise we are looking to match the current Dawn look and feel as closely as possible while using the new variables.

eugenekasimov:
### PR Summary: 

There has been changes in a way the dynamic pay buttons (Pay now, pay with PayPal etc) are rendered. Currently, we have our own selectors to style those buttons, but since the whole HTML structure has been changed we need to make adjustments. 

So far, everything works fine just because there is backward compatibility. But there are plans to remove it and then the current styling won't work. 

Here is a [doc](https://shopify.dev/docs/storefronts/themes/pricing-payments/accelerated-checkout/upgrade-accelerated-checkout) that describes CSS properties we need to use to properly style shopify pay buttons. 

There are some know issues:
- The border is not applied to dynamic buttons.
- The focus may look a bit weird on some buttons. 

Fixes #3613.

I remove old selectors we used to apply styles (shadow and border radius) to shopify pay buttons.

I added new selectors with some properties we previously used and linked them to our variables. 
You can noticed that I didn't add all available properties to selectors, we still get them with default values mentioned in this [doc](https://shopify.dev/docs/storefronts/themes/pricing-payments/accelerated-checkout/upgrade-accelerated-checkout). 

### Other considerations
It's possible I missed something and removed not all references to previous selectors, but I think I remove the most obvious ones and even if something was left it should impact the style of the buttons. 

### Demo links

- [Store](url)
- [Editor](url)

### DIFF
--- a/assets/base.css
+++ b/assets/base.css
@@ -1188,10 +1188,7 @@ deferred-media {
 .button,
 .shopify-challenge__button,
 .customer button,
-button.shopify-payment-button__button--unbranded,
-.shopify-payment-button [role='button'],
-.cart__dynamic-checkout-buttons [role='button'],
-.cart__dynamic-checkout-buttons iframe {
+button.shopify-payment-button__button--unbranded {
   --shadow-horizontal-offset: var(--buttons-shadow-horizontal-offset);
   --shadow-vertical-offset: var(--buttons-shadow-vertical-offset);
   --shadow-blur-radius: var(--buttons-shadow-blur-radius);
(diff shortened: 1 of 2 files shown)


melissaperreault:
@eugenekasimov Reading the doc, are we allowed to change the radius beyond 4px of the wallet buttons?

eugenekasimov:
> @eugenekasimov Reading the doc, are we allowed to change the radius beyond 4px of the wallet buttons? 

We are. That 4px is a default value for the radius on cart page. If we didn't apply our own values that we actually do we would have 4px for default.

melissaperreault (approved):
🚢

PR #3683, opened 2024-12-10
Dawn settings update 

katycobb:
### PR Summary: 

Our new dense UI makes the online store editor look and feel more like a pro tool, but the settings labels weren't built for this new interface. This is resulting in some too-long text which wraps and makes Dawn's settings look more crowded and tougher to scan than they should be.

This is also an opportunity to bring our settings content up to date based on our newest Polaris content uplift guidelines.

Fixes #0.

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Editor](https://admin.shopify.com/store/os2-demo/themes/174516633622/editor)

### Checklist

### DIFF
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -447,6 +447,9 @@
         },
         "change_slides_speed": {
           "label": "Change every"
+        },
+        "heading_utilities": {
+         "content": "Utilities"
         },           
         "show_social": {
           "label": "Social media icons",
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -2447,7 +2447,7 @@
         },
         "show_rating": {
           "label": "Product rating",
-          "info": "An app is required for product ratings. [Learn more](https://help.shopify.com/manual/online-store/themes/theme-structure/page-types#search-results-section-settings)"
+          "info": "An app is required for product ratings. [Learn more](https://help.shopify.com/en/manual/online-store/themes/theme-structure/page-types/search-page)"
         },
         "header__1": {
           "content": "Product card"
@@ -2666,7 +2666,7 @@
           "options__9": {
             "label": "Bottom Right"
           },
-          "label": "Content position"
+          "label": "Position"
         },
         "color_scheme": {
           "info": "Visible when container displayed."
@@ -2686,6 +2686,9 @@
             "label": "Large"
           }
         },
+        "content_header": {
+          "content": "Content"
+        },        
         "desktop_content_alignment": {
           "options__1": {
             "label": "Left"
@@ -2696,7 +2699,7 @@
           "options__3": {
             "label": "Right"
           },
-          "label": "Content alignment"
+          "label": "Alignment"
         },
         "header": {
           "content": "Mobile layout"
@@ -2711,7 +2714,7 @@
           "options__3": {
             "label": "Right"
           },
-          "label": "Content alignment"
+          "label": "Alignment"
         },
         "show_text_below": {
           "label": "Stack text below image"
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -694,11 +694,14 @@
         "layout_header": {
         "content": "Layout"
         },
+        "text_header": {
+        "content": "Text"
+        },        
         "blog": {
           "label": "Blog"
         },
         "post_limit": {
-          "label": "Posts"
+          "label": "Post count"
         }, 
         "columns_desktop": {
           "label": "Columns"
@@ -751,12 +754,15 @@
         },
         "header_collection": {
           "content": "Collection layout"
-        },        
+        },
+        "header_mobile": {
+        "content": "Mobile layout"
+        },                 
         "collection": {
           "label": "Collection"
         },
         "products_to_show": {
-          "label": "Products"
+          "label": "Product count"
         },
         "columns_desktop": {
           "label": "Columns"
@@ -812,7 +818,7 @@
           "label": "Quick add"
         },
         "columns_mobile": {
-          "label": "Columns (mobile)",
+          "label": "Columns",
           "options__1": {
             "label": "1"
           },
@@ -821,7 +827,7 @@
           }
         },
         "swipe_on_mobile": {
-          "label": "Carousel (mobile)"
+          "label": "Carousel"
         }
       },
       "presets": {
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -2875,8 +2875,7 @@
           "options__3": {
             "label": "Right"
           },
-          "label": "Desktop content position",
-          "info": "Position is automatically optimized for mobile."
+          "label": "Content position"
         },
         "content_alignment": {
           "options__1": {
@@ -2891,7 +2890,7 @@
           "label": "Content alignment"
         },
         "full_width": {
-          "label": "Make section full width"
+          "label": "Full width"
         }
       },
       "blocks": {
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -1895,7 +1895,7 @@
           "default": "Collections"
         },
         "sort": {
-          "label": "Sort collections by:",
+          "label": "Sort collections",
           "options__1": {
             "label": "Alphabetically, A-Z"
           },
@@ -1925,22 +1925,21 @@
           },
           "options__3": {
             "label": "Square"
-          },
-          "info": "Add images by editing your collections. [Learn more](https://help.shopify.com/manual/products/collections)"
+          }
         },
         "columns_desktop": {
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
         "header_mobile": {
           "content": "Mobile Layout"
         },
         "columns_mobile": {
-          "label": "Number of columns on mobile",
+          "label": "Mobile columns",
           "options__1": {
-            "label": "1 column"
+            "label": "1"
           },
           "options__2": {
-            "label": "2 columns"
+            "label": "2"
           }
         }
       }
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -2542,7 +2542,7 @@
           }
         },
         "columns_desktop": {
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
         "column_alignment": {
           "label": "Column alignment",
@@ -2573,16 +2573,16 @@
           "content": "Mobile Layout"
         },
         "columns_mobile": {
-          "label": "Number of columns on mobile",
+          "label": "Columns",
           "options__1": {
-            "label": "1 column"
+            "label": "1"
           },
           "options__2": {
-            "label": "2 columns"
+            "label": "2"
           }
         },
         "swipe_on_mobile": {
-          "label": "Enable swipe on mobile"
+          "label": "Carousel"
         }
       },
       "blocks": {
--- a/locales/en.default.schema.json
+++ b/locales/en.default.schema.json
@@ -2453,8 +2453,17 @@
       "name": "Search results",
       "settings": {
         "columns_desktop": {
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
+        "columns_mobile": {
+          "label": "Mobile columns",
+          "options__1": {
+            "label": "1 column"
+          },
+          "options__2": {
+            "label": "2 columns"
+          }
+        },        
         "image_ratio": {
           "label": "Image ratio",
           "options__1": {
@@ -2468,10 +2477,10 @@
           }
         },
         "show_secondary_image": {
-          "label": "Show second image on hover"
+          "label": "Second image on hover"
         },
         "show_vendor": {
-          "label": "Show vendor"
+          "label": "Vendor"
         },
         "show_rating": {
           "label": "Product rating",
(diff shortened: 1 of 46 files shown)


locales/en.default.schema.json
```diff
@@ -1385,8 +1390,7 @@
           "options__2": {
             "label": "Image second"
           },
-          "label": "Desktop image placement",
-          "info": "Image first is the default mobile layout."
```
Roi-Arthur, line 1389: This doesn't impact mobile at all; might still be worth mentioning it so that Merchants know to expect it?

locales/en.default.schema.json
```diff
@@ -1686,19 +1693,18 @@
               },
               "options__4": {
                 "label": "Large"
-              },
-              "info": "For best results, use an image with a 16:9 aspect ratio. [Learn more](https://help.shopify.com/manual/shopify-admin/productivity-tools/image-editor#understanding-image-aspect-ratio)"
```
Roi-Arthur, line 1690: Same comment as further up; could we maybe still mention needing to consider aspect ratio. "_Learn about aspect ratio to achieve best results_" or something?

locales/en.default.schema.json
```diff
@@ -2091,20 +2072,13 @@
             "text": {
               "label": "Text",
               "default": "Share"
-            },
-            "featured_image_info": {
-              "content": "If you include a link in social media posts, the page’s featured image will be shown as the preview image. [Learn more](https://help.shopify.com/manual/online-store/images/showing-social-media-thumbnail-images)."
-            },
-            "title_info": {
-              "content": "A store title and description are included with the preview image. [Learn more](https://help.shopify.com/manual/promoting-marketing/seo/adding-keywords#set-a-title-and-description-for-your-online-store)."
```
Roi-Arthur, line 2099: This seems like useful information to know and it's relevant where it appears?
image

locales/en.default.schema.json
```diff
@@ -2328,56 +2302,61 @@
...
+              "label": "Heading",
               "default": "Heading"
             }
           }
         }
       },
       "settings": {
         "header": {
-          "content": "Media",
-          "info": "Learn more about [media types.](https://help.shopify.com/manual/products/product-media)"
```
Roi-Arthur, line 2374: Any reason we're removing this? Seems relevant where it appears 
image

locales/en.default.schema.json
```diff
@@ -2469,8 +2445,17 @@
...
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
+        "columns_mobile": {
+          "label": "Mobile columns",
+          "options__1": {
+            "label": "1 column"
+          },
+          "options__2": {
+            "label": "2 columns"
```
Roi-Arthur, line 2456: Should this align with other instances where we only use 1 / 2, like [line 1917](https://github.com/Shopify/dawn/pull/3683/files#diff-8634bdb8283e62e6df834e80736461b1c05b2d6d669fde29e2fe1b9b7e86ed87R1917)?

locales/en.default.schema.json
```diff
@@ -2727,8 +2699,7 @@
           },
           "options__4": {
             "label": "Large"
-          },
-          "info": "For best results, use an image with a 16:9 aspect ratio. [Learn more](https://help.shopify.com/manual/shopify-admin/productivity-tools/image-editor#understanding-image-aspect-ratio)"
```
Roi-Arthur, line 2731: Similar to previous comments about mentioning aspect ratio

locales/en.default.schema.json
```diff
@@ -2755,11 +2726,10 @@
           "options__3": {
             "label": "Right"
           },
-          "label": "Mobile content alignment"
+          "label": "Content alignment"
         },
         "show_text_below": {
-          "label": "Show content below image on mobile",
-          "info": "For best results, use an image with a 16:9 aspect ratio. [Learn more](https://help.shopify.com/manual/shopify-admin/productivity-tools/image-editor#understanding-image-aspect-ratio)"
```
Roi-Arthur, line 2762: Similar to previous comments about mentioning aspect ratio

locales/en.default.schema.json
```diff
@@ -1168,7 +1160,7 @@
           "content": "Mobile layout"
```
Roi-Arthur, line 1160: Other settings use "Mobile Layout" (double capital)

locales/en.default.schema.json
```diff
@@ -3111,20 +3083,21 @@
               "label": "Subheading",
               "default": "Tell your brand's story through images"
             },
+            "header_button": {
+            "content": "Button"      
+            },
             "button_label": {
-              "label": "Button label",
-              "info": "Leave the label blank to hide the button.",
```
Roi-Arthur, line 3116: Other similar mentions weren't removed (ex [L#1317](https://github.com/Shopify/dawn/pull/3683/files#diff-8634bdb8283e62e6df834e80736461b1c05b2d6d669fde29e2fe1b9b7e86ed87R1317)).
Should this stay?

locales/en.default.schema.json
```diff
@@ -290,58 +287,54 @@
...
+          "label": "Type",
           "drawer": {
             "label": "Drawer"
           },
           "page": {
             "label": "Page"
           },
           "notification": {
-            "label": "Popup notification"
+            "label": "Popup"
```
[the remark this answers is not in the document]
katycobb: I see what you mean - it almost feels like the "popup notification" should be a separate setting, conditional to choosing the page type. I'm not introducing any breaking changes for this settings cleanup, so I can file this one for a later update. For now I'll revert the change to at least keep the "notification" bit.

locales/en.default.schema.json
```diff
@@ -443,42 +436,41 @@
         "options__8": {
           "label": "Round"
         },
-        "label": "Image shape",
-        "info": "Standard-styled cards have no borders when an image shape is active."
```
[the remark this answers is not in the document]
katycobb: Thanks for the examples! I think what's bugging me here is the help text is describing two conditions set in different spots in your theme: Your card style needs to be "standard" (set in theme settings) AND you have to have applied an image shape, and only then your border (also controlled in theme settings: may or may not be particularly noticeable) is affected. I don't know if there's a simple/informative way of describing this: IMO the current text isn't particularly helpful. 

I'm thinking it's something that's easy enough to experiment with and see the result in the preview if all the conditions align.

locales/en.default.schema.json
```diff
@@ -737,7 +732,7 @@
           "label": "Description"
         },
         "show_description": {
-          "label": "Show collection description from the admin"
+          "label": "Collection description"
```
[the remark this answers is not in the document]
katycobb: received similar feedback from Kevin + addressed!

locales/en.default.schema.json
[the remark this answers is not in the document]
Roi-Arthur: Current text does leave something to be desired, yes; I think the fact that the settings are in different places is exactly why there should be a mention of it though, because otherwise it's kind of a guessing game.
Also makes me question whether we should have both settings together, with either both being global or both being at the section level.

locales/en.default.schema.json
```diff
@@ -443,42 +436,41 @@
...
-          "info": "To display your social media accounts, link them in your [theme settings](/editor?context=theme&category=social%20media)."
+          "content": "Social media",
+          "info": "Manage social accounts in [theme settings](/editor?context=theme&category=social%20media)"
         },
         "header__2": {
           "content": "Announcements"
         },
         "show_social": {
-          "label": "Show icons on desktop"
+          "label": "Icons"
```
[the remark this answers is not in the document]
katycobb: Hm looks like this is the case for the country/region selector and language selector too. I'll look at adding info text under the heading for all three.

locales/en.default.schema.json
```diff
@@ -530,7 +522,7 @@
           "label": "Heading"
         },
         "desktop_layout": {
-          "label": "Desktop layout",
+          "label": "Layout",
```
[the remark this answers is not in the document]
katycobb: Good point. I think the right fix here would actually be renaming the options to remove the directional reference and use "first" and "last" instead. I'll make a note to consider that in a followup.

locales/en.default.schema.json
```diff
@@ -1211,8 +1203,10 @@
           },
           "options__4": {
             "label": "Large"
+          }
           },
-          "info": "For best results, use an image with a 3:2 aspect ratio. [Learn more](https://help.shopify.com/manual/shopify-admin/productivity-tools/image-editor#understanding-image-aspect-ratio)"
```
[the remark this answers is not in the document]
katycobb: Really trying to embrace the newest content guidelines which encourage the barest minimum a merchant needs to know to complete the task at hand, and relying on them using sidekick/help center to find more info if needed.

locales/en.default.schema.json
katycobb: Again leaning into the "provide only what's necessary" content guidance. IMO this all sounds like pretty standard patterns when something is shared online so I think I'm comfortable removing it!

locales/en.default.schema.json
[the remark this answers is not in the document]
katycobb: Connected with Meli for a second UX opinion on this one - we'll remove the info text for now and add a bit more context to the help docs for this setting and the media border area of the theme settings docs.

locales/en.default.schema.json
katycobb: the principle we're trying to follow is: we will gracefully handle the mobile expression of all settings without needing to draw attention to it.

locales/en.default.schema.json
```diff
@@ -619,6 +608,9 @@
           "default": "Collections",
           "label": "Heading"
         },
+         "header_layout": {
+        "content": "Layout"
+        },
```
ludoboludo, line 613: We have many references to `Layout` across the file. There could be an opportunity in consolidating them like we do for `sections.all.padding` etc. 
But it's not something that needs to happen now, it will mess with the translations we're getting back and add some more wait time.

locales/en.default.schema.json
```diff
@@ -699,27 +691,33 @@
...
+          "label": "Post count"
+        }, 
         "columns_desktop": {
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
         "show_view_all": {
-          "label": "Enable \"View all\" button if blog includes more blog posts than shown"
+          "label": "\"View all\" button",
+          "info": "Visible if blog has more posts than shown"
```
ludoboludo, line 711: Not too sure myself when it should end with a period or not.

locales/en.default.schema.json
```diff
@@ -751,17 +752,24 @@
...
+          "label": "Product count"
         },
         "columns_desktop": {
-          "label": "Number of columns on desktop"
+          "label": "Columns"
         },
         "show_view_all": {
-          "label": "Enable \"View all\" if collection has more products than shown"
+          "label": "\"View all\" button",
+          "info": "Visible if collection has more products than shown"
```
ludoboludo, line 772: Same here for the period or not.

ludoboludo (commented):
Looking good, just a few comments while we wait for the rest of the translations

locales/en.default.schema.json
```diff
@@ -178,11 +177,10 @@
       "name": "Typography",
       "settings": {
         "type_header_font": {
-          "label": "Font",
-          "info": "Selecting a different font can affect the speed of your store. [Learn more about system fonts.](https://help.shopify.com/manual/online-store/os/store-speed/improving-speed#fonts)"
+          "label": "Font"
```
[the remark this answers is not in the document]
katycobb: This info lives inside the font picker as well, which I think is actually the better placement since the merchant would have shown shown intent to change the font first 🙂

locales/en.default.schema.json
katycobb: Yes good call - it would make sense to unify it

locales/en.default.schema.json
katycobb: The new guidance is no period unless it's more than once sentence. So most of these won't have a period unless they're followed by a "Learn more" link.

locales/en.default.schema.json
```diff
@@ -443,42 +436,35 @@
...
           "label": "Change every"
         },
-        "header__1": {
-          "content": "Social media icons",
-          "info": "To display your social media accounts, link them in your [theme settings](/editor?context=theme&category=social%20media)."
-        },
-        "header__2": {
-          "content": "Announcements"
+        "heading_utilities": {
+         "content": "Utilities"
```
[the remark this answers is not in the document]
katycobb: Oop my bad! @tyleralsbury led me astray 😃. @ludoboludo if you're familiar with how to add translation notes to the code I'd love to pair on it if you have time.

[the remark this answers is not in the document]
ludoboludo: Actually it might just be not doable at present for us as JSON doesn't support comments 🤔 Maybe something ill bring to those folks and see what they say is best 😄

ludoboludo (approved):

Roi-Arthur (approved):
:ouioui:

bakura10:
We would love to migrate our themes to use a similar wording. Does Shopify has some guidelines on this?

katycobb:
Hi @bakura10, great to hear. I'll work on getting some guidelines up in the dev docs shortly.

PR #3710, opened 2025-01-28
Update quick order list to use paginated product.variants

lhoffbeck:
### PR Summary: 

Updates quick order list component to rely on liquid variant pagination (releasing early 02/2025).

**What this IS**
Product.variants is capped to return a max of 250 objects, see [docs for more details](https://shopify.dev/docs/storefronts/themes/product-merchandising/variants/support-high-variant-products). As a result, the quick order list as-is can't contain a comprehensive view of all of a product's variants.

The main changes in this PR are to (1) paginate product.variants when rendering quick order list and (2) display pagination links within quick order list. There are a number of other minor changes to refactor and simplify the component.

**What this IS NOT**
While we did fix a number of existing bugs, this is NOT intended to be a full rebuild of the quick order list. 

### Decision log

|---|---|---|---|---|
| 1 |   |   |   |   |

### Testing steps/scenarios

- [ ] Step 1

### Demo links

- [Store](url)
- [Editor](url)

### Checklist

### DIFF
--- a/assets/global.js
+++ b/assets/global.js
@@ -1193,7 +1193,7 @@ class BulkAdd extends HTMLElement {
   constructor() {
     super();
     this.queue = [];
-    this.requestStarted = false;
+    this.setRequestStarted(false);
     this.ids = [];
   }
 
@@ -1211,7 +1211,7 @@ class BulkAdd extends HTMLElement {
   }
 
   sendRequest(queue) {
-    this.requestStarted = true;
+    this.setRequestStarted(true);
     const items = {};
     queue.forEach((queueItem) => {
       items[parseInt(queueItem.id)] = queueItem.quantity;
@@ -1219,6 +1219,10 @@ class BulkAdd extends HTMLElement {
     this.queue = this.queue.filter((queueElement) => !queue.includes(queueElement));
     const quickBulkElement = this.closest('quick-order-list') || this.closest('quick-add-bulk');
     quickBulkElement.updateMultipleQty(items);
+  }
+
+  setRequestStarted(requestStarted) {
+    this._requestStarted = requestStarted;
   }
 
   resetQuantityInput(id) {
@@ -1247,15 +1251,8 @@ class BulkAdd extends HTMLElement {
     } else {
       event.target.setCustomValidity('');
       event.target.reportValidity();
+      event.target.setAttribute('value', inputValue);
       this.startQueue(index, inputValue);
-    }
-  }
-
-  getSectionsUrl() {
-    if (window.pageNumber) {
-      return `${window.location.pathname}?page=${window.pageNumber}`;
-    } else {
-      return `${window.location.pathname}`;
     }
   }
 
--- a/assets/quick-add-bulk.js
+++ b/assets/quick-add-bulk.js
@@ -18,8 +18,6 @@ if (!customElements.get('quick-add-bulk')) {
         this.listenForActiveInput();
         this.listenForKeydown();
         this.lastActiveInputId = null;
-        const pageParams = new URLSearchParams(window.location.search);
-        window.pageNumber = decodeURIComponent(pageParams.get('page') || '');
       }
 
       connectedCallback() {
@@ -55,15 +53,15 @@ if (!customElements.get('quick-add-bulk')) {
 
       listenForActiveInput() {
         if (!this.classList.contains('hidden')) {
-          this.getInput().addEventListener('focusin', (event) => event.target.select());
+          this.getInput()?.addEventListener('focusin', (event) => event.target.select());
         }
         this.isEnterPressed = false;
       }
 
       listenForKeydown() {
-        this.getInput().addEventListener('keydown', (event) => {
+        this.getInput()?.addEventListener('keydown', (event) => {
           if (event.key === 'Enter') {
-            this.getInput().blur();
+            this.getInput()?.blur();
             this.isEnterPressed = true;
           }
         });
@@ -100,6 +98,13 @@ if (!customElements.get('quick-add-bulk')) {
         });
       }
 
+      getSectionsUrl() {
+        const pageParams = new URLSearchParams(window.location.search);
+        const pageNumber = decodeURIComponent(pageParams.get('page') || '');
+
+        return `${window.location.pathname}${pageNumber ? `?page=${pageNumber}` : ''}`;
+      }
+
       updateMultipleQty(items) {
         this.selectProgressBar().classList.remove('hidden');
 
@@ -130,7 +135,7 @@ if (!customElements.get('quick-add-bulk')) {
           })
           .finally(() => {
             this.selectProgressBar().classList.add('hidden');
-            this.requestStarted = false;
+            this.setRequestStarted(false);
           });
       }
 
--- a/assets/quick-order-list.js
+++ b/assets/quick-order-list.js
@@ -1,88 +1,36 @@
-if (!customElements.get('quick-order-list-remove-button')) {
-  customElements.define(
-    'quick-order-list-remove-button',
-    class QuickOrderListRemoveButton extends BulkAdd {
-      constructor() {
-        super();
-        this.addEventListener('click', (event) => {
-          event.preventDefault();
-          this.startQueue(this.dataset.index, 0);
-        });
-      }
-    }
-  );
-}
-
-if (!customElements.get('quick-order-list-remove-all-button')) {
-  customElements.define(
-    'quick-order-list-remove-all-button',
-    class QuickOrderListRemoveAllButton extends HTMLElement {
-      constructor() {
-        super();
-        this.quickOrderList = this.closest('quick-order-list');
-        const allVariants = this.quickOrderList.querySelectorAll('[data-quantity-variant-id]');
-        const items = {};
-        let hasVariantsInCart = false;
-
-        allVariants.forEach((variant) => {
-          const cartQty = parseInt(variant.dataset.cartQuantity);
-          if (cartQty > 0) {
-            hasVariantsInCart = true;
-            items[parseInt(variant.dataset.quantityVariantId)] = 0;
-          }
-        });
-
-        if (!hasVariantsInCart) {
-          this.classList.add('hidden');
-        }
-
-        this.actions = {
-          confirm: 'confirm',
-          remove: 'remove',
-          cancel: 'cancel',
-        };
-
-        this.addEventListener('click', (event) => {
-          event.preventDefault();
-          if (this.dataset.action === this.actions.confirm) {
-            this.toggleConfirmation(false, true);
-          } else if (this.dataset.action === this.actions.remove) {
-            this.quickOrderList.updateMultipleQty(items);
-            this.toggleConfirmation(true, false);
-          } else if (this.dataset.action === this.actions.cancel) {
-            this.toggleConfirmation(true, false);
-          }
-        });
-      }
-
-      toggleConfirmation(showConfirmation, showInfo) {
-        this.quickOrderList
-          .querySelector('.quick-order-list-total__confirmation')
-          .classList.toggle('hidden', showConfirmation);
-        this.quickOrderList.querySelector('.quick-order-list-total__info').classList.toggle('hidden', showInfo);
-      }
-    }
-  );
-}
-
 if (!customElements.get('quick-order-list')) {
   customElements.define(
     'quick-order-list',
     class QuickOrderList extends BulkAdd {
+      cartUpdateUnsubscriber = undefined;
+
+      sectionsToRender = [
+        {
+          id: 'cart-icon-bubble',
+          section: 'cart-icon-bubble',
+          selector: '.shopify-section',
+        },
+        {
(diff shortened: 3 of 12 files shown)


assets/global.js
```diff
@@ -1247,18 +1257,11 @@ class BulkAdd extends HTMLElement {
...
     }
   }
 
-  getSectionsUrl() {
-    if (window.pageNumber) {
-      return `${window.location.pathname}?page=${window.pageNumber}`;
-    } else {
-      return `${window.location.pathname}`;
-    }
-  }
```
lhoffbeck, line 1260: This was only used by the QuickOrderBulk subclass, no need for it to be on the parent.

assets/quick-add-bulk.js
```diff
@@ -55,15 +53,15 @@ if (!customElements.get('quick-add-bulk')) {
...
       }
 
       listenForKeydown() {
-        this.getInput().addEventListener('keydown', (event) => {
+        this.getInput()?.addEventListener('keydown', (event) => {
           if (event.key === 'Enter') {
-            this.getInput().blur();
+            this.getInput()?.blur();
             this.isEnterPressed = true;
           }
```
lhoffbeck, line 66: 🐛-fix not related to this PR. This was throwing an error if quick add bulk was in a sold out state and no input was shown.

assets/quick-order-list.js
```diff
@@ -1,307 +1,281 @@
...
-      constructor() {
-        super();
-        this.addEventListener('click', (event) => {
-          event.preventDefault();
-          this.startQueue(this.dataset.index, 0);
-        });
-      }
-    }
-  );
-}
```
lhoffbeck, line 14: Replaced this with a simple button so that events would be processed in the same QuickOrderList queue as the other quantity updates.

assets/quick-order-list.js
```diff
@@ -1,307 +1,281 @@
...
-
-if (!customElements.get('quick-order-list-remove-all-button')) {
-  customElements.define(
-    'quick-order-list-remove-all-button',
-    class QuickOrderListRemoveAllButton extends HTMLElement {
```
lhoffbeck, line 19: This isn't actually removed, just moved after QuickOrderList to fix initialization order

assets/quick-order-list.js
```diff
@@ -1,307 +1,281 @@
...
-        this.quickOrderList.querySelector('.quick-order-list-total__info').classList.toggle('hidden', showInfo);
-      }
-    }
-  );
-}
-
 if (!customElements.get('quick-order-list')) {
   customElements.define(
     'quick-order-list',
     class QuickOrderList extends BulkAdd {
```
lhoffbeck, line 4: There's a decent amount of refactoring to QuickOrderList, we were able to simplify a decent amount of code + remove code that didn't work ([for example](https://github.com/Shopify/dawn/pull/3710/files#diff-e2e672962a587498206c828b48ac000e3569e397b960aeff5491353f78d94553L83-L84)) / work correctly ([for example](https://github.com/Shopify/dawn/pull/3710/files#r1937347233)).

assets/quick-order-list.js
```diff
@@ -1,307 +1,281 @@
...
+          });
+        });
+      }
+
+      get currentPage() {
+        return this.querySelector('.pagination-wrapper')?.dataset?.page ?? '1';
+      }
+
+      get cartVariantsForProduct() {
+        return JSON.parse(this.querySelector('[data-cart-contents]')?.innerHTML || '[]');
```
lhoffbeck, line 93: This is the secret sauce for remove all. Where we used to read variant inputs with non-zero quantities out of the DOM (which doesn't work with paginated variants), we now grab the JSON dump of this cart product ([liquid change](https://github.com/Shopify/dawn/pull/3710/files#diff-544215e8ade6cd5533a29ec01cff17d5dd80698302f9fdc9828b774306f8a198R288-R290)).

assets/quick-order-list.js
```diff
@@ -1,307 +1,279 @@
...
-              resolve();
-            })
-            .catch((e) => {
-              console.error(e);
-              reject(e);
-            });
-        });
-      }
-
       getSectionsToRender() {
```
lhoffbeck, line 132: Sections / targets got updated to [simplify cart update the same as in quick-add-bulk](https://github.com/Shopify/dawn/pull/3710/files#r1937325971) and remove the duplicate section for quick order list.

assets/quick-order-list.js
```diff
@@ -1,307 +1,279 @@
...
+
+            this.innerHTML = responseQuickOrderList.innerHTML;
+            this.initEventListeners();
+          })
+          .catch((e) => {
+            console.error(e);
+          });
+      }
+
+      renderSections(parsedState) {
```
lhoffbeck, line 186: Primary change here is to how the quickorderlist section is updated to account for pagination. Fixed a few bugs too:
- focus state is now preserved after update
- no longer ignores cart updates to this product made from other tabs

assets/quick-order-list.js
```diff
@@ -477,18 +456,51 @@ if (!customElements.get('quick-order-list')) {
...
+          remove: 'remove',
+          cancel: 'cancel',
+        };
+
+        this.addEventListener('click', (event) => {
+          event.preventDefault();
+          if (this.dataset.action === this.actions.confirm) {
+            this.toggleConfirmation(false, true);
+          } else if (this.dataset.action === this.actions.remove) {
+            const items = this.quickOrderList.cartVariantsForProduct.reduce(
```
lhoffbeck, line 486: This class was moved down from the start of the file. Only other real change is that we grab the total cart contents from `this.quickOrderList.cartVariantsForProduct` rather than reading from the DOM.

snippets/quick-order-list.liquid
```diff
@@ -281,11 +284,17 @@
...
+          {{ cart | line_items_for: product | map: "variant_id" | json }}
+        </script>
       </div>
     {%- endif -%}
   </quick-order-list>
   <template id="QuickOrderListErrorTemplate-{{ product.id }}">
-    {{ 'icon-error.svg' | inline_asset_content }}
+    <span class="svg-wrapper">
+      {{ 'icon-error.svg' | inline_asset_content }}
+    </span>
```
lhoffbeck, line 297: Fixes bug where error message would have a gigantic icon

ludoboludo (commented):
A few things I noticed while testing though they might have been existing bug/issues 🤔 

- the loading spinner when changing page seem odd to me as it's next to `view cart` though we're not updating anything about the cart ([screenshot](https://screenshot.click/31-30-i47vb-ucneu.png))
- when I go to remove all, it seem to lose my focus position and sends me back to the top of the product list - [recording](https://screenshot.click/31-31-pm98p-czn76.mp4)
- some odd scroll behaviour both on the pdp and modal - [recording](https://screenshot.click/31-42-4687b-7au82.mp4)

Oliviammarcello:
> -the loading spinner when changing page seem odd to me as it's next to view cart though we're not updating anything about the cart ([screenshot](https://screenshot.click/31-30-i47vb-ucneu.png))

> I don't have major feelings about this one way or another. I like the spinner replacing the remove all button when the page is changing to help prevent foot-guns, but we could also handle that by just disabling/enabling that button with the rest of the table. Other thing to call out is that the cart quantity could be updated when changing pages if any quantities changed in another tab.

I agree with your intent @ludoboludo, but with page jumps on load I think keeping the spinner in the "footer" area is probably best to ensure it is always visible.

snippets/quantity-input.liquid
```diff
@@ -11,7 +11,7 @@
 {% endcomment %}
 
 <quantity-input class="quantity cart-quantity">
-  <button class="quantity__button" name="minus" type="button">
+  <button class="quantity__button" name="minus" type="button" data-target="decrement-{{ variant.id }}">
```
lhoffbeck, line 14: `data-target` enables us to refocus inputs after content is updated so keyboard order doesn't get goofed up

lhoffbeck:
@ludoboludo  I just pushed up a change that simplified how we do event handling, poking around it seems like it resolved all the funky focus/scroll issues you were running into. Mind taking a look again?

oksanashopify (approved):
I tophatted changes and described workflows works as expected for me. Great work 🚀 

[questions]

1. Sorry if it's not related to changes in this PR. When we hit "remove" icon, there is no visual update for customer that quantity is being removed and customer just can continue to hit "remove" button. [Video](https://screenshot.click/03-00-5p1bx-ct7gy.mp4). Shout it have disabled or maybe loading state?

lhoffbeck:
> 1. Sorry if it's not related to changes in this PR. When we hit "remove" icon, there is no visual update for customer that quantity is being removed and customer just can continue to hit "remove" button. [Video](https://screenshot.click/03-00-5p1bx-ct7gy.mp4). Shout it have disabled or maybe loading state?

Yeah agreed, that's pretty gross. I'll flag it with the team that owns the component to see if they want to do some UX exploration 👍

assets/quick-add-bulk.js
```diff
@@ -55,15 +53,15 @@ if (!customElements.get('quick-add-bulk')) {
 
       listenForActiveInput() {
         if (!this.classList.contains('hidden')) {
-          this.getInput().addEventListener('focusin', (event) => event.target.select());
+          this.getInput()?.addEventListener('focusin', (event) => event.target.select());
```
LA1CH3, line 56: Nit but to remove some repetition in a few of these areas, you could do something like:

```ts
// or just rename `getInput` as a getter
get input() {
  return this.getInput();
}

listenForActiveInput() {
  if (!this.classList.contains('hidden') && this.input) {
  this.input.addEventListener(...);
  }
}

// same thing for `listenForKeydown`
```

assets/quick-order-list.js
```diff
@@ -1,307 +1,280 @@
...
           };
         }
 
-        if (this.getTotalBar()) {
-          this.totalBarPosition = window.innerHeight - this.getTotalBar().offsetHeight;
+        this.totalBar = this.getTotalBar();
+        if (this.totalBar) {
+          this.totalBarPosition = window.innerHeight - this.totalBar.offsetHeight;
 
           window.addEventListener('resize', () => {
```
LA1CH3, line 23: Pre-existing issue, but we should probably return the event listener here so we can unsubscribe from the resize event when the quick order list is unmounted/detached.

assets/quick-order-list.js
```diff
@@ -1,307 +1,280 @@
...
+        this.quickOrderListTable.addEventListener('keydown', this.handleSwitchVariantOnEnter.bind(this));
+
+        this.initVariantEventListeners();
+      }
+
+      initVariantEventListeners() {
+        this.allInputsArray = Array.from(this.querySelectorAll('input[type="number"]'));
+
+        this.querySelectorAll('quantity-input').forEach((qty) => {
+          const debouncedOnChange = debounce(this.onChange.bind(this), 250);
```
LA1CH3, line 78: Nit- can we store this debounce time value as a constant somewhere?

assets/quick-order-list.js
LA1CH3, line 1: I realize this is all pre-existing, but there's a huge amount of selectors just hard-coded and repeated throughout this file- would be great to store them all as constants/variables and just reference those.

assets/quick-order-list.js
```diff
@@ -318,84 +291,69 @@ if (!customElements.get('quick-order-list')) {
...
-            publish(PUB_SUB_EVENTS.cartUpdate, { source: this.quickOrderListId, cartData: parsedState });
+            this.renderSections(parsedState);
+            publish(PUB_SUB_EVENTS.cartUpdate, {
+              source: this.id,
+              cartData: parsedState,
+            });
           })
-          .catch(() => {
+          .catch((e) => {
+            console.error(e);
```
LA1CH3, line 351: Is there a way to emit errors to bugsnag from here?

LA1CH3 (commented):
Lacking storefront context but based off PR description LGTM! 

Only thing I am unsure about whether its expected or not is that for a product without paginated variants, there is no pagination (as expected) but also no "Total items", "Product subtotal", "View cart" and "Remove all"- is that expected? 


https://github.com/user-attachments/assets/4cc55580-d1cf-411f-acf9-42fca49bb151

assets/quick-order-list.js
lhoffbeck: Yeah, totally agree. I was trying to keep this PR as tight as possible to avoid introducing regressions.

lhoffbeck:
> Lacking storefront context but based off PR description LGTM!
> 
> Only thing I am unsure about whether its expected or not is that for a product without paginated variants, there is no pagination (as expected) but also no "Total items", "Product subtotal", "View cart" and "Remove all"- is that expected?
> 
>  single.variant.vs.multi.variant.sf.mp4

Yeah good callout, [this is existing behavior](https://github.com/Shopify/dawn/blob/755f6a448e25d40a8eb57550948ff3d074cf3818/snippets/quick-order-list.liquid#L121-L129). My assumption is that the original devs built it like this because there's very little (no?) customer benefit to showing a quick order list with a single variant vs the buyer just using the main product form.

PR #3723, opened 2025-02-12
Fix: Adjust `disclosure__list` dropdown position for RTL layout in header localization

heXaCo0l:
Ensured the dropdown aligns correctly and remains fully visible without causing overflow.


heXaCo0l:
I have signed the CLA!

PR #3772, opened 2025-04-23
[WIP] Updated Unit Price Support

sunblaze:
### PR Summary: 

Updates unit pricing to support new units and how they're displayed with translation support.

New units would not be displayed correctly without updating the theme.

Chose to implement it completely in liquid.

### Other considerations

- We considered adding a new filter to liquid
- We debated adding a formatted string in the unit price measurement drop

- remove css that capitalized all of unit prices (which included the units)
- item/each display for new measured_type
- m2 is now m² etc.

### Testing steps/scenarios

- [ ] Test all the variants that they display correctly

### Demo links

- [Store](https://sunblaze.myshopify.com/products/ball)
- [Editor](https://admin.shopify.com/store/sunblaze/themes/163916906552?key=snippets%2Funit-price.liquid)

### Checklist


locales/en.default.json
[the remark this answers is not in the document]
sunblaze: Yep. Plan to copy paste them once they all get translated. Crapy that it's duplication, but I think it's worth the tradeoff

snippets/price.liquid
```diff
@@ -108,20 +108,9 @@
...
-          <span class="visually-hidden">&nbsp;{{ 'accessibility.unit_price_separator' | t }}&nbsp;</span>
-          <span>
-            {%- if product.selected_or_first_available_variant.unit_price_measurement.reference_value != 1 -%}
-              {{- product.selected_or_first_available_variant.unit_price_measurement.reference_value -}}
-            {%- endif -%}
-            {{ product.selected_or_first_available_variant.unit_price_measurement.reference_unit }}
-          </span>
-        </span>
-      </small>
+      {%- if use_variant -%}
```
trishrempel, line 111: Where does `use_variant` and `target` come from?
I'm wondering why the code changed to use these instead of `product.selected_or_first_available_variant`.

locales/en.default.json
```diff
@@ -164,6 +164,47 @@
...
       },
+      "unit_price": {
+        "per_item": {
+          "single": "{{price}} each",
+          "one": "{{price}} per {{count}} item",
+          "other": "{{price}} per {{count}} items"
+        },
+        "per_unit": {
+          "single": "{{price}}/{{unit}}",
+          "one": "{{price}}/{{unit}}",
```
trishrempel, line 175: You'd want to add `{{count}}` here, because the strings between `one` and `other` need to have the same interpolation strings.

```suggestion
          "one": "{{price}}/{{count}}{{unit}}",
```

trishrempel (approved):

locales/en.default.json
```diff
@@ -164,6 +164,47 @@
         "sale_price": "Sale price",
         "unit_price": "Unit price"
       },
+      "unit_price": {
+        "per_item": {
+          "single": "{{price}} each",
+          "one": "{{price}} per {{count}} item",
```
Roi-Arthur, line 170: Would this line ever come up? Since we do
```
{% if count %}
  {%- if count == 1 -%}
    {{- 'products.product.unit_price.per_item.single' | t: price: price -}}
  {%- else -%}
    {{- 'products.product.unit_price.per_item' | t: price: price, count: count -}}
  {%- endif -%}
{% endif %}
```

sunblaze: Yes and no. One is not always 1 in translations in some languages take [Slovenian](https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html#sl), 101, 201, etc. also use it.

But @trishrempel discovered storefront render doesn't use the I18n backend to do this as it should to pluralize, so currently "one" isn't possible to be use no matter what language is used. But we want this to be future proofed for when that's fixed

Roi-Arthur: > One is not always 1 in translations in some languages take [Slovenian](https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html#sl), 101, 201, etc. also use it.

Well that would be for the Slovenian file to decide? By which I mean translators would know to give us the proper keys (whether one, two, many, other, etc…)?
The English file doesn't need that string if it's not used, though for that matter I suppose I'm actually more surprised that we have a "_single_" key than a "_one_" 🤔

Roi-Arthur: Ah fair enough about the semantic difference, thanks for those links!

sections/main-cart-items.liquid
```diff
@@ -102,7 +102,7 @@
                         <p class="caption-with-letter-spacing">{{ item.product.vendor }}</p>
                       {%- endif -%}
 
-                      <a href="{{ item.url }}" class="cart-item__name h4 break">{{ item.product.title | escape }}</a>
+                      <a href="{{ item.url }}" class="cart-item__name h4 break">{{- item.product.title | escape }}</a>
```
trishrempel, line 105: I think it makes sense to keep the dash out of here?

```suggestion
                      <a href="{{ item.url }}" class="cart-item__name h4 break">{{ item.product.title | escape }}</a>
```

trishrempel (approved):

trishrempel:
@bakura10 Thanks for the feedback, you might like [this PR](https://github.com/Shopify/dawn/pull/3805) better 😁

trishrempel:
Closing in favour of https://github.com/Shopify/dawn/pull/3805

PR #3803, opened 2025-06-05
[Cart] Support nested lines

graciebee:
### PR Summary: 

Surface Nested Cart Lines

Addresses https://github.com/Shopify/horizon/issues/4768

The [mocks provided to us by UX](https://www.figma.com/design/A7Hei3wsnqQRd4etZLe5ye/%E2%9E%95-Nested-Cart-Lines--formerly-Add-Ons-?node-id=5492-21592&p=f&t=iCiLdmxXRfpg9ZBY-0) use the indentation and size of the thumbnail to indicate that the item is nested. Therefore, I conditionally applied a class `.cart-item__nested-line` to the `tr` if the `cart.item` had a `parent`. There is also slightly less padding between the nested line and the parent line, so I updated the top padding of the nested line to be `1rem` instead of `4rem` for desktop screens, and the bottom margin to be `1.5rem` instead of `3.5rem` for mobile screens. For the cart drawer, the top padding is `1rem` instead of `1.7rem`.

### Other considerations

1. The relationship between the parent and the nested line is created at the time of adding the item to the cart, by passing along the `key` of the parent item as `parent_line_key` in the nested line. This is not something we are building natively into the themes, as the expectation is that 3p apps would handle this (with app blocks etc.). For the purposes of my test store linked below for demo-ing, I hard coded a block into my `buy-button.liquid` to assign the `Smart TV` as parent if added to the cart. If you would like to test on your own store though, you can update your `buy-button.liquid` with the following block of code, where `{ID}` is the id of the nested line product, and `{KEY}` is the key of the parent line.
```liquid
        {% comment %} line 59, within the 'product' form {% endcomment %}
        {% if product.selected_or_first_available_variant.id == {ID} %}
          <input
            type="hidden"
            name="parent_line_key"
            value="{KEY}"
          >
        {% endif %}
```

If the merchant has nested lines enabled, any nested lines will appear nested under their parents, as opposed to a flat list in the cart.

**desktop view**
image

**mobile view**
image

**cart drawer**
image

### Testing steps/scenarios
Note: the products described in this testing steps are related to the linked demo store. See the Other Considerations 2 point about using your own test shop.
- [ ] Add the Smart TV to your cart
- [ ] Navigate to the Warranty product page and add the item to your cart (Quick buy button on main page won't work)
- [ ] Navigate to cart, notice that the Warranty thumbnail is smaller and indented slightly in comparison to the other items in the cart.
- [ ] Ensure expected style is maintained across multiple screen sizes

### Demo links

- [Store](https://abc-add-ons.myshopify.com/)
- [Editor](https://admin.shopify.com/store/abc-add-ons/themes/151048814822/editor)

### Checklist

### DIFF
--- a/assets/component-cart-items.css
+++ b/assets/component-cart-items.css
@@ -65,6 +65,13 @@ cart-items .title-wrapper-with-link {
   position: relative;
 }
 
+.cart-item__nested-line .cart-item__media {
+  text-align: right;
+  .cart-item__image-container {
+    width: 50%;
+  }
+}
+
 .cart-item__link {
   display: block;
   bottom: 0;
@@ -288,6 +295,10 @@ cart-remove-button .icon-remove {
   .cart-items td {
     vertical-align: top;
     padding-top: 4rem;
+  }
+
+  .cart-items .cart-item__nested-line td {
+    padding-top: 1rem;
   }
 
   .cart-item {
--- a/assets/component-cart-items.css
+++ b/assets/component-cart-items.css
@@ -68,7 +68,7 @@ cart-items .title-wrapper-with-link {
 .cart-item__nested-line .cart-item__media {
   text-align: right;
   .cart-item__image-container {
-    width: 50%;
+    width: 66%;
   }
 }
 
--- a/assets/component-cart-items.css
+++ b/assets/component-cart-items.css
@@ -32,6 +32,14 @@ cart-items .title-wrapper-with-link {
   align-items: flex-start;
 }
 
+.cart-item__nested-line .cart-item__image-container {
+  min-width: calc(10rem / var(--font-body-scale));
+  justify-content: right;
+  img {
+    width: 66%;
+  }
+}
+
 .cart-item__image-container:after {
   content: none;
 }
@@ -63,13 +71,6 @@ cart-items .title-wrapper-with-link {
 
 .cart-item__media {
   position: relative;
-}
-
-.cart-item__nested-line .cart-item__media {
-  text-align: right;
-  .cart-item__image-container {
-    width: 66%;
-  }
 }
 
 .cart-item__link {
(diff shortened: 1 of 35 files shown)


carrotderek (commented):
🎩 works well. Is there a concern on the Cart drawer that product thumbnails don't have a border?

Screenshot 2025-06-06 at 10 39 03 AM

NathanPJF (commented):
Let's continue the discussion over in the Horizon codebase as there's some implementation items I'd like to go over. Please ping me on a PR or issue over there once it's started.

patrickDouglas:
FYI, store password is `winnie`

sections/main-cart-items.liquid
```diff
@@ -230,7 +236,15 @@
...
-                              <button class="quantity__button" name="minus" type="button">
+                              <button
+                                class="quantity__button"
+                                name="minus"
+                                type="button"
+                                {% assign can_update_quantity = item.instructions.can_update_quantity | default: true %}
+                                {% if can_update_quantity == false %}
+                                  disabled
+                                {% endif %}
+                              >
```
FCalabria, line 247: (nit) I'd move the assign outside of the button, like this:
```suggestion
                              {% assign can_update_quantity = item.instructions.can_update_quantity | default: true %}
                              <button
                                class="quantity__button"
                                name="minus"
                                type="button"
                                {% if can_update_quantity == false %}
                                  disabled
                                {% endif %}
                              >
```

This way we don't risk it being undefined in the other uses if for some reason we remove or move the button to a different place.

assets/component-cart-items.css
```diff
@@ -290,6 +302,10 @@ cart-remove-button .icon-remove {
     padding-top: 4rem;
   }
 
+  .cart-items .cart-item__nested-line td {
+    padding-top: 1rem;
```
FCalabria, line 306: This also affects the drawer's second line with the quantity selector. Are you sure it should be that way?

FCalabria (approved):
Just two comments, one is optional and the other is something for you to double check, but feel free to merge without asking for a second review on my side. Very straightforward change!

kjellr (approved):
This works in my testing, and matches the mockups I've seen. 

It _is_ weird that the cart drawer doesn't display a border around the images, while the cart page does ([as noted here](Is there a concern on the Cart drawer that product thumbnails don't have a border?)). But that's not caused by this PR so I think it's something we could tackle separately.

PR #3857, opened 2025-09-23
Add separate custom font support for heading and body

HarshankPatel9038:
- **Added checkboxes to enable custom heading and custom body fonts independently.**
- **Updated settings_schema.json with font URL and family fields for both heading and body.**
- **Improved font loader logic to handle all common font variants (100–900, normal & italic).**
- **Supports both local .woff2 fonts and Google Fonts URLs.**
- **This makes theme font customization more flexible, maintainable, and clear for merchants.**

### DIFF
--- a/config/settings_schema.json
+++ b/config/settings_schema.json
@@ -109,10 +109,29 @@
         "content": "t:settings_schema.typography.settings.header__1.content"
       },
       {
+        "type": "checkbox",
+        "id": "enable_custom_heading_font",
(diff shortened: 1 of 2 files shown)


HarshankPatel9038:
Update ThemeThis PR introduces a reusable and unified custom font loading system for both heading and body fonts. Merchants can now enable custom fonts independently for headings and body text, using either local .woff2 files (with auto-generated weight/style variants) or external Google Fonts links.

PR #3870, opened 2025-10-27
Fix assignment in render tag

graygilmore:
The only required change here is inside the render tag for `render 'swatch-input'` where we previously defined `value: value | escape`. This will throw a syntax error in v5.9.0 of Liquid using the new rigid parser.


karreiro (approved):
Thank you, @graygilmore!

EvilGenius13 (approved):

Roi-Arthur (approved):

PR #3943, opened 2026-07-26
ignore

dgopsq:
Documents how this theme is built, and gives agents a way to actually *see* their changes rather than inferring correctness from the code.

## Docs

| File | Purpose |
| --- | --- |
| `AGENTS.md` | Tool-neutral entry point — what the theme is, the rules, the preview loop, and a protocol for keeping the docs current |
| `CLAUDE.md` | Reduced to Claude Code specifics; points at `AGENTS.md` |
| `docs/ARCHITECTURE.md` | 13 numbered design decisions with rationale and accepted trade-offs |
| `docs/PREVIEW.md` | The preview loop, a page↔file map, and the `?view=` trick |

`ARCHITECTURE.md` deliberately separates **decisions** from a **drift/cleanup backlog**, so deliberate choices don't get "fixed" and accidents don't get enshrined.

## Findings recorded

Each was verified against the tree, not taken on trust:

- **Upstream tags are not ancestors of `HEAD`** — `git merge upstream/main` will not work. Every Dawn sync is a manual replay.
- **34 files differ from upstream by whitespace alone** — permanent merge conflicts for zero functional gain, caused by format-on-save with a newer Prettier than Dawn was authored with.
- **GemPages and PageFly are entirely dead** — `gp-head.liquid` and `pagefly-main-js.liquid` have zero callers, yet PageFly's translation block sits in all 30 locale files.
- **Two `{% schema %}` blocks fail `JSON.parse`** (trailing commas) — tolerated by Shopify today.
- **`locales/it.json` repurposes `powered_by_shopify` to hold the VAT number** — flagged so nobody "corrects" it.
- `scheme-white-text` is genuinely white-on-white; the dated `!important` block in `base.css` patches its symptoms rather than the scheme.

## Tooling

[`agent-browser`](https://github.com/vercel-labs/agent-browser) as a **local devDependency**, matching the repo's rule that nothing is installed globally.

```sh
npm run dev                                   # background it
npm run shot -- /products/doiy-sweetie-mug    # writes a PNG, prints the path
```

`scripts/shot.mjs` handles what this store needs every time: dismissing the cookie banner, setting the viewport, waiting for the page to settle, and reporting console errors. Flags for `--mobile` / `--tablet`, `--full`, `--dark`, and `--live`.

`npm run browser -- <cmd>` exposes the full CLI for interaction, accessibility snapshots, and `diff screenshot` visual regression.

## Doc upkeep enforcement

`AGENTS.md` carries the protocol (a trigger table, plus "replace, don't accumulate"). A `Stop` hook in `.claude/settings.json` makes it fire rather than stay aspirational — **at most once per session**, and only when theme files changed while docs did not. It explicitly says not to invent a doc change to satisfy the check, so routine edits stay quiet.

## Verification

Run against the live dev server, not just written:

- Local screenshot of a product page — the custom `variant_sku` and `qty_disclaimer` blocks render as documented
- `?view=` template switching returns a different page
- Hot reload picks up a CSS edit with no restart (reverted afterwards)
- `npm run check`: **45 warnings, 0 errors** — unchanged from before this branch

Two corrections made during the work: an initial claim that agents were blocked on Shopify OAuth was wrong (the session caches in `~/.config/shopify-cli-kit-nodejs/`, not `~/.config/shopify/`), and my first `.shopifyignore` additions used bare trailing slashes, which the CLI warns about — switched to `dir/*`, which also silenced a pre-existing warning.

## Not included

The drift backlog in `ARCHITECTURE.md` §3 is documented but **not acted on** — deleting dead GemPages/PageFly files, fixing the schema commas, and re-syncing the drifted layouts are all separate changes. This PR is docs and tooling only; no Liquid, CSS, or JS behaviour changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dgopsq:
Opened against the wrong repository by mistake — this belongs to a downstream fork of Dawn, not upstream. Already closed; apologies for the noise.

